PR #4, opened 2018-09-11
fix make test by fixing glog format

### LINKED ISSUE #3: make test fails becuase glog format
```
>> make test
go test -v -race github.com/bertinatto/ebs-csi-driver/pkg/...
# github.com/bertinatto/ebs-csi-driver/pkg/cloud/devicemanager
pkg/cloud/devicemanager/manager.go:217: Verbose.Infof format %s has arg device of wrong type *devicemanager.Device
```

leakingtapan:
Fixes #3 

Using `%v` as format style since device is a struct instead of string

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ bin
 # Vagrant
 .vagrant
 Vagrantfile
+
+# vim
+*.swp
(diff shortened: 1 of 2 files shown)


leakingtapan:
@bertinatto

Issue #11, opened 2018-09-19
Update example attacher and provisioner to use headless service. 


leakingtapan:
Currently, the [attacher](https://github.com/d-nishi/aws-ebs-csi-driver/blob/master/deploy/kubernetes/attacher.yaml) and [provisioner](https://github.com/d-nishi/aws-ebs-csi-driver/blob/master/deploy/kubernetes/provisioner.yaml) are setting **service.spec.port** to be dummy port which is unnecessary. Because attacher/provisioner should be deployed as stateful set where loading balancing is not needed. 

Instead [Headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) could  be used as best practice in the example yaml file.

ddebroy:
LGTM

PR #12, opened 2018-09-19
Update example attacher and provisioner yaml file to use headless service

leakingtapan:
Fixes: #11 

Tested on my EKS cluster that example app is deployed after recreate attacher/provisioner service as headless.


d-nishi:
@jsafrane @gnufied --- can you please review this?

PR #29, opened 2018-09-30
Add unit test for util package

leakingtapan:
Covering the parse logic for endpoint for different cases.

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestParseEndpoint(t *testing.T) {
+	testCases := []struct {
+		name      string
+		endpoint  string
+		expScheme string
+		expAddr   string
+		expErr    error
+	}{
+		{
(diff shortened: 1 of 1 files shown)


pkg/util/util_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Util test")
+}
```
bertinatto: Ginkgo is currently used only in e2e tests; unit tests are currently written using the stdlib only.

Can we do the same with this unit test to keep the consistency?

Also, the name of this function seems incorrect.

leakingtapan: Im wondering what is intention of using Ginkgo for e2e test, but stdlib for unit test?  After check more, I saw kubernetes/kubernetes code base has the same pattern.  Why not use one framework for both?

gnufied: Because ginkgo is considered more suitable for behaviour driven tests where it is easy to perform check from outside. unit tests etc use stdlib because unit tests aren't necessarily behaviour of k8s. There are some other reasons such as how tagging etc works for ginkgo tests.  For unit tests it is easy to focus and run just one file.

leakingtapan: K. updated test to use stdlib for unit testing.

pkg/util/util_test.go
```diff
@@ -0,0 +1,79 @@
...
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+)
+
+func TestUtil(t *testing.T) {
```
bertinatto: I think a better name would be something like `TestParseEndpoint`, since this is the only function being tested.

pkg/util/util_test.go
```diff
@@ -0,0 +1,79 @@
...
+			expScheme: "tcp",
+			expAddr:   "/127.0.0.1",
+		},
+		{
+			name:      "valid tcp endpoint",
+			endpoint:  "tcp:///127.0.0.1",
+			expScheme: "tcp",
+			expAddr:   "/127.0.0.1",
+		},
+	}
```
bertinatto: It would be nice to test an invalid endpoint as well.

PR #34, opened 2018-10-03
Wait for volume status in e2e test

bertinatto:
These are the bits missing in the unfinished e2e test.

I ran this test in a loop for about 20 minutes and it didn't flake.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -18,13 +18,14 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/bertinatto/ebs-csi-driver/pkg/cloud"
 	csi "github.com/container-storage-interface/spec/lib/go/csi/v0"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -48,10 +49,11 @@ var (
 
 var _ = Describe("EBS CSI Driver", func() {
 
-	It("Should create->attach->stage->mount volume and check if it is writable, then unmount->unstage->detach->delete and check disk is deleted", func() {
+	It("Should create, attach, stage and mount volume, check if it's writable, unmount, unstage, detach, delete, and check if it's deleted", func() {
 
+		r1 := rand.New(rand.NewSource(time.Now().UnixNano()))
 		req := &csi.CreateVolumeRequest{
-			Name:               "volume-name-e2e-test",
+			Name:               fmt.Sprintf("volume-name-e2e-test-%d", r1.Uint64()),
 			CapacityRange:      stdCapRange,
 			VolumeCapabilities: stdVolCap,
 			Parameters:         nil,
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_test.go
```diff
@@ -155,15 +140,17 @@ func testAttachWriteReadDetach(volumeID, volName, nodeID string, readOnly bool)
...
-		err := ioutil.WriteFile(testFile, []byte(testFileContents), 0644)
+		err := ioutil.WriteFile(testFile, testFileContents, 0644)
 		Expect(err).To(BeNil(), "Failed to write file")
+		// Read the file and check if content is correct
+		data, err := ioutil.ReadFile(testFile)
+		Expect(err).To(BeNil(), "Failed to read file")
+		Expect(data).To(Equal(testFileContents), "File content is incorrect")
 	}
 
 	// Unmount Disk
```
leakingtapan: Should move the umount to be right after mount disk succeeds. So that any failures in side file read/write won't cause file is not umounted.

tests/e2e/e2e_test.go
```diff
@@ -177,28 +164,40 @@ func testAttachWriteReadDetach(volumeID, volName, nodeID string, readOnly bool)
...
 }
 
-func waitForVolumes(params *ec2.DescribeVolumesInput, nVolumes int) {
+func waitForVolumeState(volumeID, state string) {
 	backoff := wait.Backoff{
-		Duration: 60 * time.Second,
-		Factor:   1.2,
-		Steps:    21,
+		Duration: 1 * time.Second,
+		Factor:   1.8,
```
leakingtapan: This value looks a bit tricky. Why is 1.8 better than 1.2? Could you add some docs around this?

tests/e2e/e2e_test.go
```diff
@@ -62,37 +64,20 @@ var _ = Describe("EBS CSI Driver", func() {
...
-		descParams := &ec2.DescribeVolumesInput{
-			Filters: []*ec2.Filter{
-				{
-					Name:   aws.String("tag:" + cloud.VolumeNameTagKey),
-					Values: []*string{aws.String(req.GetName())},
-				},
-			},
-		}
-		waitForVolumes(descParams, 1 /* number of expected volumes */)
+		waitForVolumeState(volume.Id, "available")
```
leakingtapan: Feels to me its better to move the `watiForVolume` creation and deletion inside driver `CreateVolume` and `DeleteVolume` so that the volume is guaranteed to be available/removed after the call completes. This also void the case where `ControllerPublishVolume` is called during volume is still being created. 

We could do similar thing for attach (saw your TODO comment on `AttachDisk`) and detach.

bertinatto: I agree that we need to have similar functionality in the `cloud` package (e.g., when creating, deleting, attaching and detaching volumes). However, it seems like the implementation of these functions in the driver should be a lot more sophisticated than this. For instance, the function that waits for the volume attachment status in Kubernetes has some magic [1] that I don't quite understand yet and can be tricky to write (in the right way).

Also, assuming that `CreateDisk` only returns when the volume is ready, e.g., with `available` status, I think it's the task of the test case to make sure that the volume is indeed created here.

[1] https://github.com/bertinatto/kubernetes/blob/5fdb0c7177010711ca7b463c7e1d4f32a163a344/pkg/cloudprovider/providers/aws/aws.go#L1862

tests/e2e/e2e_test.go
bertinatto: Good catch, thanks.

tests/e2e/e2e_test.go
leakingtapan: Got ya. We can punt this to #38 when we have time to properly implement this.

Issue #17, opened 2018-09-25
[Bug] Controller Service uses node AZ to create volume


leakingtapan:
Currently, [AWS cloud interface](https://github.com/d-nishi/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L142) used by EBS CSI controller service is instantiated with the same AZ as where controller service is running. This causes the issue that EBS volume is only created at the AZ where controller service runs. 

For example, if there are three nodes, each in us-west-2a, us-west-2b and us-west-2c. And controller service runs in us-west-2b while application is deployed to us-west-2a by scheduler. During volume creation, the new volume will always be created in us-west-2b which causes application deployment to fail. 

One way to fix this is using node selector to always deploy application in the same AZ as controller. Or use static provision where PV/PVC created statically. However, both approach is non-ideal.

leakingtapan:
@ddebroy thanks for the input. Will take a look at topology awareness framework.

PR #42, opened 2018-10-04
Implement topology awareness support for dynamic provisioning

leakingtapan:
Changes includes:
1. return topology in NodeGetInfo response eg. accessible_topology =
    {"com.amazon.aws.csi.ebs/zone": "us-west-2a"}
2. consumes topology requirement in CreateVolume request and creates
    volume using the given zone while favoring preferred topologies
3. pick a random zone when topology requirement is not provided
    (this is the case of dynamic provisioning without delayed binding)
4. Add unit test for pickAvailabilityZone()
5. Add test case for CreateDisk with zone

Fixes: #17


leakingtapan:
Build will be fixed by #43

pkg/cloud/cloud.go
```diff
@@ -187,9 +204,21 @@ func (c *cloud) CreateDisk(volumeName string, diskOptions *DiskOptions) (*Disk,
...
+		err  error
+	)
+	if diskOptions.AvailabilityZone == nil {
+		zone, err = c.pickRandomAvailabilityZone()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		zone = *diskOptions.AvailabilityZone
+	}
```
bertinatto: What about making `DiskOptions.AvailabilityZone` a `string` instead? Then you can pick a randon AZ if it it's an emtpy string.

Currently it's declared as a pointer, then we make it a string (through `zone`), then make it a pointer again (line 221 below).

pkg/cloud/cloud.go
```diff
@@ -78,16 +80,26 @@ var (
 	ErrAlreadyExists = errors.New("Resource already exists")
 )
 
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
```
bertinatto: nit: personally I'd prefer to move this into `pickRandomAvailabilityZone`, but that's OK if you want to leave it here.

pkg/cloud/cloud.go
```diff
@@ -98,6 +110,9 @@ type EC2 interface {
 	DetachVolume(input *ec2.DetachVolumeInput) (*ec2.VolumeAttachment, error)
 	AttachVolume(input *ec2.AttachVolumeInput) (*ec2.VolumeAttachment, error)
 	DescribeInstances(input *ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error)
+
+	// Get all the zones for current region
+	DescribeAvailabilityZones(input *ec2.DescribeAvailabilityZonesInput) (*ec2.DescribeAvailabilityZonesOutput, error)
```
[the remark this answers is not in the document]
leakingtapan: removed this. I also added a link to aws.ec2 go sdk at top of the EC2 interface, since this interface is just a facade of the full ec2 client.

pkg/cloud/cloud.go
```diff
@@ -187,9 +204,21 @@ func (c *cloud) CreateDisk(volumeName string, diskOptions *DiskOptions) (*Disk,
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		zone = *diskOptions.AvailabilityZone
+	}
+
 	request := &ec2.CreateVolumeInput{
-		AvailabilityZone:  aws.String(m.GetAvailabilityZone()),
+		AvailabilityZone:  aws.String(zone),
```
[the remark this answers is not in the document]
leakingtapan: It does.

pkg/cloud/cloud.go
leakingtapan: Good point. I gave it a second thought and I feels this is more clear that double nil and empty are the same invalid case and this remove the extra logic to do `stringPtr()`. updated the code.

pkg/cloud/cloud_test.go
```diff
@@ -42,8 +42,23 @@ func TestCreateDisk(t *testing.T) {
...
+			},
+			expErr: nil,
+		},
+		{
+			name:       "success: normal with provided zone",
+			volumeName: "vol-test-name",
+			diskOptions: &DiskOptions{
+				CapacityBytes:    util.GiBToBytes(1),
+				Tags:             map[string]string{VolumeNameTagKey: "vol-test"},
+				AvailabilityZone: stringPtr("us-west-2"),
```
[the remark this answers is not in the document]
leakingtapan: Once updated `AvailabilityZone` to be string, `stringPtr` is not needed

pkg/cloud/cloud.go
leakingtapan: rand just need to be seeded once. so leave it at the init func.

leakingtapan:
Addressed the comments and also rebased

PR #44, opened 2018-10-05
Add update-gomock script to generate mock files for testing

leakingtapan:


leakingtapan:
Build will be fixed by #43

PR #51, opened 2018-10-10
Update manifests so driver is registered correctly

### LINKED ISSUE #50: Manifests files no longer work
After `KubeletPluginsWatcher` graduated to beta [1], this feature comes
enabled by default.

Basically, it replaces the older driver registration
mechanism, where the sidecar container used register the driver in the
Kubelet. With `KubeletPluginsWather` enabled, this registrarion is done
by the Kubelet itself.

As a result of this, the currently sample manifests no longer work.

[1] https://github.com/kubernetes/kubernetes/pull/68200/files

/kind bug
/assign

bertinatto:
Fixes #50

/kind bug
/assign @leakingtapan


bertinatto:
Sure, do you mean something like this?

Issue #52, opened 2018-10-10
Cannot attach volume to instance in a different AZ


bertinatto:
PR https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/42
changed the `CreateDisk` behaviour by creating a volume in a random AZ
if such isn't provided. However, the volume might be created in an AZ where there's no driver
running.

This is the error that I'm getting:

```
Warning  FailedAttachVolume  84s  attachdetach-controller  (combined from similar events): AttachVolume.Attach failed for volume "pvc-44c2c835-cc91-11e8-93bb-0e3fe2b5fad4" : rpc error: code = Internal desc = Could not attach volume "vol-06899fef1736f24b9" to node "i-0a7e7ab68218ef0e0": could not attach volume "vol-06899fef1736f24b9" to node "i-0a7e7ab68218ef0e0": InvalidVolume.ZoneMismatch: The volume 'vol-06899fef1736f24b9' is not in the same availability zone as instance 'i-0a7e7ab68218ef0e0'
```

leakingtapan:
@bertinatto this issue could be resolved, right?

PR #53, opened 2018-10-10
Don't pick random AZ on volume creation

bertinatto:
This partly reverts 760663c to prevent issue #52. This is not meant to
be the definite solution; it just buys us some time while keeping the
driver functional.

/kind bug
/assign @leakingtapan

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -20,8 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/rand"
-	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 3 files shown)


pkg/cloud/cloud.go
```diff
@@ -464,17 +455,3 @@ func (c *cloud) getInstance(ctx context.Context, nodeID string) (*ec2.Instance,
...
-	if err != nil {
-		return "", err
-	}
-
-	var zones []string
-	for _, zone := range output.AvailabilityZones {
-		zones = append(zones, *zone.ZoneName)
-	}
-
-	return zones[rand.Int()%len(zones)], nil
```
leakingtapan: nit: could you remove the rand seed in func init() since rand is not needed?

leakingtapan:
/lgtm
Thanks for the quick fix. one small comment.

PR #58, opened 2018-10-12
Wait for attachment state after attach/detach

### LINKED ISSUE #38: Implement wait for volume attach/detach
As noted [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L263)

This is to ensure `AttachVolume` is fully completed before proceeding to mount. 

We will also need to do similar thing for `DetachVolume`, `CreateVolume` and `DeleteVolume`

bertinatto:
Also, upsdate e2e test.

Fixes #38.

/kind feature
/assign @leakingtapan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,8 @@ test-sanity:
 
 .PHONY: test-e2e
 test-e2e:
-	go test -v ./tests/e2e/...
+	go test -c ./tests/e2e/... -o bin/e2e.test && \
+	sudo -E bin/e2e.test -ginkgo.v
 
 .PHONY: image
 image:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -30,7 +30,7 @@ test-sanity:
 
 .PHONY: test-e2e
 test-e2e:
-	go test -v ./tests/e2e/...
+	ginkgo -v ./tests/e2e/...
```
leakingtapan: Im wondering why need to replace `go test` with `ginkgo`?

bertinatto: I need to enable Ginkgo's verbosity *and* I also need to run the test with admin privileges due to the `mount` command. For some reason, executing `go test -ginkgo.v` with `sudo` didn't work for me (I think it's related to `go modules`), so I called `ginkgo` directly.

In any case, I improved these bits by compiling the test and then calling it with `sudo`.

PR #61, opened 2018-10-14
Add missing capability and clusterrole permission to enable topology awareness scheduling

leakingtapan:
Did manual testing on 1.12 and able to verify that volume and pod are co-scheduled in us-east-1c while provisioner is in us-east-1a.

### DIFF
--- a/deploy/kubernetes/v1.12+/node.yaml
+++ b/deploy/kubernetes/v1.12+/node.yaml
@@ -27,6 +27,9 @@ rules:
   - apiGroups: ["storage.k8s.io"]
     resources: ["volumeattachments"]
     verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["csi.storage.k8s.io"]
+    resources: ["csinodeinfos"]
+    verbs: ["get", "list", "watch", "update"]
 
 ---
 
--- a/deploy/kubernetes/v1.12+/provisioner.yaml
+++ b/deploy/kubernetes/v1.12+/provisioner.yaml
@@ -16,6 +16,9 @@ rules:
     resources: ["secrets"]
     verbs: ["get", "list"]
   - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
     resources: ["persistentvolumes"]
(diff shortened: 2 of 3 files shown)


bertinatto:
Have you had the chance to test this with a single node only?

I get this error:

```
Warning  FailedScheduling  1s (x3 over 5s)   default-scheduler  0/1 nodes are available: 1 node(s) had volume node affinity conflict.
```

I added `volumeBindingMode: WaitForFirstConsumer` to the storage class and started getting this error:

```
  Warning  FailedMount             6s (x25 over 8s)  kubelet, ip-172-18-15-56.ec2.internal  MountVolume.NodeAffinity check failed for volume "pvc-679f9e80-d084-11e8-8509-0ecfa83a9ab2" : No matching NodeSelectorTerms
```

leakingtapan:
@bertinatto I just tried single node and couldn't reproduce the error. Could you share more info about how do you run the dirver? like your pod, sc manifest file. And for single node I assume you are running with only one worker node (not including master node), right?

bertinatto:
I used the manifest files from `deploy/kubernetes/v1.12+`. The only change I made was in `secrets.yaml`.

> And for single node I assume you are running with only one worker node (not including master node), right?

No, by single node I meant all-in-one (master + worker in the same node).

I believe I figure out what the problem is. I created a k8s PR to fix it: https://github.com/kubernetes/kubernetes/pull/69870

bertinatto:
As described in the PR above, a gate should be enabled. Also, we need to create the CR manually.

We should probably update the documentation as part of this PR. What do you think?

I'm also wondering if the driver would still work with v1.11 (after applying this patch). Have you had the chance to try it in v1.11?

leakingtapan:
> We should probably update the documentation as part of this PR. What do you think?

I created a PR to update the Setup doc: kubernetes-csi/docs#65 and @verult will consolidate my PR since he has a similar PR kubernetes-csi/docs#64

> I'm also wondering if the driver would still work with v1.11 (after applying this patch). Have you had the chance to try it with v1.11?

Topology is a v1.12+ feature. So it won't work on version <= v1.11 anyway. 

I will send our another PR to keep manifest update after this is merged, external-provisioner v0.4.1 is cut and I test it again

PR #68, opened 2018-10-17
Remove hard-coded region in e2e test

### LINKED ISSUE #66: [Bug] e2e test failed when running in a region other than us-east-1
/kind bug

The e2e test ec2 client has a [hard coded](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/tests/e2e/setup_test.go#L38) region, which causes e2e test to fail when it runs in a region other than us-east-1

bertinatto:
Fixes #66.

/king bug
/assign @leakingtapan

### DIFF
--- a/tests/e2e/setup_test.go
+++ b/tests/e2e/setup_test.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


bertinatto:
@leakingtapan, I only have access to `us-east-1`, could you check if this works in another region?

leakingtapan:
tested on us-west-2 with no issue.

/lgtm

PR #69, opened 2018-10-17
Switch to Alpine 3.8 as base image

### LINKED ISSUE #59: Switch to use AmazonLinux for driver container base image
https://hub.docker.com/_/amazonlinux/

bertinatto:
Fixes #59.

Image size decreases considerably:

```
quay.io/bertinatto/ebs-csi-driver           testing            b73128d25c5d   2 minutes ago    30.2MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,8 @@ WORKDIR /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver
 ADD . .
 RUN CGO_ENABLED=0 GOOS=linux go build -o bin/aws-ebs-csi-driver ./cmd/aws-ebs-csi-driver
 
-FROM registry.fedoraproject.org/fedora-minimal
+FROM alpine:3.8
+RUN apk add --no-cache ca-certificates e2fsprogs
 COPY --from=builder /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/bin/aws-ebs-csi-driver /bin/aws-ebs-csi-driver
-RUN microdnf install -y e2fsprogs && microdnf clean all
 
 ENTRYPOINT ["/bin/aws-ebs-csi-driver"]


bertinatto:
@leakingtapan, what do you think about this?

I don't have any problems using any other base image, like Amazon Linux. However, Alpine seemed to be the lighter alternative.

Dockerfile
```diff
@@ -17,8 +17,8 @@ WORKDIR /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver
 ADD . .
 RUN CGO_ENABLED=0 GOOS=linux go build -o bin/aws-ebs-csi-driver ./cmd/aws-ebs-csi-driver
 
-FROM registry.fedoraproject.org/fedora-minimal
+FROM alpine:3.8
 COPY --from=builder /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/bin/aws-ebs-csi-driver /bin/aws-ebs-csi-driver
-RUN microdnf install -y e2fsprogs && microdnf clean all
+RUN apk add --no-cache ca-certificates e2fsprogs
```
leakingtapan: Just realize one existing issue, could you switch `RUN apk add...` with `COPY` so that docker build doesn't generate new mid layer each time we make changes to driver?

leakingtapan: I mean switching the line order to be like:
```
RUN apk add --no-cache ca-certificates e2fsprogs
COPY --from=builder /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/bin/aws-ebs-csi-driver /bin/aws-ebs-csi-driver
```

Because each line in docker file generates a new layer of docker image, we could avoid keep generating the `apk add` layer if it is before `COPY` line

bertinatto: Gotcha... Done

PR #77, opened 2018-10-23
Return early when detaching an already detached volume

leakingtapan:
TODOs:
 - [x] ~Fix unit test~

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -327,6 +327,7 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
 
 	if !device.IsAlreadyAssigned {
 		glog.Warningf("DetachDisk called on non-attached volume: %s", volumeID)
+		return nil
 	}
 
 	request := &ec2.DetachVolumeInput{
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -250,22 +250,31 @@ func TestAttachDisk(t *testing.T) {
 
 func TestDetachDisk(t *testing.T) {
 	testCases := []struct {
-		name     string
-		volumeID string
-		nodeID   string
-		expErr   error
+		name             string
+		volumeID         string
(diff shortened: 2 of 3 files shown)


pkg/cloud/cloud.go
```diff
@@ -311,6 +311,7 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
 
 	if !device.IsAlreadyAssigned {
 		glog.Warningf("DetachDisk called on non-attached volume: %s", volumeID)
+		return nil
```
bertinatto: The in-tree implementation doesn't return early, and I think there's a reason for that: the `device` object isn't really synchronized with the actual state of the world.

However, since `GetDevice()` loads all devices in use by the instance, I don't think there would be a problem here. 

I think the safest option would be to call `DetachVolumeWithContext` and wait for the volume state. However, I don't think this patch is incorrect, so I'm OK with merging it.

We might also want to take the chance and `Release()` the device only if

leakingtapan: Ah. Let me take some time to read through the in-tree implementation first and understand more.

leakingtapan:
The unit test is failed because, with this new change, `TestDetachDisk` will require a volume to be attached before calling `DetachDisk` since it relies on the internal state of device manage. Will need to refactor this test a bit in order to test this new behavior. Will research in-tree behavior first, then fix the test.

leakingtapan:
@bertinatto since `GetDevice()` reads all the devices attached to the node by the time of detaching. It should be ok to return early when Driver knows the volume ID invalid.

leakingtapan:
@gnufied, for your background, here is the equivalent logic in cloudprovider aws: https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L2107-L2108

We are wondering why it doesn't return immediately?

leakingtapan:
/cc @jsafrane 
Do you have any comment on this?

PR #78, opened 2018-10-24
Rename e2e test to be integ test

leakingtapan:


leakingtapan:
Prow hook is working!

tests/integ/integ_test.go
```diff
@@ -12,7 +12,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package e2e
+package integ
```
bertinatto: I'd be +1 to use `integration`, since it's not too long.

leakingtapan: sure.

PR #91, opened 2018-11-01
Swtich to use amazon linux 2 as base image

leakingtapan:


leakingtapan:
Switch to Alinux2 as alpine requires distribution approval from amazon

leakingtapan:
Ref #59

Issue #83, opened 2018-10-29
Add golint to verification


leakingtapan:
Should be added to `hack/verify-all` and similar to `gofmt`

leakingtapan:
There are several choices outside:
1. golint: https://github.com/golang/lint this is used by upstream k/k however, the error is pretty verbose and most of the complaints are documentation issues regarding to public interface. Since AWS EBS CSI driver implements the CSI Spec, there is not real reason to document the same interface
2. golangci-lint: https://github.com/golangci/golangci-lint used by [aws alb ingress controller](https://github.com/kubernetes-sigs/aws-alb-ingress-controller), it gives more useful error and has a set of configurable checker modules. One drawback is it is slow on megacheck.
3, go-tools: https://github.com/dominikh/go-tools

I decided to go with golangci-lint since it is more useful

PR #95, opened 2018-11-05
Add golint verification

leakingtapan:
Fixes: #83 

Ran integration test manual and verified that the test is passed

### DIFF
--- a/tests/integration/setup_test.go
+++ b/tests/integration/setup_test.go
@@ -36,7 +36,6 @@ import (
 
 const (
 	endpoint = "tcp://127.0.0.1:10000"
-	region   = "us-east-1"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


tests/integration/setup_test.go
```diff
@@ -58,7 +57,12 @@ var _ = BeforeSuite(func() {
 	ebs, err = cloud.NewCloud()
 	Expect(err).To(BeNil(), "Set up Cloud client failed with error")
 	drv = driver.NewDriver(ebs, nil, endpoint)
-	go drv.Run()
+	go func() {
+		err := drv.Run()
+		if err != nil {
+			fmt.Println("Driver exits: ", err)
```
bertinatto: Can we use something like: `log.Fatalf("Failed to start CSI driver: %v, err")`

leakingtapan: Since gomega `Expect` is used in integration test, I changed to use `Expect(err).To(BeNil())`. Let me know if you think `log` is needed

PR #121, opened 2018-11-17
Add CHANGELOG for v0.1.0 alpha release

leakingtapan:
Fixes: #79


leakingtapan:
Updated README and CHANGELOG by removing alpha in the version string. I think its clean enough if we use 0.1.0 as alpha release for kubernetes.

PR #122, opened 2018-11-21
Update to CSI 1.0

bertinatto:

/kind feature

This PR updates the driver to the new CSI 1.0 spec.


bertinatto:
/hold
I'll take the chance and remove the `vendor` dir instead of updating it.

bertinatto:
/hold cancel

Apparently we have to keep the `vendor` directory because the tests aren't using go 1.11 yet.

leakingtapan:
Is this PR tested against v1.13?

deploy/kubernetes/v1.12+/attacher.yaml
```diff
@@ -27,6 +27,9 @@ rules:
   - apiGroups: ["storage.k8s.io"]
     resources: ["volumeattachments"]
     verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["csi.storage.k8s.io"]
+    resources: ["csinodeinfos"]
+    verbs: ["get", "list", "watch"]
```
leakingtapan: Since I cannot to comment there, do we still need `resource: ["events"]` at line 20? It seems not needed per: https://github.com/kubernetes-csi/external-attacher/blob/9da8c6d20d58750ee33d61d0faf0946641f50770/deploy/kubernetes/rbac.yaml

deploy/kubernetes/v1.12+/attacher.yaml
```diff
@@ -78,7 +81,7 @@ spec:
             capabilities:
               add: ["SYS_ADMIN"]
             allowPrivilegeEscalation: true
-          image: quay.io/k8scsi/csi-attacher:v0.4.1
+          image: quay.io/k8scsi/csi-attacher:v1.0-canary
```
leakingtapan: Do you know what's the difference between v1.0.0 and v1.0-canary?

bertinatto: Yep, v1.0.0 has a bug where it reports the wrong version: https://github.com/kubernetes/kubernetes/issues/71378

There's not a v1.0.1 yet, but v1.0-canary is supposed to get all minor updates.

deploy/kubernetes/v1.12+/attacher.yaml
bertinatto: I'm not aware about events. I'll try to remove it and see if it still works.

bertinatto: Removed "events" and no problem happened. Updated.

bertinatto:
I just changed the base branch for this PR from `master` to `next`.

PR #126, opened 2018-11-28
Wait for volume to become available

### LINKED ISSUE #65: Implement wait for volume create/delete
Similar to #38

bertinatto:

/kind bug

This patch waits for the volume to become `available` when `CreateVolume`
is called. This prevents the CO from trying to attach the volume when
it's still being created.

Fixes #65.

/assign @leakingtapan


bertinatto:
/hold cancel

Actually, no. The real problem was that `gomock` wasn't generating the correct import paths.

bertinatto:
I just changed the base branch for this PR from `master` to `next`.

pkg/cloud/mocks/mock_ec2.go
```diff
@@ -1,144 +1,128 @@
-// Code generated by MockGen. DO NOT EDIT.
```
leakingtapan: Did we use a different version of MockGen so that the content are different from the previous generated code? Which MockGen are you using? Could you add some documentation about this in [docs](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/docs) folder

bertinatto: I think it was the same version as before, but apparently it had some incompatibilities with go modules. I had to update the imports of this file manually (not good).

I updated the deps to their last minor versions and now it works fine.

PR #127, opened 2018-11-28
Remove unsupported MULTI_NODE_READER_ONLY volume capability

### LINKED ISSUE #123: Is MULTI_NODE_READER_ONLY Volume Capability required?
According to [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md):
```
// Can be published as readonly at multiple nodes simultaneously.
MULTI_NODE_READER_ONLY = 3;
```
MULTI_NODE_READER_ONLY should be used for volume that can be published as readonly on multiple nodes, and this is not supported by EBS yet but got [set by the driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/driver.go#L71)

Should this be removed?

/cc @bertinatto

bertinatto:

/kind bug
/assign leakingtapan

Fixes #123.

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -67,9 +67,6 @@ func NewDriver(cloud cloud.Cloud, mounter *mount.SafeFormatAndMount, endpoint st
 			{
 				Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
 			},
-			{
-				Mode: csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY,
-			},
 		},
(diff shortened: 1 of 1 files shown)


leakingtapan:
Did you get a chance to test if still works after removing the capability?

leakingtapan:
Its weird that removing the line causes coverage drops. This might blocks code from merged. Feel free to merge if auto merge doesn't happen

bertinatto:
Yeah, it feels awkward. This has happened in other PRs as well.

I understand that removing code means less code being exercised during tests. At the same time the coverage drops. I'd expect the report tool would be smarter in this sense.

Issue #25, opened 2018-09-28
Implement Snapshot support


leakingtapan:
Driver implementation should support following operations:
 - [x] CreateSnapshot
 - [x] DeleteSnapshot
 - [ ] ListSnapshots
 - [x] CreateVolume from snapshot

References:
* https://kubernetes.io/docs/concepts/storage/volume-snapshots/
* https://kubernetes.io/docs/concepts/storage/volume-snapshot-classes/
* https://kubernetes-csi.github.io/docs/Implement-Snapshot-Feature.html

bertinatto:
CC @tsmetana

leakingtapan:
Punt the feature to beta release

tsmetana:
First version (does not implement snapshot listing yet): https://github.com/tsmetana/aws-ebs-csi-driver/tree/snapshots-updated

It's based on PR#122 (CSI-1.0) so I will wait with my PR.

bertinatto:
@tsmetana: we just merged my PR into `next` branch. If you want to proceed with you PR you can point it against `next` as well.

leakingtapan:
/assign @dkoshkin
Dimitri will help debugging issues related to snapshot

tsmetana:
OK. I'm not sure which is the right place to discuss this... What is happening with the snapshots: in the waitForSnapshotCreate there is an truncated exponential backoff with a condition function that basically polls for the snapshot unitl it's "completed" or an error occurs. The polling loop uses ec2.DescribeSnapshotsWithContext(ctx, ...). After few iterations this call fails with GRPC error ("DeadlineExceeded" iirc). 

Looks like the issue is that the CreateSnapshot CSI call should be blocking (by specification) but the GRPC connection times out prematurely -- it's unknown how long time a snapshot creation might take. The controllers (external snapshotter in our case) should be able to handle that -- for example external attacher distinguishes the fatal and transient errors like this: https://github.com/kubernetes-csi/external-attacher/blob/master/pkg/connection/connection.go#L256

I guess similar mechanism needs to be implemented also in the external snapshotter, otherwise this patch will not work...

tsmetana:
The issues I mentioned in the previous comment were worked around in the patch that's been recently merged.

leakingtapan:
@tsmetana just to confirm that is `ListSnapshot` is used by k8s yet, so it is not implemented?

tsmetana:
[...]

PR #131, opened 2018-11-30
Add snapshots support

tsmetana:
This is the first version of volume snapshotting support.

It implements:
* CREATE_DELETE_SNAPSHOT capability support (i.e. no snapshot listing so far). 
* create volume from snapshot

I did only basic testing of the feature so far, but it looks to be working fine. I will add some examples and documentation of the feature.

Ref Issue #25

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -73,6 +74,8 @@ const (
 const (
 	// VolumeNameTagKey is the key value that refers to the volume's name.
 	VolumeNameTagKey = "CSIVolumeName"
+	// SnapshotNameTagKey is the key value that refers to the snapshot's name.
+	SnapshotNameTagKey = "CSIVolumeSnapshotName"
 )
 
 var (
@@ -109,7 +112,21 @@ type DiskOptions struct {
 	Encrypted        bool
 	// KmsKeyID represents a fully qualified resource name to the key to use for encryption.
 	// example: arn:aws:kms:us-east-1:012345678910:key/abcd1234-a123-456a-a12b-a123b4cd56ef
-	KmsKeyID string
+	KmsKeyID   string
+	SnapshotID string
+}
+
+// Snapshot represents an EBS volume snapshot
+type Snapshot struct {
+	SnapshotID     string
+	SourceVolumeID string
+	Size           int64
+	CreationTime   time.Time
+}
+
+// SnapshotOptions represents parameters to create an EBS volume
+type SnapshotOptions struct {
+	Tags map[string]string
 }
 
 // EC2 abstracts aws.EC2 to facilitate its mocking.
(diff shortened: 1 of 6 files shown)


pkg/cloud/cloud.go
```diff
@@ -396,6 +427,81 @@ func (c *cloud) IsExistInstance(ctx context.Context, nodeID string) bool {
...
+		Tags:         tags,
+	}
+	request := &ec2.CreateSnapshotInput{
+		VolumeId:          aws.String(volumeID),
+		DryRun:            aws.Bool(false),
+		TagSpecifications: []*ec2.TagSpecification{&tagSpec},
+		Description:       aws.String(descriptions),
+	}
+
+	res, err := c.ec2.CreateSnapshotWithContext(ctx, request)
```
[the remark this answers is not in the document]
tsmetana: You're right... I've never seen an AWS disk where the snapshot would take that long. The change in the external controller that simplified the phase handling (the actual snapshotting is usually quick but then some backends take a long time to move the snapshot however the original disk is OK to be used). I'll figure this out.

pkg/cloud/cloud.go
```diff
@@ -396,6 +427,81 @@ func (c *cloud) IsExistInstance(ctx context.Context, nodeID string) bool {
...
+
+	ec2snapshot, err := c.getSnapshot(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.ec2SnapshotResponseToStruct(ec2snapshot), nil
+}
+
+// Helper method conerting EC2 snapshot type to the internal struct
```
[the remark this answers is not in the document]
tsmetana: OK. Handling the nil case is definitely safer.

suphanatnack:
มใใ

tsmetana:
Rebased with some changes to the backoff logic.

pkg/driver/controller.go
```diff
@@ -124,6 +126,20 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		Encrypted:        isEncrypted,
 		KmsKeyID:         kmsKeyId,
 	}
+
+	// Shall we restore a snapshot?
```
leakingtapan: Do we still need this comment? If the snapshot ID is provide, the create volume API call will create the volume from snapshot, right?

pkg/driver/controller.go
```diff
@@ -290,11 +306,57 @@ func (d *Driver) isValidVolumeCapabilities(volCaps []*csi.VolumeCapability) bool
...
+	opts := &cloud.SnapshotOptions{
+		Tags: map[string]string{cloud.SnapshotNameTagKey: snapshotName},
+	}
+	snapshot, err = d.cloud.CreateSnapshot(ctx, volumeID, opts)
+
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Could not create snapshot %q: %v", snapshotName, err)
+	}
+	csiSnapshot, err := newCreateSnapshotResponse(snapshot)
+	return csiSnapshot, err
```
leakingtapan: NIT:
```suggestion
	return newCreateSnapshotResponse(snapshot)
```

pkg/cloud/cloud.go
```diff
@@ -549,3 +680,56 @@ func isAWSErrorVolumeNotFound(err error) bool {
...
+				return true, fmt.Errorf("unexpected State of newly created AWS EBS snapshot %v: %q", snapshotID, *snapshot.State)
+			}
+		}
+		return false, nil
+	}
+
+	// Truncated exponential backoff: if the exponential backoff times-out, just keep polling using the longest interval
+	err := wait.ExponentialBackoff(backoff, conditionFunc)
+	if err == wait.ErrWaitTimeout {
+		timeout := time.Duration(backoff.Duration.Seconds() * math.Pow(backoff.Factor, float64(backoff.Steps)))
```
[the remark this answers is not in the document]
tsmetana: Kuberentes API is supposed to be "intent based" so as long as there exists a VolumeSnapshot object the API should try to create it (no timeout). The idea here is not to extend the wait period too much so in case the operation succeeds/fail we are able to inform the user. Since it's unknown how long may the operation take I only used the exponential backoff for the first few poll iterations and then stopped to prolong the interval and make it constant.

pkg/driver/controller.go
tsmetana: Yes. Will remove it with reabse.

pkg/driver/controller.go
tsmetana: OK. Will change this.

leakingtapan:
Thx @tsmetana for rebasing the change. We are merging this to snapshot branch to bake it a bit more and fix some timeout issue. After that, we will merge the final change into master

@dkoshkin

PR #135, opened 2018-12-03
Refactor the code

leakingtapan:
1. Split `NewDriver` into two methods `NewDriver` and `NewMockDriver`.
So that one is only takes in endpoint and serves as production uses and
the other create driver with mock dependencies
2. Move controller capability, node capability and volume capability out
of driver struct since they are constant that should never change for a
specific version of driver

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -28,6 +28,23 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var (
+	// volumeCaps represents how the volume could be accessed.
+	// It is SINGLE_NODE_WRITER since EBS volume could only be
+	// attached to a single node at any given time.
+	volumeCaps = []csi.VolumeCapability_AccessMode{
+		{
+			Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+		},
+	}
+
+	// controllerCaps represents the capability of controller service
+	controllerCaps = []csi.ControllerServiceCapability_RPC_Type{
+		csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
+		csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME,
+	}
+)
+
 func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	glog.V(4).Infof("CreateVolume: called with args %+v", *req)
 	volName := req.GetName()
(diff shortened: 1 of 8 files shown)


pkg/driver/driver.go
```diff
@@ -46,35 +46,33 @@ type Driver struct {
...
+}
+
+// NewMockDriver creates a new mock driver used for testing
+func NewMockDriver(endpoint string) *Driver {
+	cloud := cloud.NewFakeCloudProvider()
+	return &Driver{
+		endpoint: endpoint,
+		nodeID:   cloud.GetMetadata().GetInstanceID(),
+		cloud:    cloud,
+		mounter:  NewFakeMounter(),
```
bertinatto: Can we move this to `fakes.go`? Also, we might want to use `NewFakeDriver` to keep consistency with other functions.

pkg/driver/controller.go
```diff
@@ -28,6 +28,21 @@ import (
...
+var (
+	// EBS volume capability. It could only be read/write on a single node at any given time.
+	volumeCaps = []csi.VolumeCapability_AccessMode{
+		{
+			Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+		},
+	}
+
+	// controller capability
+	controllerCaps = []csi.ControllerServiceCapability_RPC_Type{
```
bertinatto: I'm not sure if making these fields global is a better practice. It'd make it difficult to create a driver with custom capabilities for testing, for example.

Another option would be to create functions to set up and return the caps, and use these function in `New*Driver()`.

If we want to keep it like this, we may want to name these vars like `defaultVolume...` and use these names in the comments (e.g., `defaultVolumeCapabilities represents the EBS volume...`.

pkg/driver/controller_test.go
```diff
@@ -298,7 +303,12 @@ func TestDeleteVolume(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Logf("Test case: %s", tc.name)
-		awsDriver := NewDriver(cloud.NewFakeCloudProvider(), NewFakeMounter(), "")
+		awsDriver := &Driver{
+			endpoint: "",
+			nodeID:   "",
+			cloud:    cloud.NewFakeCloudProvider(),
+			mounter:  NewFakeMounter(),
+		}
```
bertinatto: Shouldn't use `NewMockDriver` here?

pkg/driver/driver.go
leakingtapan: good catch. done

pkg/driver/controller_test.go
leakingtapan: yep. updated.

pkg/driver/controller.go
leakingtapan: > I'm not sure if making these fields global is a better practice. It'd make it difficult to create a driver with custom capabilities for testing, for example.

I feel we will never create a CSI EBS driver with a different capability eg, volume capability of `MULTI_NODE_MULTI_WRITER` since that is limited by EBS; or we will create a driver whose controller service capability is configurable, eg `CREATE_DELETE_VOLUME` is configurable since we always want dynamic provisioning. 

I kept variable name but updated the comments of those variable though.

leakingtapan:
@bertinatto yes, its going to be merged into master

pkg/driver/controller.go
bertinatto: I don't feel like moving those vars to a global scope would make the code better, even though they might not be ever changed.

I agree that initializing them in `NewDriver` feels messy, but I'd rather create functions that return the default values and use those functions in `NewDriver` (if I'm not mistaken GCP driver does something like that).

However, I don't want to block this PR from being merged, so feel free to merge it if you think it's OK as is.

bertinatto:
@leakingtapan: ah, OK. I just noticed that `next` is *your* branch, not the target branch.

Issue #21, opened 2018-09-26
Add version command to ebs-csi-driver cmd


leakingtapan:
`./aws-ebs-csi-driver --version` should return version with following information in Json formation including:
* git commit sha
* driver version
* go compiler version
* build date
* Platform OS and Arch

PR #136, opened 2018-12-04
Add version flag

dkoshkin:
Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/21

This PR adds `--version` flag at the top level command.
Running `aws-ebs-csi-driver --version` will pretty print the JSON, with the same format as `kubectl version`.

This PR does not add support for printing the info in any other format. 

**What testing is done?** 
Locally built the image and tested:
```
➜  aws-ebs-csi-driver git:(version) ✗ docker run -it amazon/aws-ebs-csi-driver:0.1.0-alpha --version
{
  "driverVersion": "0.1.0-alpha",
  "gitCommit": "c7b3689df8a5cdf212ca80f9f0332ac3d518eefc",
  "buildDate": "2018-12-04T18:45:21Z",
  "goVersion": "go1.11.1",
  "compiler": "gc",
  "platform": "linux/amd64"
}
```
I am not aware of a way to unit test when using `-ldflags`, if anyone has some pointers would be glad to add a unit test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+PKG=github.com/kubernetes-sigs/aws-ebs-csi-driver
 IMAGE=amazon/aws-ebs-csi-driver
 VERSION=0.1.0-alpha
+GIT_COMMIT?=$(shell git rev-parse HEAD)
+BUILD_DATE?=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
+LDFLAGS?="-X ${PKG}/pkg/driver.driverVersion=${VERSION} -X ${PKG}/pkg/driver.gitCommit=${GIT_COMMIT} -X ${PKG}/pkg/driver.buildDate=${BUILD_DATE}"
 
 .PHONY: aws-ebs-csi-driver
 aws-ebs-csi-driver:
 	mkdir -p bin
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -14,11 +14,14 @@
 
 IMAGE=amazon/aws-ebs-csi-driver
 VERSION=0.1.0-alpha
+GIT_COMMIT?=$(shell git rev-parse HEAD)
+BUILD_DATE?=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
+LDFLAGS?="-X github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.driverVersion=${VERSION} -X github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.gitCommit=${GIT_COMMIT} -X github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.buildDate=${BUILD_DATE}"
```
leakingtapan: Is it possible to create a variable for `github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver` and refer it from within `LDFLAGS`? This line is long and, when people makes changes, it could be error-prune if we have it in three places

pkg/driver/version.go
```diff
@@ -0,0 +1,43 @@
+package driver
```
leakingtapan: Could you add a COPY RIGHT HEADER here?

leakingtapan:
@dkoshkin thanks for the PR. For the unit test, could you create a `version_test.go` and just test the struct and methods within the file to get us some coverage?

Issue #137, opened 2018-12-04
Incompatibale example manifest files on master branch


leakingtapan:
/kind bug

**What happened?**
I deployed ./deploy/kubernetes/v1.12+ driver on my v1.12 cluster, csi-driver-registrar fails to start

Due to follow error:
```
Events:
  Type     Reason       Age                  From                                     Message
  ----     ------       ----                 ----                                     -------
  Normal   Scheduled    3m10s                default-scheduler                        Successfully assigned default/csi-node-zt2v2 to ip-172-20-127-156.ec2.internal
  Warning  FailedMount  67s                  kubelet, ip-172-20-127-156.ec2.internal  Unable to mount volumes for pod "csi-node-zt2v2_default(dbced501-f7f9-11e8-81f1-0a75e9a76798)": timeout expired waiting for
volumes to attach or mount for pod "default"/"csi-node-zt2v2". list of unmounted volumes=[registration-dir]. list of unattached volumes=[kubelet-dir plugin-dir registration-dir device-dir csi-node-sa-token-blrn
x]
  Warning  FailedMount  62s (x9 over 3m10s)  kubelet, ip-172-20-127-156.ec2.internal  MountVolume.SetUp failed for volume "registration-dir" : hostPath type check failed: /var/lib/kubelet/plugins_registry/ is not a directory
```

**What you expected to happen?**
v1.12 manifest should still work with v1.12 cluster

leakingtapan:
This is caused by back ward incompatible changes made on CSI with regard to k/k v1.13. So that v1.13 manifests can not run on v1.12 cluster.

PR #138, opened 2018-12-04
Partially revert PR #122 and add latest manifest files

leakingtapan:

Fixes bugs: #137

I use latest instead of v1.13 to highlight that is the latest version
under development which might have issues. And any other versioned
manifest files should not break.

**What testing is done?** 
Tested on my v1.12 cluster and deployed sample app, verified it works


leakingtapan:
Ref: #122

PR #149, opened 2018-12-15
Update docs

d-nishi:
Update README

Update README for better usability of the driver

**What testing is done?** 
Tested using the driver with KOPS and k8s v1.12


docs/README.md
```diff
@@ -61,38 +52,49 @@ There are several optional parameters that could be passed into `CreateVolumeReq
...
 | "kmsKeyId"        |                  |          | The full ARN of the key to use when encrypting the volume. When not specified, the default KMS key is used |
 
-### Topology
-`topology.ebs.csi.aws.com/zone` is the only topology key that represents the availability zone of which a volume is accessible.
+## Prerequisites
+### Kubernetes
+* Kubernetes 1.12+ is required. Although this driver should work with any other container orchestration system that implements the CSI specification, so far it has only been tested in Kubernetes.
 
-To enable topology support on kuberetes, make sure `CSINodeInfo` and `CSIDriverRegistry` feature flags are enabled on both kubelet and kube-apiserver and `CSINodeInfo` CRD is installed on the cluster following [Enabling CSINodeInfo](https://kubernetes-csi.github.io/docs/Setup.html#enabling-csinodeinfo).
+* Enable the flag `--allow-privileged=true` in the manifest entries of kubelet and kube-apiserver.
```
leakingtapan: For the manifest, ru referring to kops cluster manifest? I think we don't need to mention manifest since not every one is using kops. Probably we can say something like:
```
Enable the flag `--allow-privileged=true` for kubelet and kube-apiserver.
```

docs/README.md
```diff
@@ -61,38 +52,49 @@ There are several optional parameters that could be passed into `CreateVolumeReq
...
 
-And *external-provisioner* must have the togology feature gate enabled with `--feature-gates=CSINodeInfo=true`
+* Add `--feature-gates=CSINodeInfo=true,CSIDriverRegistry=true` in the manifest entries of kubelet and kube-apiserver. This is required to enable topology support of EBS volumes in Kubernetes.
 
-## Installation
-### Kubernetes
-Under the directory [deploy/kubernetes](../deploy/kubernetes), there are a few manifest files that are needed to deploy the CSI driver along with sidecar containers. If you are using Kubernetes v1.12, use the manifest files under [deploy/kubernetes/v1.12](../deploy/kubernetes/v1.12); for kubernetes v1.10 and v1.11, use the files under [deploy/kubernetes/v1.[10,11]](../deploy/kubernetes/v1.[10,11]).
+* Install the `CSINodeInfo` CRD on the cluster using the instructions provided here: [Enabling CSINodeInfo](https://kubernetes-csi.github.io/docs/Setup.html#enabling-csinodeinfo).
+
+* Ensure that the feature gate is enabled as `--feature-gates=CSINodeInfo=true` when using *external-provisioner*
```
leakingtapan: Just notice a typo of the previous doc, we need to fix the feature gate flag:
```
* Ensure that the feature gate is enabled as `--feature-gates=Topology=true` when using *external-provisioner*
```
Ref: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/latest/provisioner.yaml#L101

docs/README.md
```diff
@@ -61,38 +52,49 @@ There are several optional parameters that could be passed into `CreateVolumeReq
...
 
-In this example we'll use Kubernetes v1.12. First of all, edit the `deploy/kubernetes/v1.12/secrets.yaml` file and add AWS credentials of the IAM user. It's a best practice to only grant required permission to the driver. A sample IAM policy can be found in [example-iam-policy.json](example-iam-policy.json).
+* Please refer to [kubernetes CSI docs](https://kubernetes-csi.github.io/docs/Home.html) for general CSI driver setup instructions on kubernetes.
+
+## Setup
+### Kubernetes
+* Use the manifest files under the directory [deploy/kubernetes](../deploy/kubernetes), needed to deploy the CSI driver and sidecar containers. If you are using Kubernetes v1.12, use the manifest files under [deploy/kubernetes/v1.12](../deploy/kubernetes/v1.12); for kubernetes v1.10 and v1.11, use the files under [deploy/kubernetes/v1.[10,11]](../deploy/kubernetes/v1.[10,11]).
```
leakingtapan: After read it again, I feel its better to use numeric list instead of bullet point here since it requires user to follow each step listed here in order

docs/README.md
d-nishi: Manifest is not a word specific to kops. Manifests is a generic term valid for all Kubernetes resources / objects whether the cluster is set by kops or some other tool.

docs/README.md
d-nishi: I see. Sure will update this.

docs/README.md
d-nishi: It really does not matter much but happy to update as well.

docs/README.md
mitchellhuang: This flag is now deprecated no?

corbanvilla: correct, I just ran into that as well

PR #155, opened 2018-12-21
Make requests idempotent

sreis:

Bug fix.

The CSI spec requires most operations to be idempotent. 

Currently, some of the requests rely on AWS API to achieve this. An example is CreateVolume. If by any chance AWS API request latency increases the CO might issue the same request that will fail. This leads the CO to send more CreateVolume requests and the possibility of creating multiple disks for the same request. (NOTE: This bus (multiple disks being created) is fixed in another PR but the underlying issue is still not fixed).

Other requests might take more time to complete. An example is NodeStageVolume. If the volume was just created it will need to be formatted for the mount to succeed. If the format takes too much time the CO can send another NodeStageVolume request that will fail because the previous one is still in progress.

This PR introduces a new class to handle inflight requests. It relies on the fact that the Golang protobuf implements the Stringer interface for all structures that it generates. We can use that string to compare requests and keep a list of the ones that are inflight.

**What testing is done?** 

Added unit tests for the new internal data structure to manage in flight requests.

### TODO
- Make more requests idempotent

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -23,6 +23,7 @@ import (
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/glog"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/internal"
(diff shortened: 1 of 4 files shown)


sreis:
Sorry for the noise. I'll reopen it when it's ready.

PR #156, opened 2018-12-21
Fix integration Test

leakingtapan:
Fix issues detect during enabling integration test

Ref #153

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -521,7 +521,7 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
 			case "available":
 				return true, nil
 			case "creating":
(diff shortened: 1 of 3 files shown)


gyuho:
lgtm thanks for fix!

pkg/cloud/cloud.go
```diff
@@ -521,7 +521,7 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
 			case "available":
 				return true, nil
 			case "creating":
-				return false, fmt.Errorf("volume %s is still being created", volumeID)
+				return false, nil
```
bertinatto: Why do you need to remove the error? Since this is in a `wait.Poll` loop, it's useful to have the error message in case the condition times out. Otherwise you only get a time out error.

PR #157, opened 2018-12-21
Fix an error when creating a volume with topology feature gate

dkoshkin:
Bug fix

* In the waitForVolume case "creating" incorrectly returned an error. This would then return from `cloud.CreateDisk`, causing a possible race condition. Where the function driver.CreateVolume could create another disk with the same `com.amazon.aws.csi.volume` tag. Eventually erroring with `ErrMultiDisks`. This also exposed a bug where the `AvailabilityZone` and `fsType` were not properly set. Which only surfaced when setting `--feature-gates=Topology=true`.
* Added `volumeBindingMode: WaitForFirstConsumer` in the StorageClass because since the flag `--feature-gates=Topology=true` is set and it is required for that feature.

**What testing is done?** 
Manual testing and new unit tests

Error `volume vol-0fff33f4c0834292a is still being created` returned
```
$ kubectl describe pvc

Events:
  Type       Reason                 Age                From                                                                    Message
  ----       ------                 ----               ----                                                                    -------
  Warning    ProvisioningFailed     39s                persistentvolume-controller                                             storageclass.storage.k8s.io "slow" not found
  Warning    ProvisioningFailed     29s                ebs.csi.aws.com_csi-provisioner-0_87553ac8-0499-11e9-a00f-06f965927db1  failed to provision volume with StorageClass "slow": rpc error: code = Internal desc = Could not create volume "pvc-0ba471eb-0482-11e9-b7b3-22df1a95de60": failed to get an available volume in EC2: volume vol-0fff33f4c0834292a is still being created
  Normal     ExternalProvisioning   18s (x3 over 33s)  persistentvolume-controller                                             waiting for a volume to be created, either by external provisioner "ebs.csi.aws.com" or manually created by system administrator
  Normal     Provisioning           14s (x2 over 33s)  ebs.csi.aws.com_csi-provisioner-0_87553ac8-0499-11e9-a00f-06f965927db1  External provisioner is provisioning volume for claim "default/claim1"
  Normal     ProvisioningSucceeded  14s                ebs.csi.aws.com_csi-provisioner-0_87553ac8-0499-11e9-a00f-06f965927db1  Successfully provisioned volumepvc-0ba471eb-0482-11e9-b7b3-22df1a95de60
Mounted By:  app-5f7dccccc6-wzn9g
```
Then when the PV is created in Kubernetes, `values` field was empty for `matchExpressions`:
```
$ kubectl get pv -o yaml

    nodeAffinity:
      required:
        nodeSelectorTerms:
        - matchExpressions:
          - key: topology.ebs.csi.aws.com/zone
            operator: In
            values:
            - ""
```
 And the pod would not properly run:
```
$ kubectl describe pods

Events:
  Type     Reason                  Age                 From                                                Message
  ----     ------                  ----                ----                                                -------
  Warning  FailedScheduling        54s (x2 over 54s)   default-scheduler                                   pod has unbound immediate PersistentVolumeClaims
  Normal   Scheduled               30s                 default-scheduler                                   Successfully assigned default/app-5f7dccccc6-t9z8s to kube-node-1-kubelet.devkubernetes01.mesos
  Normal   SuccessfulAttachVolume  26s                 attachdetach-controller                             AttachVolume.Attach succeeded for volume "pvc-0ba471eb-0482-11e9-b7b3-22df1a95de60"
  Warning  FailedMount             20s (x25 over 22s)  kubelet, kube-node-1-kubelet.devkubernetes01.mesos  MountVolume.NodeAffinity check failed for volume "pvc-0ba471eb-0482-11e9-b7b3-22df1a95de60" : No matching NodeSelectorTerms
```

This only really surfaced itself when setting `--feature-gates=Topology=true` since the pod needs to have the proper `nodeAffinity` set in PV spec.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -86,14 +86,17 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		}
 	}
 
+	volumeParams := req.GetParameters()
+	fsType := volumeParams["fsType"]
+
 	// volume exists already
 	if disk != nil {
+		disk.FsType = fsType
 		return newCreateVolumeResponse(disk), nil
 	}
 
 	// create a new volume
 	zone := pickAvailabilityZone(req.GetAccessibilityRequirements())
-	volumeParams := req.GetParameters()
 	volumeType := volumeParams["type"]
 	iopsPerGB := 0
 	if volumeType == cloud.VolumeTypeIO1 {
@@ -125,7 +128,6 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Could not create volume %q: %v", volName, err)
 	}
-	fsType := volumeParams["fsType"]
 	disk.FsType = fsType
 	return newCreateVolumeResponse(disk), nil
 }
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -18,12 +18,18 @@ package driver
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
-	csi "github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+)
+
+const (
+	expZone   = "us-west-2b"
+	expFsType = "ext2"
 )
 
 func TestCreateVolume(t *testing.T) {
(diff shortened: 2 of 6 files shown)


pkg/driver/controller_test.go
```diff
@@ -215,6 +221,49 @@ func TestCreateVolume(t *testing.T) {
 				VolumeContext: map[string]string{"fsType": ""},
 			},
 		},
+		{
+			name: "success volume exists and contains VolumeContext and AccessibleTopology",
```
```suggestion
			name: "success when volume exists and contains VolumeContext and AccessibleTopology",
```

pkg/driver/controller.go
```diff
@@ -86,14 +86,17 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		}
 	}
 
+	volumeParams := req.GetParameters()
+	fsType := volumeParams["fsType"]
+
 	// volume exists already
 	if disk != nil {
+		disk.FsType = fsType
```
bertinatto: I'm wondering if it's necessary for us to return the FS type as an attribute. The reason is that we don't do any validation with it, we simple send it back to the CO.

For example, if the client sends in an invalid FS type, we return success. Later on it'll fail when staging, but I'm wondering if we should return an error right here. Another option would be to ignore it (other drivers seems to do that).

@leakingtapan, do you know if it's strictly necessary that we return this info back to the CO?

leakingtapan: We need to pass `fsType` back since it is needed by `NodeStageVolume` later [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L100). 


> Another option would be to ignore it (other drivers seems to do that).

In other driver, what's the behavior once a invalid fsType is ignored? Does it fall back to default option?

leakingtapan:
The change /lgtm

@dkoshkin lets take a look at @bertinatto 's comment before merge.

pkg/driver/controller.go
bertinatto: > We need to pass fsType back since it is needed by NodeStageVolume later here.

From what I understand from CreateVolume specs, that wouldn't be required. 

I took a quick look in GCP and it seems like they don't handle fsType as well. It'd be nice to test this.

It'd be nice to have this tested, but we can probably do that in another PR.

dkoshkin: > From what I understand from CreateVolume specs, that wouldn't be required.
It's not required in `CreateVolume` but then when the volume is formatted in `NodeStageVolume` the fsType is missing and it would just use the default right?

leakingtapan: @bertinatto it is not required to be CSI compliant, however, it is required to be EBS in-tree complicant since `fsType` is supported there. See [here](https://kubernetes.io/docs/concepts/storage/storage-classes/#aws-ebs). And this explains why it is not handled by GCE PD CSI driver since their in-tree doesn't have this. See [here](https://kubernetes.io/docs/concepts/storage/storage-classes/#gce-pd)

@dkoshkin Yep. When `fsType` is missing the default one will be used in node stage phase [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L102)

So from a EBS point of view, we would still need `fsType` passed into `NodeStageVolume` from `CreateVolume`

leakingtapan:
@dkoshkin you might need to rebase to pick up the fixes for integ test

Issue #158, opened 2018-12-25
Update driver manifest for driver-registrar


leakingtapan:
**Is your feature request related to a problem? Please describe.**
The CSI sidecar driver registrar is split into two side cars: cluster-driver-registrar and node-driver-registrar. EBS CSI [node manifest file](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/latest/node.yaml) should be updated in according to this

**References**
* cluster-driver-registrar: https://github.com/kubernetes-csi/cluster-driver-registrar
* node-driver-registrar: https://github.com/kubernetes-csi/node-driver-registrar

dkoshkin:
I have some code I've been testing to do this, you can assign to me if you like.

PR #171, opened 2019-01-07
CSI image version and deployment manifests updates 

dkoshkin:
Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/158

This PR replaces the deprecated https://github.com/kubernetes-csi/driver-registrar with:
- cluster-driver-registrar: https://github.com/kubernetes-csi/cluster-driver-registrar
- node-driver-registrar: https://github.com/kubernetes-csi/node-driver-registrar
---

* Because the `driver-registrar` is now replaced with 2 different images one `cluster-driver-registrar` and one `node-driver-registrar` it no longer makes sense to have both of those pods in the `csi-node` DS as `node-driver-registrar` is a cluster level component and should register the driver once, ie in a single pod.
To support this, this PR also merged the `attacher.yaml` and `provisioner.yaml` into a single `controller.yaml` `StatefulSet` (this is also consistent how other CSI drivers including DOs and GCPs does it)
I also went through all the RBAC policies and tested them to confirm there aren't any extra/missing ones.
* Replaced `quay.io/k8scsi/csi-attacher:v1.0-canary` and `external-provisioner` images with a stable latest release `v1.0.1` - having version predictability is always good - Interested on everyone else's thoughts on this.
* Set `priorityClassName` which requires the pods to run in `kube-system` namespace.
* I also removed directories for versions pre v1.13, Kubernetes `v1.13` uses CSI 1.0 and although this driver has a release https://github.com/kubernetes-sigs/aws-ebs-csi-driver/releases with support for CSI 0.3, that release has some bugs and users can still use that version for older k8s versions but they should be encouraged to use newer, more stable versions. - Interested on everyone else's thoughts on this.
* Added double quotes to the `secret.yaml` file, this allows users to rely on EC2 instance roles and still run `kubectl apply -f` without having to remove that file.

**What testing is done?** 
Locally ran e2e and sanity tests.


deploy/kubernetes/latest/controller.yaml
```diff
@@ -0,0 +1,233 @@
...
+              value: /csi/csi.sock
+            - name: KUBE_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - name: plugin-dir
+              mountPath: /csi
+        - name: csi-provisioner
+          securityContext:
```
leakingtapan: Same here

deploy/kubernetes/latest/node.yaml
```diff
@@ -121,6 +80,33 @@ spec:
               mountPath: /csi
             - name: device-dir
               mountPath: /dev
+        - name: node-driver-registrar
+          securityContext:
```
leakingtapan: Is this needed for node-driver-registrar?

deploy/kubernetes/latest/node.yaml
```diff
@@ -121,6 +80,33 @@ spec:
...
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "rm -rf /registration/ebs.csi.aws.com-reg.sock /csi/csi.sock"]
+          env:
+            - name: ADDRESS
+              value: /csi/csi.sock
+            - name: DRIVER_REG_SOCK_PATH
+              value: /var/lib/kubelet/plugins/ebs.csi.aws.com/csi.sock
+            - name: KUBE_NODE_NAME
```
leakingtapan: Which container needs this env? ebs-plugin is reading it from ec2 metadata service already [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/driver.go#L59).

deploy/kubernetes/latest/controller.yaml
```diff
@@ -0,0 +1,233 @@
...
+metadata:
+  name: csi-controller-sa
+  namespace: kube-system
+
+---
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-provisioner-runner-role
```
leakingtapan: NIT: 
```suggestion
  name: external-provisioner-role
```

deploy/kubernetes/latest/controller.yaml
```diff
@@ -0,0 +1,233 @@
...
+  kind: ClusterRole
+  name: external-provisioner-runner-role
+  apiGroup: rbac.authorization.k8s.io
+
+---
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-attacher-runner-role
```
leakingtapan: NIT: 
```suggestion
  name: external-attacher-role
```

deploy/kubernetes/latest/controller.yaml
```diff
@@ -0,0 +1,233 @@
...
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["get", "list", "watch", "create", "update", "patch"]
+  - apiGroups: ["csi.storage.k8s.io"]
+    resources: ["csinodeinfos"]
```
leakingtapan: Seem `csinodeinfos` is only needed by attacher role

leakingtapan:
There are a lot of details here, will do another around of review

deploy/kubernetes/latest/controller.yaml
dkoshkin: These are coming from: https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml

I don't really liking the naming either and if you are ok with changing it I can do that.

deploy/kubernetes/latest/controller.yaml
dkoshkin: Same here: https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml

deploy/kubernetes/latest/controller.yaml
```diff
@@ -0,0 +1,233 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
dkoshkin: What do you mean here, don't we want to keep a version in the repo so a user can deploy the driver?

deploy/kubernetes/latest/controller.yaml
leakingtapan: Yeah, we can decide whatever we like, in this case shorter name feel better to me.

deploy/kubernetes/latest/node.yaml
```diff
@@ -2,46 +2,26 @@ apiVersion: v1
...
-  namespace: default
+  namespace: kube-system
 
 ---
 
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: csi-node
-  namespace: default
```
leakingtapan: Why the node permission are all removed? If this is the case, Im wondering do we still need the service account of node daemon set?

deploy/kubernetes/latest/controller.yaml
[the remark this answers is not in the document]
leakingtapan: I mean move all the manifest files under `deploy/kubernetes/latest` to be under `deploy/kubernetes/`

dkoshkin:
Ok @leakingtapan I think this is ready for a final review.

PR #181, opened 2019-01-20
Update driver manifest files

leakingtapan:
/bug 

1) rename controller daemon set name and node stateful set name. This
resolves name collision when deploying multiple CSI drivers.
2) remove secrets from node ebs-plugin since it is not needed
3) update sample manifest files with better names

**What testing is done?** 
e2e test passed

### DIFF
--- a/deploy/kubernetes/controller.yaml
+++ b/deploy/kubernetes/controller.yaml
@@ -108,15 +108,15 @@ roleRef:
 kind: StatefulSet
 apiVersion: apps/v1beta1
 metadata:
-  name: csi-controller
+  name: ebs-csi-controller
   namespace: kube-system
 spec:
-  serviceName: csi-controller
+  serviceName: ebs-csi-controller
   replicas: 1
   template:
     metadata:
       labels:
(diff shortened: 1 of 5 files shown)


leakingtapan:
@dkoshkin

deploy/kubernetes/controller.yaml
```diff
@@ -108,15 +108,15 @@ roleRef:
 kind: StatefulSet
 apiVersion: apps/v1beta1
 metadata:
-  name: csi-controller
+  name: ebs-controller
   namespace: kube-system
```
dkoshkin: What do you think of `ebs-csi-controller`.

deploy/kubernetes/node.yaml
```diff
@@ -1,16 +1,16 @@
 kind: DaemonSet
 apiVersion: apps/v1beta2
 metadata:
-  name: csi-node
+  name: ebs-node
```
dkoshkin: Sam as above `ebs-node` is confusing, `ebs-csi-node`

PR #183, opened 2019-01-21
Check if target path is mounted before unmounting

### LINKED ISSUE #175: NodeUnstageVolume Idempotency
Why are we concerned about `NodeUnStageVolume` here? I thought CO should guarantee unstage is called after stage for the same target path. So that these two operations should never happen concurrently. 

But this also makes me wondering how should be handle the idempotency of `NodeUnstageVolume` itself? We currently do NOT check whether the target path exists on disk before doing unmount. This will cause issue if the same target path is unmounted twice and the second call will fail since it is already unmount.

_Originally posted by @leakingtapan in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/163_

sreis:

Bug fix. Closes #175

`NodeUnstageVolume` doesn't check if target path was already unmounted.

**What testing is done?** 

We don't have unit tests for the node service so testing was done manually.

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -149,8 +149,29 @@ func (d *Driver) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolu
 		return nil, status.Error(codes.InvalidArgument, "Staging target not provided")
 	}
 
+	// Check if target directory is a mount point. GetDeviceNameFromMount
+	// given a mnt point, finds the device from /proc/mounts
+	// returns the device name, reference count, and error code
+	dev, refCount, err := mount.GetDeviceNameFromMount(d.mounter, target)
+	if err != nil {
+		msg := fmt.Sprintf("failed to check if volume is mounted: %v", err)
+		return nil, status.Error(codes.Internal, msg)
+	}
+
+	// From the spec: If the volume corresponding to the volume_id
+	// is not staged to the staging_target_path, the Plugin MUST
+	// reply 0 OK.
+	if refCount == 0 {
+		klog.V(5).Infof("NodeUnstageVolume: %s target not mounted", target)
+		return &csi.NodeUnstageVolumeResponse{}, nil
+	}
+
+	if refCount > 1 {
+		klog.Warningf("NodeUnstageVolume: found %d references to device %s mounted at target path %s", refCount, dev, target)
+	}
+
 	klog.V(5).Infof("NodeUnstageVolume: unmounting %s", target)
-	err := d.mounter.Interface.Unmount(target)
+	err = d.mounter.Interface.Unmount(target)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Could not unmount target %q: %v", target, err)
 	}
--- a/pkg/driver/node_test.go
+++ b/pkg/driver/node_test.go
@@ -202,11 +202,12 @@ func TestNodeStageVolume(t *testing.T) {
 				t.Fatalf("Expected error %v, got no error", tc.expErrCode)
 			}
 
-			if len(tc.expActions) > 0 && !reflect.DeepEqual(fakeMounter.Log, tc.expActions) {
+			// if fake mounter did anything we should
+			// check if it was expected
+			if len(fakeMounter.Log) > 0 && !reflect.DeepEqual(fakeMounter.Log, tc.expActions) {
 				t.Fatalf("Expected actions {%+v}, got {%+v}", tc.expActions, fakeMounter.Log)
 			}
-
-			if len(tc.expMountPoints) > 0 && !reflect.DeepEqual(fakeMounter.MountPoints, tc.expMountPoints) {
+			if len(fakeMounter.MountPoints) > 0 && !reflect.DeepEqual(fakeMounter.MountPoints, tc.expMountPoints) {
 				t.Fatalf("Expected mount points {%+v}, got {%+v}", tc.expMountPoints, fakeMounter.MountPoints)
 			}
 		})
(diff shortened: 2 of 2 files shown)


sreis:
Test failed with ```"msg":"failed to dial (instance might not be ready yet)"```.

leakingtapan:
@sreis did you do any manual testing to mimic the situation?

pkg/driver/node.go
```diff
@@ -149,8 +149,25 @@ func (d *Driver) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolu
 		return nil, status.Error(codes.InvalidArgument, "Staging target not provided")
 	}
 
+	// Check if target directory is a mount point. GetMountRefs
+	// finds all mount references to the source, returns a list of
```
leakingtapan:
```suggestion
	// finds all mount references to the target, returns a list of
```

pkg/driver/node.go
```diff
@@ -149,8 +149,25 @@ func (d *Driver) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolu
...
+	refs, err := d.mounter.Interface.GetMountRefs(target)
+	if err != nil {
+		msg := fmt.Sprintf("failed to check if volume is mounted: %v", err)
+		return nil, status.Error(codes.Internal, msg)
+	}
+
+	// From the spec: If the volume corresponding to the volume_id
+	// is not staged to the staging_target_path, the Plugin MUST
+	// reply 0 OK.
+	if len(refs) == 0 {
```
leakingtapan: Could it happen that `len(refs) > 1`? And what should we do?

sreis: If we follow the GetMountRefs code we'll end up [at this function](https://github.com/kubernetes/kubernetes/blob/master/pkg/util/mount/mount_linux.go#L1283-L1290) 
```// This function assumes source cannot be device.
// Some filesystems may share a source name, e.g. tmpfs. And for bind mounting,
// it's possible to mount a non-root path of a filesystem, so we need to use
// root path and major:minor to represent mount source uniquely.
``` 
To answer your question I don't see a way of this happening because we don't support `tmpfs` and afaik that type of bind mounting. So, if `len(refs) > 1` it means `target` is being managed by something else other than this driver. 

I understand the concern with handling that code path but if we return an error the CO will keep retrying. Are you ok with logging or do you prefer to return an error in that scenario?

leakingtapan: Let's return an error for now

leakingtapan: Lets log the error for now

leakingtapan:
/lgtm

Checking on the failing integration test

leakingtapan:
This is caused by: https://github.com/kubernetes/test-infra/pull/10934/

leakingtapan:
@sreis testing on EC2 instance the test is actually stuck on detaching:
```
6056 ± ./hack/test-integration.sh                                                                                                                                                ⏎ ✹
=== RUN   TestIntegration
Running Suite: AWS EBS CSI Driver Integration Tests
===================================================
Random Seed: 1548394228
Will run 1 of 1 specs

EBS CSI Driver
  Should create, attach, stage and mount volume, check if it's writable, unmount, unstage, detach, delete, and check if it's deleted
  /home/ubuntu/ebs-csi/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/integration/integration_test.go:52
Creating volume with name "volume-name-integration-test-14653969182178890307"
Waiting for 1 volumes with ID "vol-04bc17a570a8d694e"
Attaching volume "vol-04bc17a570a8d694e" to node "i-0c5166f6e3363507d"
Checking if attachment state of volume "vol-04bc17a570a8d694e" is "attached"
Staging volume "vol-04bc17a570a8d694e" to path "/tmp/volume-name-integration-test-14653969182178890307/stage"
E0125 05:30:39.062107    3575 mount_linux.go:151] Mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t ext4 -o defaults /dev/xvdbl /tmp/volume-name-integration-test-14653969182178890307/stage
Output: mount: wrong fs type, bad option, bad superblock on /dev/xvdbl,
       missing codepage or helper program, or other error
       In some cases useful info is found in syslog - try
       dmesg | tail  or so


Publishing volume "vol-04bc17a570a8d694e" to path "/tmp/volume-name-integration-test-14653969182178890307/mount"
Writing and reading a file
Unpublishing volume "vol-04bc17a570a8d694e" from path "/tmp/volume-name-integration-test-14653969182178890307/mount"
Unstaging volume "vol-04bc17a570a8d694e" from path "/tmp/volume-name-integration-test-14653969182178890307/stage"
Detaching volume "vol-04bc17a570a8d694e" from node "i-0c5166f6e3363507d"
```

The volume is not unstage correctly.

leakingtapan:
/lgtm cancel

The `GetMountRefs` is not doing its job. It returns empty list while the target is still mounted. Have you tested on your environment?

sreis:
@leakingtapan Are you sure `target` is still mounted?

Also, I'm not liking the following sequence of messages. The [spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodeunstagevolume) says `NodeUnstageVolume` should be called **before** `NodeUnpublishVolume`. And it's not what it's doing here.

```
Unpublishing volume "vol-04bc17a570a8d694e" from path "/tmp/volume-name-integration-test-14653969182178890307/mount"
Unstaging volume "vol-04bc17a570a8d694e" from path "/tmp/volume-name-integration-test-14653969182178890307/stage"
Detaching volume "vol-04bc17a570a8d694e" from node "i-0c5166f6e3363507d"
```

I'll try to debug this.

sreis:
I spent most of the time trying to run the integration tests locally and couldn't figure out why `GetMountRefs` works when I deployed it and not in the integration tests.  

So, I changed the approach and used `mount.GetDeviceNameFromMount` instead. This function checks if there's a device mounted at a target path and the number of mount points pointing to said device. On paper `GetMountRefs` should be doing the same thing but it seems to be excluding the `target` path  from the search.

leakingtapan:
@sreis you will need to add some test once #187 is merged to get coverage check happy

sreis:
@leakingtapan Added the test. PTAL

leakingtapan:
Thanks for the quick turn around. Will take a look

Issue #180, opened 2019-01-18
Add the ability to provide arbitrary EBS volume tags


dkoshkin:
**Is your feature request related to a problem? Please describe.**
With dynamically provisioned volumes there is always a danger of having orphaned EBS volumes. With current implementation, the only way to tie an EBS volume to a resource in Kubernetes is via the `CSIVolumeName` tag that gets automatically applied or with `volumeHandle` in the PV spec. However, both of these will not work if the Kubernetes resource is deleted and the EBS volume persists (either due the `ReclaimPolicy` or some bug where the volume is not deleted).

Providing a mechanism for operators to tag created volumes with custom tags, would help them in managing those EBS volumes, with this feature a user would be able to tag the volume with team, app, billing code, or whatever other tag they might already be using with other EBS volumes.

**Describe the solution you'd like in detail**
A new field can be added in the `StorageClass` `parameters` tag:
```
parameters:
  type: gp2 #gp2, io1, sc1, st1 - io1 requires 'iopsPerGB' to be set
  fsType: ext4 # ext4, ext3, ext2
  encrypted: "false"  # false/true
  additionalTags: "team=kubernetes,app=myapp,important"
```
The driver would perform a single validation:
* `additionalTags` does not contain `CSIVolumeName`, as that is reserved to be used by the driver.

Otherwise tags would be passed along to the AWS API request and if any of those validations fail, that error will be propagated back to the user via Kubernetes events.

**Describe alternatives you've considered**
Another option would be for the user to pass these tags in the PVC spec via an annotation. AFAIK this is not supported by the CSI spec.

**Additional context**
I might be wrong here but I believe the AWS cloud-provider supports this feature.

frittentheke:
This feature also has some nice security implications.... I personally would like my provisioners IAM policy to require certain tags to be applied and then only allow the attacher to attach such volumes in their policy.

Otherwise I could i.e. hack into a worker node and reattach i.e. an etcd volume from another, totally unrelated EC2 (yeah, I know about encryption at rest for etcd 😉)

Just best practice on AWS: https://aws.amazon.com/de/premiumsupport/knowledge-center/iam-policy-tags-restrict/

mgoodness:
[...]

PR #185, opened 2019-01-22
New parameter additionalTags to tag EBS volumes

dkoshkin:
New Feature
Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/180

This PR enables tagging created EBS volumes with custom tags. Most of the plumbing was already there since a tag is used to identify the volume, exposed a new parameter  `additionalTags` and added unit test cases.

**What testing is done?** 
* Unit test cases
* Locally tested on a cluster


leakingtapan:
/close 
For further discussion

PR #198, opened 2019-02-01
Remove vendor directory

leakingtapan:
Fixes: #198


leakingtapan:
There are some issue in aws-k8s-test, which we use for integration testing.

leakingtapan:
This should fix the issue https://github.com/aws/aws-k8s-tester/pull/41

PR #209, opened 2019-02-13
Don't wait for the snapshot to be completed

tsmetana:
This patch changes the way the snapshots are being created: Originally the driver waited for the snapshot to complete. This is wrong since the snapshot creation may take a really long time. Also the gRPC would most likely timeout (the call is meant to be blocking) and cause all kinds of other problems.

Now the driver retruns the snapshot as soon as it appears in EC2 with ReadyToUse state in `false` and updates the state accordingly when the controller calls the SnapshotCreate again with the same parameters.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -122,6 +121,7 @@ type Snapshot struct {
 	SourceVolumeID string
 	Size           int64
 	CreationTime   time.Time
+	ReadyToUse     bool
 }
 
 // SnapshotOptions represents parameters to create an EBS volume
@@ -509,10 +509,6 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
 	if res == nil {
 		return nil, fmt.Errorf("nil CreateSnapshotResponse")
 	}
-	err = c.waitForSnapshotCreate(ctx, res.SnapshotId)
-	if err != nil {
-		return nil, err
-	}
 
 	return c.ec2SnapshotResponseToStruct(res), nil
 }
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -732,7 +740,7 @@ func (c *cloud) waitForSnapshotCreate(ctx context.Context, snapshotID *string) e
 			case "completed":
 				return true, nil
 			case "pending":
-				return false, nil
+				return true, nil
```
leakingtapan: Since we are always returning true here, do we still need this `waitForSnapshotCreate` when creating snapshot? Can we just check the snapshot state returned from the `CreateSnapshotWithContext`?

pkg/cloud/cloud.go
```diff
@@ -554,12 +555,19 @@ func (c *cloud) ec2SnapshotResponseToStruct(ec2Snapshot *ec2.Snapshot) *Snapshot
...
 	snapshotSize := util.GiBToBytes(aws.Int64Value(ec2Snapshot.VolumeSize))
-	return &Snapshot{
+	snapshot := &Snapshot{
 		SnapshotID:     aws.StringValue(ec2Snapshot.SnapshotId),
 		SourceVolumeID: aws.StringValue(ec2Snapshot.VolumeId),
 		Size:           snapshotSize,
 		CreationTime:   aws.TimeValue(ec2Snapshot.StartTime),
 	}
+	if aws.StringValue(ec2Snapshot.State) == "completed" {
+		snapshot.ReadyToUse = true
```
leakingtapan:
```suggestion
snapshot.ReadyToUse = aws.StringValue(ec2Snapshot.State) == "completed"
```

pkg/cloud/cloud.go
tsmetana: Yes, this is rather paranoid -- I only wanted to be sure `DescribeSnapshotsWithContext` can find the newly created snapshot.

pkg/cloud/cloud.go
tsmetana: Will change it.

tsmetana:
OK. I've removed the waiting method.

PR #210, opened 2019-02-13
Add snapshot examples

tsmetana:
Added external snapshotter to the controller yaml, updated readme and deployment example.

### DIFF
--- a/deploy/kubernetes/controller.yaml
+++ b/deploy/kubernetes/controller.yaml
@@ -105,6 +105,56 @@ roleRef:
 
 ---
 
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-snapshotter-role
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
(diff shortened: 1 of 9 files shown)


deploy/kubernetes/controller.yaml
```diff
@@ -23,6 +23,18 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["get", "list", "watch", "create", "update", "patch"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
```
leakingtapan: We should create a separate clusterrole just for snapshotter, and tie it with the `csi-controller-sa` service account. Refer to https://github.com/kubernetes-csi/external-snapshotter/blob/01bd7f356e6718dee87914232d287631655bef1d/deploy/kubernetes/rbac.yaml

examples/kubernetes/volume_scheduling/snapshot.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: snapshot.storage.k8s.io/v1alpha1
```
leakingtapan: Could you create a separate dir `snapshot` under `examples/kubernetes` with all the necessary manifest files just for this use case? I am trying to modularize the examples so that user can just pick whichever example they care about. It will be great if you also add some README.md under the snapshot dir with some instructions about how to execute the example.

PR #219, opened 2019-02-19
Add unit tests for ControllerPublish and ControllerUnpublish requests

### LINKED ISSUE #199: Implement Unit Test for ControllerPublishVolume and ControllerUnpublishVolume
- [x] ControllerPublishVolume
- [x] ControllerUnpublishVolume

sreis:

Increases test coverage. 

Increases test coverage.

Closes #199 

**What testing is done?** 

Add unit tests for ControllerPublish and ControllerUnpublish requests.

### DIFF
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -268,7 +268,7 @@ func TestCreateVolume(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			awsDriver := NewFakeDriver("", NewFakeMounter())
+			awsDriver := NewFakeDriver("", NewFakeCloudProvider(), NewFakeMounter())
 
 			resp, err := awsDriver.CreateVolume(context.TODO(), tc.req)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


leakingtapan:
travis is having issue these days:
```
{"message":"Couldn't find a repository matching this job.","error":true}
The command "$GOPATH/bin/goveralls -coverprofile=profile.cov -service=travis-ci" exited with 1.
```

pkg/driver/controller_test.go
```diff
@@ -429,3 +429,204 @@ func TestPickAvailabilityZone(t *testing.T) {
 		})
 	}
 }
+
+func TestControllerPublishVolume(t *testing.T) {
+	fakeCloud := NewFakeCloudProvider()
```
leakingtapan: Could you create a `mock_cloud` using gomock? With that, we could truly unit test controller service without depending on the fake cloud implementation. 

Here is something similar I did for another driver: https://github.com/aws/aws-fsx-csi-driver/blob/master/pkg/driver/controller_test.go#L224

sreis: That's a good idea but maybe we can do it in a follow up PR and introduce it in all the required tests. wdyt?

leakingtapan: Fair enough.

leakingtapan:
@sreis could you rebase. thx

sreis:
@leakingtapan I've rebased PTAL.

pkg/driver/controller_test.go
leakingtapan: #220

Issue #159, opened 2018-12-25
Integration with k8s liveness probe


leakingtapan:
**Is your feature request related to a problem? Please describe.**
See https://github.com/kubernetes-csi/livenessprobe for more details

PR #225, opened 2019-02-26
Add liveness probe

leakingtapan:
Fixes: #159 

**What testing is done?**

### DIFF
--- a/deploy/kubernetes/controller.yaml
+++ b/deploy/kubernetes/controller.yaml
@@ -197,6 +197,18 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
+          ports:
+            - name: healthz
+              containerPort: 9808
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 10
+            timeoutSeconds: 3
+            periodSeconds: 10
+            failureThreshold: 5
         - name: cluster-driver-registrar
           imagePullPolicy: Always
           image: quay.io/k8scsi/csi-cluster-driver-registrar:v1.0.1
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/controller.yaml
```diff
@@ -197,6 +198,19 @@ spec:
...
+              protocol: TCP
+          livenessProbe:
+            failureThreshold: 5
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 10
+            timeoutSeconds: 3
+            periodSeconds: 2
+            failureThreshold: 1
```
dkoshkin: `failureThreshold: 1` might be too aggressive, in their example they actually have 2:
```
failureThreshold: 5
...
failureThreshold: 1
```

deploy/kubernetes/controller.yaml
```diff
@@ -238,16 +252,25 @@ spec:
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-snapshotter
           image: quay.io/k8scsi/csi-snapshotter:v1.0.1
+          imagePullPolicy: Always
```
bertinatto: We might think about changing these policies to `IfNotPresent` so the whole deployment doesn't fail if `quay.io` can't be reached.

deploy/kubernetes/controller.yaml
dkoshkin: Looks like the test failures could have been caused because the the node driver restarted:
```
Logging pods the kubelet thinks is on node ip-172-20-50-17.us-west-2.compute.internal
Feb 26 02:54:34.076: INFO: kube-proxy-ip-172-20-50-17.us-west-2.compute.internal started at <nil> (0+0 container statuses recorded)
Feb 26 02:54:34.076: INFO: ebs-csi-node-xn2c8 started at 2019-02-26 02:33:06 +0000 UTC (0+3 container statuses recorded)
Feb 26 02:54:34.076: INFO: 	Container ebs-plugin ready: true, restart count 6
```

All 4 failures were on the volumes that take a long time to format and even if the formatting is attempted a second time, the 15 min timeout is not enough time.

deploy/kubernetes/controller.yaml
dkoshkin: 👍 for all the tagged images, for `amazon/aws-ebs-csi-driver:latest` we should just remove `imagePullPolicy` and let Kubernetes do its thing and try to pull it since it has the `:latest` tag.

deploy/kubernetes/controller.yaml
leakingtapan: Notice the driver is keep being restarted even without any traffic due to probe timeout. Update the `periodSeconds` to 10s for less frequent probe, `failureThreshold` to 5 for higher failure tolerance

PR #229, opened 2019-02-27
fix a bug when capacity range is nil

leakingtapan:
This bug could happen when `req.GetCapacityRange()` is nil but it is still used later for `req.GetCapacityRange().GetLimitBytes()`

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -54,16 +54,16 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.InvalidArgument, "Volume name not provided")
 	}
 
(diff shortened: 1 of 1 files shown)


leakingtapan:
@dkoshkin its not mentioned clearly from CSI spec: https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume It only says `CapacityRange` is an optional field. So we should take it as an assumption of the driver.

BTW, one interesting statement is from the spec:
```
  // ... If
  // specified it MUST always be honored, even when creating volumes
  // from a source; which MAY force some backends to internally extend
  // the volume after creating it.
```
We should test it in our snapshot restore story at least manually

bertinatto:
/lgtm

Good catch, @leakingtapan!

PR #236, opened 2019-03-07
Update image for runtime deployment

leakingtapan:


dkoshkin:
We should also include the `liveness-probe` in the diagram right?

leakingtapan:
It will be used here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/wiki/Architecture-Diagrams

leakingtapan:
> We should also include the `liveness-probe` in the diagram right?

Updated

PR #238, opened 2019-03-12
Add example for block volume

leakingtapan:


examples/kubernetes/block-volume/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Edit [Persistence Volume Claim Spec](./specs/raw-claim.yaml)
+Make sure the `volumeMode` is `Block`.
+
+### Edit [Application Pod](./specs/pod.yaml)
+Make sure the pod is consuming the PVC with the defined name and `volumeDevices` is used instead of `volumeMounts`.
+
+### Deploy the Application
+```sh
+>> kubectl apply -f examples/kubernetes/block-volume/specs/storageclass.yaml
```
dkoshkin: For this bash example, since it doesn't have an output can you remove `>>` to make it easier to copy and paste.

examples/kubernetes/block-volume/README.md
```diff
@@ -0,0 +1,39 @@
...
+>> kubectl apply -f examples/kubernetes/block-volume/specs/raw-claim.yaml
+>> kubectl apply -f examples/kubernetes/block-volume/specs/pod.yaml
+```
+
+### Access Block Device
+After the objects are created, verify that pod is running:
+
+```sh
+>> kubectl get pods
```
dkoshkin:
```suggestion
$ kubectl get pods
```
I believe the convention is to use an `$` in examples.

PR #239, opened 2019-03-14
Update driver README

leakingtapan:


docs/README.md
```diff
@@ -1,48 +1,29 @@
...
+| v0.1.0                                 | yes   | no     |
 
 ## Features
-### Capabilities
-The list of supported driver capabilities:
-* Identity Service: **CONTROLLER_SERVICE** and **ACCESSIBILITY_CONSTRAINTS**
-* Controller Service: **CREATE_DELETE_VOLUME** and **PUBLISH_UNPUBLISH_VOLUME**
-* Node Service: **STAGE_UNSTAGE_VOLUME**
+The following CSI interfaces are implemented:
+* Controller Service: CreateVolume, DeleteVolume, ControllerPublishVolume, ControllerUnpublishVolume, ControllerGetCapabilities, ValidateVolumeCapabilities, CreateSnapshot, DeleteSnapshot
```
bertinatto: Are these actually called interfaces? Shouldn't be RPC call?

docs/README.md
```diff
@@ -55,51 +36,75 @@ There are several optional parameters that could be passed into `CreateVolumeReq
...
 
-4. Install the `CSINodeInfo` CRD on the cluster using the instructions provided here: [Enabling CSINodeInfo](https://kubernetes-csi.github.io/docs/csi-node-info-object.html#enabling-csinodeinfo).
-
-5. Please refer to [kubernetes CSI docs](https://kubernetes-csi.github.io/docs/) for general CSI driver setup instructions on kubernetes.
+## Kubernetes Version Compability Matrix
+| AWS EBS CSI Driver \ Kubernetes Version| v1.12 | v1.13 |
+|----------------------------------------|-------|-------|
+| master branch                          | no    | yes   |
+| v0.2.0                                 | no    | yes   |
+| v0.1.0                                 | yes   | yes   |
```
bertinatto: One may think that we still support (backport fixes) v0.1.0. Do we?

If not, should we signal somehow that these version are no longer maintained?

docs/README.md
```diff
@@ -55,51 +36,75 @@ There are several optional parameters that could be passed into `CreateVolumeReq
...
-
-5. Deploy the csi-provisioner, csi-attacher and csi-node manifests to the cluster in one step:
+Edit the [secret manifest](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/secret.yaml) using your favorite text editor. The secret should have [enough IAM permission](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/example-iam-policy.json) to create EBS volume. Then deploy the secret:
+```sh
+kubectl apply -f deploy/kubernetes/secret.yaml
+```
 
+Install the `CSINodeInfo` CRD on the cluster:
+```sh
+kubectl create -f https://raw.githubusercontent.com/kubernetes/csi-api/release-1.13/pkg/crd/manifests/csinodeinfo.yaml
```
bertinatto: This shouldn't be necessary anymore starting from 1.14, as these CRDs became built-in objects.

docs/README.md
leakingtapan: I am not planning to backport fixes. But I still have to maintain it since EKS only supports 1.11 for now. And driver 0.2 requires 1.13. We could deprecate v0.1.0 after EKS launches 1.13.

docs/README.md
leakingtapan: Since people might use it for version < 1.14, updated to only do this if this is the case.

Issue #265, opened 2019-04-03
Move variable declaration out of Makefile


bertinatto:
Some variables in the Makefile [1] are filled with the results of some commands (jq, curl, etc.). Those commands are always executed, even when I'm executing targets that don't need those variables.

As an example, this is what I get when I try to build the driver (`golang` image doen't have `jq`):

```
$ make image-release 
docker build -t quay.io/bertinatto/ebs-csi-driver:testing .
Sending build context to Docker daemon  104.2MB
Step 1/8 : FROM golang:1.11.4-stretch as builder
 ---> dd46c1256829
Step 2/8 : WORKDIR /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver
 ---> Using cache
 ---> 07654df5902f
Step 3/8 : ADD . .
 ---> 28a8fd7c4a1c
Step 4/8 : RUN make
 ---> Running in 84bcbb7a2ad6
mkdir -p bin
/bin/sh: 1: jq: not found
(23) Failed writing body
/bin/sh: 1: jq: not found
(23) Failed writing body
/bin/sh: 1: jq: not found
(23) Failed writing body
CGO_ENABLED=0 GOOS=linux go build -ldflags "-X github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.driverVersion=testing -X github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.gitCommit=549b9ec38566744aa5b2b00a4fca2a89240f8347 -X github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.buildDate=2019-04-03T15:28:57Z -s -w" -o bin/aws-ebs-csi-driver ./cmd/
```

It would be better to move the integration test code to another script under `hack/`, then call it from the Makefile. Another option is to move those vars under their specific targets
[1] https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/Makefile#L23-L29

CC @leakingtapan

leakingtapan:
We could move those integration specific variable/commands to [this](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/test-integration.sh) file.

PR #267, opened 2019-04-04
Refactor integration test script

leakingtapan:
Fixes: #265


bertinatto:
What's the difference between this and `test-integration` script? 

Could we have a single script for running integration tests?

leakingtapan:
The test-integration script is running on the ec2 instance. It’s called by aws-k8s-tester, which is called by run-integration-test

bertinatto:
Okay, maybe we can rename them to make this more obvious?

/lgtm

Issue #220, opened 2019-02-20
Refactor controller unit test with proper mock


leakingtapan:
Proper mocking should be used for controller unit testing instead of using the current [fake cloud provider](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/fakes.go) that has fake logic that are implemented for sanity testing purpose.

**Goal**
* move fake cloud provider inside `tests/sanity` package and make sure it is only used there
* auto generate cloud provider using gomock. See [./hack/update-gomock](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/update-gomock) as an example
* update controller_test.go and node_test.go to use the generated cloud provider mock.

**Reference PR**

Could you create a `mock_cloud` using gomock? With that, we could truly unit test controller service without depending on the fake cloud implementation. 

Here is something similar I did for another driver: https://github.com/aws/aws-fsx-csi-driver/blob/master/pkg/driver/controller_test.go#L224

_Originally posted by @leakingtapan in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/219_

bertinatto:
+1

The fake cloud provider object was introduced way before I started using gomock, so we should definitely get rid of it in unit tests.

zacharya:
If no one has taken this up, I'd like to contribute

leakingtapan:
@zacharya I updated the issue description. let me know if you have any questions, will be happy to clarify more.

zacharya:
I submitted #269 for this.  Let me know if you have any issues with the implementation or test refactors.

PR #269, opened 2019-04-05
Refactor controller unit test with proper mock

zacharya:
Refactor Cloud interface unit tests to use GoMock

- Addresses #220 .  
- One thing to discuss - moving `cloud.FakeCloudProvider` to `sanity.fakeCloudProvider` required exporting `cloud.metadata` due to the dependency.  I don't see any issues with this, but I figured I'd bring it up in case some of the maintainers see problems.

**What testing is done?** 
Both `make test` and `make test-sanity` are running fine.


leakingtapan:
Im okay with making metadata struct public.

pkg/driver/controller_test.go
```diff
@@ -48,327 +55,752 @@ func TestCreateVolume(t *testing.T) {
...
+					VolumeId: "invalid-volume-name",
+				}
+				expResp := &csi.DeleteVolumeResponse{}
+
+				ctx := context.Background()
+				mockCtl := gomock.NewController(t)
+				defer mockCtl.Finish()
+
+				mockCloud := mocks.NewMockCloud(mockCtl)
+				mockCloud.EXPECT().DeleteDisk(gomock.Eq(ctx), gomock.Eq(req.VolumeId)).Return(true, nil)
```
leakingtapan: I think this case should return `cloud.ErrNotFound` for invalid volume ID.

pkg/driver/controller_test.go
```diff
@@ -48,327 +55,752 @@ func TestCreateVolume(t *testing.T) {
...
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					t.Fatalf("Unexpected error: %v", srvErr.Code())
+				}
+				if !reflect.DeepEqual(resp, expResp) {
+					t.Fatalf("Expected resp to be %+v, got: %+v", expResp, resp)
+				}
 			},
```
leakingtapan: And could you add a test case for `DeleteDisk` failed?

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+				if _, err := awsDriver.CreateSnapshot(context.Background(), req); err != nil {
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					if srvErr.Code() != expErrCode {
+						t.Fatalf("Expected error code %d, got %d message %s", expErrCode, srvErr.Code(), srvErr.Message())
+					}
+					return
+				}
```
leakingtapan: Should add an else branch to fail the test if `err == nil`

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+
+				awsDriver := controllerService{cloud: mockCloud}
+				resp, err := awsDriver.CreateSnapshot(context.Background(), req)
+				if err != nil {
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+				}
+				if expErrCode != codes.OK {
```
leakingtapan: This line is always true?
```suggestion
				if err != codes.OK {
```

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+				}
+
+				mockCloud.EXPECT().GetSnapshotByName(gomock.Eq(ctx), gomock.Eq(extraReq.GetName())).Return(mockSnapshot, nil)
+				_, err = awsDriver.CreateSnapshot(ctx, extraReq)
+				if err != nil {
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					if srvErr.Code() != extraExpErrCode {
```
leakingtapan: inline the `extraExpErrCode` for readability since its actual codes.Ok

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+				}
+				snap := resp.GetSnapshot()
+				if snap == nil {
+					t.Fatalf("Expected snapshot %v, got nil", expSnapshot)
+				}
+
+				mockCloud.EXPECT().GetSnapshotByName(gomock.Eq(ctx), gomock.Eq(extraReq.GetName())).Return(mockSnapshot, nil)
+				_, err = awsDriver.CreateSnapshot(ctx, extraReq)
+				if err != nil {
+					srvErr, ok := status.FromError(err)
```
leakingtapan: Could we failed right away if `err!=nil`?

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					if srvErr.Code() != extraExpErrCode {
+						t.Fatalf("Expected error code %d, got %d message %s", expErrCode, srvErr.Code(), srvErr.Message())
+					}
+					return
+				}
+				if extraExpErrCode != codes.OK {
```
```suggestion
				if err != codes.OK {
```

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					if srvErr.Code() != expErrCode {
+						t.Fatalf("Expected error code %d, got %d message %s", expErrCode, srvErr.Code(), srvErr.Message())
+					}
+					return
+				}
+				if expErrCode != codes.OK {
```
```suggestion
				if err != codes.OK {
```

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+		name     string
+		testFunc func(t *testing.T)
 	}{
 		{
-			name:       "success normal",
-			req:        &csi.DeleteSnapshotRequest{},
-			expErrCode: codes.OK,
+			name: "success normal",
+			testFunc: func(t *testing.T) {
+				expErrCode := codes.OK
```
leakingtapan: NIT: inline `expErrCode`

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					if srvErr.Code() != expErrCode {
+						t.Fatalf("Expected error code %d, got %d message %s", expErrCode, srvErr.Code(), srvErr.Message())
+					}
+					return
+				}
+				if expErrCode != codes.OK {
```
```suggestion
				if err != codes.OK {
```

pkg/driver/controller_test.go
```diff
@@ -432,166 +864,299 @@ func TestPickAvailabilityZone(t *testing.T) {
...
+					t.Fatalf("Expected error %v, got no error", expErrCode)
+				}
+			},
 		},
 		{
 			name: "success not found",
-			req: &csi.DeleteSnapshotRequest{
-				SnapshotId: "xxx",
+			testFunc: func(t *testing.T) {
+				expErrCode := codes.OK
```
leakingtapan: NIT: inline `expErrCode`

pkg/driver/controller_test.go
```diff
@@ -600,193 +1165,390 @@ func TestControllerPublishVolume(t *testing.T) {
...
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					if srvErr.Code() != expErrCode {
+						t.Fatalf("Expected error code %d, got %d message %s", expErrCode, srvErr.Code(), srvErr.Message())
+					}
+					return
+				}
+				if expErrCode != codes.OK {
```
leakingtapan: This check seem problematic. Why do we need this check when `err == nil` from `if _, err := awsDriver.ControllerPublishVolume(ctx, req); `?

pkg/driver/controller_test.go
```diff
@@ -600,193 +1165,390 @@ func TestControllerPublishVolume(t *testing.T) {
...
```
leakingtapan: same above

pkg/driver/controller_test.go
```diff
@@ -600,193 +1165,390 @@ func TestControllerPublishVolume(t *testing.T) {
...
```
leakingtapan: same above

pkg/driver/controller_test.go
```diff
@@ -600,193 +1165,390 @@ func TestControllerPublishVolume(t *testing.T) {
...
```
```suggestion
				if err == nil {
				   t.Fatalf("Expected error %v, got no error", expErrCode) 
				}
```

pkg/driver/controller_test.go
```diff
@@ -600,193 +1165,390 @@ func TestControllerPublishVolume(t *testing.T) {
...
```
leakingtapan: Same above

pkg/driver/controller_test.go
```diff
@@ -600,193 +1165,390 @@ func TestControllerPublishVolume(t *testing.T) {
...
```
leakingtapan: same above

pkg/driver/node_test.go
```diff
@@ -762,9 +808,9 @@ func TestNodeGetInfo(t *testing.T) {
 	}
 }
 
-func newTestNodeService(metadata cloud.MetadataService, mounter mount.Interface) nodeService {
+func newTestNodeService(cloud cloud.Cloud, mounter mount.Interface) nodeService {
```
leakingtapan: I think we can just pass in Metadata here since it is changed to be public:

```suggestion
func newTestNodeService(cloud cloud.MetadataService, mounter mount.Interface) nodeService {
```

So that it saves us from creating mock cloud in the test as well.

zacharya:
Thanks for the review.  I'll knock these out first thing tomorrow.

pkg/driver/node_test.go
zacharya: Good catch.  Done and was able to remove all mock cloud creation.

pkg/driver/controller_test.go
zacharya: Done

pkg/driver/controller_test.go
zacharya: Done

pkg/driver/controller_test.go
zacharya: Done

pkg/driver/controller_test.go
zacharya: Done

pkg/driver/controller_test.go
zacharya: Done

pkg/driver/controller_test.go
zacharya: I agree.  I think this one was just trying to port the dynamic error handling logic to the single test without enough context.  Removed.

pkg/driver/controller_test.go
zacharya: Done

zacharya:
@leakingtapan I think I've touched on all of your suggestions.  I went through each test case with the goal of making each more readable (rather than bring the dynamic error handling logic over from the previous state).  Let me know if you see anything that still needs fixing.

leakingtapan:
The code a lot readable and more test covered now Thank @zacharya 

/lgtm
/approve

bertinatto:
Just a heads up about this PR: we should always squash commits as much as possible.

leakingtapan:
+1 Forgot to remind @zacharya of doing this.

zacharya:
@bertinatto @leakingtapan Will do next time

PR #271, opened 2019-04-05
Add expand support

### LINKED ISSUE #146: Add Volume Resizing Support
Volume Resizing Support includes:

* Should support all filesystem types: ext2, ext3, ext4, xfs. It should using different resize tool for different filesystem.

* And for different instance types including nvme.

* And for filesystems that requires resizing partition table.

This should be compatiable with in-tree volume feature 

**Extra Context**
CSI Resize KEP: https://github.com/kubernetes/enhancements/pull/780

**Ref**
* https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/
* https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/recognize-expanded-volume-linux.html

bertinatto:

/kind feature

This PR adds partial support for volume expanding.

It doesn't add the resizer to the manifest file on purpose, so we can keep working on this without exposing the feature to people that try out the driver.

Fixes #146.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -160,6 +160,8 @@ type EC2 interface {
 	CreateSnapshotWithContext(ctx aws.Context, input *ec2.CreateSnapshotInput, opts ...request.Option) (*ec2.Snapshot, error)
 	DeleteSnapshotWithContext(ctx aws.Context, input *ec2.DeleteSnapshotInput, opts ...request.Option) (*ec2.DeleteSnapshotOutput, error)
 	DescribeSnapshotsWithContext(ctx aws.Context, input *ec2.DescribeSnapshotsInput, opts ...request.Option) (*ec2.DescribeSnapshotsOutput, error)
+	ModifyVolumeWithContext(ctx aws.Context, input *ec2.ModifyVolumeInput, opts ...request.Option) (*ec2.ModifyVolumeOutput, error)
+	DescribeVolumesModificationsWithContext(ctx aws.Context, input *ec2.DescribeVolumesModificationsInput, opts ...request.Option) (*ec2.DescribeVolumesModificationsOutput, error)
 }
 
 type Cloud interface {
@@ -168,6 +170,7 @@ type Cloud interface {
 	DeleteDisk(ctx context.Context, volumeID string) (success bool, err error)
 	AttachDisk(ctx context.Context, volumeID string, nodeID string) (devicePath string, err error)
 	DetachDisk(ctx context.Context, volumeID string, nodeID string) (err error)
+	ResizeDisk(ctx context.Context, volumeID string, reqSize int64) (newSize int64, err error)
 	WaitForAttachmentState(ctx context.Context, volumeID, state string) error
 	GetDiskByName(ctx context.Context, name string, capacityBytes int64) (disk *Disk, err error)
 	GetDiskByID(ctx context.Context, volumeID string) (disk *Disk, err error)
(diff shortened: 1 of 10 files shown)


bertinatto:
The sanity error is already fixed in csi-test: https://github.com/kubernetes-csi/csi-test/commit/455ebbfab40bee10001a4c3341accc628ed7be2d

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
-		if awsError.Code() == "InvalidSnapshot.NotFound" {
-			return true
+	return isAWSError(err, "InvalidSnapshot.NotFound")
+}
+
+// ResizeDisk resizes an EBS volume in GiB increments,
+// rouding up to the next GiB if reqSize isn't an even GiB.
+// It returns the new size or a negative number if the size couldn't be determined.
+func (c *cloud) ResizeDisk(ctx context.Context, volumeID string, reqSizeBytes int64) (int64, error) {
+	// If current size isgreater or equal, than there's nothing to be done here
```
leakingtapan:
```suggestion
	// If current size is greater or equal, then there's nothing to be done here
```

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
+			aws.String(volumeID),
+		},
+	}
+	volume, err := c.getVolume(ctx, request)
+	if err != nil {
+		return -1, err
+	}
+
+	// AWS resizes in chunks of GiB (not GB)
+	reqSizeGiB := util.RoundUpGiB(reqSizeBytes)
```
leakingtapan: NIT: use `newSizeGiB` in comparison to `oldSizeGiB` to reduce confusion.

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
+func (c *cloud) ResizeDisk(ctx context.Context, volumeID string, reqSizeBytes int64) (int64, error) {
+	// If current size isgreater or equal, than there's nothing to be done here
+	request := &ec2.DescribeVolumesInput{
+		VolumeIds: []*string{
+			aws.String(volumeID),
+		},
+	}
+	volume, err := c.getVolume(ctx, request)
+	if err != nil {
+		return -1, err
```
leakingtapan: I feels its simpler if the caller just need to check on err instead of both the err and the returned size. 
```suggestion
		return 0, err
```

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
-	if awsError, ok := err.(awserr.Error); ok {
-		// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html
-		if awsError.Code() == "InvalidSnapshot.NotFound" {
-			return true
+	return isAWSError(err, "InvalidSnapshot.NotFound")
+}
+
+// ResizeDisk resizes an EBS volume in GiB increments,
+// rouding up to the next GiB if reqSize isn't an even GiB.
+// It returns the new size or a negative number if the size couldn't be determined.
```
leakingtapan:
```suggestion
// It returns the new size or error if failed to resize disk
```

pkg/driver/controller_test.go
```diff
@@ -1719,3 +1719,94 @@ func TestControllerUnpublishVolume(t *testing.T) {
...
+				},
+			},
+			newSize:  -1,
+			expError: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.Background()
```
leakingtapan: We recently refactor the test to create a `testFunc` for each test cases, so that the `testFunc` body is cleaner and easier to maintain when isolated between different test cases. Could you do the same? See `TestControllerUnpublishVolume` as an example.

pkg/driver/node.go
```diff
@@ -228,6 +230,35 @@ func (d *nodeService) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
 }
 
+func (d *nodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
+	klog.V(4).Infof("NodeExpandVolume: called with args %+v", *req)
+	volumeID := req.GetVolumeId()
+	if len(volumeID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
+	}
+
+	// TODO: range specified
```
leakingtapan: Is this still needed?

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
+	if waitErr != nil {
+		// Return negative number to avoid describing the volume
+		// to get the actual volume size after the modification failed.
+		return -1, waitErr
+	}
+
+	return modVolSizeGiB, nil
+}
+
+func (c *cloud) getVolumeModification(ctx context.Context, volumeID string) (*ec2.VolumeModification, error) {
```
leakingtapan:
```suggestion
func (c *cloud) getLatestVolumeModification(ctx context.Context, volumeID string) (*ec2.VolumeModification, error) {
```

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
+	output, err := c.ec2.ModifyVolumeWithContext(ctx, req)
+	if err != nil {
+		if isAWSErrorIncorrectModification(err) {
+			m, err2 := c.getVolumeModification(ctx, volumeID)
+			if err2 != nil {
+				return -1, fmt.Errorf("could not get modification state of volume %q: %v", volumeID, err)
+			}
+
+			// Check for idempotency
+			state := aws.StringValue(m.ModificationState)
```
leakingtapan: I think we need to move the idempotency check to controller.go. That's where we handles idempotency for `CreateVolume` and `ControllerPublish` so far. And cloud.go should be a simply interface to AWS service to not handle too much other complexities.

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
+	waitErr := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		m, err := c.getVolumeModification(ctx, volumeID)
+		if err != nil {
+			return false, err
+		}
+		if aws.StringValue(m.ModificationState) == ec2.VolumeModificationStateCompleted || aws.StringValue(m.ModificationState) == ec2.VolumeModificationStateOptimizing {
+			modVolSizeGiB = aws.Int64Value(m.TargetSize)
+			return true, nil
+		}
+		return true, nil
```
leakingtapan: Should we return false here? Otherwise, `modVolSizeGiB` will be 0 in this case?

pkg/cloud/cloud.go
```diff
@@ -765,27 +768,130 @@ func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
...
+	})
+	if waitErr != nil {
+		// Return negative number to avoid describing the volume
+		// to get the actual volume size after the modification failed.
+		return -1, waitErr
```
leakingtapan: Can we return 0 instead since there is `waitErr` returned anyway?

pkg/driver/controller_test.go
```diff
@@ -1719,3 +1719,94 @@ func TestControllerUnpublishVolume(t *testing.T) {
 		t.Run(tc.name, tc.testFunc)
 	}
 }
+
+func TestExpandVolume(t *testing.T) {
```
```suggestion
func TestControllerExpandVolume(t *testing.T) {
```

pkg/driver/controller_test.go
bertinatto: hm, I agree that having a main test function might make things hard to read sometimes,  but the `testFunc` from other tests has lots of duplicated boilerplate code.

It seems like just adding the `testFunc` added to this file around 200 lines of boilerplate code (d6a4017a). IMO that pattern is more appropriate when changes take 1 or 2 lines.

I can change this test if you feel sure about it, though.

pkg/cloud/cloud.go
bertinatto: Fixed

pkg/cloud/cloud.go
bertinatto: Fixed

pkg/cloud/cloud.go
bertinatto: I removed the comment because it's misleading.

bertinatto:
> It seems I cannot find the change for manifest to add the resizer? could you add the resizer in manifest?

I was hoping to leave the manifest out for now to avoid exposing the feature before it's ready (see the PR's description). However, I understand it might be hard to test it, so I just added it.

pkg/driver/node.go
```diff
@@ -227,6 +230,38 @@ func (d *nodeService) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
+	devicePath := strings.TrimSpace(string(output))
+	if len(devicePath) == 0 {
+		return nil, status.Errorf(codes.Internal, "Could not get valid device for mount path: %q", req.GetVolumePath())
+	}
+
+	// TODO: refactor Mounter to expose a mount.SafeFormatAndMount object
+	r := resizefs.NewResizeFs(&mount.SafeFormatAndMount{
+		Interface: mount.New(""),
+		Exec:      mount.NewOsExec(),
+	})
```
bertinatto: @leakingtapan: we might need to refactor the `Mounter` interface, but I'd like to avoid doing that in this PR.

leakingtapan:
> I was hoping to leave the manifest out for now to avoid exposing the feature before it's ready (see the PR's description). However, I understand it might be hard to test it, so I just added it.

Good point. How about we add the resizer into the helm chat and make it configurable (like what I did for [volume scheduling](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/aws-ebs-csi-driver/values.yaml#L17).

The advantage is we can keep the manifest to only expose the stable features and use helm chart to expose alpha/unstable features. The drawback is we need to careful to keep both in sync

pkg/driver/node.go
leakingtapan: I think the [Mounter interface](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/mount.go) already implements `mount.SafeFormatAndMount`? Could you try:
```
	r := resizefs.NewResizeFs(&d.mounter)
```

pkg/driver/controller_test.go
leakingtapan: Don't want to block this PR. Ok with live with shared test function.

pkg/driver/node.go
bertinatto: `cannot use &d.mounter (type *Mounter) as type *mount.SafeFormatAndMount in argument to resizefs.NewResizeFs`

`NewResizeFs` won't take `d.mounter` as argument because `SafeFormatAndMount` is a struct (not an interface).

pkg/driver/node.go
```diff
@@ -227,6 +230,39 @@ func (d *nodeService) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
+
+	// TODO: lock per volume ID to have some idempotency
```
bertinatto: @leakingtapan: I opened #307 to discuss this. I think the idempotency of the node service should be fixed in a separate PR.

bertinatto:
> > I was hoping to leave the manifest out for now to avoid exposing the feature before it's ready (see the PR's description). However, I understand it might be hard to test it, so I just added it.
> 
> Good point. How about we add the resizer into the helm chat and make it configurable (like what I did for [volume scheduling](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/aws-ebs-csi-driver/values.yaml#L17). )
> 
> The advantage is we can keep the manifest to only expose the stable features and use helm chart to expose alpha/unstable features. The drawback is we need to careful to keep both in sync

Sounds good to me. Would it be acceptable to leave this out of this PR, though? Unfortunately I have a limited bandwidth at the moment, so I won't be able to tinker with helm soon.

leakingtapan:
I can do the helm part.

Could you rebase the PR before the merge?

bertinatto:
> I can do the helm part.

Should I remove the manifest commit then?

> Could you rebase the PR before the merge?

I think it's already rebased.

leakingtapan:
Let's remove the manifest and squash the commits into one.
Sorry for the confusion

bertinatto:
> Let's remove the manifest and squash the commits into one.
> Sorry for the confusion

Done

Issue #177, opened 2019-01-10
Investigate Deploying Multiple CSI Driver Best Practises


leakingtapan:
Several Points:
* sidecars should be packaged with its own driver and we should allow duplicating them across different CSI drivers. 
* Each sidecar container only supports only one CSI driver and they usually are tested, packaged and deployed along with a specific CSI driver release.
* each driver should have its own set of service accounts, cluster roles and cluster role bindings

leakingtapan:
The above point is good enough as of now.
/close

PR #277, opened 2019-04-09
Rename driver sa, clusterrole and clusterrole binding to be driver specific

leakingtapan:
Ref: #177 

/cc @dkoshkin @bertinatto


bertinatto:
/lgtm

Since there could N CSI driver deployments in the cluster.

Issue #278, opened 2019-04-14
Make Storageclass Parameters Case Insensitive


leakingtapan:
/kind bug

**What happened?**
Ref: https://github.com/kubernetes/kubernetes/issues/75941

PR #283, opened 2019-04-19
Make storageclass paramter case in-sensitive

leakingtapan:
Fixes: #278 

/cc @bertinatto

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -19,6 +19,7 @@ package driver
 import (
 	"context"
 	"strconv"
+	"strings"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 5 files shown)


pkg/driver/controller.go
```diff
@@ -99,8 +100,35 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+			}
+		case EncryptedKey:
+			if value == "true" {
+				isEncrypted = true
+			} else {
+				isEncrypted = false
+			}
+		case KmsKeyIdKey:
+			kmsKeyId = value
+		}
```
bertinatto: Shouldn't we explicitly default to an `codes.InvalidArgument` error if we get something unexpected?

Do you know what the in-tree plugin does in this case? If it simply ignores the parameter, we may do the same.

pkg/driver/controller.go
```diff
@@ -99,8 +100,35 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	var (
+		fsType      string
+		volumeType  string
+		iopsPerGB   int
+		isEncrypted bool
+		kmsKeyId    string
+	)
+
+	for key, value := range req.GetParameters() {
+		switch strings.ToLower(key) {
```
bertinatto: Just to clarify, do we support a lower case `fsType` in-tree?  What about the value?

pkg/driver/controller.go
leakingtapan: It does return errors: https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/awsebs/aws_util.go#L180

Updated

pkg/driver/controller.go
leakingtapan: We do support lower case `fstype` in-tree as the key, that's [this PR](https://github.com/kubernetes/kubernetes/issues/75941) fixes for in-tree.

bertinatto:
Nice fix, good job on keeping in-tree and driver synced.

The changes LGTM, but I'd like to see a test for this. If you want to get this in ASAP, we can merge this and can create an Issue for the current cycle.

leakingtapan:
@bertinatto added unit test. PTAL

Issue #233, opened 2019-03-05
Implement ListSnapshot


leakingtapan:
**Is your feature request related to a problem? Please describe.**
As a follow up to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/25

/cc @tsmetana

zacharya:
Is anyone working on this? I'd like to take it on.

zacharya:
I'm ready to put in a PR on this, just have one question:

The [CSI spec](https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go#L3198) allows `ListSnapshotsRequest.MaxEntries` to be anything 0 or greater, but the AWS API requires `DescribeSnapshotsInput.MaxResults` to be between 5 and 1000.  I'm setting `DescribeSnapshotsInput.MaxResults` if `ListSnapshotsRequest.MaxEntries` is greater than 0. If it's between 1 and 4, I set it to 5 and just trim the slice before returning to facilitate compliance with the CSI spec.

Do we want to do the same thing with values greater than 1000?  One approach there would be to facilitate multiple calls, but there is currently no upper limit to `ListSnapshotsRequest.MaxEntries` in the spec other than `int32` bounds.  Another approach is to just let the AWS API behave [as documented](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#DescribeSnapshotsInput) and we can return 1000 entries and a `NextToken`.

leakingtapan:
> but the AWS API requires `DescribeSnapshotsInput.MaxResults` to be between 5 and 1000. I'm setting

What's the `DescribeSnapshotsInput` behavior when maxResults is less than 5?

> I set it to 5 and just trim the slice before returning to facilitate compliance with the CSI spec.

After trim, the next time `ListSnapshots`is called, is the elements between 2 and 4 going to be returned?

> Another approach is to just let the AWS API behave as documented and we can return 1000 entries and a NextToken
I feel this way is better. @bertinatto how do you think?

zacharya:
When it's less than 5, it will return an `InvalidParameterValue` error.  In this way it differs from being greater than 1000 where it will just truncate and return a `NextToken` value that starts with 1001.  We could return an error, but then there would be AWS-specific CSI behavior leaking through.

[...]

PR #286, opened 2019-04-25
Implementing ListSnapshots

zacharya:
Fixes: #233 

This implements ListSnapshots from the CSI spec (not yet used by Kubernetes).

**What testing is done?** 
Unit tests have been written for `driver` and `cloud` packages that handle both happy path and error cases.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -95,6 +95,9 @@ var (
 	// ErrMultiSnapshots is returned when multiple snapshots are found
 	// with the same ID
 	ErrMultiSnapshots = errors.New("Multiple snapshots with the same name found")
+
+	// ErrInvalidMaxResults is returned when a MaxResults pagination parameter is between 1 and 4
+	ErrInvalidMaxResults = errors.New("MaxResults parameter must be 0 or greater than or equal to 5")
 )
 
 // Disk represents a EBS volume
@@ -142,7 +145,7 @@ type SnapshotOptions struct {
 // ec2ListSnapshotsResponse is a helper struct returned from the AWS API calling function to the main ListSnapshots function
 type ec2ListSnapshotsResponse struct {
 	Snapshots []*ec2.Snapshot
-	NextToken string
+	NextToken *string
 }
 
 // EC2 abstracts aws.EC2 to facilitate its mocking.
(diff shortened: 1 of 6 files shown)


pkg/cloud/cloud.go
```diff
@@ -542,6 +559,44 @@ func (c *cloud) GetSnapshotByName(ctx context.Context, name string) (snapshot *S
 	return c.ec2SnapshotResponseToStruct(ec2snapshot), nil
 }
 
+func (c *cloud) ListSnapshots(ctx context.Context, volumeID string, maxResults int64, nextToken string) (listSnapshotsResponse *ListSnapshotsResponse, err error) {
+	describeSnapshotsInput := &ec2.DescribeSnapshotsInput{}
+
+	if maxResults >= 5 {
```
leakingtapan: Since `controller.ListSnapshots` already has logic to guarantee `maxResults >= 5`, as `cloud` is an internal struct, I feel we can just set `describeSnapshotsInput.MaxResults = aws.Int64(maxResults)` without checking `if maxResults >= 5 {`.

pkg/cloud/cloud.go
```diff
@@ -542,6 +559,44 @@ func (c *cloud) GetSnapshotByName(ctx context.Context, name string) (snapshot *S
 	return c.ec2SnapshotResponseToStruct(ec2snapshot), nil
 }
 
+func (c *cloud) ListSnapshots(ctx context.Context, volumeID string, maxResults int64, nextToken string) (listSnapshotsResponse *ListSnapshotsResponse, err error) {
```
leakingtapan: Could you add some documentation for this function and clarify that `maxResults==0` means no restriction?

pkg/cloud/cloud_test.go
```diff
@@ -647,6 +648,165 @@ func TestGetSnapshotByName(t *testing.T) {
...
+
+				if _, err := c.ListSnapshots(ctx, "", 0, ""); err != nil {
+					if err != ErrNotFound {
+						t.Fatalf("Expected error %v, got %v", ErrNotFound, err)
+					}
+				} else {
+					t.Fatalf("Expected error, got none")
+				}
+			},
+		},
```
leakingtapan: Could you add a test case where the `volumeId` is set?

pkg/cloud/cloud.go
zacharya: I just pushed another commit that enabled ListSnapshot sanity testing.  In order to satisfy the unit tests from the sanity testing packages, I had to push the logic down into the `cloud` package so controller was able to handle 0 < `MaxEntries` < 5.  Let me know how you want to proceed with this.  It's mostly cosmetic as we're just using `fakeCloudProvider` anyways.

leakingtapan: If that's what the sanity test checks, im okay with the change.

PR #288, opened 2019-04-30
Update sanity tests

bertinatto:

/kind bug

Update CSI sanity library and fix the fake cloud implementation to return an error if snapshot doesn't exist. This is a requirement to test #271.


bertinatto:
I just removed the `WIP` label to unblock #271.

/assign @leakingtapan 
PTAL

go.mod
```diff
@@ -47,7 +47,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/json-iterator/go v1.1.5 // indirect
-	github.com/kubernetes-csi/csi-test v1.1.0
+	github.com/kubernetes-csi/csi-test v2.0.0+incompatible
```
leakingtapan: Do you know what are the incompatible changes introduced by this csi-test v2.0.0 vs v1.1.0

bertinatto: Yes: https://github.com/kubernetes-csi/csi-test/releases/tag/v2.0.0

We should be good.

PR #289, opened 2019-05-03
Add max number of volumes that can be attached to an instance

### LINKED ISSUE #39: Populate MaxVolumesPerNode in NodeGetInfoResponse
This is to help CO to decide how many volume maximum could be attached to the node without going over the limit. 

Reference:
https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetinfo

bertinatto:

/king feature

Without having this field set, k8s will interpret that the node can handle any amount of volumes. I'll create an issue to customize this number per instance type.

Closes #39

/assign @leakingtapan

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
@@ -44,6 +45,13 @@ const (
 
 	// default file system type to be used when it is not provided
 	defaultFsType = FSTypeExt4
+
+	// defaultMaxEBSVolumes is the maximum number of volumes that an AWS instance can have attached.
+	// More info at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html
+	defaultMaxEBSVolumes = 39
+
+	// defaultMaxEBSNitroVolumes is the limit of volumes for some smaller instances, like c5 and m5.
+	defaultMaxEBSNitroVolumes = 25
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/driver/node.go
```diff
@@ -44,6 +44,10 @@ const (
 
 	// default file system type to be used when it is not provided
 	defaultFsType = FSTypeExt4
+
+	// maxVolumesPerNode is the maximum number of volumes that an AWS instance can have attached.
+	// More info at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html
+	maxVolumesPerNode = 39
```
leakingtapan: Since NVME volume only allow 28 attachment, and some instance type has EBS as default volume. Feels its safer to set the value to 27 to begin with to accommodate worse case.

bertinatto: I added some logic to return the volumes limit based on the instance type.

bertinatto: PTAL and let me know if it makes sense.

pkg/driver/node.go
```diff
@@ -464,6 +473,16 @@ func findNvmeVolume(findName string) (device string, err error) {
 	return resolved, nil
 }
 
+// getVolumesLimit returns the limit of volumes that the node supports
+func (d *nodeService) getVolumesLimit() int64 {
+	ebsNitroInstanceTypeRegex := "^[cmr]5.*|t3|z1d"
+	instanceType := d.metadata.GetInstanceType()
+	if ok, _ := regexp.MatchString(ebsNitroInstanceTypeRegex, instanceType); ok {
```
leakingtapan: Feels its less error prune if we just maintain a hard coded list of nitro instance instead of regex. And define a function that checks if it is nitro based:

```go
func isNitroInstance(instanceType string) boolean 
```

How do you think?

See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances for a full list of Nitro instances.

bertinatto: I'm OK with that, but I'm not sure if I understand that page well enough to come up with a reliable list. If nothing, the regex above comes from the in-tree implementation, so it seems to be a reasonable good start.

leakingtapan:
Since the implementation aligns with in-tree implementation, let's merged it in. And we could refactor both later.

/lgtm

Issue #287, opened 2019-04-30
Return error if snapshot doesn't exist


bertinatto:
/kind bug

**What happened?**

It seems like we're not returning a `status.NotFound` error when `CreateVolume` is given a non-existent snapshot ID.

https://github.com/kubernetes-csi/csi-test/blob/e1f2421c7f0a0cc7f9942fea1192392f253891db/pkg/sanity/controller.go#L738

CC @tsmetana @leakingtapan

bertinatto:
My guess is that AWS will return an `InvalidSnapshot.NotFound` error. In that case, we just have to  return an `cloud.ErrNotFound` in `CreateDisk`; and in `CreateVolume` return an `status.NotFound`.

tsmetana:
The strange thing is that the tesuite complains about controller returning `nil` not a wrong error. I'll take a look at this.

bertinatto:
> The strange thing is that the test suite complains about controller returning nil not a wrong error. I'll take a look at this.

Yes, I fixed that in #288. However, in addition to fixing the fake cloud implementation (ie #288), we need to make sure this is fixed in `cloud.go` as well.

tsmetana:
> Yes, I fixed that in #288. However, in addition to fixing the fake cloud implementation (ie #288), we need to make sure this is fixed in `cloud.go` as well.

OK. So we miss the error "translation". Thanks for explaining.

PR #293, opened 2019-05-07
Return correct error when restoring non-existent snapshot

tsmetana:
This should fix the issue #287: return correct error code when restoring a volume referencing non-existent snapshot.


tsmetana:
@bertinatto I hope this is what you had on mind. :)

bertinatto:
The test for this is at #288.

Issue #280, opened 2019-04-16
Use aws native wait function for volume


leakingtapan:
**Is your feature request related to a problem?/Why is this needed**
The driver need to wait for volume creation after it is created; wait for volume attachment after it is attached to some instance. We currently implement the functionality ourselves, but instead the aws natvie wait functions could be used:

* ec2.[WaitUntilVolumeAvailableWithContext](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.WaitUntilVolumeAvailableWithContext) -> [waitForVolume](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L653)
* ~ec2.[WaitUntilVolumeInUseWithContext](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.WaitUntilVolumeInUseWithContext) -> [WaitForAttachmentStatus](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L387)~

/feature

bertinatto:
+1

zacharya:
I'd like to take this on if there are no objections

PR #296, opened 2019-05-14
Replace custom wait function with AWS native wait functions

zacharya:
Fixes: #280 

This PR replaces custom wait functions in the `cloud` package with AWS-native wait functions, namely:
`ec2.WaitUntilVolumeInUseWithContext` and `ec2.WaitUntilVolumeAvailableWithContext`

**What testing is done?** 
Unit tests have been written to handle errors returned from the AWS-native wait functions.  Also, unit tests for `cloud.CreateDisk`, `cloud.AttachDisk` and `cloud.DetachDisk` have been refactored so that test functions are defined per test case.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -32,7 +32,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	dm "github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/devicemanager"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 )
 
@@ -160,6 +159,8 @@ type EC2 interface {
 	CreateSnapshotWithContext(ctx aws.Context, input *ec2.CreateSnapshotInput, opts ...request.Option) (*ec2.Snapshot, error)
 	DeleteSnapshotWithContext(ctx aws.Context, input *ec2.DeleteSnapshotInput, opts ...request.Option) (*ec2.DeleteSnapshotOutput, error)
 	DescribeSnapshotsWithContext(ctx aws.Context, input *ec2.DescribeSnapshotsInput, opts ...request.Option) (*ec2.DescribeSnapshotsOutput, error)
+	WaitUntilVolumeAvailableWithContext(ctx aws.Context, input *ec2.DescribeVolumesInput, opts ...request.WaiterOption) error
+	WaitUntilVolumeInUseWithContext(ctx aws.Context, input *ec2.DescribeVolumesInput, opts ...request.WaiterOption) error
 }
 
 type Cloud interface {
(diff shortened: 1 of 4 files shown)


pkg/cloud/cloud.go
```diff
@@ -408,47 +409,17 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
...
-				return true, nil
-			}
-		}
-		return false, nil
+	describeVolumesInput := &ec2.DescribeVolumesInput{
+		VolumeIds: []*string{aws.String(volumeID)},
+	}
+	switch state {
+	case "attached":
+		return c.ec2.WaitUntilVolumeInUseWithContext(ctx, describeVolumesInput)
```
leakingtapan: By double looking at the method's implementation, it is checking for if `volumes[].State` is `InUse` state, and it is different to our current check on `Volume.Attachment[].State`. Is it okay?

pkg/cloud/cloud.go
```diff
@@ -408,47 +409,17 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
...
+	case "detached":
+		return c.ec2.WaitUntilVolumeAvailableWithContext(ctx, describeVolumesInput)
```
leakingtapan: Same as here

pkg/cloud/cloud.go
zacharya: Yes this is a great point.  I think it comes down to whether the `attaching` attachment state would return as `InUse` just as `attached` does and whether that matters to us.  I'll try and test via EC2.

zacharya: I confirmed that while a volume is in the `attaching` state, that its state is reflected as `InUse` (see below).  This might be a problem for `AttachDisk` in that the native `WaitUntilVolumeInUseWithContext` function could return success before a problem happens during attachment.  I think we will want to stay with the custom check on `Volume.Attachment[].State` for the attachments.  For `CreateDisk` and `DeleteDisk` I think we can use the AWS native functions.  What do you think?

```
{
    "Volumes": [
        {
            "AvailabilityZone": "xxx",
            "Attachments": [
                {
                    "AttachTime": "2019-05-23T21:53:04.000Z",
                    "InstanceId": "xxx",
                    "VolumeId": "xxx",
                    "State": "attaching",
                    "DeleteOnTermination": true,
                    "Device": "/dev/xvda"
                }
            ],
            "Encrypted": false,
            "VolumeType": "gp2",
            "VolumeId": "xxx",
            "State": "in-use",
...
```

leakingtapan: sounds good to me. Let's just do this for `CreateVolume` then, we probably don't need the one for `DeleteVolume` since we are not waiting for it even now.

zacharya:
@leakingtapan sorry this one totally fell of my radar. Is this still needed? I’m happy to pick this back up and finish within a couple days. If not no worries and apologies.

leakingtapan:
No worries. I'm de-prioritizing this task for 0.4 release. we can pick this up later.

Issue #99, opened 2018-11-12
Create Helm chart for driver


leakingtapan:
Currently, user has to rely on the sample manifest files under deploy/kubernetes to deploy EBS CSI driver.  

As a user who is looking for production ready EBS CSI driver to deploy on kubernetes, he shouldn't need to think about editing manifest file or asking best practices of editing it. We could leverage helmchart to make this a lot easier for user.  

https://hub.helm.sh/

leakingtapan:
Punt to next cycle given we are approaching code freeze and configuration option requirements for helm chat is not yet clear.

PR #303, opened 2019-06-05
Add helm chart

leakingtapan:
Fixes: #99


leakingtapan:
@bertinatto good point, I don't have a good answer for this as well. created a new issue to track this.

Issue #309, opened 2019-06-20
snapshot not found when using ListSnapshot 


leakingtapan:
/kind bug

**What happened?**

> I'm trying to recreate the VolumeSnapshotContents and VolumeSnapshot objects, so that I can create a PVC using them as a datasource. Both the VolumeSnapshotContents and VolumeSnapshot objects are successfully created and linked in Kubernetes.

> However, when I do `k get volumesnapshot -o yaml -n demo`, I get the following message in the status: `'Failed to check and update snapshot: failed to list snapshot content snapcontent-81551962-92b6-11e9-bb60-02c4fcac63ba: "can not find snapshot for snapshotID snap-0cb40b2e0dd1ab464"'`

```
k logs statefulset/ebs-csi-controller --container ebs-plugin -n kube-system
I0619 19:38:59.753261       1 controller.go:389] ListSnapshots: called with args snapshot_id:"snap-0cb40b2e0dd1ab464"
I0619 19:38:59.818175       1 controller.go:397] ListSnapshots: snapshot not found, returning with success
```

If I look in the AWS CLI:
```
% aws ec2 describe-snapshots --snapshot-ids snap-0cb40b2e0dd1ab464

{
    "Snapshots": [
        {
            "Description": "Created by AWS EBS CSI driver for volume vol-0fe35ebd485acaab3",
            "Tags": [
                {
                    "Value": "snapshot-81551962-92b6-11e9-bb60-02c4fcac63ba",
                    "Key": "CSIVolumeSnapshotName"
                }
            ],
            "Encrypted": false,
            "VolumeId": "vol-0fe35ebd485acaab3",
            "State": "completed",
            "VolumeSize": 4,
            "StartTime": "2019-06-19T17:20:16.511Z",
            "Progress": "100%",
            "OwnerId": "285010428092",
            "SnapshotId": "snap-0cb40b2e0dd1ab464"
        }
    ]
}
```

Using the tag, I see it too:

```
% aws ec2 describe-snapshots --filter "Name=tag:CSIVolumeSnapshotName,Values=snapshot-81551962-92b6-11e9-bb60-02c4fcac63ba"

{
    "Snapshots": [
        {
            "Description": "Created by AWS EBS CSI driver for volume vol-0fe35ebd485acaab3",
            "Tags": [
                {
                    "Value": "snapshot-81551962-92b6-11e9-bb60-02c4fcac63ba",
[...]

PR #311, opened 2019-06-27
Fix ListSnapshot bug to use GetSnapshotById

leakingtapan:
Fixes: #309

### DIFF
--- a/examples/kubernetes/snapshot/specs/snapshot-import/volume-snapshot-content.yaml
+++ b/examples/kubernetes/snapshot/specs/snapshot-import/volume-snapshot-content.yaml
@@ -1,0 +1,12 @@
+apiVersion: snapshot.storage.k8s.io/v1alpha1
+kind: VolumeSnapshotContent
+metadata:
+  name: static-snapshot-content
+spec:
+  csiVolumeSnapshotSource:
+    driver: ebs.csi.aws.com 
+    snapshotHandle: snap-0fba4d7649d765c50
+  volumeSnapshotRef:
+    kind: VolumeSnapshot
+    name: static-snapshot-demo
+    namespace: default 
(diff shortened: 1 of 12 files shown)


examples/kubernetes/snapshot/specs/snapshot-import/volume-snapshot-content.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: snapshot.storage.k8s.io/v1alpha1
+kind: VolumeSnapshotContent
+metadata:
+  name: static-snapshot-content
+spec:
+  csiVolumeSnapshotSource:
+    driver: ebs.csi.aws.com 
+    snapshotHandle: snap-0fba4d7649d765c50
```
bertinatto: Do you want to leave this handle here just as an example? Or would you rather leave it empty?

tests/sanity/fake_cloud_provider.go
```diff
@@ -187,6 +187,15 @@ func (c *fakeCloudProvider) GetSnapshotByName(ctx context.Context, name string)
...
 
+func (c *fakeCloudProvider) GetSnapshotById(ctx context.Context, snapshotID string) (snapshot *cloud.Snapshot, err error) {
+	ret, exists := c.snapshots[snapshotID]
+	if !exists {
+		return nil, cloud.ErrNotFound
+	}
+
+	return ret.Snapshot, nil
+}
+
```
bertinatto: It looks like you also have to change the `GetSnapshotByName` above, since it's comparing the the snapshot name with the ID.

Also, we should probably not use the snapshot ID as name in line 148 above.

examples/kubernetes/snapshot/specs/snapshot-import/volume-snapshot-content.yaml
leakingtapan: I would prefer to leave the handle as an example so that people know the format of it in case they put in some other string (maybe snapshot name) and it fails to work.

tests/sanity/fake_cloud_provider.go
leakingtapan: good catch. updated

PR #322, opened 2019-07-25
Update manifest for Kubernetes 1.15

jsafrane:
`deploy/kubernetes` does not work since Kubernetes 1.14:

* The cluster-driver-registrar sidecar container was not released in 1.14 and the version from 1.13 does not work. It is now formally deprecated, so create `CSIDriver` instance during installation.
* Bumped image versions to 1.15.
* Fixed also couple of issues with the yaml file along the way.

### DIFF
--- a/deploy/kubernetes/manifest.yaml
+++ b/deploy/kubernetes/manifest.yaml
@@ -84,13 +84,16 @@ roleRef:
 ---
 
 kind: StatefulSet
-apiVersion: apps/v1beta1
+apiVersion: apps/v1
 metadata:
   name: ebs-csi-controller
   namespace: kube-system
 spec:
   serviceName: ebs-csi-controller
   replicas: 1
+  selector:
+    matchLabels:
+      app: ebs-csi-controller
   template:
     metadata:
       labels:
@@ -139,7 +142,7 @@ spec:
             periodSeconds: 10
             failureThreshold: 5
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.1.0
+          image: quay.io/k8scsi/csi-provisioner:v1.3.0
           args:
             - --provisioner=ebs.csi.aws.com
             - --csi-address=$(ADDRESS)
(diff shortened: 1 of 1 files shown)


leakingtapan:
> deploy/kubernetes does not work since Kubernetes 1.14 

Could you clarify on what is not working on 1.14? I have e2e tests testing against 1.14, and don't find any obvious issue.

deploy/kubernetes/manifest.yaml
```diff
@@ -193,7 +155,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-attacher
-          image: quay.io/k8scsi/csi-attacher:v1.1.0
+          image: quay.io/k8scsi/csi-attacher:v1.2.0
```
leakingtapan: same as above question

jsafrane:
> Could you clarify on what is not working on 1.14

As I wrote above, cluster-driver-registrar sidecar container was not released in 1.14 and the version from 1.13 does not work. It most probably does not create CSIDriver instance.

deploy/kubernetes/manifest.yaml
```diff
@@ -167,20 +141,8 @@ spec:
...
-            - --v=5
-          env:
-            - name: ADDRESS
-              value: /csi/csi.sock
-          volumeMounts:
-            - name: socket-dir
-              mountPath: /csi
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.1.0
+          image: quay.io/k8scsi/csi-provisioner:v1.3.0
```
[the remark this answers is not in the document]
jsafrane: Kubernetes docs claim that 1.3.0 works with 1.14
https://kubernetes-csi.github.io/docs/external-provisioner.html

deploy/kubernetes/manifest.yaml
jsafrane: https://kubernetes-csi.github.io/docs/external-attacher.html

PR #325, opened 2019-07-31
Remove the deprecated connection-timeout flag for liveness probe

leakingtapan:
**Is this a bug fix or adding new feature?**
The connection-timeout is deprecated: https://github.com/kubernetes-csi/livenessprobe/blob/v1.1.0/cmd/livenessprobe/main.go#L41

PR #327, opened 2019-08-06
Update helm manifest for 1.15

wongma7:

**What is this PR about? / Why do we need it?** keep the helm manifest up to date with the deploy/kubernetes manifest updated in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/322

**What testing is done?** ran helm install like
helm install --name aws-ebs-csi-driver \
    --set enableVolumeScheduling=true \
    --set enableVolumeResizing=true \
    --set enableVolumeSnapshot=true \
    --set image.repository=$IMAGE_NAME \
    --set image.tag=$IMAGE_TAG \
    ./aws-ebs-csi-driver

and all containers go Running without issue.

### DIFF
--- a/aws-ebs-csi-driver/templates/manifest.yaml
+++ b/aws-ebs-csi-driver/templates/manifest.yaml
@@ -81,35 +81,6 @@ roleRef:
   name: ebs-external-attacher-role
   apiGroup: rbac.authorization.k8s.io
 
----
-
-kind: ClusterRole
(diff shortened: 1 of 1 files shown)


leakingtapan:
Thanks for the change

leakingtapan:
The verify/sanity/unit failures are being fixed at: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/325 e2e tests seems are having similar issue too.

Issue #324, opened 2019-07-29
EBS CSI Driver doesn't respect fstype from existing storage


shanesiebken:
/kind bug

**What happened?**
The AWS EBS CSI Driver fails to mount pre-existing (in-tree) storage with CSIMigration / CSIMigrationAWS feature flags where the fstype on the storage is different than ext4 (the driver's default).

**What you expected to happen?**
The AWS CSI Driver should successfully read the in-tree storage fstype and mount the volume correctly.

**How to reproduce it (as minimally and precisely as possible)?**

1. Create a persistent volume with the in-tree aws cloud-provider with a volume fstype other than "ext4".
2. Install the EBS csi driver, and enable CSIMigration and CSIMigrationAWS feature gates.
3. Attempt to attach the existing storage to a pod.

**Anything else we need to know?**:
Looking at the requests coming from the storage shim (CSIMigration code), it looks like the fstype is referenced in a different part of the request than "attributes", which is where node.go looks for it at [line 175](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L175). Another case for shimmed storage could likely be added here.

**Environment**
- Kubernetes version (use `kubectl version`): 1.15.0
- Driver version: commit sha:2aed4b5de3976c8e551b492137c58be642e05897

leakingtapan:
Another potential issue is [NodePublishVolume](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L443) is not taking the fstype as well

shanesiebken:
@leakingtapan I think addressing [this issue](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/326) would solve this issue as well. I believe the fstype is passed in the volume capability from the Migration shim.

Issue #326, opened 2019-08-06
Driver does not format volumes as xfs


jsafrane:
/kind bug

**What happened?**
This storage class should provision XFS volumes:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2019-08-06T14:27:30Z"
  name: e2e-volume-4295-ebs.csi.aws.com-sc
  resourceVersion: "20116"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/e2e-volume-4295-ebs.csi.aws.com-sc
  uid: 529e3b50-b856-11e9-9486-06385e99b400
parameters:
  csi.storage.k8s.io/fstype: xfs
provisioner: ebs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

It provisions this PV (truncated):
```
  spec:
    csi:
      driver: ebs.csi.aws.com
      fsType: xfs
      volumeAttributes:
        fstype: ""
        storage.kubernetes.io/csiProvisionerIdentity: 1565101200642-8081-ebs.csi.aws.com
      volumeHandle: vol-censored
```

Such PV is then formatted as ext4.

**What you expected to happen?**
The PV is formatted as xfs.

**Anything else we need to know?**:
There are actually two issues:

- CreateVolume call should check `csi.storage.k8s.io/fstype` parameter. I am not sure it's actually passed to the CSI driver, maybe it's passed via `VolumeOptions`.

- NodeStage should use the fsType passed in `NodeStageVolumeRequest.volume_capability.MountVolume.fs_type` instead of the one in volume context. The driver does not need to carry the filesystem type in the volume context at all. 

**Environment**
- Kubernetes version (use `kubectl version`): 1.14
- Driver version: `amazon/aws-ebs-csi-driver:latest` (as of today)

leakingtapan:
/assign

Yep. Confirming that the driver doesn't respect `csi.storage.k8s.io/fstype` yet. This is something we need to fix

leakingtapan:
/close

As the PR is merged

PR #335, opened 2019-08-11
Fix the bug by passing fstype correctly

leakingtapan:
Fixes: #324 #326 

/cc @wongma7 @jsafrane @shanesiebken 

This fixes a bug to use `fstype` inside volume capacity for NodeStageVolume and NodePublishVolume call. So that pre-created persistence volume's fstype is used when mounting the volume correctly.

This change also removed the usage of `fstype` in driver's storage class parameter in favor of reserved `csi.storage.k8s.io/fstype` parameter.

**What testing is done?** 
Tested on Kops cluster with a pre-created xfs EBS volume and consume it with `fstype` set in the PV spec.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -103,7 +103,6 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	var (
-		fsType      string
 		volumeType  string
 		iopsPerGB   int
 		isEncrypted bool
@@ -112,8 +111,8 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	for key, value := range req.GetParameters() {
 		switch strings.ToLower(key) {
-		case FsTypeKey:
-			fsType = value
+		case "fstype":
+			klog.Warning("\"fstype\" is deprecated, please use \"csi.storage.k8s.io/fstype\" instead")
 		case VolumeTypeKey:
 			volumeType = value
 		case IopsPerGBKey:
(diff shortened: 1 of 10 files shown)


pkg/driver/node.go
```diff
@@ -117,6 +117,8 @@ func (d *nodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
+	fsType := volCap.GetMount().GetFsType()
```
jsafrane: Please check for nil from `GetMount()` and empty string from `GetFsType()`

leakingtapan: done

wongma7:
something wrong with coveralls https://status.coveralls.io/ we use in travis

leakingtapan:
I notice removing the `fstype` in the driver could cause `CreateVolume` to fail since this key becomes unknown to the driver. Adding the key back with deprecation message.

pkg/driver/controller.go
```diff
@@ -488,9 +485,7 @@ func newCreateVolumeResponse(disk *cloud.Disk) *csi.CreateVolumeResponse {
 		Volume: &csi.Volume{
 			VolumeId:      disk.VolumeID,
 			CapacityBytes: util.GiBToBytes(disk.CapacityGiB),
-			VolumeContext: map[string]string{
-				FsTypeKey: disk.FsType,
-			},
+			VolumeContext: map[string]string{},
```
wongma7: OK: external-provisioner will fill in volume.spec.csi.fstype for us

wongma7:
/lgtm

re-adding

leakingtapan:
merged manually since the coverage drop is due to the deprecation message added

Issue #192, opened 2019-01-28
Report test result as independent test case in Testgrid for E2E tests


leakingtapan:
**Is your feature request related to a problem? Please describe.**
Our e2e tests result are reported under [testgrid](https://testgrid.k8s.io/sig-aws-ebs-csi-driver#e2e-test-single-az). However, it only reports the aggregated overall result. We should make report each individual test result for each test case in testgrid.

See [gce pd](https://testgrid.k8s.io/sig-gcp-compute-persistent-disk-csi-driver#Kubernetes%20Master%20Driver%20Stable) as an example.

**Describe the solution you'd like in detail**
TBD

PR #340, opened 2019-08-12
Output junit to ARTIFACTS for testgrid

wongma7:

**What is this PR about? / Why do we need it?** test-grid should have tests separated by testcase instead of just Overall

fix https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/192

**What testing is done?** wait for the test to complete for this PR and check test-grid

### DIFF
--- a/hack/run-e2e-test
+++ b/hack/run-e2e-test
@@ -98,7 +98,7 @@ helm install --name aws-ebs-csi-driver \
 # Run the test
 go get -u github.com/onsi/ginkgo/ginkgo
 export KUBECONFIG=$HOME/.kube/config
-ginkgo -p -nodes=$NODES -v --focus="$FOCUS" tests/e2e
+ginkgo -p -nodes=$NODES -v --focus="$FOCUS" tests/e2e -- -report-dir=$ARTIFACTS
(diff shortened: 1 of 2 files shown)


wongma7:
doesn't seem to work. i think ginkgo is eating the report-dir arg? In EFS, I don't call ginkgo I call go test.

leakingtapan:
>doesn't seem to work. i think ginkgo is eating the report-dir arg? In EFS, I don't call ginkgo I call go test.

The `report-dir` flag is bring in as flag by `framework.HandleFlags()`

wongma7:
Got it, need to put non-ginkgo args after --. (We have framework.handleflags already.)
```
To pass arguments/custom flags down to your test suite:

$ ginkgo -- <PASS-THROUGHS>

Note: the -- is important! Only arguments following -- will be passed to your suite. To parse arguments/custom flags in your test suite, declare a variable and initialize it at the package-level:
```
https://onsi.github.io/ginkgo/#running-tests

leakingtapan:
override the failure since [coverall](https://status.coveralls.io/) is having issue. And this only impact out test coverage

PR #345, opened 2019-08-15
Update to use kubernetes 1.15 as dependency

leakingtapan:
Since k8s 1.15+ uses go mod for vendering dependencies, it causes problem for dependencies using `replace` directive.

Add update-gomod.sh to update go.mod when consuming k/k as go module

Ref: https://github.com/kubernetes/kubernetes/issues/79384


wongma7:
Nice! TBH I kind of hate that we have to do this, I almost want to stay with dep, but dep is super slow. Anyway it's k/k's fault. Will go mod commands like tidy and vendor still work?

leakingtapan:
> Will go mod commands like tidy and vendor still work?

go mod vendor still works that's how I get to test this change locally. go mod tidy seems to still works. But I haven't tested it enough

PR #359, opened 2019-08-29
Use new test framework for test orchestration

leakingtapan:
Implement test framework to test orchestration. This code will be ported to https://github.com/aws/aws-k8s-tester. Testing before the port. 

This is an effort to deprecate run-e2e-test.sh

TODOs:
* ~handle ARTIFACTS~

/cc @wongma7


tester/test-config.yaml
```diff
@@ -0,0 +1,79 @@
+cluster:
+  kops:
+    zones: us-west-2a,us-west-2b,us-west-2c
+    nodeCount: 3
+    nodeSize: c5.large
+    kuberneterVersion: 1.14
```
wongma7: kuberneterVersion -> kubernetesVersion

tester/pkg/cluster.go
```diff
@@ -0,0 +1,35 @@
+package pkg
```
wongma7: not a fan of package name pkg...should refactor into subdirs of /pkg/ at some point

tester/pkg/config.go
```diff
@@ -0,0 +1,45 @@
...
+	IamPolicies       string `yaml:"iamPolicies"`
+	//KubeApiServer     KubeApiServer `yaml:"kubeApiServer"`
+	//Kubelet           Kubelet       `yaml:"kubelet"`
+}
+
+type KubeApiServer struct {
+	FeatureGates FeatureGates `yaml:"featureGates"`
+}
+
+type FeatureGates struct {
```
wongma7: prefer to keep it map[string]string (same as kops config https://github.com/kubernetes/kops/blob/b0860f7738f8f1c38e288910494ef6382e4dd7fc/pkg/apis/kops/v1alpha1/componentconfig.go#L410) otherwise we'll be chasing every feature

In the future we could use the kops API directly, rather than duplicating its fields here...something like:
```
// kops.ClusterSpec is https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/cluster.go#L47
type KopsCluster struct {
    kops.ClusterSpec spec `yaml:"clusterSpec"` 
}
```
Then run `kops create cluster --config=FILE_FOR_KopsCluster.spec`

As opposed to creating kops cluster, getting the spec file, then appending to it.

But it's probably not worth the trouble to figure it out now.

tester/pkg/kops.go
```diff
@@ -0,0 +1,258 @@
...
+		// create cluster
+		err := c.createCluster()
+		if err != nil {
+			return err
+		}
+
+		c.waitForCreation()
+
+		// wait for cluster creation to success
+		// or return err if failed
```
wongma7: TODO to have a timeout from waitForCreation

tester/pkg/steps.go
```diff
@@ -0,0 +1,48 @@
...
+	testCmd.Stdin = os.Stdin
+	testCmd.Stderr = os.Stderr
+	err := testCmd.Run()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getAwsAccountId(region string) (string, error) {
```
wongma7: what will we use this for

tester/framework.go
```diff
@@ -0,0 +1,105 @@
...
+		tErr := t.tearDown.Run()
+		if tErr != nil {
+			fmt.Printf("failed to tear down cluster: %v", tErr)
+		}
+		return err
+	}
+
+	err = t.test.Run()
+	t.uninstall.Run()
+	t.tearDown.Run()
```
wongma7: TODO capture signal and do tearDown?

tester/pkg/config.go
leakingtapan: Good point on kops.ClusterSpec. We should try it later.
I created a FeatureGates string, which we will append to the cluster spec for now.

tester/pkg/steps.go
leakingtapan: removed

tester/framework.go
leakingtapan: added TODO

tester/pkg/kops.go
leakingtapan: added timeout

tester/pkg/cluster.go
leakingtapan: we can find a better directory to put when moving to tester

tester/migration-test-config.yaml
```diff
@@ -0,0 +1,134 @@
...
+  # TODO known test failures to skip temporarily
+  # - should not allow expansion of pvcs without AllowVolumeExpansion property
+  #   - Test passes but cleanup fails, need https://github.com/kubernetes/kubernetes/pull/81107
+  # - (block volmode) Verify if offline PVC expansion works / should resize volume when PVC is edited while pod is using it
+  #   - NodeExpand for BlockVolumes not well-defined, need more investigation and possibly https://github.com/container-storage-interface/spec/issues/380
+  # - should provision storage with mount options
+  #   - Known bug, need https://github.com/kubernetes/kubernetes/pull/80191 but not yet in a patch release
+  cd ./tests/e2e-migration
+  go test -v -timeout 0 ./... -kubeconfig=$HOME/.kube/config -report-dir=$ARTIFACTS -ginkgo.focus="\[ebs-csi-migration\]" -ginkgo.skip="\[Disruptive\]\
+    |suld.not.allow.expansion\
```
wongma7: somehow the skip string got messed up

leakingtapan: fixed.

PR #375, opened 2019-09-16
Return success if instance or volume are not found

### LINKED ISSUE #330: ControllerUnpublish should return no error on success.
/kind bug

**What happened?**
The driver returns `codes.Internal` when detaching a volume from deleted node or detaching a deleted volume (deleted outside of the CSI driver). As AWS detaches volumes when a node is deleted and deleted volume must have been detached before, returning success is more appropriate.

When `codes.Internal` is returned, CO will retry again, introducing endless loop between CO and the driver.

See https://github.com/container-storage-interface/spec/issues/373 for discussion.

bertinatto:

/kind bug

From the CSI specs;

```
If the volume corresponding to the volume_id or the node corresponding to node_id cannot be found by the Plugin and the volume can be safely regarded as ControllerUnpublished from the node, the plugin SHOULD return 0 OK.
```

Fixes #330

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -401,6 +401,11 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
 
 	_, err = c.ec2.DetachVolumeWithContext(ctx, request)
 	if err != nil {
+		if isAWSErrorIncorrectState(err) ||
+			isAWSErrorInvalidAttachmentNotFound(err) ||
+			isAWSErrorVolumeNotFound(err) {
+			return ErrNotFound
+		}
 		return fmt.Errorf("could not detach volume %q from node %q: %v", volumeID, nodeID, err)
 	}
 
@@ -685,6 +690,9 @@ func (c *cloud) getInstance(ctx context.Context, nodeID string) (*ec2.Instance, 
 	for {
 		response, err := c.ec2.DescribeInstancesWithContext(ctx, request)
 		if err != nil {
+			if isAWSErrorInstanceNotFound(err) {
+				return nil, ErrNotFound
+			}
 			return nil, fmt.Errorf("error listing AWS instances: %q", err)
 		}
 
@@ -804,11 +812,32 @@ func isAWSErrorIncorrectModification(err error) bool {
 	return isAWSError(err, "IncorrectModificationState")
 }
 
+// isAWSErrorInstanceNotFound returns a boolean indicating whether the
+// given error is an AWS InvalidInstanceID.NotFound error. This error is
+// reported when the specified instance doesn't exist.
+func isAWSErrorInstanceNotFound(err error) bool {
+	return isAWSError(err, "InvalidInstanceID.NotFound")
+}
+
 // isAWSErrorVolumeNotFound returns a boolean indicating whether the
 // given error is an AWS InvalidVolume.NotFound error. This error is
 // reported when the specified volume doesn't exist.
 func isAWSErrorVolumeNotFound(err error) bool {
 	return isAWSError(err, "InvalidVolume.NotFound")
+}
+
+// isAWSErrorIncorrectState returns a boolean indicating whether the
+// given error is an AWS IncorrectState error. This error is
+// reported when the resource is not in a correct state for the request.
+func isAWSErrorIncorrectState(err error) bool {
+	return isAWSError(err, "IncorrectState")
+}
+
+// isAWSErrorInvalidAttachmentNotFound returns a boolean indicating whether the
+// given error is an AWS InvalidAttachment.NotFound error. This error is reported
+// when attempting to detach a volume from an instance to which it is not attached.
+func isAWSErrorInvalidAttachmentNotFound(err error) bool {
+	return isAWSError(err, "InvalidAttachment.NotFound")
 }
 
 // isAWSErrorSnapshotNotFound returns a boolean indicating whether the
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -265,6 +265,9 @@ func (d *controllerService) ControllerUnpublishVolume(ctx context.Context, req *
 	}
 
 	if err := d.cloud.DetachDisk(ctx, volumeID, nodeID); err != nil {
+		if err == cloud.ErrNotFound {
+			return &csi.ControllerUnpublishVolumeResponse{}, nil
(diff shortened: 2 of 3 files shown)


pkg/cloud/cloud.go
```diff
@@ -405,6 +405,9 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
 
 	_, err = c.ec2.DetachVolumeWithContext(ctx, request)
 	if err != nil {
+		if isAWSErrorVolumeNotFound(err) {
```
leakingtapan: Is the `ErrNotFound` returned from `c.getInstance(ctx, nodeID)`enough to cover the issue reported by @jsafrane ? Why do we still need this check?

Also, regarding driver is returning internal error, should we check for `IncorrectState` when detaching an volume that is already detached, and return success? It will get following error in this case:
```
An error occurred (IncorrectState) when calling the DetachVolume operation: Volume 'vol-0c1cd2b52306edf24'is in the 'available' state.
```

AtkinsChang:
@leakingtapan 
any progress on this pr?
This may address the volume state bug when node failure.
If aws instant is terminated, the volume state will transit to `available` causing following error:
```
rpc error: code = Internal desc = Could not detach volume "vol-xxxxxxxxxxxxxxxxx" from node "i-xxxxxxxxxxxxxxxxx": could not detach volume "vol-xxxxxxxxxxxxxxxxx" from node "i-xxxxxxxxxxxxxxxxx": IncorrectState: Volume 'vol-xxxxxxxxxxxxxxxxx ' is in the 'available' state.
status code: 400, request id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

leakingtapan:
@bertinatto could you look at my comment about this fix?

pkg/cloud/cloud.go
leakingtapan: > That's a good point. Are you sure we get a IncorrectState in this case? Looking at the docs, it seems like we get a InvalidAttachment.NotFound instead.

I tested using awscli as follows:
```
aws ec2 detach-volume --volume-id vol-0d63967ab4999f4b6 --region us-east-1 --instanc
e-id i-03376857fb7981add

An error occurred (IncorrectState) when calling the DetachVolume operation: Volume 'vol-0d63967ab4999f4b6'is in the 'available' state.
```
And similar result with a simple go program:
```
func main() {
	c, _ := cloud.NewCloud("us-east-1")
	err := c.DetachDisk(context.Background(), "vol-0d63967ab4999f4b6", "i-03376857fb7981add")

	fmt.Println(err)
}
```

Although I don't know when `InvalidAttachment.NotFound` could happen, how about we check for both?

leakingtapan: @jsafrane how is the check done in cloud provider now?

bertinatto: > Although I don't know when InvalidAttachment.NotFound could happen, how about we check for both?

I think that makes sense. I changed the code to reflect that.

From the docs:

```
InvalidAttachment.NotFound | Indicates an attempt to detach a volume from an instance to which it is not attached.
```

Perhaps `InvalidAttachment.NotFound` is returned when the volume is attached to another instance?

bertinatto: > @jsafrane how is the check done in cloud provider now?

Apparently it doesn't special-case any error; it just returns it:

https://github.com/kubernetes/kubernetes/blob/5bac42bff9bfb9dfe0f2ea40f1c80cac47fc12b2/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L2392-L2395

That makes sense because the in-tree plugin doesn't any requirement to be idempotent.

leakingtapan: > Perhaps InvalidAttachment.NotFound is returned when the volume is attached to another instance?
Yep, just tested and confirmed you are right 👍

leakingtapan: > > @jsafrane how is the check done in cloud provider now?
> 
> Apparently it doesn't special-case any error; it just returns it:
> 
> https://github.com/kubernetes/kubernetes/blob/5bac42bff9bfb9dfe0f2ea40f1c80cac47fc12b2/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L2392-L2395
> 
> That makes sense because the in-tree plugin doesn't any requirement to be idempotent.

I'm thinking how about we move the logic of:

```
if isAWSErrorIncorrectState(err) ||
			isAWSErrorInvalidAttachmentNotFound(err) ||
			isAWSErrorVolumeNotFound(err) {
			return ErrNotFound
		}
```
To controller.go `ControllerUnpublishVolume` ?
This has several advantages that: 1) keep the cloud provider simpler and cleaner 2) align with current cloud provider implementation so that we can unify the two cloud providers (legacy cloud provider and the one used in CSI driver) easier See @jsafrane 's suggestion [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/393) 3) keep the CSI specific error handling in CSI controller

bertinatto: > I'm thinking how about we move the logic of:
> 
> ```
> if isAWSErrorIncorrectState(err) ||
> 			isAWSErrorInvalidAttachmentNotFound(err) ||
> 			isAWSErrorVolumeNotFound(err) {
> 			return ErrNotFound
> 		}
> ```
> 
> To controller.go `ControllerUnpublishVolume` ?
> This has several advantages that: 1) keep the cloud provider simpler and cleaner 2) align with current cloud provider implementation so that we can unify the two cloud providers (legacy cloud provider and the one used in CSI driver) easier See @jsafrane 's suggestion [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/393) 3) keep the CSI specific error handling in CSI controller

A drawback would be that the controller will have to understand AWS-specific error codes.

As an alternative, we could implement a helper function in the cloud package (be it the legacy one or this one) to evaluate whether the error means "OK" in an idempotent context. This method would then be used in the CSI driver only, not in the in-tree plugin.

In any case, whatever the solution we choose, we should implement it equally across all the other methods in the cloud package. Unfortunately I don't have the cycles for that at the moment.

leakingtapan: k. I can leave with the current change for now. We will reconsider this when unifying legacy cloud provider [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/393)

PR #386, opened 2019-10-07
Update aws sdk version to enable EKS IAM for SA

leakingtapan:
Update aws sdk version to enable EKS IAM for SA


deploy/kubernetes/overlays/dev/kustomization.yaml
```diff
@@ -5,3 +5,4 @@ bases:
 images:
 - name: amazon/aws-ebs-csi-driver
   newTag: latest
+  newName: chengpan/aws-ebs-csi-driver
```
bertinatto: Was this intentional?

leakingtapan: yep. since it is under dev overlay and I feel it could be handy for local testing. Let me know if you think otherwise

PR #388, opened 2019-10-10
Update go mod and mount library

leakingtapan:
Update go mod and mount library


leakingtapan:
Failure is due to external-snapshotter is still using an old version of apimachinery [here](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/client/clientset/versioned/typed/volumesnapshot/v1alpha1/volumesnapshot_client.go#L84). The the `serializer.DirectCodecFactory` is renamed to `WithoutConvenstionCodecFactory` in 1.16.1 k8s release [here](https://github.com/kubernetes/kubernetes/blob/v1.16.1/staging/src/k8s.io/apimachinery/pkg/runtime/serializer/codec_factory.go#L305)

external-snapshotter needs to be updated first

PR #420, opened 2019-11-25
Added tolerations for node support

zerkms:
**Is this a bug fix or adding new feature?** kind of both

At the moment the `tolerations` value is defined yet never used. Without it it's impossible to deploy ebs-csi-driver (the node component) to nodes with taints (which is necessary if you specialise nodes)

**What testing is done?** 
I modified the chart locally and generated using different tolerations, including empty, default and with extra custom tolerations.

### DIFF
--- a/aws-ebs-csi-driver/templates/manifest.yaml
+++ b/aws-ebs-csi-driver/templates/manifest.yaml
@@ -338,13 +338,12 @@ spec:
       hostNetwork: true
       priorityClassName: system-node-critical
       tolerations:
-        - key: CriticalAddonsOnly
-          operator: Exists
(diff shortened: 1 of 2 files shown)


leakingtapan:
Is there any toleration other than `CriticalAddonsOnly` you want to use?

zerkms:
@leakingtapan yep, but those are specific to my cluster: it's only used solely to run different workloads on different machines.

zerkms:
It looks like AWS API has outage at this very moment so I believe the failure caused by it.

/retest

zerkms:
Team, do I need to provide here anything else?

PR #438, opened 2020-01-10
Implement different driver modes and AWS Region override for controller service

### LINKED ISSUE #378: Support create pvc in other regions
/kind feature

**What happened?**
When I create pvc use aws-ebs-csi-driver,if the pvc in Tokyo region always get the error like:
```
could not create volume in EC2: InvalidZone.NotFound: The zone 'ap-northeast-1c' does not exist.
or
could not create volume in EC2: InvalidZone.NotFound: The zone 'ap-northeast-1a' does not exist.
```
The reason for this phenomenon is cause by multiple region.
After multiple tests, I can confirm the aws-ebs-csi-driver can only create pvc in the same region with controller.When the region of pvc and controller is different,the ebs-csi-node can report it's own zone to controller,but the controller can't recognize the zone in other region.Then we get the error:```The zone does not exist```.

Can create cross-regional pvc in kubernetes cluster on aws machine by aws-ebs-csi-driver.
Create a cross-regional kubernetes cluster,and create a pvc in a different region with ebs-csi-controller pod.

**Environment**
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:54Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:05:50Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}

- Driver version:
latest alpha

rfranzke:
Feature

Fixes #378 

This PR allows to only run the controller service of the driver (and not the node service). Also, it allows to configure the AWS region instead of always looking it up via the AWS EC2 metadata service.

This enables the use-case of running the CSI controllers (csi-provisioner, csi-attacher, etc., + the driver controller) separately outside of the cluster, and not necessarily on an AWS EC2 instance.

**What testing is done?** 
I ran the CSI controllers + EBS driver separately outside of the Kubernetes cluster (with the newly introduced flags of this PR). Also, I ran the EBS driver daemon set inside the cluster (together with the registrar) without the flags (i.e., with standard mode). I was successfully able to provision and deprovision EBS volumes.

### DIFF
--- a/aws-ebs-csi-driver/templates/manifest.yaml
+++ b/aws-ebs-csi-driver/templates/manifest.yaml
@@ -222,10 +222,11 @@ spec:
         - name: ebs-plugin
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           args:
+            - controller
             - --endpoint=$(CSI_ENDPOINT)
+            {{ include "aws-ebs-csi-driver.extra-volume-tags" . }}
             - --logtostderr
             - --v=5
-            {{ include "aws-ebs-csi-driver.extra-volume-tags" . }}
           env:
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
@@ -241,6 +242,10 @@ spec:
                   name: aws-secret
                   key: access_key
                   optional: true
+            {{- if .Values.region }}
+            - name: AWS_REGION
+              value: {{ .Values.region }}
+            {{- end }}
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
@@ -353,6 +358,7 @@ spec:
             privileged: true
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           args:
+            - node
             - --endpoint=$(CSI_ENDPOINT)
             - --logtostderr
             - --v=5
--- a/aws-ebs-csi-driver/values.yaml
+++ b/aws-ebs-csi-driver/values.yaml
@@ -68,3 +68,9 @@ affinity: {}
 #   key1: value1
 #   key2: value2
 extraVolumeTags: {}
+
+# AWS region to use. If not specified then the region will be looked up via the AWS EC2 metadata
+# service.
+# ---
+# region: us-east-1
+region: ""
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -1,0 +1,110 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/cmd/options"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver"
+
+	"k8s.io/klog"
+)
+
+// Options is the combined set of options for all operating modes.
+type Options struct {
+	DriverMode driver.Mode
+
+	*options.ServerOptions
+	*options.ControllerOptions
+	*options.NodeOptions
+}
+
+// used for testing
+var osExit = os.Exit
+
+// GetOptions parses the command line options and returns a struct that contains
+// the parsed options.
+func GetOptions(fs *flag.FlagSet) *Options {
+	var (
+		version = fs.Bool("version", false, "Print the version and exit.")
+
+		args = os.Args[1:]
+		mode = driver.AllMode
+
+		serverOptions     = options.ServerOptions{}
+		controllerOptions = options.ControllerOptions{}
+		nodeOptions       = options.NodeOptions{}
+	)
+
+	serverOptions.AddFlags(fs)
+	klog.InitFlags(fs)
+
+	if len(os.Args) > 1 {
+		cmd := os.Args[1]
+
+		switch {
+		case cmd == string(driver.ControllerMode):
+			controllerOptions.AddFlags(fs)
+			args = os.Args[2:]
+			mode = driver.ControllerMode
+
+		case cmd == string(driver.NodeMode):
+			nodeOptions.AddFlags(fs)
+			args = os.Args[2:]
+			mode = driver.NodeMode
+
+		case cmd == string(driver.AllMode):
+			controllerOptions.AddFlags(fs)
+			nodeOptions.AddFlags(fs)
+			args = os.Args[2:]
+
+		case strings.HasPrefix(cmd, "-"):
+			controllerOptions.AddFlags(fs)
+			nodeOptions.AddFlags(fs)
+			args = os.Args[1:]
+
+		default:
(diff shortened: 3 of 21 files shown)


leakingtapan:
@rfranzke This is a nice features! I like the way this PR is heading towards. Several comments:
1. Regarding adding the `--region` flag, IIUC, this is similar to what's asked 
 in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/378.  I think a better way to handle this is letting controller service to figure out the region through volume topology instead of either passing the region as cli flag or reading from IMDS. The way it will works is during provisioning in the `CreateVolume` method, controller service will read the region from the zone [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L149), and create a new cloud client using the region, the volume will be created with the per region client.
2. Regarding the `--run-as-controller` flag, there are three modes to run the driver 1) controller service only 2) node service only and 3) both (the current mode). I would like a more expressive way to define the three modes than `--run-as-controller`. One way I'm thinking is we use subcommand to describe what service to run. To run just the controller service, we do:
```/bin/aws-ebs-csi-driver controller```. To run just the node service, we do: ```/bin/aws-ebs-csi-driver node```. The benefit is we can cleanly separate the flags that matters to each service independently. Such that `extra-volume-tags` will only be a flag that is valid to the `aws-ebs-csi-driver controller` subcommand. And if no subcommand is provided, we will run the driver as current mode (both controller and service).

**Comment by rfranzke** [comment]
@leakingtapan Thanks for your feedback, I’ve updated the PR based on your suggestions wrt driver mode (`all` vs. `controller` vs. `node`) and implemented the subcommands.

Regarding the first suggestion wrt region: 

> The way it will works is during provisioning in the CreateVolume method, controller service will read the region from the zone [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L149), and create a new cloud client using the region, the volume will be created with the per region client.

I don’t think this is possible, unfortunately. Checking the link you referred, I could certainly move the statement [up and create a new cloud client with the region](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L98), but this only covers the `CreateVolume` case. What if a volume [shall be deleted](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L191-L207)? As far as I can see no region information is provided in the [`DeleteVolumeRequest`](https://github.com/container-storage-interface/spec/blob/HEAD/lib/go/csi/csi.pb.go#L1730-L1741) object but only the volume id. Where would I get the correct region from?

Another concern was that, even if the `DeleteVolume` case is solvable somehow, then we would have to remove the [default cloud client](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L62-L75) as it tries to talk to the metadata service (which will not work if the controller does not run on an AWS EC2 instance). Instead, we would have to check in every function where `d.cloud` is used if it was initialized already, and if not yet then initialize it. (Yes, this is not a problem/argument against it, just something I noticed and wanted to mention).

Though, I have changed it from being a command line parameter to an environment variable (analogous to [this code](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L222-L225) and as `AWS_REGION` is a pretty well-known environment variable for the use-case of overwriting the region).

Do you see an option to get rid of the “static” region overwriting via the env variable? I’m no CSI expert and not too familiar with the code overall, so chances are high that I’m missing something.

Happy to hear your feedback now!

PS: I don’t think the failing tests were related to my changes, were they?

**Comment by leakingtapan** [comment]
> I don’t think this is possible, unfortunately. Checking the link you referred, I could certainly move the statement [up and create a new cloud client with the region](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L98), but this only covers the `CreateVolume` case. What if a volume [shall be deleted](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L191-L207)? As far as I can see no region information is provided in the [`DeleteVolumeRequest`](https://github.com/container-storage-interface/spec/blob/HEAD/lib/go/csi/csi.pb.go#L1730-L1741) object but only the volume id. Where would I get the correct region from?

It's a good point for `DeleteVolume`. In my opinion, the CSI spec should have a way to pass in topology information (eg. zone) into `DeleteVolume` call. So that volume deletion could be handled correctly for volumes across different zones. Symmetrically, this seems necessary as a reverse operation of `CreateVolume`.

After a second thought, I think we don't need to block this by the above issue. Using `AWS_REGION` as an override to the IMDS region is acceptable to me. As we can solve the above problem later (if CSI spec) can support this. In the ideal world, the region resolution ordering will be 1) automatic resolution using topology 2) using AWS_REGION 3) using IMDS. And with this PR, we are adding option 2) which sounds good to me :)

**Comment by leakingtapan** [comment]
Created an issue in CSI: https://github.com/container-storage-interface/spec/issues/408

File: aws-ebs-csi-driver/templates/manifest.yaml [hunk 1]
```diff
@@ -214,10 +214,13 @@ spec:
         - name: ebs-plugin
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           args:
+            {{- if .Values.driverMode }}
+            - {{ .Values.driverMode }}
```
leakingtapan: How about just hard code `controller` here? since it always going to be controller service. And, for the sake of completeness, could you update the daemonset to use the `node` subcommand to always start it as node service?

aws-ebs-csi-driver/values.yaml
```diff
@@ -48,3 +48,17 @@ affinity: {}
...
+# region: us-east-1
+region: ""
+
+# Driver mode. There are three modes:
+# * "all" which is the standard/default mode (starting both the controller and the node service)
+# * "controller" which only starts the controller service
+# * "node" which only starts the node service
+# If no mode is specified then the "all" mode will be used.
+# ---
+# mode: controller
```
leakingtapan: not needed anymore? since we hard code in the above

cmd/options.go
```diff
@@ -0,0 +1,108 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
leakingtapan: add some unit test for this? Also, since we are not running unit test for `./cmd` package yet due to main.go was the only file earlier, as we are adding more file, could you make changes to run unit test for `./cmd` as part of the `make test` target.

pkg/driver/driver.go
```diff
@@ -60,9 +74,19 @@ func NewDriver(options ...func(*DriverOptions)) (*Driver, error) {
 	}
 
 	driver := Driver{
-		controllerService: newControllerService(&driverOptions),
-		nodeService:       newNodeService(),
-		options:           &driverOptions,
+		options: &driverOptions,
+	}
+
+	switch driverOptions.mode {
```
leakingtapan: We probably should selectively register the gRPC service for each mode as well at `driver.Run()`:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/9ca95cea4aaba3b03ae5346520070e629509f748/pkg/driver/driver.go#L94-L96

leakingtapan:
Since coverage dropped, could you add some unit tests?

rfranzke:
@leakingtapan can you have another look, please?

leakingtapan:
@rfranzke I see the coverage is still drop. Could you check where the coverage is missing? Once the coverage is back, you are good to go. Thx

rfranzke:
@leakingtapan please check now :)

PR #441, opened 2020-01-22
Schedule the EBS CSI DaemonSet on all nodes by default

pcfens:
Fixes a bug.

I think that the EBS CSI DaemonSet should schedule on all nodes regardless of taints, similar to what's done in the VPC CNI. Exceptions can still be made by re-rendering the helm chart in aws-ebs-csi-driver.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/420#issuecomment-558409004 suggests the previous version was specific to one installation. By making the default behavior to run everywhere we err on the side of things working.

I'm running these changes on my production cluster without issues.

### DIFF
--- a/aws-ebs-csi-driver/values.yaml
+++ b/aws-ebs-csi-driver/values.yaml
@@ -57,8 +57,7 @@ resources: {}
 nodeSelector: {}
 
 tolerations:
-  - key: CriticalAddonsOnly
-    operator: Exists
+  - operator: Exists
 
 affinity: {}
 
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -19,8 +19,7 @@ spec:
       hostNetwork: true
       priorityClassName: system-node-critical
       tolerations:
-        - key: CriticalAddonsOnly
-          operator: Exists
+        - operator: Exists
       containers:
         - name: ebs-plugin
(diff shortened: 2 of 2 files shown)


deploy/kubernetes/base/controller.yaml
```diff
@@ -19,9 +19,6 @@ spec:
         beta.kubernetes.io/os: linux
       serviceAccount: ebs-csi-controller-sa
       priorityClassName: system-cluster-critical
-      tolerations:
```
leakingtapan: Why changing controller deployment is needed too? I think you are only changing node service daemon set

pcfens: It seems odd to have a specific use case toleration for the controller, so I removed it too. Is there a community preference about controllers running on tainted nodes by default that should be followed here?

leakingtapan: I don't remember the exact reason why this was added. But I believe this is due to similar reason as [here](https://github.com/kubernetes/kubernetes/pull/46748). So that the driver could still be scheduled onto the node which is tainted with `CriticalAddonsOnly`

pcfens: I didn't realize that was upstream convention (thanks for sending that along). I've added toleration back to the controller.

Issue #428, opened 2019-12-11
Driver name can't be changed. 


MartinForReal:
**Is your feature request related to a problem? Please describe.**
I need to run multiple CSI drivers on the same Kubernetes cluster because the worker nodes in my cluster are provisioned under user account (which is different from the account which provisions master node ). So I can't provision new volume using the same CSI driver due to access issues.  

**Describe the solution you'd like in detail**
If the driver name can be configured. Then I can run one CSI driver instance for each user.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/94107a3c633141b7834d022c60c76af5e64eec77/pkg/driver/driver.go#L31

leakingtapan:
@MartinForReal the CSI driver name needs to be unique and stable, it's not supposed to be configured dynamically. And CSI driver controller service deployment is supposed to be one per cluster.

For you use case, is it possible for you to assign the CSI driver a role that can assume role in other account?

MartinForReal:
@leakingtapan This might be reasonable but I can hardly agree. 

> controller service deployment is supposed to be one per cluster.

This might not be true because if I use different storage implementation on the same cluster. There will be more than one CSI controller on the same cluster available. 

And the driver identification is stable after the controller daemon is started. Here we treat each controller as a unique storage provider even though these controllers are using the same implementation. And CSI has support for this scenario.

And we can set a default value to support migration.

leakingtapan:
Different storage implementation comes with different csi drivers. You can see that all aws csi driver has different driver name, efs.csi.aws.com, ebs.csi.aws.com, which enables you to deploy them all in the same cluster. 

> And CSI has support for this scenario.

Where do u see this is documented?

MartinForReal:
@leakingtapan As you said in the last comment. We can have multiple instances of the driver if we can change the name. such as ebs.csi.aws.com.us-east1, ebs.csi.aws.com.cn-west2, ebs.csi.aws.com.cn-west-user1

PR #465, opened 2020-03-20
support add drivername for different account

zanghao2:
* fix:#428. Each account has its own drivername

* In our situation each account has its own work nodes and storage, all the work nodes manged by one cluster. 
* We want each account uses only its own storage
* If all accounts use one aws-ebs-csi-driver, the volume can not attached to right node(We have test this situation).

**What testing is done?**
* I have test on my own cluster.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -30,6 +30,7 @@ func main() {
 
 	drv, err := driver.NewDriver(
 		driver.WithEndpoint(options.ServerOptions.Endpoint),
+		driver.WithDriverName(options.ServerOptions.DriverName),
 		driver.WithExtraVolumeTags(options.ControllerOptions.ExtraVolumeTags),
 		driver.WithMode(options.DriverMode),
 	)
--- a/cmd/options/server_options.go
+++ b/cmd/options/server_options.go
@@ -26,8 +26,12 @@ import (
 type ServerOptions struct {
 	// Endpoint is the endpoint that the driver server should listen on.
 	Endpoint string
+
+	// DriverName for different account
+	DriverName string
 }
 
 func (s *ServerOptions) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&s.Endpoint, "endpoint", driver.DefaultCSIEndpoint, "Endpoint for the CSI driver server")
+	fs.StringVar(&s.DriverName, "drivername", driver.DefaultCSIDriverName, "DriverName for the CSI driver server")
 }
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -41,4 +41,5 @@ const (
 // constants for default command line flag values
 const (
 	DefaultCSIEndpoint = "unix://tmp/csi.sock"
+	DefaultCSIDriverName= "ebs.csi.aws.com"
 )
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -40,8 +40,11 @@ const (
 )
 
 const (
-	DriverName  = "ebs.csi.aws.com"
-	TopologyKey = "topology." + DriverName + "/zone"
+	TopologyKey = "topology." + "ebs.csi.aws.com" + "/zone"
+)
+
+var (
+	DriverName  string
 )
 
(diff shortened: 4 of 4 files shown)


leakingtapan:
Have you considered granting a single EBS csi driver cross account permission so that I can create volumes in multiple accounts? As I mentioned in the issue, this will break csi migration 

/cc @msau42 how do you think?

msau42:
Even though this is not the intended use case, this allows multiple instances of the same driver to run on the same node, which will break some features like volume limits.

This will also break csi migration, but that could be called out as a caveat for doing this.

For this particular use case, why not give each account its own cluster instead of sharing a cluster and trying to duplicate all the system workloads?

zanghao2:
@msau42 @leakingtapan  Thank you for your reply. My scenario is that all account  nodes are managed on a cluster. I don't want to create a cluster for each account, but also want each account to have a aws-ebs-csi-driver separately and I don't know whether it is feasible to grant a single EBS csi driver cross account.

zanghao2:
> Thanks for your pull request. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.**
> 
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.
> 
> * If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address.  Check your existing CLA data and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * If you signed the CLA as a corporation, please sign in with your organization's credentials at https://identity.linuxfoundation.org/projects/cncf to be authorized.
> * If you have done the above and are still having issues with the CLA being reported as unsigned, please log a ticket with the Linux Foundation Helpdesk: https://support.linuxfoundation.org/
> * Should you encounter any issues with the Linux Foundation Helpdesk, send a message to the backup e-mail support address at: [login-issues@jira.linuxfoundation.org](mailto:login-issues@jira.linuxfoundation.org)

msau42:
I'm wondering if other system-level controllers, like the aws cloud controller, will have similar issues. I'm not sure how well all these controllers were designed to have more than one operating in a single cluster.

leakingtapan:
Yep. Feel the same that it’s a problem for all the controller that acts on cloud provider resources.

MartinForReal:
The good news is that we can deploy multiple alb ingress controllers for each account. And cloud provider and PVC CNI plugin will work as long as roles are attached to EC2 instances. 

And we can utilize different types of storage if more than one type of storage provider is defined.

leakingtapan:
> The good news is that we can deploy multiple alb ingress controllers for each account. And cloud provider and PVC CNI plugin will work as long as roles are attached to EC2 instances. As far as I can see, all the other controllers will work in this scenario.

I'm curious how do multiple alb ingress controllers decide on which one to act on the newly created ingress object, for example. Same question applies to CCM where a service object is created, how do multiple CCMs decide on which one to act on the object.

/cc @M00nF1sh

MartinForReal:
In my scenario,  we use the VPC CNI plugin as the pod network provider. And alb Ingress controller will create an lb using ak/sk instead of delegating responsibility to cloud provider。We are running the alb-ingress controller in IP mode which means that alb backend will connect to pod without any DNAT. If alb-ingress-controller is not in IP mode, services exposed by ingress should be the type of NodePort(which will not use cloud provider). And we can use ingress-class to classify ingress object. I agree that LoadBalancer services can only be accessible using lb where the master resides.  But do we really need LoadBalancer service in case that we've already had the alb-ingress-controller? I don't think that exposing middleware directly to outside world is reasonable.

superbrothers:
I also plan to create a cluster consisting of EC2 instances across multiple accounts, which brings me to this issue.

Since the ebs-csi-driver does not support multiple accounts yet, I am considering handling this by deploying a renamed driver for each account.

I'm not sure why this issue is closed, but if the change is acceptable, I will create a pull request.

PR #467, opened 2020-03-20
Add tolerations and taints to Snapshot statefulset, remove ^M

Quarky9:
Add same tolerations and taints to snapshot set as the main controller

Allow deployment of optional control along main via helm chart

**What testing is done?** 
Deployed private test cluster using forked helm chart


aws-ebs-csi-driver/templates/statefulset.yaml
```diff
@@ -1,26 +1,36 @@
...
-      serviceAccount: ebs-snapshot-controller
-      containers:
-        - name: snapshot-controller
-          image: quay.io/k8scsi/snapshot-controller:v2.0.1
-          args:
-            - --v=5
-            - --leader-election=false
-{{- end }}
+{{- if .Values.enableVolumeSnapshot }}
+#Snapshot controller
```
msau42: The snapshot controller should be deployed as part of Kubernetes cluster deployment, not the CSI driver. The snapshot controller handles all operations across all CSI drivers.

Quarky9: Hi,
I understand that the snapshot controller can be used for various CSI implementation within the same cluster, but in terms of placement within the cluster:

How is that different from the main ebs-csi-controller ?
Both are part of the extended control plane of each cluster and we run them on the controller nodes.
Thx for any clarification ;-)

msau42: For example, let's say you deploy all these csi drivers in your cluster: ebs, efs, lustre, ceph, xyz, etc.

Even though you have 1+ csi driver in the cluster, only 1 snapshot-controller should be deployed. Therefore snapshot controller deployment needs to live outside of any one csi driver.

leakingtapan: What @msau42 says makes sense if that how snapshot controller was designed. The situation now is there is no such kubernetes cluster deployment creates the snapshot controller on aws, and we need a place to create the controller

leakingtapan: It will be great if kops can create it after cluster creation./cc @justinsb

msau42: In the meantime, can we at least make the snapshot controller deployment optional with clear documentation that you should only have one deployed per cluster?

msau42: I would also caution against putting the snapshot CRD definitions in any sort of automated driver deployment method. Deletion of a CRD means all the objects also get deleted, which means all the snapshots get deleted.

Issue #470, opened 2020-03-23
Add Kubernetes e2e tests of the driver


jsafrane:
Kubernetes allows to run its e2e tests against a CSI driver: https://github.com/kubernetes/kubernetes/tree/master/test/e2e/storage/external

It will go through all Kubernetes features - dynamic provisioning, topology, subpath, volume expansion, snapshots, cloning... It allows a CSI driver to select features to tests (e.g. no cloning for AWS EBS).

What is necessary:

1. Kubernetes cluster.
2. Upstream tests, see the link above. Basically run `ginkgo -focus=...` in a specific kubernetes/kubernetes dir.
3. Test manifest describing what to tests (see readme linked above, simple yaml file).

From the other tests in this repo, 1. and 3. look easy, however, getting upstream tests at point 2. may be complicated. Is `tester/migration-test-config.yaml` something similar?

/feature

leakingtapan:
Are you trying to test CSI migration with the k/k external test suites or test EBS CSI with the external test? Either way the tester config file [single-az-config.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/tester/single-az-config.yaml), tells the tester what cluster to spin up and what test to run.

If you want to download k/k source and compile it, you could put it inside the `test` block.

PR #473, opened 2020-03-24
WIP: Add external Kubernetes tests

jsafrane:
Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/470

Kubernetes offers so called external tests, where it can run e2e tests against an installed CSI drivers.
See https://github.com/kubernetes/kubernetes/tree/master/test/e2e/storage/external

As consequence, I renamed `tests/e2e-migration`  to `tests/e2e-kubernetes`, because both external & migration tests are there now.

Work in progress:
 * Update to Kubernetes 1.17


jsafrane:
I tested it in my AWS account - it installs 1.15 cluster, runs AWS EBS driver there and runs 1.16 tests against the driver.

The test don't pass - as it runs 1.16 tests, expecting 1.16 cluster. I haven't been able to install Kubernetes 1.16 with `k8s-e2e-tester`. In addition, 1.17 or even 1.18 cluster + tests would be very helpful, 1.16 is pretty old.

Looking at https://github.com/aws/aws-k8s-tester, it [hardcodes kops 1.14.alpha](https://github.com/aws/aws-k8s-tester/blob/master/e2e/tester/pkg/kops.go#L107). Is there any plan to move to 1.17?

Whole k8s-e2e-tester is quite cryptic, it does not log anything and I can't even get Kubernetes logs from it.

cc @leakingtapan @wongma7

tester/external-config.yaml
```diff
@@ -0,0 +1,75 @@
+cluster:
+  kops:
+    stateFile: s3://jsafrane-k8s-kops-csi-e2e
```
leakingtapan: Fix the s3 path

tester/external-config.yaml
```diff
@@ -0,0 +1,75 @@
...
+  IMAGE_NAME=$AWS_ACCOUNT_ID.dkr.ecr.us-west-2.amazonaws.com/aws-ebs-csi-driver
+  docker build -t $IMAGE_NAME:$IMAGE_TAG .
+  docker push $IMAGE_NAME:$IMAGE_TAG
+
+install: |
+  set -x
+  echo "Deploying driver"
+  # install helm
+  OS_ARCH=$(go env GOOS)-amd64
+  helm_name=helm-v2.14.1-$OS_ARCH.tar.gz
```
leakingtapan: Use helm3?

tests/e2e-kubernetes/manifest.yaml
```diff
@@ -0,0 +1,31 @@
...
+SnapshotClass:
+  FromName: true
+DriverInfo:
+  Name: ebs.csi.aws.com
+  SupportedSizeRange:
+    Min: 1Gi
+    Max: 16Ti
+  SupportedFsType:
+    xfs: {}
+    ext4: {}
```
leakingtapan: add ext2 and ext3 since they are supported too

tests/e2e-kubernetes/README.md
```diff
@@ -0,0 +1,23 @@
...
+* External CSI tests: `FOCUS=External.Storage`.
+* CSI migration tests: `FOCUS=ebs-csi-migration`.
+
+```sh
+go test -v -timeout 0 ./... -kubeconfig=$HOME/.kube/config -report-dir=$ARTIFACTS -ginkgo.focus="$FOCUS" -ginkgo.skip="\[Disruptive\]" -gce-zone=us-west-2a
+```
+
+# Update dependencies
+```sh
+go mod edit -require=k8s.io/kubernetes@v1.15.3
```
leakingtapan: Which k/k version supports external tests? Update this accordingly?

jsafrane: The version of installed kubernetes and tests should be the same. Running newer version of the tests against older kubernetes will test features that are not enabled by default or run regression tests for bugs that were not fixed yet. That's why 1.15.

As I wrote above, we need 1.17 cluster or even 1.18 to run some real tests. But I don't know how to install it with k8s-e2e-tester.

tester/external-config.yaml
jsafrane: I'd prefer all config.yamls to be the same and the rest uses 2.14.

tester/external-config.yaml
jsafrane: Fixed

tests/e2e-kubernetes/manifest.yaml
jsafrane: I am not sure it brings any real value. It only doubles number of tests. It's served by ext4 kernel module anyway.

leakingtapan:
To make sure migration test also works
/test pull-aws-ebs-csi-driver-migration-test-latest

ayberk:
/reopen
/remove-lifecycle rotten

I think we still need this? cc @wongma7 @jsafrane

wongma7:
Yes we do. The e2e-migration test setup is not good enough, it's failing since inline migration fix is not in yet, so can't be presubmit, and anyway we should run the same tests with migration on AND off.

I can rebase and pick this up from jsafrane, I revamped the run-e2e-test script recently and it doesn't rely on the tester binary so it should be easier to keep up with k/k test/e2e

AndyXiangLi:
Hi @wongma7, are you still actively working on this issue?

wongma7:
/close

thanks all, we picked this up in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/726

PR #475, opened 2020-03-26
Generate deployment manifests from helm chart

### LINKED ISSUE #305: Investigate how to keep helm chart in sync with deploy manifest
We created Helm chart for the driver [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/303). Since we still have a deploy/kubernetes/manifest.yaml, this creates a problem about how to sync the two.

Potential solutions:
* deprecate the deploy/kubernetes/manifest.yaml (not ideal)
* Evaluate Kustomize 
* etc.

**Additional Context**
* We have aws-alb-ingress-controler that is doing the same thing, I will check on that side how is this problem solved there.

### LINKED ISSUE #469: Helm chart image registries different to kustomize manifests
/kind bug

**What happened?**

The Helm chart is referencing images in Dockerhub and quay.io.

The kustomize manifests appear to be referencing images in:
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks 

They should be the same presumably?

Compare the image versions in the Helm chart with the output of
`kubectl kustomize "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=master"`

I'm happy to submit a fix if necessary.

**Environment**
- Driver version:

krmichel:
/feature

This fixes #305, fixes #469 

- Helm
  - Update helpers to have a subset of standard labels to use as selectorLabels
  - Split apart rbac and service account templates to be single resources to deal with inconsistency in `helm template`
  - Update/Add labels on resources to use helpers
  - Replace deprecated `serviceAccount` with `serviceAccountName`
  - Add minor special handling if the release name is `kustomize`
- Manifest
  - Started generating manifests from helm chart
    - Split apart rbac manifest as part of the generation process
  - Updated kustomization.yaml files
- Makefile
  - Added target to download helm
  - Added `generate-kustomize` target

Tested by generating manifests with `kubectl kustomize` for stable, alpha, and dev before generating manifests and then again after.  These manifests were compared to look for only expected differences (new labels and small reconciliations).


krmichel:
@leakingtapan, Here is the PR for generating the manifests from the helm chart.

krmichel:
fixes #469

krmichel:
This has passed tests.  @bertinatto , @d-nishi , @leakingtapan: anyone want to take a look at it?

bertinatto:
Thank you for your PR, @krmichel! I'll leave this to @leakingtapan as he is our Helm expert.

krmichel:
@leakingtapan Any chance you can take a look at this soon?

Makefile
```diff
@@ -84,3 +91,24 @@ push-release:
 .PHONY: push
 push:
 	docker push $(IMAGE):latest
+
+.PHONY: generate-kustomize
+generate-kustomize: bin/helm
```
leakingtapan, line 96: This seems a bit over complicated. Is it possible to simply this step? I imagine it will be hard to update this list of command, whenever there is some changes or refactoring to the helm templates. I think it is okay to restructure the layer out of kustomize file structure to make the generation earlier and more maintainable. Maybe combine some of the kustomize files into one for each layer? As they are all auto generated.

deploy/kubernetes/base/node.yaml
```diff
@@ -25,8 +30,9 @@ spec:
         - name: ebs-plugin
           securityContext:
             privileged: true
-          image: amazon/aws-ebs-csi-driver:latest
+          image: "amazon/aws-ebs-csi-driver:latest"
```
leakingtapan, line 33: NIT: redundant quota?

Makefile
krmichel: Yeah, I don't like it much either.  What I ran into is that when running helm template for files with more than one resource those resources weren't rendered in a consistent order.  I could run the target twice in a row and get the same resources but in a different order which makes it difficult to manage in source control.  I would be open if anyone has a way to produce reliably reproducible output and keep the alpha features separate in the folder structure.

deploy/kubernetes/base/node.yaml
krmichel: I didn't add them to the helm chart and was trying to keep actual change to a minimum for this, but I will remove the quotes from the chart.

docs/README.md
```diff
@@ -158,5 +158,13 @@ Dependencies are managed through go module. To build the project, first turn on
 * Build image and push it with latest tag: `make image && make push`
 * Build image and push it with release tag: `make image-release && make push-release`
 
+### Helm and manifests
+The helm chart for this project is in the `aws-ebs-csi-driver` directory.  The manifests for this project are in the `deploy/kubernetes` directory.  All of the manifests except kustomize patches are generated by running `helm template`.  This keeps the helm chart and the manifests in sync.
+
+When updating the helm chart:
+* Generate manifests: `make generate-kustomize`
```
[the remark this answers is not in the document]
krmichel: I will look into running a linter.  This one, at first glance, looks like it could be promising.
https://github.com/instrumenta/kubeval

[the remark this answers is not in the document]
krmichel: I added kubeval

Makefile
leakingtapan: Given the limitation with helm. I think I can live with it for now, and we can follow up with this  as a separate issue.

TBBle: I just noticed this. It should have been fixed in Helm 3.1.0: https://github.com/helm/helm/pull/6842. Weird you were seeing this with Helm 3.1.2 here.

krmichel: Yeah, I saw that helm issue when I was working on this but still saw issues with resources moving around in the processed output on multiple runs without changing anything.

PR #478, opened 2020-04-05
Correct golint warning

gliptak:
```
aws-ebs-csi-driver/pkg/driver/controller.go
Line 418: warning: if block ends with a return statement, so drop this else and outdent its block (golint)
```


gliptak:
@bertinatto @wongma7 please /ok-o-test

leakingtapan:
Please update the description with what’s the lint error being fixed here? And I’m curious why the lint error isn’t showing as part of the prow job ?

gliptak:
@leakingtapan updated description

I cannot speak to prow (rules)

gliptak:
@leakingtapan would coverage decrease prevent this from merging?

leakingtapan:
Seems the coverage drop is not related to ur change. I will just merge it. Thx for the fix!

PR #486, opened 2020-04-22
docs/README: add missing "--namespace" flag to "helm" command

gyuho:


wongma7:
/lgtm
looks like all the templates have `namespace: kube-system` set...but I don't see the harm in the helm install command being explicit!

TBBle:
Would it be worthwhile putting a check in the NOTES.txt that causes a Helm render error if you _don't_ provide a namespace? With Helm v3, the --namespace parameter also sets the namespace used to hold the secrets with the chart data, and so if you set this differently, `helm ls` will report a different namespace to where all the resources are.

Conversely, does the EBS-CSI driver _have_ to be in kube-system? The other option would be to fix the templates to use `{{.Release.Namespace}}` so that they are installed wherever the user specifies, which would allow using, e.g. ebs-csi-system, as a namespace for the driver and all its resources.

wongma7:
No, it doesn't have to be in kube-system. I *think* helm best practice is to not specify `namespace:` in the templates at all then the namespace is determined by the --namespace flag or whatever your current kube context is. But I don't keep up with helm and I have to defer to people who actually use it.

TBBle:
Ah, you are correct, excluding the namespace does do the right thing, and honours the `--namespace` flag, and the default Helm chart created by `helm create` does not have `namespace` metadata fields in its objects.

PR #488, opened 2020-04-23
Add nodeAffinity to avoid Fargate worker nodes

bgsilvait:
bug fix

Add nodeAffinity in the daemonset to avoid schedule ebs-csi-node pods on Fargate worker nodes, as Fargate does not support ebs.

If there is a pod running in Fargate, the daemoset tries to place ebs-csi-node pod in the same node, with causes the pod to Pending state.

**What testing is done?** 
Apply the original with kubectl apply -k "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=master" -> pods Pending.

Adding the nodeAffinity, all pods in the Pending state on Fargate are deleted and new are only created at EC2 nodes

### DIFF
--- a/aws-ebs-csi-driver/templates/daemonset.yaml
+++ b/aws-ebs-csi-driver/templates/daemonset.yaml
@@ -20,6 +20,15 @@ spec:
       annotations: {{ toYaml .Values.node.podAnnotations | nindent 8 }}
       {{- end }}
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: eks.amazonaws.com/compute-type
+                operator: NotIn
+                values:
+                - fargate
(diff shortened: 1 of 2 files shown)


bgsilvait:
I signed it

wongma7:
/lgtm

This is the same way we avoid scheduling kube-proxy there

bgsilvait:
@wongma7 I see that 2 e2e tests failed:

[ebs-csi-e2e] [single-az] Dynamic Provisioning should create a volume on demand with volume type "sc1" and fs type "ext2" 

and 

[ebs-csi-e2e] [single-az] Snapshot should create a pod, write and read to it, take a volume snapshot, and create another pod from the snapshot 

I don't see on how my changes with nodeAffinity can cause this errors and yes is the same way used to avoid scheduling kube-proxy on Fargate

bgsilvait:
@leakingtapan Ok, helm chart fixed.

Issue #492, opened 2020-05-03
Fargate node cannot run ebs-csi-controller


literalice:
/kind bug

**What happened?**

When the `ebs-csi-controller` pods are bound to fargate nodes, the pods stay in `CrashLoopBackOff` status.

**What you expected to happen?**

The `ebs-csi-controller` pods should successfully run.

**How to reproduce it (as minimally and precisely as possible)?**

1. Creates a fargate profile for `kube-system` namespace.
2. Deploys the csi driver like this:

```
kubectl apply -k "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=master"
```

**Anything else we need to know?**:

I'm going to submit a pull request which fixes `compute-type` of the deployment like this:

```yaml
# deploy/kubernetes/base/controller.yaml
# ...
spec:
  replicas: 2
  selector:
    matchLabels:
      app: ebs-csi-controller
  template:
    metadata:
      annotations:
        eks.amazonaws.com/compute-type: ec2  # I'll add this annotation.
      labels:
        app: ebs-csi-controller
```

Error log is:

```
$ kubectl logs -p ebs-csi-controller-6cd475bdbb-5wkvs -n kube-system -c ebs-plugin
I0503 10:08:32.978758       1 driver.go:43] Driver: ebs.csi.aws.com Version: 0.4.0
panic: could not get metadata from AWS: EC2 instance metadata is not available

goroutine 1 [running]:
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.newControllerService(...)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/controller.go:63
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.NewDriver(0x7ffc9c21faed, 0x30, 0x3, 0x3, 0x0)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/driver.go:47 +0x237
main.main()
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/cmd/main.go:45 +0x15f
```

Fargate node cannot access the EC2 instance metadata endpoint and cannot run the pod normally.

**Environment**
- Kubernetes version (use `kubectl version`): v1.15.11-eks-af3caf
- Driver version: ebs.csi.aws.com Version: 0.4.0

[...]

PR #496, opened 2020-05-05
Binds controller pods to ec2 nodes: #492

literalice:

Bug fix.
Fixes: #492

The controller pods have to run on EC2 nodes even if `kube-system` namespace uses fargate profile.

**What testing is done?** 

1. Created a fargate profile for `kube-system` namespaces
2. Deployed with: `kubectl apply -k deploy/kubernetes/overlays/stable/`
3. Made sure the controller pods are on EC2 nodes

### DIFF
--- a/deploy/kubernetes/base/controller.yaml
+++ b/deploy/kubernetes/base/controller.yaml
@@ -16,6 +16,8 @@ spec:
       app.kubernetes.io/name: aws-ebs-csi-driver
   template:
     metadata:
+      annotations:
+        eks.amazonaws.com/compute-type: ec2
       labels:
         app: ebs-csi-controller
         app.kubernetes.io/name: aws-ebs-csi-driver


deploy/kubernetes/base/controller.yaml
```diff
@@ -16,6 +16,8 @@ spec:
       app.kubernetes.io/name: aws-ebs-csi-driver
   template:
     metadata:
+      annotations:
+        eks.amazonaws.com/compute-type: ec2
```
wongma7, line 20: I don't think this is guaranteed to work, need to use a selector or affinity, what if we copied https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/488/files?

literalice: Thank you for your suggestion. I tried it, but it led to `Pending` status of the pods and got a error event:

```
$ kubectl get events -n kube-system

LAST SEEN   TYPE      REASON              OBJECT                                     MESSAGE
<unknown>   Warning   FailedScheduling    pod/ebs-csi-controller-5685df7966-2bsfv    MatchNodeSelector failed: Fargate profile system cannot satisfy pod's node selector/affinity requirements
```

If we create a fargate profile for namespaces, I think any pods in the namespaces should have affinities to nodes which have fargate-node's labels.

wongma7: I think I understand, so kube-system is part of a fargate profile but we still want certain pods in that namespace like this controller or coredns to run on ec2.

Plus, the reason the DaemonSet needs to use affinity is because while fargate on eks already knows to ignore DaemonSets, kube-scheduler doesn't so it will create pointless pods scheduled to fargate nodes.

wongma7:
/lgtm
/approve

Sorry for the delay, thanks

leakingtapan:
/hold
This change limits where the controller could be run. What if want to run it on prem? Or on non-eks environment? How do we support such cases?

wongma7:
@leakingtapan since the `eks.amazonaws.com/compute-type` annotation is specific to eks on fargate so it shouldn't block those use-cases. 

IN the future it *might* be possible to run this controller on fargate, and maybe we won't want to block it anymore , but it's simple enough to document if you want to run something on fargate you can remove the annotation (like here with coredns) https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-gs-coredns

PR #519, opened 2020-06-04
added service account name on node daemonset

### LINKED ISSUE #518: Helm chart node DaemonSet missing service account
/kind bug

**What happened?**
With zero EBS permissions on the node EC2 IAM instance profile, and using only role annotations for EBS permissions on the service accounts, I am getting errors on the node DaemonSet. 

Node DaemonSet has permissions to interact with AWS EBS APIs.

Remove all EBS permissions from the EC2 instance profile for worker nodes.

RogerReed:
Bug fix

This fixes #518 

**What testing is done?** 
Running the Helm chart with this tweak in our clusters

### DIFF
--- a/aws-ebs-csi-driver/templates/node.yaml
+++ b/aws-ebs-csi-driver/templates/node.yaml
@@ -29,6 +29,7 @@ spec:
                 operator: NotIn
                 values:
                 - fargate
+      serviceAccountName: ebs-csi-controller-sa
       nodeSelector:
         kubernetes.io/os: linux
         kubernetes.io/arch: amd64


aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -29,6 +29,7 @@ spec:
                 operator: NotIn
                 values:
                 - fargate
+      serviceAccountName: ebs-csi-controller-sa
```
leakingtapan, line 32: Why does the daemonset require controller SA? What's the error you see in the original issue: #518?

Issue #347, opened 2019-08-15
Allow overriding node attach limits as driver CLI parameter


gnufied:
Given all the complexities around how limits will be calculated on the node, it is a good idea to just allow an admin to override the limits when starting the csi-ebs driver.

leakingtapan:
Not sure how is this going to work if the worker nodes are of different instance type where each has a different attach limit. I think this should be a per node attributes.

gnufied:
The `CSINode` object where limits are actually saved are *per-node* basis, so we should be able to support this as per-node attribute. 

Are you referring to how will this be set during deploy time?

gnufied:
The flow of this code is:

csi-ebs-driver-on-node --> store-limit-> CSINode --> scheduler uses CSINode object during counting volumes.

So it is perfectly possible to support this on per-node basis. It is another problem, how will tools like kops will surface this configuration parameter though..

leakingtapan:
Probably I missed something. How could cluster operator set this new CLI parameter differently for drivers on different node? Since it is deployed as daemon set, I am assuming the pods in daemon set are identical

otterley:
On EC2 Nitro instances, the number of attachable EBS volumes is not a fixed value that can be calculated and supplied when the CSI driver is launched, regardless of instance type, that is correct throughout the lifetime of the node.  Because the formula is _MaxEbsAttachableVolumes = MaxAttachments - AttachedENIs_, where _MaxAttachments_ depends on the EC2 instance type, the value must be recalculated each time a new ENI is attached or detached from the instance.

leakingtapan:
Agree that this proposal is very suboptimal

gnufied:
> Agree that this proposal is very suboptimal

Designing a generic mechanism to calculate number of attachments as shared resource is kind of a tricky problem. There are various ways overrides can be configured such as pre-allocating a number for ENIs or percentage of attachments(and subtracting it from max). It would be nice if AWS had a API for determining `MaxEbsAttachableVolumes` an instance supports, so as we can remove the hacks in this driver and in-tree.

rfranzke:
[...]

PR #522, opened 2020-06-08
Allow volume attach limit overwrite via command line parameter

rfranzke:
/kind feature

This PR is a part of https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/347 and proceeds with the way suggested in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/347#issuecomment-638177835: In order to allow existing clusters that are leveraging/relying on the `KUBE_MAX_PD_VOLS` feature for in-tree provisioners in Kubernetes (https://kubernetes.io/docs/concepts/storage/storage-limits/#custom-limits) to migrate to CSI, the EBS CSI driver is supporting the `--volume-attach-limit-overwrite` flag now until a more sophisticated solution can be implemented.

PS: I was adding a few missing doc strings along the way and reduced package/function name stuttering according to the [official Go suggestions](https://blog.golang.org/package-names). Hope that's fine.

### DIFF
--- a/cmd/options/node_options.go
+++ b/cmd/options/node_options.go
@@ -21,6 +21,18 @@ import (
 )
 
 // NodeOptions contains options and configuration settings for the node service.
-type NodeOptions struct{}
+type NodeOptions struct {
+	// VolumeAttachLimit specifies the value that shall be reported as "maximum number of attachable volumes"
+	// in CSINode objects. It is similar to https://kubernetes.io/docs/concepts/storage/storage-limits/#custom-limits
+	// which allowed administrators to specify custom volume limits by configuring the kube-scheduler. Also, each AWS
+	// machine type has different volume limits. By default, the EBS CSI driver parses the machine type name and then
+	// decides the volume limit. However, this is only a rough approximation and not good enough in most cases.
+	// Specifying the volume attach limit via command line is the alternative until a more sophisticated solution presents
+	// itself (dynamically discovering the maximum number of attachable volume per EC2 machine type, see also
+	// https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/347).
+	VolumeAttachLimit int64
+}
 
-func (s *NodeOptions) AddFlags(fs *flag.FlagSet) {}
+func (o *NodeOptions) AddFlags(fs *flag.FlagSet) {
+	fs.Int64Var(&o.VolumeAttachLimit, "volume-attach-limit", -1, "Value value for the maximum number of volumes attachable for all nodes. If the flag is not specified then the value is approximated from the instance type.")
+}
--- a/docs/design.md
+++ b/docs/design.md
@@ -235,3 +235,15 @@ Example 2: `AWS_REGION=us-west-1 /bin/aws-ebs-csi-driver controller --extra-volu
 
 - `node`: This will only start the node service of the CSI driver.\
 Example: `/bin/aws-ebs-csi-driver node --endpoint=unix://...`
+
+## Custom volume limits
+
+For the Kubernetes in-tree volume provisioners (including the `kubernetes.io/aws-ebs` provisioner) it was possible for administrators to provide a custom volume limit overwrite (see https://kubernetes.io/docs/concepts/storage/storage-limits/#custom-limits).
+This solution is not working with CSI any longer.
+As part of [#347](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/347) we discuss how we can implement a sophisticated computation of the volume attach limit per node (e.g., based on the used machine types and already attached network interfaces).
+However, it turns out that such optimal implementation is not easily achievable.
+Each AWS machine type has different volume limits.
+Today, the EBS CSI driver parses the machine type name and then decides the volume limit.
+Unfortunately, this is only a rough approximation and not good enough in most cases.
+In order to allow existing clusters that are leveraging/relying on this feature to migrate to CSI, the EBS CSI driver is supporting the `--volume-attach-limit` flag.
+Specifying the volume attach limit via command line is the alternative until a more sophisticated solution presents itself (dynamically discovering the maximum number of attachable volume per EC2 machine type).
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -53,9 +53,10 @@ type Driver struct {
 }
 
 type DriverOptions struct {
-	endpoint        string
-	extraVolumeTags map[string]string
-	mode            Mode
+	endpoint          string
+	extraVolumeTags   map[string]string
+	mode              Mode
+	volumeAttachLimit int64
 }
 
 func NewDriver(options ...func(*DriverOptions)) (*Driver, error) {
@@ -81,10 +82,10 @@ func NewDriver(options ...func(*DriverOptions)) (*Driver, error) {
 	case ControllerMode:
 		driver.controllerService = newControllerService(&driverOptions)
 	case NodeMode:
-		driver.nodeService = newNodeService()
+		driver.nodeService = newNodeService(&driverOptions)
 	case AllMode:
 		driver.controllerService = newControllerService(&driverOptions)
-		driver.nodeService = newNodeService()
+		driver.nodeService = newNodeService(&driverOptions)
 	default:
 		return nil, fmt.Errorf("unknown mode: %s", driverOptions.mode)
 	}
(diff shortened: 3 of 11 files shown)


rfranzke:
cc @gnufied

hardikdr:
@leakingtapan @wongma7 It would be helpful if you could also take a quick look.

It seems like a reasonable change, to not break the user experience with CSI migration on AWS/EKS. Basically the `KUBE_MAX_PD_VOLS` on kube-scheduler  is ineffective with CSI-drivers and leaves no way for end-user to overwrite  max-attachable-volumes. 
- This PR at least helps in keeping existing experience intact.

wongma7:
Giving @leakingtapan a chance to chime in as he had some objections to this but I agree it's necessary to make the limit configurable somehow and a flag seems the best way.

rfranzke:
@leakingtapan Do you have some time to check the PR? Any thoughts on it?

rfranzke:
Hm, what's the general review process now? As @leakingtapan seems to be busy/inactive, can another owner/approver help to progress with this PR? @bertinatto @jsafrane @gnufied

cmd/options/node_options.go
```diff
@@ -21,6 +21,15 @@ import (
...
+	// in CSINode objects. It is similar to https://kubernetes.io/docs/concepts/storage/storage-limits/#custom-limits
+	// which allowed administrators to specify custom volume limits. Note that this solution is just a "stop-gap"
+	// solution until a more sophisticated solution can be implemented (dynamically discovering the maximum number of
+	// attachable volume per EC2 machine type, see also https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/347).
+	VolumeAttachLimitOverwrite int64
+}
 
-func (s *NodeOptions) AddFlags(fs *flag.FlagSet) {}
+func (o *NodeOptions) AddFlags(fs *flag.FlagSet) {
+	fs.Int64Var(&o.VolumeAttachLimitOverwrite, "volume-attach-limit-overwrite", -1, "Overwrite value for the maximum number of attachable volumes for all nodes")
```
gnufied, line 34: I would prefer this is simply called `volume-attach-limit` integer, so as if not specified the documentation will say - value is derived from instance type.

cmd/options/server_options.go
```diff
@@ -28,6 +28,6 @@ type ServerOptions struct {
 	Endpoint string
 }
 
-func (s *ServerOptions) AddFlags(fs *flag.FlagSet) {
-	fs.StringVar(&s.Endpoint, "endpoint", driver.DefaultCSIEndpoint, "Endpoint for the CSI driver server")
+func (o *ServerOptions) AddFlags(fs *flag.FlagSet) {
+	fs.StringVar(&o.Endpoint, "endpoint", driver.DefaultCSIEndpoint, "Endpoint for the CSI driver server")
```
gnufied, line 32: same as above. I am not sure why this change was done.

pkg/driver/driver.go
```diff
@@ -40,37 +40,43 @@ const (
...
-	options *DriverOptions
+	options *Options
 }
 
-type DriverOptions struct {
-	endpoint        string
-	extraVolumeTags map[string]string
-	mode            Mode
+// Options contains configuration options for the driver.
+type Options struct {
```
gnufied, line 59: Just checking was `DriverOptions` to `Options` change made to make `govet` etc happy?  having this change included in the same commit makes it harder to review the actual code that this PR proposes. If you really want to make this change then please do that in a different commit.

rfranzke:
@gnufied I've dropped all unrelated changes.

cmd/options/node_options.go
```diff
@@ -21,6 +21,15 @@ import (
 )
 
 // NodeOptions contains options and configuration settings for the node service.
-type NodeOptions struct{}
+type NodeOptions struct {
+	// VolumeAttachLimit specifies the value that shall be reported as "maximum number of attachable volumes"
+	// in CSINode objects. It is similar to https://kubernetes.io/docs/concepts/storage/storage-limits/#custom-limits
+	// which allowed administrators to specify custom volume limits. Note that this solution is just a "stop-gap"
+	// solution until a more sophisticated solution can be implemented (dynamically discovering the maximum number of
```
gnufied, line 28: hmm, I am not sure about documenting this as "stop-gap" where no other solution exists or being worked on. I would drop that wording and document it as such until an alternative presents itself and then we can deprecate this.

docs/design.md
```diff
@@ -235,3 +235,14 @@ Example 2: `AWS_REGION=us-west-1 /bin/aws-ebs-csi-driver controller --extra-volu
...
+This solution is not working with CSI any longer.
+As part of [#347](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/347) we discuss how we can implement a sophisticated computation of the volume attach limit per node (e.g., based on the used machine types or already attached network interfaces).
+However, it turns out that such optimal implementation is not easily achievable. 
+In order to allow existing clusters that are leveraging/relying on this feature to migrate to CSI, the EBS CSI driver is supporting the `--volume-attach-limit` flag.
+It works similar to earlier supported the `KUBE_MAX_PD_VOLS` environment variable.
+If provided, the EBS driver will report the configured value as "maximum number of attachable volumes" in the `CSINode` objects during driver registration. 
+Please consider using this flag as "stop gap" solution that is probably going to be replaced in the future with a more intelligent way of getting to know the attachment limits.
```
gnufied, line 248: same as above about "stop gap" wording.

cmd/options/node_options.go
rfranzke: Hm, but isn't it exactly like this? It's a stop-gap solution because the alternative would be to not have any possibility to migrate to CSI in case one depends on the Kubernetes feature allowing to overwrite the volume limits. Also, nobody is working on the "real" solution, or? The last update from the ticket is that the easiest implementation would require AWS to expose this information, e.g. via metadata service.

Anyways, what do you suggest to write here instead? Shall I add that the flag is planned to be deprecated in the future when a better solution is implemented?

rfranzke:
Btw, any pointers on how to get the e2e tests fixed? They fail unrelatedly, see https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/522/pull-aws-ebs-csi-driver-e2e-multi-az/1275303611722633217/build-log.txt.

```
$ curl https://sum.golang.org/lookup/github.com/aws/aws-k8s-tester@v1.3.10-0.20200623051819-1bafc137b062 -I
HTTP/2 410
```

cmd/options/node_options.go
gnufied: I think `KUBE_MAX_PD_VOLS` environment variable should work for CSI drivers too - can you open a github issue in kubernetes/kubernetes repo.

rfranzke: @gnufied I've rephrased this part and also the section in `docs/design.md`.

> I think KUBE_MAX_PD_VOLS environment variable should work for CSI drivers too - can you open a github issue in kubernetes/kubernetes repo.

According to @hardikdr it doesn't work, but he is currently reproducing the issue an then will report it to k/k. But yes, independent of this it's still reasonable to have it configurable as you initially suggested in #347.

gnufied:
Looks good to me but we might want to wait for @leakingtapan feedback too. 

/lgtm
/hold

cmd/options/node_options.go
hardikdr: The env-var `KUBE_MAX_PD_VOLS` doesn't have any effect on max volumes attachments, checked on k8s 1.18 with CSI drivers. But I guess, this should better be handled at the CSI-drivers than scheduler.

pkg/driver/node.go
```diff
@@ -514,6 +516,9 @@ func findNvmeVolume(findName string) (device string, err error) {
 
 // getVolumesLimit returns the limit of volumes that the node supports
 func (d *nodeService) getVolumesLimit() int64 {
+	if d.driverOptions.volumeAttachLimit > 0 {
```
leakingtapan, line 519: I think ` d.driverOptions.volumeAttachLimit >= 0` is better, as 0 attach limit is a valid case that ppl might need, and the default -1 will be the option to enable fallback to instance type

sjmiller609:
this is configurable in the helm chart [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/271046d1be7507cc21717dbaf96e62d3de235700/charts/aws-ebs-csi-driver/templates/node.yaml#L67)

PR #526, opened 2020-06-23
Chart option to disable default toleration of all taints

risinger:
bug fix
I would like to use a taint to prevent EBS CSI Driver components from running on a node. In the helm chart, the Controller Service and Node Service currently tolerate all taints. I believe this was introduced by mistake during a merge, since neither of the parents include the removal of the key in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/0c243cd29eff03686e481241634ed75813dcec62.


leakingtapan:
See: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/122 and https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/441 for the reason why it was removed originally

leakingtapan:
> I would like to use a taint to prevent EBS CSI Driver components from running on a node.

What's ur use case for this?

risinger:
Thanks for the links. What's the point of `.Values.tolerations` and `.Values.node.tolerations` if all taints are tolerated by default, and the user is unable to disable that without using kustomize? I would personally prefer (and expect) to manually specify any tolerations on services in my cluster than have them automatically tolerate all taints. To me at least, it's a logical requirement that pods I want to run on tainted nodes will need to tolerate that taint. 

My current use case: I'm running a cluster with both public and private nodes. I want to specify exactly which pods are allowed to run on the public nodes to limit exposed attack surface. No pods running on the public nodes require persistent storage so I do not want the CSI Node Service or the CSI Controller Service to run on those public nodes. 

If you or others feel strongly that tolerating all taints should be the default, I could work with a toggle for it instead.

wongma7:
i am inclined to agree with @risinger, seems a bit heavy handed to default tolerate all taints. more control should be given to the admin to decide whether a pod runs on a given node or not. If they anticipate a node will need to run stateful pods then the onus is on them to ensure ebs pod is colocated there

/ok-to-test

wongma7:
with that said, tolerating all taints is the 'backwards compatible' way since in the non-csi world, the in-tree aws ebs plugin will be ready in kubelet whether you want it or not.

risinger:
I updated the default behavior to maintain backward compatibility. 
Please let me know if there is something else you'd like to see before merge @leakingtapan @wongma7

leakingtapan:
Fixes: #591

risinger:
Initial 2 failures look like the same io2 flake on #593 
Seems like this is the most relevant ticket https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/545. Happy to create a more specific issue on request. 

After merging master, there's a critical failure almost immediately. Not sure if flake or what. I'll give it one more shot. 

/retest

risinger:
Thanks @wongma7. 
The merge from master wiped the approval. Can I get another /lgtm?

krmichel:
@leakingtapan This is clean now but your lgtm got removed when master got merged in

Issue #529, opened 2020-07-08
Migrated AWS EBS volumes don't have tags


jsafrane:
/kind bug

**What happened?**
In-tree awes-ebs provisioner creates volumes with these tags:
```
kubernetes.io/cluster/jsafrane-10235-vgdv6: owned     
Name:                                       jsafrane-10235-vgdv6-dynamic-pvc-447cc711-bb65-4b4d-836d-a822e4e77e43     
kubernetes.io/created-for/pv/name:          pvc-447cc711-bb65-4b4d-836d-a822e4e77e43     
kubernetes.io/created-for/pvc/name:         myclaim     
kubernetes.io/created-for/pvc/namespace:    default     
```

When migration to CSI is enabled, provisioned volumes don't have such tags, only
```
CSIVolumeName: pvc-4e4cb311-3907-4192-bc58-cde8ea112392
```

**What you expected to happen?**
When requested on cmdline, the driver should add also the "legacy" tags to EBS volumes.

**Anything else we need to know?**:
- Driver version: today's master

jsafrane:
* The cluster ID is hard to get, I can imagine a new driver option `--kubernetes-cluster-id=XXX` or let `--extra-volume-tags` accept `kubernes.io` prefix as suggested in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/450, so Kubernetes vendor can add it.

* PV / PVC namespace/name: `CreateVolume` request, if enabled via `external-provisioner --extra-create-metadata=true`, gets PVC name+namespace + PV name.
  * The driver can either *always* tag volumes with pv/pvc names when it gets them from the provisioner (i.e. someone has deployed the external-provisioner with `--extra-create-metadata=true`. This would be my preferred solution.
  * Or there is a new cmdline options in the driver, say `--legacy-tags=true`, that then explicitly enables volume tags.

wongma7:
> let --extra-volume-tags accept kubernes.io prefix as suggested in #450, so Kubernetes vendor can add it.

I prefer this.

> always tag volumes with pv/pvc names when it gets them from the provisioner (i.e. someone has deployed the external-provisioner with --extra-create-metadata=true.

Also prefer this, I can't think of a reason someone would need a driver cmdline option to disable these tags.

jsafrane:
[...]

PR #530, opened 2020-07-10
Add tags that the in-tree volume plugin uses

jsafrane:
Bugfix, to reach 100% compatibility with the in-tree volume plugin, the CSI driver should create also the same tags on dynamically provisioned volumes.

* When the external-provisioner passes PV/PVC name/namespace via `CreateVolumeRequest.Parameters`, use those to tag volumes (i.e. the provisioner must run with `--extra-create-metadata=true`).
* When `--k8s-tag-cluster-id` was provided on cmdline, use those to tag the provisioned volumes as owned by the cluster.

Both set of tags are optional.

Fixes: #529

**What testing is done?** 
* Tested with Kubernetes 1.18 and external-provisioner 1.6.

### DIFF
--- a/cmd/options/controller_options.go
+++ b/cmd/options/controller_options.go
@@ -27,8 +27,12 @@ type ControllerOptions struct {
 	// ExtraVolumeTags is a map of tags that will be attached to each dynamically provisioned
 	// volume.
 	ExtraVolumeTags map[string]string
+	// ID of the kubernetes cluster. This is used only to create the same tags on volumes that
+	// in-tree volume volume plugin does.
+	KubernetesClusterID string
 }
 
 func (s *ControllerOptions) AddFlags(fs *flag.FlagSet) {
 	fs.Var(cliflag.NewMapStringString(&s.ExtraVolumeTags), "extra-volume-tags", "Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
+	fs.StringVar(&s.KubernetesClusterID, "k8s-tag-cluster-id", "", "ID of the Kubernetes cluster used for tagging provisioned EBS volumes (optional).")
 }
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -36,6 +36,16 @@ const (
 
 	// KmsKeyId represents key for KMS encryption key
 	KmsKeyIDKey = "kmskeyid"
+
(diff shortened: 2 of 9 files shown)


jsafrane:
cc @wongma7 @leakingtapan @bertinatto - PTAL

cmd/options/controller_options.go
```diff
@@ -27,8 +27,12 @@ type ControllerOptions struct {
...
 	// volume.
 	ExtraVolumeTags map[string]string
+	// ID of the kubernetes cluster. This is used only to create the same tags on volumes that
+	// in-tree volume volume plugin does.
+	ClusterID string
 }
 
 func (s *ControllerOptions) AddFlags(fs *flag.FlagSet) {
 	fs.Var(cliflag.NewMapStringString(&s.ExtraVolumeTags), "extra-volume-tags", "Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
+	fs.StringVar(&s.ClusterID, "cluster-id", "", "ID of the Kubernetes cluster (optional).")
```
bertinatto, line 37: I'm afraid `--cluster-id` alone doesn't represent what this flag is really doing (adding labels with cluster ID to the volumes).

Also, I'd make the flag explicitly mention k8s (in addition to the help message already mentioning it). Something like `k8s-cluster-id-volume-tags`. It's similar to the flag above, plus it emphasizes it's k8s-specific.

pkg/driver/constants.go
```diff
@@ -36,6 +36,51 @@ const (
...
+	// the external provisioner sidecar is started with --extra-create-metadata=true and
+	// thus provides such metadata to the CSI driver.
+	PVCNamespaceTag = "kubernetes.io/created-for/pvc/namespace"
+
+	// PVNameTag is tag applied to provisioned EBS volume for backward compatibility
+	// with in-tree volume plugin. Value of the tag is PV name. It is applied only when
+	// the external provisioner sidecar is started with --extra-create-metadata=true and
+	// thus provides such metadata to the CSI driver.
+	PVNameTag = "kubernetes.io/created-for/pv/name"
 )
```
bertinatto, line 84: Note: in the future we should consider moving CO-specific constants to their own packages/files.

cmd/options/controller_options.go
jsafrane: Renamed to `--k8s-tag-cluster-id=` with description "ID of the Kubernetes cluster used for tagging provisioned EBS volumes (optional)"

Issue #439, opened 2020-01-13
Raw Block device attach failure


kvalliy:
/kind bug

What happened?
First time I attach a raw block PVC to a pod , it works fine.Once the pod usign the raw block device is up and running , I delete the pod and then try to recreate the pod with the same PVC , this time the PVC fails to mount with the following error
```
  Type     Reason                  Age                From                                     Message
  ----     ------                  ----               ----                                     -------
  Normal   Scheduled               36s                default-scheduler                        Successfully assigned default/app to ip-172-19-133-106.ec2.internal
  Warning  FailedAttachVolume      27s (x5 over 35s)  attachdetach-controller                  AttachVolume.Attach failed for volume "pvc-4297a6ff-33ec-11ea-b252-0a6741012e41" : rpc error: code = AlreadyExists desc = Resource already exists
  Normal   SuccessfulAttachVolume  19s                attachdetach-controller                  AttachVolume.Attach succeeded for volume "pvc-4297a6ff-33ec-11ea-b252-0a6741012e41"
  Warning  FailedMapVolume         2s (x3 over 4s)    kubelet, ip-172-19-133-106.ec2.internal  MapVolume.SetUp failed for volume "pvc-4297a6ff-33ec-11ea-b252-0a6741012e41" : rpc error: code = Internal desc = Could not remove mount target "/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-4297a6ff-33ec-11ea-b252-0a6741012e41": remove /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-4297a6ff-33ec-11ea-b252-0a6741012e41: device or resource busy
```
I am able to reproduce this consistently every time I delete and recreate a pod using raw block PVC , seems like the initial delete doesn't cleanup properly on the node

What you expected to happen?
I expect the pod to be mounted and go to a running state 

How to reproduce it (as minimally and precisely as possible)?
Create a dynamically provisioned PVC 
Create a pod to use that PVC
Once the pod is running delete the pod and recreate it back on the same node

Anything else we need to know?:

Environment
Amazon EKS
Kubernetes version (use kubectl version):
v1.14.8
Driver version:
v0.4.0

kvalliy:
on debugging further on the node , I see that the reason the plugin is unable to delete the device file is because it seems to be mounted to devtmpfs on the node 
```
[...]

PR #531, opened 2020-07-16
Update features state

ialidzhikov:
/kind documentation

Update the features state as currently they wrong state that given features are still alpha.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -68,9 +68,9 @@ Following sections are Kubernetes specific. If you are Kubernetes user, use foll
 * **Dynamic Provisioning** - uses persistence volume claim (PVC) to request the Kuberenetes to create the EBS volume on behalf of user and consumes the volume from inside container. Storage class's **allowedTopologies** could be used to restrict which AZ the volume should be provisioned in. The topology key should be **topology.ebs.csi.aws.com/zone**.
 * **Mount Option** - mount options could be specified in persistence volume (PV) to define how the volume should be mounted.
 * **NVMe** - consume NVMe EBS volume from EC2 [Nitro instance](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances).
-* **Block Volume** (beta since 1.14) - consumes the EBS volume as a raw block device for latency sensitive application eg. MySql
-* **Volume Snapshot** (alpha) - creating volume snapshots and restore volume from snapshot.
-* **Volume Resizing** (alpha) - expand the volume size.
+* **[Block Volume](https://kubernetes-csi.github.io/docs/raw-block.html)** - consumes the EBS volume as a raw block device for latency sensitive application eg. MySql. The corresponding CSI feature (`CSIBlockVolume`) is GA since Kubernetes 1.18.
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -68,9 +68,9 @@ Following sections are Kubernetes specific. If you are Kubernetes user, use foll
 * **Dynamic Provisioning** - uses persistence volume claim (PVC) to request the Kuberenetes to create the EBS volume on behalf of user and consumes the volume from inside container. Storage class's **allowedTopologies** could be used to restrict which AZ the volume should be provisioned in. The topology key should be **topology.ebs.csi.aws.com/zone**.
 * **Mount Option** - mount options could be specified in persistence volume (PV) to define how the volume should be mounted.
 * **NVMe** - consume NVMe EBS volume from EC2 [Nitro instance](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances).
-* **Block Volume** (beta since 1.14) - consumes the EBS volume as a raw block device for latency sensitive application eg. MySql
-* **Volume Snapshot** (alpha) - creating volume snapshots and restore volume from snapshot.
-* **Volume Resizing** (alpha) - expand the volume size.
+* **Block Volume** ([beta since 1.14](https://kubernetes-csi.github.io/docs/raw-block.html)) - consumes the EBS volume as a raw block device for latency sensitive application eg. MySql
```
bertinatto, line 71: Seems like this is GA since 1.18 (the linked page must be outdated as well):

https://github.com/kubernetes/kubernetes/blob/43fbe17dc682b703b0b663935a797244378f0aaf/pkg/features/kube_features.go#L283-L287

ialidzhikov: Nice catch, updated the feature state here and proposed https://github.com/kubernetes-csi/docs/pull/350.

msau42: Is the AWS implementation of block volume considered GA? There is at least one known issue is hit very often on a pod restart on the same node. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/439

bertinatto: I don't think there's any feature considered GA in the EBS CSI driver currently, as the dirver itself ([should be beta still](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/milestones?state=closed).

Note that the graduation level of these features are related to k8s, not the driver.

msau42: It might be good to be more explicit about what the feature status is at the driver level, and what it is at the k8s level. For example, nvme listed here is not a k8s feature.

leakingtapan: Since the updated doc is under kubernetes section, it makes sense to update it this way. But yeah we need to be explicit about driver level feature support (which is missing from README). We need to prioritize the fix of $439

ialidzhikov: > It might be good to be more explicit about what the feature status is at the driver level, and what it is at the k8s level. For example, nvme listed here is not a k8s feature.

Good point, @msau42 . I now clarified that the feature status in the description corresponds to the CSI feature status.

ialidzhikov:
@bertinatto , could you please recheck the PR? Thanks in advance.

bertinatto:
/lgtm

Thanks for contributing, @ialidzhikov! Please let us know if you want to work on https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/531#discussion_r458528338 (driver-level feature support docs) and need help!

Issue #545, opened 2020-08-12
Fix e2e-test-single-az failures


wongma7:
/kind bug

**What happened?** tests have been incredibly failureprone as of late https://testgrid.k8s.io/sig-aws-ebs-csi-driver#e2e-test-single-az&width=20

**What you expected to happen?** tests should pass consistently

**How to reproduce it (as minimally and precisely as possible)?** create a PR and let the CI run tests

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

wongma7:
Need to dump csi driver logs to aid in debugging.

wongma7:
Most failures are sc1 and st1 which are slow hdd with minimum 500GiB size. Not sure how long formatting should take, of course formatting 500GiB does not mean all 500GiB needs to be written, but let's say it does then in a best case scenario of 250MiB/s writing all 500GiB would take 34 mins.

io1 tests also fail a bunch and it could be because we provision with bare minimum iops of 100.

Doesn't really explain why the tests have been so flaky only over last couple months

Anyway I do not see the value in testing every volume type fs type combination so I will remove all but gp2 and one of io2/io1 + xfs

ayberk:
I'm closing this as I think it doesn't happen anymore. @wongma7 to confirm.

/close

Issue #272, opened 2019-04-05
Build production container image from distroless


leakingtapan:
**Is your feature request related to a problem? Please describe.**
https://github.com/GoogleContainerTools/distroless

PR #551, opened 2020-08-27
Change the base image to debian-base

vpnachev:
I would say it is a security improvement.

When amazonlinux is used as base image we are detecting multiple critical vulnerabilities in the image, some of them are 
1. CVE-2016-4658
1. CVE-2017-7376
1. CVE-2016-4448
1. CVE-2016-1834
1. CVE-2017-16997
1. CVE-2017-18269
1. CVE-2019-5482
1. CVE-2019-5481
1. CVE-2019-3822
...
And many others.

Change the base image to ~alpine linux~ debian-base-amd64 to reduce the number of reported security vulnerabilities and also the size of the image from:
```
$ docker image ls amazon/aws-ebs-csi-driver:latest
REPOSITORY                  TAG                 IMAGE ID            CREATED             SIZE
amazon/aws-ebs-csi-driver   latest              ddab7a7bfa60        3 seconds ago       464MB # amazonlinux as base image
```
to
```
$ docker image ls amazon/aws-ebs-csi-driver:latest
REPOSITORY                  TAG                 IMAGE ID            CREATED             SIZE
amazon/aws-ebs-csi-driver   latest              17fafe83d963        19 seconds ago      36.7MB # alpinelinux as base image
```

```
$  docker image ls amazon/aws-ebs-csi-driver:latest
REPOSITORY                  TAG                 IMAGE ID            CREATED             SIZE
amazon/aws-ebs-csi-driver   latest              f2432a80d7ef        2 seconds ago       167MB # debian-base as base image
```
**What testing is done?**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.14.1-stretch as builder
+FROM golang:1.14.1-stretch AS builder
 WORKDIR /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver
 
 # Cache go modules
@@ -23,8 +23,14 @@ RUN go mod download
 ADD . .
 RUN make
 
-FROM amazonlinux:2
+FROM amazonlinux:2 AS amazonlinux
 RUN yum install ca-certificates e2fsprogs xfsprogs util-linux -y
+COPY --from=builder /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/bin/aws-ebs-csi-driver /bin/aws-ebs-csi-driver
+
+ENTRYPOINT ["/bin/aws-ebs-csi-driver"]
+
+FROM k8s.gcr.io/build-image/debian-base:v2.1.3 AS debian-base
+RUN clean-install ca-certificates e2fsprogs mount udev util-linux xfsprogs
 COPY --from=builder /go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/bin/aws-ebs-csi-driver /bin/aws-ebs-csi-driver
 
 ENTRYPOINT ["/bin/aws-ebs-csi-driver"]
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,6 @@
 module github.com/kubernetes-sigs/aws-ebs-csi-driver
 
 require (
-	github.com/aws/aws-k8s-tester/e2e/tester v0.0.0-20200731204948-ad5df86f9de5 // indirect
 	github.com/aws/aws-sdk-go v1.29.11
 	github.com/container-storage-interface/spec v1.2.0
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f // indirect
@@ -9,7 +8,6 @@ require (
 	github.com/elazarl/goproxy v0.0.0-20181111060418-2ce16c963a8a // indirect
 	github.com/golang/mock v1.4.1
 	github.com/golang/protobuf v1.3.3
-	github.com/gregjones/httpcache v0.0.0-20181110185634-c63ab54fda8f // indirect
 	github.com/kubernetes-csi/csi-test v2.0.0+incompatible
 	github.com/kubernetes-csi/external-snapshotter/v2 v2.0.1
 	github.com/onsi/ginkgo v1.10.2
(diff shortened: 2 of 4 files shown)


bertinatto:
Thank you for your PR, @vpnachev.

This is a really good point. @leakingtapan @matthew are you OK with that?

Even more interesting would be to have distroless as the base image. There's an issue for that here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/272. A this has been considered in the PD driver as well: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/177.

msau42:
Ref discussion why K8s chose debian over alpine as a base image: https://github.com/kubernetes/kubernetes/issues/40248#issuecomment-280781931

I'm actually using debian-slim as the base image, and it dropped the image size to 48mb

danbeaulieu:
+1 to distroless if the base layer needs to change. In the cloud native space alpine is an outlier since many projects have adopted slim images or distroless. Adding another non-standard distro to infrastructure complicates container supply chain management.

vpnachev:
/retest

Switching to distroless would require a major refactoring in the build system, i.e. to be switched to bazel, because of the dependencies like `e2fsprogs`, `xfsprogs` and `util-linux` which is not the focus of this PR.

What I can do is to use use the same image as for the GCP CSI driver https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/e1670c165e72bca32c778b884656b03ff1818386/Dockerfile#L25. What do you think?

vpnachev:
Actually, the more widely used base image for CSI components is us.gcr.io/k8s-artifacts-prod/build-image/debian-base-amd64:v2.1.0.

wongma7:
I would prefer to maintain both a Dockerfile.amazonlinux (amazon still need an AL2 based image) and a debian one.

vpnachev:
Dockerfile.amazonlinux has been added.

vpnachev:
A mitigation for https://github.com/fvbommel/util/issues/6 has been implemented.

Dockerfile
```diff
@@ -23,8 +23,8 @@ RUN go mod download
 ADD . .
 RUN make
 
-FROM amazonlinux:2
-RUN yum install ca-certificates e2fsprogs xfsprogs util-linux -y
+FROM us.gcr.io/k8s-artifacts-prod/build-image/debian-base-amd64:v2.1.0
```
leakingtapan, line 26: is there a multi arch image for this? I would prefer stick with multi-arch image to make multi arch image build easier.

go.mod
```diff
@@ -51,6 +49,7 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.17.3
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.17.3
 	k8s.io/sample-controller => k8s.io/sample-controller v0.17.3
+	vbom.ml/util => github.com/fvbommel/util v0.0.2 // Mitigate https://github.com/fvbommel/util/issues/6
```
leakingtapan, line 52: What is this dependency for?

gnufied: It is being pulled via k8s dependency chain.  Unit tests are failing atm in this repo without the fix:

```
Activated service account credentials for: [pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com]
+ make test
go test -v -race ./cmd/... ./pkg/...
go: k8s.io/kubernetes@v1.17.3 requires
	k8s.io/kubectl@v0.0.0 requires
	vbom.ml/util@v0.0.0-20160121211510-db5cfe13f5cc: unrecognized import path "vbom.ml/util": https fetch: Get "https://vbom.ml/util?go-get=1": dial tcp: lookup vbom.ml on 10.63.240.10:53: no such host
make: *** [Makefile:67: test] Error 1
+ EXIT_VALUE=2
+ set +o xtrace
```

vpnachev: Ref: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/551#issuecomment-682387518

Dockerfile
ialidzhikov: `k8s.gcr.io/build-image/debian-base` (the same image but using the vanity domain) is multi arch.

```suggestion
FROM k8s.gcr.io/build-image/debian-base:v2.1.3
```

vpnachev: Thanks. I have done it.

vpnachev:
/test pull-aws-ebs-csi-driver-e2e-single-az

However, I think the failure of this test is related to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/545

ialidzhikov:
@leakingtapan , @bertinatto did you managed to recheck this PR after the recent changes?

wongma7:
@vpnachev looks like Dockerfile.amazonlinux got dropped. There is some red tape preventing me from pushing a non-AL2 based image to the official amazon Docker Hub https://hub.docker.com/r/amazon/aws-ebs-csi-driver. So for the next release which I want to make this week, we might still need to base it on AL2. What I can do is partially revert this, make the release, then resubmit it and try to get an exception to push debian-based image to hub.docker.com/u/amazon in time for the next release. WDYT?

wongma7:
Actually I guess all I need to do is change the github action to build with target al2, I see now your Dockerfile allows for that. Thanks!!

Issue #498, opened 2020-05-07
Attach volume fails after resizing completed


msau42:
/kind bug

**What happened?**
Running the "verify offline volume expansion" e2e test.

After controller resize finishes, a pod is created and we try to attach the volume, but attach fails because it still thinks the volume is being modified. This continues for at least 5 minutes after the resizing has finished, at which point the test times out.

```
May  7 21:19:58.215: INFO: At 2020-05-07 21:14:42 +0000 UTC - event for standard-rwoc4mps: {external-resizer ebs.csi.aws.com } FileSystemResizeRequired: Require file system resize of volume on node
May  7 21:19:58.215: INFO: At 2020-05-07 21:14:44 +0000 UTC - event for security-context-00009429-f857-4280-a75f-710af14c0a0b: {attachdetach-controller } FailedAttachVolume: AttachVolume.Attach failed for volume "pvc-683dd3d3-4858-4471-858e-b9cd72eb2b43" : rpc error: code = Internal desc = Could not attach volume "vol-085a666d487519e08" to node "i-0968acda4b0e832b6": could not attach volume "vol-085a666d487519e08" to node "i-0968acda4b0e832b6": OperationNotPermitted: Cannot attach volume vol-085a666d487519e08 when it is in modification state: MODIFYING
        status code: 400, request id: c73d69dd-73c9-45c5-9b53-5bd5346a3196
...
May  7 21:19:58.215: INFO: At 2020-05-07 21:18:54 +0000 UTC - event for security-context-00009429-f857-4280-a75f-710af14c0a0b: {attachdetach-controller } FailedAttachVolume: AttachVolume.Attach failed for volume "pvc-683dd3d3-4858-4471-858e-b9cd72eb2b43" : rpc error: code = Internal desc = Could not attach volume "vol-085a666d487519e08" to node "i-0968acda4b0e832b6": could not attach volume "vol-085a666d487519e08" to node "i-0968acda4b0e832b6": OperationNotPermitted: Cannot attach volume vol-085a666d487519e08 when it is in modification state: MODIFYING
```

csi resizer and csi attacher are running on the same node:

csi-resizer logs:
```
[...]

PR #552, opened 2020-08-28
Make EBS controllerexpansion idempotent

gnufied:
Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/498

This PR contains following fixes:
- When describeVolume returns volume size to be greater than requested size, we also verify if volume has any modifications pending. I have observed that, AWS can set targetSize on volume even with pending volume modifications. This is the root cause of EBS resize flake.
- When volume modifications check return volume resizing to be completed, we also verify volume size by describing volume. This is to ensure that - we minimize impact of eventual consistency by making two different API calls.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -59,7 +59,10 @@ var (
 		VolumeTypeST1,
 		VolumeTypeStandard,
 	}
-	VolumeNotBeingModified = fmt.Errorf("volume is not being modified")
+
+	volumeModificationDuration   = 1 * time.Second
+	volumeModificationWaitFactor = 1.7
+	volumeModificationWaitSteps  = 10
 )
 
 // AWS provisioning limits.
@@ -115,6 +118,9 @@ var (
 
 	// ErrInvalidMaxResults is returned when a MaxResults pagination parameter is between 1 and 4
 	ErrInvalidMaxResults = errors.New("MaxResults parameter must be 0 or greater than or equal to 5")
+
+	// VolumeNotBeingModified is returned if volume being described is not being modified
+	VolumeNotBeingModified = fmt.Errorf("volume is not being modified")
 )
 
 // Disk represents a EBS volume
@@ -799,13 +805,6 @@ func isAWSError(err error, code string) bool {
 	return false
 }
 
-// isAWSErrorIncorrectModification returns a boolean indicating whether the given error
-// is an AWS IncorrectModificationState error. This error means that a modification action
-// on an EBS volume cannot occur because the volume is currently being modified.
-func isAWSErrorIncorrectModification(err error) bool {
-	return isAWSError(err, "IncorrectModificationState")
-}
-
 // isAWSErrorInstanceNotFound returns a boolean indicating whether the
 // given error is an AWS InvalidInstanceID.NotFound error. This error is
 // reported when the specified instance doesn't exist.
@@ -900,23 +899,12 @@ func (c *cloud) ResizeDisk(ctx context.Context, volumeID string, newSizeBytes in
 	}
 
 	klog.Infof("expanding volume %q to size %d", volumeID, newSizeGiB)
-	var mod *ec2.VolumeModification
 	response, err := c.ec2.ModifyVolumeWithContext(ctx, req)
 	if err != nil {
-		if !isAWSErrorIncorrectModification(err) {
-			return 0, fmt.Errorf("could not modify AWS volume %q: %v", volumeID, err)
-		}
-
-		m, modFetchError := c.getLatestVolumeModification(ctx, volumeID)
-		if modFetchError != nil {
-			return 0, modFetchError
-		}
-		mod = m
+		return 0, fmt.Errorf("could not modify AWS volume %q: %v", volumeID, err)
 	}
 
-	if mod == nil {
-		mod = response.VolumeModification
-	}
+	mod := response.VolumeModification
 
 	state := aws.StringValue(mod.ModificationState)
 	if volumeModificationDone(state) {
@@ -955,9 +943,9 @@ func (c *cloud) checkDesiredSize(ctx context.Context, volumeID string, newSizeGi
 // waitForVolumeSize waits for a volume modification to finish and return its size.
 func (c *cloud) waitForVolumeSize(ctx context.Context, volumeID string) (int64, error) {
 	backoff := wait.Backoff{
-		Duration: 1 * time.Second,
-		Factor:   1.7,
-		Steps:    10,
+		Duration: volumeModificationDuration,
+		Factor:   volumeModificationWaitFactor,
+		Steps:    volumeModificationWaitSteps,
 	}
 
 	var modVolSizeGiB int64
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -56,6 +56,7 @@ var (
 		VolumeTypeST1,
 		VolumeTypeStandard,
 	}
+	VolumeNotBeingModified = fmt.Errorf("volume is not being modified")
```
bertinatto, line 59: Could you move this down below, together with other errors?

pkg/cloud/cloud.go
```diff
@@ -855,26 +861,60 @@ func (c *cloud) ResizeDisk(ctx context.Context, volumeID string, newSizeBytes in
 	newSizeGiB := util.RoundUpGiB(newSizeBytes)
 	oldSizeGiB := aws.Int64Value(volume.Size)
 
+	latestMod, modFetchError := c.getLatestVolumeModification(ctx, volumeID)
+
+	if latestMod != nil && modFetchError == nil {
+		state := aws.StringValue(latestMod.ModificationState)
+		if state == ec2.VolumeModificationStateModifying {
+			return oldSizeGiB, fmt.Errorf("volume %q is still being expanded to size %d", volumeID, newSizeGiB)
```
bertinatto, line 869: In order to make this function idempotent, we shouldn't return an error if the volume modification is in progress, no?

It feels like it should return success _if the volume is already being modified to newSizeGiB_, but the volume is being modified to a value _different than newSizeGiB_, then it's OK to return an error?

pkg/cloud/cloud.go
```diff
@@ -855,26 +861,60 @@ func (c *cloud) ResizeDisk(ctx context.Context, volumeID string, newSizeBytes in
...
+		}
+	}
+
+	// if there was an error fetching volume modifications and it was anything other than VolumeNotBeingModified error
+	// that means we have an API problem.
+	if modFetchError != nil && modFetchError != VolumeNotBeingModified {
+		klog.Errorf("error fetching volume modifications for %q: %v", volumeID, modFetchError)
+		return oldSizeGiB, fmt.Errorf("error fetching volume modifications for %q: %v", volumeID, modFetchError)
```
bertinatto, line 877: nit: no need to log the error since we're returning it (it'll show up in the logs anyway)

pkg/cloud/cloud.go
wongma7: that doesn't seem right either, then it would return that resize has succeeded which is a lie. 

attach/controllerpublish behaviour is to wait for the attach to complete if it's found to be in progress. The same could be done here

gnufied: Yep - was just about to push a change that made the whole RPC call block until operation is finished.

bertinatto: > that doesn't seem right either, then it would return that resize has succeeded which is a lie.

That's a good point, but currently we're doing the same thing in the first call (because it considers the volume resized once the operation is in the Optimizing state). So if Optimizing mean success for the first call, it should also mean success for the subsequent calls.

> attach/controllerpublish behaviour is to wait for the attach to complete if it's found to be in progress. The same could be done here

+1

> Yep - was just about to push a change that made the whole RPC call block until operation is finished.

By waiting for the modification to be in the Completed state? If so, can/should we change the behaviour in-tree as well, to be consistent with the CSI driver?

pkg/cloud/cloud.go
gnufied: removed

pkg/cloud/cloud.go
gnufied: Alright modified so that each subsequent resize call will now wait for previous resize to finish before returning. 

> By waiting for the modification to be in the Completed state? If so, can/should we change the behaviour in-tree as well, to be consistent with the CSI driver?

No not "Completed" state. I reverted the code to "optimizing" *and* "completed" check, because the main reason this was breaking was because `describeVolume` can return updated size before volume has really finished resizing. "optimized" vs "completed" state had no impact on the outcome.

wongma7: double checking, doc says size can take a few seconds after volume is in optimizing state, and I guess it does not matter because even if fs resize is attempted prematurely, the fs resize will just get retried and must eventually succeed.

"Size changes usually take a few seconds to complete and take effect after a volume is in the Optimizing state. "
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-volume-modifications.html

pkg/cloud/cloud.go
```diff
@@ -868,16 +899,17 @@ func (c *cloud) ResizeDisk(ctx context.Context, volumeID string, newSizeBytes in
...
 	response, err := c.ec2.ModifyVolumeWithContext(ctx, req)
 	if err != nil {
 		if !isAWSErrorIncorrectModification(err) {
 			return 0, fmt.Errorf("could not modify AWS volume %q: %v", volumeID, err)
 		}
 
-		m, err := c.getLatestVolumeModification(ctx, volumeID)
-		if err != nil {
-			return 0, err
+		m, modFetchError := c.getLatestVolumeModification(ctx, volumeID)
```
wongma7, line 910: is checking the ModificationState immediately after calling ModifyVolume still necessary, won't waitForVolumeSize do the same thing?

pkg/cloud/cloud.go
msau42: We don't return size in fs resize, it just blindly tries to resize to the full disk size, whatever it may be at the time. So is it possible that we won't fs expand to the intended size if we don't wait?

wongma7: shoot you are right. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/9b33867eb74b03f6525166cc84c4b68132aa9278/vendor/k8s.io/kubernetes/pkg/util/resizefs/resizefs_linux.go#L38

I am probably misreading the doc, the correct meaning must be that "optimizing" means the size change has already taken effect and only the performance change is still in progress.

gnufied: > Size changes usually take a few seconds to complete and take effect after a volume is in the Optimizing state.

tbh - the wording here is imprecise in EBS docs. We have been told by EBS engineers that volume is "ready-to-use"after it enters "optimizing" state but if we wait for "VolumeModificationComplete" state then, it can take really long time to complete resize operation and afaict there is no intermediate state between "optimizing" and "complete" (although docs do imply that there is a intermediate state where size changes take effect few seconds after volume enters "optimizing" state).

wongma7 (approved):

pkg/cloud/cloud.go
gnufied: I have removed most of this code. Still keeping some of the check in `VolumeModification` object after modification, so as we can return success immediately if volume was previously expnaded to same size already..

pkg/cloud/cloud.go
gnufied: fixed. good catch.

wongma7:
lgtm
/retest

wongma7:
@bertinatto this look good to u?

bertinatto (commented):
/lgtm

Issue #553, opened 2020-08-31
Bring back the vendor directory


gnufied:
Compared to every other repository in k8s org, EBS CSI driver does not vendor the dependencies. This creates problems like - https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/552/pull-aws-ebs-csi-driver-unit/1300182099357077505/build-log.txt

When we removed vendor via  - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/202 the argument was, it will result in "consistent build image". Isn't this other way around? Whereas having vendored depenencies and using those depenencies during build time will result in consistent images?

/kind bug

cc @bertinatto @leakingtapan @wongma7

leakingtapan:
IIRC the "consistent build image" was referring to the usage of go.mod/go.sum which lock down all the dependency version, compared to go vendor dir could potentially modified without being noticed. But anyway, what's the advantage of vendor dir? I think Golang is moving toward go mod and vendor dir will be deprecated.

gnufied:
I have added verification check so as any direct modification to vendor files will be detected and will cause tests to fail. I think that should address concerns about vendored files being silently modified.  If anything - vendored dependencies make builds more reproducible.

leakingtapan:
Fair enough. I feels this is more golang issue on whether vendor dir or go mod should be preferred. Let’s use vendor then if all k8s projects are doing the same thing

PR #555, opened 2020-09-02
Vendor deps

gnufied:
This makes several fixes to aws-ebs-driver tooling:

1. It brings back vendoring (fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/553)
2. It adds phony target to bin/aws-ebs-driver binary because we aren't rebuilding it 
3. It add verification for vendoring
4. it fixes issue with vbox.ml disappearing from internet
5. It ensures that we are using vendored files when building the binary and image.

cc @leakingtapan @wongma7 @bertinatto


hack/verify-vendor.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
bertinatto, line 1: hmm.... this is not being called from `hack/verify-all`, but that script isn't used anywhere as well.... Maybe not in this PR, but we'll have to do some cleanup at some point

hack/verify-vendor.sh
```diff
@@ -0,0 +1,60 @@
...
+            fi
+            ;;
+        *) echo "skipping check, dep >= 0.5 required";;
+    esac
+elif [ -f go.mod ]; then
+    echo "Repo uses 'go mod'."
+    # shellcheck disable=SC2235
+    if [ "${JOB_NAME}" ] &&
+           ( [ "${JOB_TYPE}" != "presubmit" ] ||
+                 [ "$( (git diff "${PULL_BASE_SHA}..HEAD" -- go.mod go.sum vendor release-tools;
```
bertinatto, line 34: If we aren't going to depend on [1] we can remove a few things, like the `Gopkg.tom` check on line 17 and `release-tools` from here.

[1] https://github.com/kubernetes-csi/csi-release-tools/blob/master/verify-vendor.sh

Makefile
```diff
@@ -28,7 +28,7 @@ GOBIN=$(shell pwd)/bin
 
 .PHONY: bin/aws-ebs-csi-driver
 bin/aws-ebs-csi-driver: | bin
-	CGO_ENABLED=0 GOOS=linux go build -ldflags ${LDFLAGS} -o bin/aws-ebs-csi-driver ./cmd/
+	CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags ${LDFLAGS} -o bin/aws-ebs-csi-driver ./cmd/
```
bertinatto, line 31: nit: use GOFLAGS instead?

hack/verify-vendor.sh
gnufied: removed.

hack/verify-vendor.sh
gnufied: added to verify-all

Makefile
```diff
@@ -104,6 +105,14 @@ push-release:
 push:
 	docker push $(IMAGE):latest
 
+.PHONY: test-vendor
+test: test-vendor
+verify: test-vendor
+test-vendor:
```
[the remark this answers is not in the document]
gnufied: renamed.

Makefile
gnufied: if it is all the same, I would like to keep this to explicit flag.

bertinatto:
@leakingtapan LGTM, do you have any comments?

hack/verify-vendor.sh
leakingtapan: looks make verify only does linting now, we need to bring back verify-all to `make verify` target

gnufied: yes, I don't think anyone is calling `./hack/verify-all` directly. I was wondering if it being configured via prow though? Have we configured it there? the `./hack/verify-all` script is broken anyways though - the file `./hack/verify-golint` does not even exist.

wongma7 (approved):
lgtm on my side

hack/verify-vendor.sh
gnufied: I have fixed `make verify` so that it calls `./hack/verify-all`

PR #561, opened 2020-09-15
Create volumes in outpost for outpost instances

ayberk:
Bug Fix

[Outposts Documentation](https://docs.aws.amazon.com/outposts/latest/userguide/what-is-outposts.html) for background.

Currently, we're running oblivious to outposts. Volumes are created in the parent region instead of the outpost itself. In order to fix it, we need to pass the `outpostArn` to [CreateVolume](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html).

I'm leveraging the `outpost-arn` instance metadata to determine if we're in an outpost instance. There are two possible responses from the endpoint:

- The outpost-arn if it's an outpost instance
- `404` if it's a regular non-outpost instance.

I've verified the behavior using [this test code](https://gist.github.com/ayberk/102a8995569d5e9002bbb27db209b42e) on a test outpost.

**What testing is done?** 
Tested on an actual outpost instance and verified it works.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -122,6 +122,7 @@ type Disk struct {
 	CapacityGiB      int64
 	AvailabilityZone string
 	SnapshotID       string
+	OutpostArn       string
 }
 
 // DiskOptions represents parameters to create an EBS volume
@@ -131,6 +132,7 @@ type DiskOptions struct {
 	VolumeType       string
 	IOPSPerGB        int
 	AvailabilityZone string
+	OutpostArn       string
 	Encrypted        bool
 	// KmsKeyID represents a fully qualified resource name to the key to use for encryption.
 	// example: arn:aws:kms:us-east-1:012345678910:key/abcd1234-a123-456a-a12b-a123b4cd56ef
@@ -277,6 +279,12 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		TagSpecifications: []*ec2.TagSpecification{&tagSpec},
 		Encrypted:         aws.Bool(diskOptions.Encrypted),
 	}
+
+	// EBS doesn't handle empty outpost arn, so we have to include it only when it's non-empty
+	if len(diskOptions.OutpostArn) > 0 {
+		request.OutpostArn = aws.String(diskOptions.OutpostArn)
+	}
+
 	if len(diskOptions.KmsKeyID) > 0 {
 		request.KmsKeyId = aws.String(diskOptions.KmsKeyID)
 		request.Encrypted = aws.Bool(true)
@@ -311,7 +319,9 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		return nil, fmt.Errorf("failed to get an available volume in EC2: %v", err)
 	}
 
-	return &Disk{CapacityGiB: size, VolumeID: volumeID, AvailabilityZone: zone, SnapshotID: snapshotID}, nil
+	outpostArn := aws.StringValue(response.OutpostArn)
+
+	return &Disk{CapacityGiB: size, VolumeID: volumeID, AvailabilityZone: zone, SnapshotID: snapshotID, OutpostArn: outpostArn}, nil
 }
 
 func (c *cloud) DeleteDisk(ctx context.Context, volumeID string) (bool, error) {
@@ -479,6 +489,7 @@ func (c *cloud) GetDiskByName(ctx context.Context, name string, capacityBytes in
 		CapacityGiB:      volSizeBytes,
 		AvailabilityZone: aws.StringValue(volume.AvailabilityZone),
 		SnapshotID:       aws.StringValue(volume.SnapshotId),
+		OutpostArn:       aws.StringValue(volume.OutpostArn),
 	}, nil
 }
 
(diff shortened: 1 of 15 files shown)


pkg/cloud/cloud.go
```diff
@@ -269,13 +271,26 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
-	request := &ec2.CreateVolumeInput{
-		AvailabilityZone:  aws.String(zone),
-		Size:              aws.Int64(capacityGiB),
-		VolumeType:        aws.String(createType),
-		TagSpecifications: []*ec2.TagSpecification{&tagSpec},
-		Encrypted:         aws.Bool(diskOptions.Encrypted),
+	var request *ec2.CreateVolumeInput
+
+	// EBS doesn't handle empty outpost arn, so we have to include it only when it's non-empty
+	if len(diskOptions.OutpostArn) > 0 {
```
wongma7, line 277: could declare request outside this if and then set request.OutpostArn in the if to reduce some of the duplication.

ayberk:
/retest

Ayberk

pkg/cloud/cloud.go
```diff
@@ -498,6 +508,7 @@ func (c *cloud) GetDiskByID(ctx context.Context, volumeID string) (*Disk, error)
 		VolumeID:         aws.StringValue(volume.VolumeId),
 		CapacityGiB:      aws.Int64Value(volume.Size),
 		AvailabilityZone: aws.StringValue(volume.AvailabilityZone),
+		OutpostArn:       aws.StringValue(volume.OutpostArn),
```
wongma7, line 511: does GetDiskByName also need this change? IT looks like GetDiskByName can get called in CreateVolume

wongma7: could verify this with a test case where GetDiskByName does return something instead of errnotfound

wongma7:
lgtm but i think there is an edgecase where GetDiskByName returns an existing volume that needs tobe handled.
/hold
for end to end test with real outpost

pkg/cloud/cloud.go
ayberk: Great catch! If the volume exists, we would return the existing disk without the outpost arn. I've updated `GetDiskByName` and added a test case. Should be good now.

wongma7:
/lgtm
nice work!!!
Please post testing result, like the output of `kubectl get pv -o yaml <name of some outpost pv created by the driver>` and/or `kubectl get csinodeinfo -o yaml <name of some outpost node where the csi driver is installed>` for sanity check then let's unhold and merge.. If you create a cluster with *only* outpost nodes, it should be possible to reuse the e2e tests, if there is a mix of outpost with non outpost nodes then I guess it will be random whether the test pods get scheduled to the outpost nodes or not...

ayberk:
I made minor changes and was able to run the dynamic provisioning example on a test outpost. Volume was attached properly.

```
ubuntu@ip-10-0-1-194:~/aws-ebs-csi-driver/examples/kubernetes/dynamic-provisioning$ kubectl describe pv pvc-ecbd15ab-0a37-4910-972d-2f9a1487a8e6
Name:              pvc-ecbd15ab-0a37-4910-972d-2f9a1487a8e6
Labels:            <none>
Annotations:       pv.kubernetes.io/provisioned-by: ebs.csi.aws.com
Finalizers:        [kubernetes.io/pv-protection external-attacher/ebs-csi-aws-com]
StorageClass:      ebs-sc
Status:            Bound
Claim:             default/ebs-claim
Reclaim Policy:    Delete
Access Modes:      RWO
VolumeMode:        Filesystem
Capacity:          4Gi
Node Affinity:
  Required Terms:
    Term 0:        topology.ebs.csi.aws.com/partition in [aws]
                   topology.ebs.csi.aws.com/account-id in [foo]
                   topology.ebs.csi.aws.com/outpost-id in [bar]
                   topology.ebs.csi.aws.com/zone in [us-east-1d]
                   topology.ebs.csi.aws.com/region in [us-east-1]
Message:
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            ebs.csi.aws.com
    FSType:            ext4
    VolumeHandle:      vol-0372dc90eb689430d
    ReadOnly:          false
    VolumeAttributes:      storage.kubernetes.io/csiProvisionerIdentity=1601421491105-8081-ebs.csi.aws.com
Events:                <none>
```

```
ubuntu@ip-10-0-1-194:~/aws-ebs-csi-driver/examples/kubernetes/dynamic-provisioning$ kubectl describe node ip-10-0-1-194
Name:               ip-10-0-1-194
Roles:              master
Labels:             
                    beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/os=linux
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=ip-10-0-1-194
                    kubernetes.io/os=linux
                    node-role.kubernetes.io/master=
                    topology.ebs.csi.aws.com/account-id=foo
                    topology.ebs.csi.aws.com/outpost-id=bar
                    topology.ebs.csi.aws.com/partition=aws
                    topology.ebs.csi.aws.com/region=us-east-1
                    topology.ebs.csi.aws.com/zone=us-east-1d
Annotations:        csi.volume.kubernetes.io/nodeid: {"ebs.csi.aws.com":"i-0954d4e23b3bf2da0"}
                    kubeadm.alpha.kubernetes.io/cri-socket: /var/run/dockershim.sock
                    node.alpha.kubernetes.io/ttl: 0
                    projectcalico.org/IPv4Address: 10.0.1.194/24
                    projectcalico.org/IPv4VXLANTunnelAddr: 192.168.245.0
                    volumes.kubernetes.io/controller-managed-attach-detach: true
```

sandeep173254:
@ayberk I am trying to do a poc with outpost but every time the volume is created in the parent region. this is my ec2 instance metadata . Any pointers would be helpful 
[test.json](https://github.com/user-attachments/files/16182686/test.json)

Issue #560, opened 2020-09-11
Migration tests are constantly failing


jsafrane:
/kind bug

**What happened?**
https://testgrid.k8s.io/sig-aws-ebs-csi-driver#migration-test-latest is completely red.

From its logs:

```
/home/prow/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/e2e-migration /home/prow/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver
go: k8s.io/kubernetes@v1.16.1 requires
	vbom.ml/util@v0.0.0-20160121211510-db5cfe13f5cc: unrecognized import path "vbom.ml/util": https fetch: Get "https://vbom.ml/util?go-get=1": dial tcp: lookup vbom.ml on 10.63.240.10:53: no such host
./hack/run-e2e-test: line 115: /home/prow/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/bin/ginkgo: No such file or directory

```

I did not dig deep into this, just few observations:

1. Kubernetes 1.16 is really old
2. I would assume that after #555 no remote module fetch is necessary.

wongma7:
yeah the migration test has its own go.mod, I guess to isolate it from the kubernetes libs of the driver itself so we can test a newer version without updating libs of driver: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/tests/e2e-migration. So need to repeat the same vbox.ml fix to stop trying to pull from there
/assign

wongma7:
isolating the go.mod is probably not needed anymore,since the lib version being used by the driver is in fact newer and definitely contains the migration tests already

wongma7:
They are now running but there are still some failures. In this one all the tests passed but it still reported failure for some reason... https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-aws-ebs-csi-driver-migration-test-latest/1306841052396654592

wongma7:
They are failing now because, based on my primitive method of checking, calls are still going to the in-tree driver instead of csi despite migration being enabled. Not sure why, I think anyway we should flip the 'MigrationComplete' flag


Port forwarding is not yet ready
Forwarding from 127.0.0.1:10252 -> 10252
Forwarding from [::1]:10252 -> 10252
Handling connection for 10252
Port forwarding is succeeded
Handling connection for 10252
Handling connection for 10252
[...]

PR #562, opened 2020-09-15
Remove migration go.mod

wongma7:

**What is this PR about? / Why do we need it?** fix (maybe) https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/560
since the whole repo already has a dependency on k8s.io/kubernetes, it seems unnecessary to have dedicated this go.mod for pulling in the in-tree tests. The test should use the repo's vendored dependency on k8s.io/kubernetes instead of its own go.mod one

**What testing is done?** None, the periodic job is failing anyway so will need to see if it starts succeeding or at least failing in a different way.


wongma7:
/retest 

2 io2 tests timed out

wongma7:
@jsafrane ptal this should fix the migration test or at least get it to run again

Issue #566, opened 2020-09-18
Enable user-provided tags on volume snapshots


chrishenzie:
**Is your feature request related to a problem? Please describe.**
Right now we have an `--extra-volume-tags` flag for enabling user-supplied tags on provisioned EBS volumes. I would like to extend this to also tag snapshots of these EBS volumes.

**Describe the solution you'd like in detail**
- Add an `--extra-snapshot-tags` flag for user-supplied tags on EBS snapshots
- Pass these tags in the `cloud.CreateSnapshot()` call

**Describe alternatives you've considered**
- Reuse the existing `--extra-volume-tags` flag and pass the same tags to both provisioned EBS volumes and snapshots.
  - Optionally, we could rename this flag to be more reflect that it affects more than just volumes
    - This would be a backwards incompatible change, any thoughts around this?
- Provision snapshots using the tags that are already set on the EBS volume
  - Potential problem, if tags are set on the volume outside of K8s, these would be propagated

@msau42

msau42:
I prefer reusing --extra-volume-tags since I'm not sure when you wouldn't want to add the tag to both. But agree that it would be a breaking change to rename it. cc @leakingtapan @wongma7

PR #568, opened 2020-09-24
Apply extra volume tags to EBS snapshots

chrishenzie:

Adding a new feature, see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/566

This PR takes the `--extra-volume-tags` that are applied to provisioned EBS volumes in `CreateVolume()`, and also adds them provisioned snapshots via `CreateSnapshot()`.

**What testing is done?** 

- Included a unit test that asserts the `snapshotOptions` received by `cloud.CreateSnapshot()` includes the supplied 
`extraVolumeTags`
- Included a unit test that asserts the `snapshotOptions` received by `cloud.CreateSnapshot()` includes the cluster ID if that driver option is enabled (See PR feedback)
- Included a unit test that asserts the request object sent to the AWS API is correct
  - This test caught a bug where we were passing a pointer to a loop variable which resulted in duplicate tags being passed to the API
- Performed manual end to end testing in an AWS cluster, created a VolumeSnapshot resource and saw the provisioned snapshot have the additional tags applied to it

### DIFF
--- a/cmd/options/controller_options.go
+++ b/cmd/options/controller_options.go
@@ -24,15 +24,19 @@ import (
 
 // ControllerOptions contains options and configuration settings for the controller service.
 type ControllerOptions struct {
+	// ExtraTags is a map of tags that will be attached to each dynamically provisioned
+	// resource.
+	ExtraTags map[string]string
 	// ExtraVolumeTags is a map of tags that will be attached to each dynamically provisioned
 	// volume.
+	// DEPRECATED: Use ExtraTags instead.
 	ExtraVolumeTags map[string]string
-	// ID of the kubernetes cluster. This is used only to create the same tags on volumes that
-	// in-tree volume volume plugin does.
+	// ID of the kubernetes cluster.
 	KubernetesClusterID string
 }
 
 func (s *ControllerOptions) AddFlags(fs *flag.FlagSet) {
-	fs.Var(cliflag.NewMapStringString(&s.ExtraVolumeTags), "extra-volume-tags", "Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
+	fs.Var(cliflag.NewMapStringString(&s.ExtraTags), "extra-tags", "Extra tags to attach to each dynamically provisioned resource. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
+	fs.Var(cliflag.NewMapStringString(&s.ExtraVolumeTags), "extra-volume-tags", "DEPRECATED: Please use --extra-tags instead. Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
 	fs.StringVar(&s.KubernetesClusterID, "k8s-tag-cluster-id", "", "ID of the Kubernetes cluster used for tagging provisioned EBS volumes (optional).")
 }
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -533,7 +533,9 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
 
 	var tags []*ec2.Tag
 	for key, value := range snapshotOptions.Tags {
-		tags = append(tags, &ec2.Tag{Key: &key, Value: &value})
+		copiedKey := key
+		copiedValue := value
+		tags = append(tags, &ec2.Tag{Key: &copiedKey, Value: &copiedValue})
 	}
 	tagSpec := ec2.TagSpecification{
 		ResourceType: aws.String("snapshot"),
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -199,7 +199,7 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		volumeTags[resourceLifecycleTag] = ResourceLifecycleOwned
 		volumeTags[NameTag] = d.driverOptions.kubernetesClusterID + "-dynamic-" + volName
 	}
-	for k, v := range d.driverOptions.extraVolumeTags {
+	for k, v := range d.driverOptions.extraTags {
 		volumeTags[k] = v
 	}
 
@@ -441,9 +441,22 @@ func (d *controllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		klog.V(4).Infof("Snapshot %s of volume %s already exists; nothing to do", snapshotName, volumeID)
 		return newCreateSnapshotResponse(snapshot)
 	}
-	opts := &cloud.SnapshotOptions{
-		Tags: map[string]string{cloud.SnapshotNameTagKey: snapshotName},
+
+	snapshotTags := map[string]string{
+		cloud.SnapshotNameTagKey: snapshotName,
 	}
+	if d.driverOptions.kubernetesClusterID != "" {
+		resourceLifecycleTag := ResourceLifecycleTagPrefix + d.driverOptions.kubernetesClusterID
+		snapshotTags[resourceLifecycleTag] = ResourceLifecycleOwned
+		snapshotTags[NameTag] = d.driverOptions.kubernetesClusterID + "-dynamic-" + snapshotName
+	}
+	for k, v := range d.driverOptions.extraTags {
+		snapshotTags[k] = v
+	}
+	opts := &cloud.SnapshotOptions{
+		Tags: snapshotTags,
+	}
+
 	snapshot, err = d.cloud.CreateSnapshot(ctx, volumeID, opts)
 
 	if err != nil {
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -58,7 +58,7 @@ type Driver struct {
 
 type DriverOptions struct {
 	endpoint            string
-	extraVolumeTags     map[string]string
+	extraTags           map[string]string
 	mode                Mode
 	volumeAttachLimit   int64
 	kubernetesClusterID string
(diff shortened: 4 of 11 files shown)


chrishenzie:
I did some further digging, and observe the following:

- Create a volume, the number of tags do not matter
- Create a snapshot with zero tags, everything works
- Create a snapshot with one tag, everything works
- Create a snapshot with more than one tag, you get the `ValidationException` error above

This API behavior seems unintuitive to me. I could not find any references to this specific `ValidationException` error on duplicate keys.

Maybe a workaround is to update the snapshot with tags after it has been created, but that seems like a bandaid.

pkg/driver/controller.go
```diff
@@ -440,6 +440,9 @@ func (d *controllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	opts := &cloud.SnapshotOptions{
 		Tags: map[string]string{cloud.SnapshotNameTagKey: snapshotName},
 	}
+	for k, v := range d.driverOptions.extraVolumeTags {
```
msau42, line 443: Do you also want to add cluster id tag, similar to how it's done for CreateVolume?

msau42 (commented):
generally lgtm
/assign @bertinatto @wongma7 
especially regarding reusing the "extra-volume-tags" arg.

pkg/driver/controller.go
chrishenzie: Added this and an additional unit test to cover this case.

pkg/cloud/cloud.go
```diff
@@ -521,8 +522,14 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
...
 	for key, value := range snapshotOptions.Tags {
-		tags = append(tags, &ec2.Tag{Key: &key, Value: &value})
+		copiedKey := key
+		copiedValue := value
+		tags = append(tags, &ec2.Tag{Key: &copiedKey, Value: &copiedValue})
 	}
+	// Enforce ordering for unit tests.
+	sort.SliceStable(tags, func(i, j int) bool {
+		return *tags[i].Key < *tags[j].Key
+	})
```
bertinatto, line 532: Is this sorting really necessary? If so, can we potentially move it to the test function?

pkg/driver/controller.go
```diff
@@ -437,9 +437,22 @@ func (d *controllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		klog.V(4).Infof("Snapshot %s of volume %s already exists; nothing to do", snapshotName, volumeID)
 		return newCreateSnapshotResponse(snapshot)
 	}
+
+	snapshotTags := map[string]string{
+		cloud.SnapshotNameTagKey: snapshotName,
+	}
+	if d.driverOptions.kubernetesClusterID != "" {
```
bertinatto, line 444: Can you also update the usage string of `k8s-tag-cluster-id` in `controller_options.go`?

bertinatto (commented):
Overal LGTM, just 2 nits.

I'd like to propose we mark the `extra-volume-tags` option as `Deprecated` and make it an alias to a new option with a more representative name. Later on we can remove `extra-volume-tags` and keep only the new one.

pkg/cloud/cloud.go
chrishenzie: Removed in favor of a custom matcher for this AWS input type. Let me know if this is what you were thinking.

pkg/driver/controller.go
chrishenzie: I removed the sentence about feature-parity with the in-tree driver since the in-tree driver does not tag snapshots. Let me know if the new sentence is sufficient.

chrishenzie:
@bertinatto One concern I have with this general `--extra-tags` approach is that it's a little less flexible for customers.

Suppose customers want different tags for snapshots than they do volumes. With this solution, they will need to combine their tags for snapshots and volumes into one string, which could potentially exceed the limit of the max # of tags.

Since this driver only provisions either EBS volumes or snapshots, it seems like another alternative would be to introduce `--extra-snapshot-tags` and use that for just snapshots. If users want the same tags applied to both, they can provide the same value for both flags.

cmd/options/controller_options.go
```diff
@@ -24,15 +24,16 @@ import (
...
+	ExtraTags map[string]string
 	// ID of the kubernetes cluster. This is used only to create the same tags on volumes that
 	// in-tree volume volume plugin does.
 	KubernetesClusterID string
 }
 
 func (s *ControllerOptions) AddFlags(fs *flag.FlagSet) {
-	fs.Var(cliflag.NewMapStringString(&s.ExtraVolumeTags), "extra-volume-tags", "Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
+	fs.Var(cliflag.NewMapStringString(&s.ExtraTags), "extra-tags", "Extra tags to attach to each dynamically provisioned resource. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
+	fs.Var(cliflag.NewMapStringString(&s.ExtraTags), "extra-volume-tags", "DEPRECATED: Please use --extra-tags instead. Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
```
bertinatto, line 36: If both flags are set, is it possible that the deprecated one will overwrite the other? Is there a way we can give `extra-tags` priority?

Perhaps we can:

1. Keep `ExtraVolumeTags` above
2. Keep the `driver.WithExtraVolumeTags` line in `cmd/main.go` 
3.  Keep the `WithExtraVolumeTags` in `driver.go` and change it to only set `o.extraTags` if it hasn't been set yet.

Does that makes sense?

bertinatto (commented):
`

cmd/options/controller_options.go
bertinatto: The other changes LGTM.

bertinatto:
> @bertinatto One concern I have with this general `--extra-tags` approach is that it's a little less flexible for customers.
> 
> Suppose customers want different tags for snapshots than they do volumes. With this solution, they will need to combine their tags for snapshots and volumes into one string, which could potentially exceed the limit of the max # of tags.
> 
> Since this driver only provisions either EBS volumes or snapshots, it seems like another alternative would be to introduce `--extra-snapshot-tags` and use that for just snapshots. If users want the same tags applied to both, they can provide the same value for both flags.

Sorry, I missed this comment. That's a valid point, but I'm not sure if we need this flexibility at the moment (I don't have any use cases in mind). To me it feels like using a single flag is OK at this point, but we can always split it if we come across a strong use case.

cmd/options/controller_options.go
chrishenzie: Yeah that makes sense. Let me know if [this captures it](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/568/commits/9b993086ea47c565b6e945927879df3d809501a2).

cmd/options/controller_options.go
```diff
@@ -24,15 +24,18 @@ import (
 
 // ControllerOptions contains options and configuration settings for the controller service.
 type ControllerOptions struct {
+	// ExtraTags is a map of tags that will be attached to each dynamically provisioned
+	// resource.
+	ExtraTags map[string]string
 	// ExtraVolumeTags is a map of tags that will be attached to each dynamically provisioned
 	// volume.
 	ExtraVolumeTags map[string]string
```
bertinatto, line 32: nit: could you add a comment [like this](https://github.com/bertinatto/kubernetes/blob/604569482fc34d010a33339f000e75ecbbf345c6/cmd/kubelet/app/options/options.go#L156)?

pkg/driver/driver.go
```diff
@@ -146,9 +146,17 @@ func WithEndpoint(endpoint string) func(*DriverOptions) {
...
+	return func(o *DriverOptions) {
+		o.extraTags = extraTags
+	}
+}
+
 func WithExtraVolumeTags(extraVolumeTags map[string]string) func(*DriverOptions) {
 	return func(o *DriverOptions) {
-		o.extraVolumeTags = extraVolumeTags
+		if o.extraTags == nil {
+			o.extraTags = extraVolumeTags
```
bertinatto, line 158: nit: could you add a warning saying this flag is deprecated, that it will be removed in the future and that the the user should use `--extra-tags` instead?

bertinatto (commented):
Just 2 nits regarding the deprecated flag. Other than that, I think we can merge this!

I'm not sure how many people out there is using this flag, so IMO it's worth to be very "loud" about it :+1:

chrishenzie:
@bertinatto I've squashed these into one commit.

See here for the deprecation comment:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/cfda96946ec0329121d1d143dffe02953ce35f5a/cmd/options/controller_options.go#L32

And here for the warning:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/cfda96946ec0329121d1d143dffe02953ce35f5a/pkg/driver/driver.go#L161-L164

I added an extra check for `extraVolumeTags != nil` so we don't trip this warning when a user doesn't supply `--extra-volume-tags`.

pkg/driver/driver.go
```diff
@@ -150,9 +150,18 @@ func WithEndpoint(endpoint string) func(*DriverOptions) {
...
+func WithExtraTags(extraTags map[string]string) func(*DriverOptions) {
+	return func(o *DriverOptions) {
+		o.extraTags = extraTags
+	}
+}
+
 func WithExtraVolumeTags(extraVolumeTags map[string]string) func(*DriverOptions) {
 	return func(o *DriverOptions) {
-		o.extraVolumeTags = extraVolumeTags
+		if o.extraTags == nil && extraVolumeTags != nil {
```
bertinatto, line 161: Ideally we'd have 2 checks: 1 for printing the warning if `extraVolumeTags != nil` and the other to overwrite `o.extraTags`. But I think this is fine for now.

bertinatto (commented):
Thank you for working on this @chrishenzie!

/lgtm

bertinatto:
/test pull-aws-ebs-csi-driver-e2e-single-az

Failures seem unrelated to this change.

bertinatto:
/retest
@wongma7 @leakingtapan the AZ job seems fairly broken [1], do you have info on that?

[1] https://testgrid.k8s.io/sig-aws-ebs-csi-driver#e2e-test-single-az

chrishenzie:
Ran `make test` locally, did not encounter these vendor version issues that prow is experiencing.

bertinatto (commented):
/lgtm

PR #577, opened 2020-10-03
[helm] add tag options and update csi-provisioner

kcking:
Feature
Add options to create EBS volume tags that the in-tree plugin creates
**What testing is done?** 
Tested within our EKS 1.17 cluster, confirmed that new volumes have correct names/tags.

### DIFF
--- a/aws-ebs-csi-driver/Chart.yaml
+++ b/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.7.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.6.0
+version: 0.6.1
 kubeVersion: ">=1.13.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
--- a/aws-ebs-csi-driver/templates/controller.yaml
+++ b/aws-ebs-csi-driver/templates/controller.yaml
@@ -50,6 +50,9 @@ spec:
             {{- if .Values.extraVolumeTags }}
               {{- include "aws-ebs-csi-driver.extra-volume-tags" . | nindent 12 }}
             {{- end }}
+            {{- if .Values.k8sTagClusterId }}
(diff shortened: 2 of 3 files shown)


kcking:
Check CLA again

leakingtapan (commented):
/ok-to-test
Could you update the chat version as well?

kcking:
Sure thing, just bumped it to `0.6.1`

jwesolowski-rms:
@kcking this looks great. @leakingtapan Is it possible to get this merged in assuming all looks fine to you?

jwesolowski-rms (approved):

jwesolowski-rms (changes requested):
Whoops. Just saw one thing. @kcking for consistency, I'm thinking adding the new `k8sTagClusterId` and `extraCreateMetadata` values to the values.yaml with a default and comment would be helpful. This would be similar to what we have for [enableVolumeScheduling](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/aws-ebs-csi-driver/values.yaml#L38-L39).

kcking:
Sure thing. I'm thinking a default of `kubernetes` for `k8sTagClusterId` as I think the in-tree plugin creates volumes with a prefix of `kubernetes`. Alternatively we could not specify by default but leave it defined in a comment

Then should we default `extraCreateMetadata` to true or false?

jwesolowski-rms:
I think I would lean towards mimicking the defaults in the controller and csi-provisioner itself:
-  `k8sTagClusterId` defaults to `""` ([link](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/44ec7de2c953b52c9fd39e85ca2183aab6465d32/cmd/options/controller_options.go#L37))
- `extraCreateMetadata` defaults to `false` ([link](https://github.com/kubernetes-csi/external-provisioner/blob/8b0707649212d770624008edbd127f312121aff9/cmd/csi-provisioner/csi-provisioner.go#L72))

kcking:
@jwesolowski-rms thank you for the links! Just added them

jwesolowski-rms (approved):
@kcking thanks for the modifications! Looks perfect.

jwesolowski-rms:
@leakingtapan any update on this?

Issue #516, opened 2020-06-01
Install with Helm 3 fails


danil-smirnov:
/kind bug

**What happened?**

When use a command mentioned in README with Helm 3 I get error `Error: unknown flag: --name`:
```
helm install \
    --namespace kube-system \
    --name aws-ebs-csi-driver \
    --set enableVolumeScheduling=true \
    --set enableVolumeResizing=true \
    --set enableVolumeSnapshot=true \
    https://github.com/kubernetes-sigs/aws-ebs-csi-driver/releases/download/v0.5.0/helm-chart.tgz
```
**What you expected to happen?**

Chart should be installed

**How to reproduce it (as minimally and precisely as possible)?**

Try the command above with Helm 3.

**Anything else we need to know?**:

The correct command to install the cart in Helm 3 should be as follows:
```
helm upgrade --install aws-ebs-csi-driver \
    --namespace kube-system \
    --set enableVolumeScheduling=true \
    --set enableVolumeResizing=true \
    --set enableVolumeSnapshot=true \
    https://github.com/kubernetes-sigs/aws-ebs-csi-driver/releases/download/v0.5.0/helm-chart.tgz
```

**Environment**
- Kubernetes version (use `kubectl version`):

Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-21T14:51:23Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.8-eks-e16311", GitCommit:"e163110a04dcb2f39c3325af96d019b4925419eb", GitTreeState:"clean", BuildDate:"2020-03-27T22:37:12Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}

- Helm version:

version.BuildInfo{Version:"v3.1.2", GitCommit:"d878d4d45863e42fd5cff6743294a11d28a9abce", GitTreeState:"clean", GoVersion:"go1.13.8"}

- Driver version:

N/A

bastienbc:
Hi I cannot install the helm chart with helm 2 either.
[...]

PR #578, opened 2020-10-04
Fixing Helm install command

danil-smirnov:
See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/516


danil-smirnov:
CLA signed

wongma7:
Works for me, thanks.

helm 3 has been out for a while now so I think it's safe to assume most are using it and we don't need to include helm 2 instructions anymore.

/lgtm
/approve

Issue #575, opened 2020-10-02
io1 volumes with <100 iops fail to provision


kcking:
/kind bug

**What happened?**
provisioning `io1` volume with < 100 IOPS fails

**What you expected to happen?**
provision the volume with a min of 100 IOPS instead of failing, as this was the behavior of the in-tree EBS plugin

**How to reproduce it (as minimally and precisely as possible)?**
provision an `io1` volume with < 100 IOPS

**Anything else we need to know?**:
Maybe this is intended behavior, but was surprising to me as we were migrating our infra over from in-tree to this driver

**Environment**
- Kubernetes version (use `kubectl version`): 1.17
- Driver version: 0.6

wongma7:
I think it is better to throw an error than silently accept and correct an invalid input.

Can you share your StorageClass? Do you trigger this situation by setting a low iopsPerGB value in it? Since even with CSI migration enabled there is no guarantee of compatibility between in-tree and CSI StorageClasses, I don't think we are violating compatibility by tweaking the validation around iopsPerGB.

Open to discussion.

msau42:
I agree that it's better to error than silently round up, so that users are not surprised when they get their bill.

For CSI migration, you could preserve in-tree behavior by rounding up when you translate the in-memory StorageClass parameters.

ayberk:
I, too, concur that we should keep this behavior, but at the same time we shouldn't surprise migrating from the in-tree plugin. Maybe it's worth having a section in the README to note differences for migration.

kcking:
Totally hear the point of a surprise bill. Yes the storage class in question had a very low IOPS/GB and was a 16GB volume because it was a testing environment.

I'm not using storage class migration because AFAICT it is not enabled in k8s 1.17 (the current latest for EKS). Instead I had my k8s database operator (foundationdb) migrate the cluster to the new CSI storage class, but ran into errors because the CSI driver could not bring up any io1 volumes with that configuration. I had to manually intervene and fix the storage class (I just changed it to gp2).

Perhaps this edge case is a rare one, but documentation regarding the io1 storage class difference between the in-tree plugin and the csi driver might have saved me from this suprise :)

wongma7:
[...]

PR #582, opened 2020-10-07
Document behavior wrt minimum and maximum iops

wongma7:

**What is this PR about? / Why do we need it?** unlike in-tree, we error if the iops of a volume is below min or above max. Document it to avoid surprises like https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/575

**What testing is done?** N/A

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -33,7 +33,7 @@ There are several optional parameters that could be passed into `CreateVolumeReq
 |-----------------------------|-----------------------------------|----------|---------------------|
 | "csi.storage.k8s.io/fsType" | xfs, ext2, ext3, ext4             | ext4     | File system type that will be formatted during volume creation |
 | "type"                      | io1, io2, gp2, sc1, st1,standard  | gp2      | EBS volume type     |
-| "iopsPerGB"                 |                                   |          | I/O operations per second per GiB. Required when io1 or io2 volume type is specified |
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -33,7 +33,7 @@ There are several optional parameters that could be passed into `CreateVolumeReq
 |-----------------------------|-----------------------------------|----------|---------------------|
 | "csi.storage.k8s.io/fsType" | xfs, ext2, ext3, ext4             | ext4     | File system type that will be formatted during volume creation |
 | "type"                      | io1, io2, gp2, sc1, st1,standard  | gp2      | EBS volume type     |
-| "iopsPerGB"                 |                                   |          | I/O operations per second per GiB. Required when io1 or io2 volume type is specified |
+| "iopsPerGB"                 |                                   |          | I/O operations per second per GiB. Required when io1 or io2 volume type is specified. If this value multiplied by the size of a requested volume produces a value below the minimum or above the maximum IOPs allowed for the volume type, AWS will return an error and volume creation will fail |
```
ayberk, line 36: We should probably add a link to the limits: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html

Funnily enough I don't see the minimum IOPS on the chart.

ayberk (approved):

docs/README.md
wongma7: added a link.

Yeah it's buried below "io1 and io2 volumes can range in size from 4 GiB to 16 TiB. You can provision from 100 IOPS ..."

PR #583, opened 2020-10-08
Fix-helm-private-registry-issues

JonathanLachapelle:

This is a bug fix, the image pull secret is not taken into consideration

Currently, when using a private registry, it's not possible to use the imagePullSecret because the variable wasn't used in the controller.yaml, node.yaml and statefulset.yaml

**What testing is done?** 

Deploying using a private registry and secret

### DIFF
--- a/aws-ebs-csi-driver/values.yaml
+++ b/aws-ebs-csi-driver/values.yaml
@@ -29,6 +29,11 @@ sidecars:
     repository: quay.io/k8scsi/csi-node-driver-registrar
     tag: "v1.1.0"
 
+snapshotController:
+  image:
+    repository: quay.io/k8scsi/snapshot-controller
+    tag: "v2.1.1"
+
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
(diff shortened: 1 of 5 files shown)


aws-ebs-csi-driver/values.yaml
```diff
@@ -29,6 +29,11 @@ sidecars:
     repository: quay.io/k8scsi/csi-node-driver-registrar
     tag: "v1.1.0"
 
+snapshotController:
```
wongma7, line 32: should this be considered a sidecar like the other images? imo yes

wongma7: My bad, it is not a sidecar, and we aren't supposed to bei nstalling it. REF: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/635

wongma7:
Actually I think we should avoid installing the controller altogether. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/635 

Sorry, I got confused myself since there is a snapshot controller and a sidecar... not the same thing. We are responsible only for the sidecar, the controller being in the helm chart is a mistake that I'll fix . In meantime, you can install it manually

PR #593, opened 2020-10-16
Set CSIMigrationAWSComplete for migration tests

wongma7:

**What is this PR about? / Why do we need it?** IDK if checking metrics' plugin_name https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/560#issuecomment-698543689 is a valid way of testing whether migration is working or not, better to set CSIMigrationAWSComplete to turn off the in-tree plugin entirely.

**What testing is done?** N/A, the migration suite is currently failing

### DIFF
--- a/tester/migration-test-config.yaml
+++ b/tester/migration-test-config.yaml
@@ -13,6 +13,7 @@ cluster:
             CSIBlockVolume: "true"
             CSIMigration: "true"
             CSIMigrationAWS: "true"
+            CSIMigrationAWSComplete: "true"
             ExpandCSIVolumes: "true"
             VolumeSnapshotDataSource: "true"
             CSIInlineVolume: "true"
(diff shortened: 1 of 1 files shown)


wongma7:
Since translation happens before the volume is added to desired state of world and so before any operations are done on it, https://github.com/kubernetes/kubernetes/blob/e23d83eead3b5ae57731afb0209f4a2aaa4009dd/pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go#L583 I would expect that if migration is enabled the metrics should have plugin_name ebs.csi.aws.com NOT kubernetes.io/aws-ebs ...

wongma7:
io2 flake

wongma7:
io2 again, i am checking if the account has some kind of volume type specific limit ???

/retest

Issue #591, opened 2020-10-15
DaemonSet Tolerations empty


groodt:
/kind bug

**What happened?**
Rendering the manifests like this:
`kubectl kustomize "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=v0.7.0"`

Results in this resource:
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  labels:
    app.kubernetes.io/name: aws-ebs-csi-driver
  name: ebs-csi-node
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app: ebs-csi-node
      app.kubernetes.io/name: aws-ebs-csi-driver
  template:
    metadata:
      labels:
        app: ebs-csi-node
        app.kubernetes.io/name: aws-ebs-csi-driver
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: eks.amazonaws.com/compute-type
                operator: NotIn
                values:
                - fargate
      containers:
      - args:
        - node
        - --endpoint=$(CSI_ENDPOINT)
        - --logtostderr
        - --v=5
        env:
        - name: CSI_ENDPOINT
          value: unix:/csi/csi.sock
        image: amazon/aws-ebs-csi-driver:v0.7.0
        livenessProbe:
          failureThreshold: 5
          httpGet:
            path: /healthz
            port: healthz
          initialDelaySeconds: 10
          periodSeconds: 10
          timeoutSeconds: 3
        name: ebs-plugin
        ports:
        - containerPort: 9808
          name: healthz
          protocol: TCP
        securityContext:
          privileged: true
        volumeMounts:
        - mountPath: /var/lib/kubelet
          mountPropagation: Bidirectional
          name: kubelet-dir
        - mountPath: /csi
          name: plugin-dir
        - mountPath: /dev
          name: device-dir
      - args:
        - --csi-address=$(ADDRESS)
        - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
        - --v=5
        env:
        - name: ADDRESS
          value: /csi/csi.sock
        - name: DRIVER_REG_SOCK_PATH
          value: /var/lib/kubelet/plugins/ebs.csi.aws.com/csi.sock
        image: quay.io/k8scsi/csi-node-driver-registrar:v1.1.0
[...]

PR #594, opened 2020-10-17
Fix default toleration

krmichel:
bug fix
This fixes issue #591

### DIFF
--- a/aws-ebs-csi-driver/templates/controller.yaml
+++ b/aws-ebs-csi-driver/templates/controller.yaml
@@ -32,7 +32,8 @@ spec:
       affinity: {{ toYaml . | nindent 8 }}
       {{- end }}
       tolerations:
-        - operator: Exists
+        - key: CriticalAddonsOnly
+          operator: Exists
       {{- with .Values.tolerations }}
 {{ toYaml . | indent 8 }}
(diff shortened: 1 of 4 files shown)


krmichel:
/assign @leakingtapan
Can I get an ok to test for this?

leakingtapan:
The taint was CriticalAddonOnly earlier but got removed due to some issue. Let me find it out

krmichel:
> The taint was CriticalAddonOnly earlier but got removed due to some issue. Let me find it out

Okay sounds good.  That is kind of what I thought but I looked for a minute and couldn't find the commit where it got removed and thought maybe I just botched a merge.  Let me know what you find.  I can certainly remove it and make it so tolerations only show up if the user adds their own.

leakingtapan:
here is the link: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/526#issuecomment-649998806

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/526 should be the right change, extra tolerations should be specified through `.Values.tolerations`. And if you care about turning of all toleration, please set `.Values.tolerateAllTaints` to false

krmichel:
Closing in favor of #526

PR #596, opened 2020-10-20
add sidecar container support for arm64 for kubernetes +1.17

muhyousri:

Although the ebs-csi-plugin is updated to support arm -   almost all the CSI container container are outdated 

Note : I added a new base to give a new option as this will not be supported by older kubernetes version , it's only supported starting v.1.17 , feel free to restructure it as you see

1- I Updated the image version according to the documentation here
https://kubernetes-csi.github.io/docs/sidecar-containers.html

2- Updated Arguments changed in the new version of some of the sidecar containers 

https://github.com/kubernetes-csi/external-provisioner/commit/69ce21cc29e02c082194643df1d920c7e96fb026#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5
https://github.com/kubernetes-csi/external-attacher/commit/58bdbaf5929905494754504a5475f081d26a2ed6#diff-db1b4e664fc5c6203530702df4e7c6eade0fd481ef5badffbc0bb4a92648d36eR114

3- Updated the attacher cluster role according to the new permissions required by the new version

https://github.com/kubernetes-csi/external-attacher/blob/master/CHANGELOG/CHANGELOG-3.0.md#changelog-since-v220

**What testing is done?** 

Tested on v1.17 ,v1.18

### DIFF
--- a/deploy/kubernetes/base-arm64/clusterrole-attacher.yaml
+++ b/deploy/kubernetes/base-arm64/clusterrole-attacher.yaml
@@ -1,0 +1,26 @@
+---
+# Source: aws-ebs-csi-driver/templates/clusterrole-attacher.yaml
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ebs-external-attacher-role
+  labels:
+    app.kubernetes.io/name: aws-ebs-csi-driver
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["csi.storage.k8s.io"]
+    resources: ["csinodeinfos"]
+    verbs: ["get", "list", "watch"]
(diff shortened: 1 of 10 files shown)


leakingtapan:
/ok-to-test

Is the k8s.gcr.io/sig-storage/ images pullable without auth into gcp? If this is the case, I would replace the existing base and helm images with the k8s.gcr one. And we don't need to worry about keep different version of yamls in sync

leakingtapan:
@muhyousri also please sign the CLA to get PR merged. thx

muhyousri:
Hello @leakingtapan , Thanks for looking into it 

The images are pullable without authentication , the issue is the latest versions of the sidecar containers will not work with versions earlier that 1.17 , so replacing it will cause the driver to fail on older clusters , that's why I created a new base to maybe referred in the docs as the version working on arm64 for +1.17 as the current one will not work on arm

leakingtapan:
The latest v0.7.0 already supports v1.18 cluster and earlier. So we will make v0.8.0 support v1.17+, which is safe to me. Please make sure to update the helm chart required cluster version to reflect this

ayberk:
Superseded by #707.

Thanks!

/close

Issue #537, opened 2020-07-30
VolumeSnapshotClass not in  "snapshot.storage.k8s.io/v1beta1"


PSjoe:
/kind bug

**What happened?**
Following the directions for CSI volume snapshots:  https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/snapshot 

Step 1 returns with:  
`error: unable to recognize "specs/classes/snapshotclass.yaml": no matches for kind "VolumeSnapshotClass" in version "snapshot.storage.k8s.io/v1beta1"`

**What you expected to happen?**
VolumeSnapshotClass should be created.

**How to reproduce it (as minimally and precisely as possible)?**
Run through example deployment in aws-ebs-csi-driver/examples/kubernetes/snapshot/ on a newly instantiated EKS cluster running 1.17.

**Environment**
- Kubernetes version (use `kubectl version`):
`    Server Version: GitVersion:"v1.17.6-eks-4e7f64", 
    Client Version:  GitVersion:"v1.17.7-eks-bffbac"`
- Driver version:
` kubectl apply -k "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=master"`
Also attempted with the 0.5.0 tag with the same results.

CSI driver appears to be running:  

    kubectl get pods --all-namespaces`
    NAMESPACE     NAME                                  READY   STATUS    RESTARTS   AGE
     kube-system   aws-node-tkfvb                        1/1     Running   0          42h
     kube-system   coredns-55c5fcd78f-kj828              1/1     Running   0          43h
     kube-system   coredns-55c5fcd78f-mlqms              1/1     Running   0          43h
     kube-system   ebs-csi-controller-78bc69cb98-7bfcm   4/4     Running   0          3s
     kube-system   ebs-csi-controller-78bc69cb98-w9fc9   4/4     Running   0          3s
     kube-system   ebs-csi-node-vg4j5                    3/3     Running   0          3s
     kube-system   kube-proxy-cx66g                      1/1     Running   0          42h`

And according to this:  https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/   the feature gate for VolumeSnapshotDataSource should be enabled by default in 1.17.

kingli-crypto:
I am facing the same issue. I believe this is due to VolumeSnapshotDataSource is not enabled in EKS 1.17.
[...]

PR #607, opened 2020-11-05
Update README.md

robisoh88:
No
Volume Snapshots README.md is missing steps to install Volume Snapshot beta CRDs (volumesnapshotclass, volumesnapshotcontents, volumesnapshots). Following the Volume Snapshot example alone does not work. Need to update the document. See below for more info.
**What testing is done?** 

Following the guide without having CRDs EBS CSI Driver fails with "VolumeSnapshotClass not in "snapshot.storage.k8s.io/v1beta1""(see [issue](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/537#issuecomment-692076110)). Moreover the EBS CSI Driver with Alpha features is the correct driver that need to be installed. Installing the stable version of the ebs csi driver will not include external snapshotter sidecar, snapshot controller  or service account/RBAC. Pull request is to mention this in the Volume Snapshot example README.

### DIFF
--- a/examples/kubernetes/snapshot/README.md
+++ b/examples/kubernetes/snapshot/README.md
@@ -10,7 +10,8 @@ This driver implements basic volume snapshotting functionality using the [extern
 
 1. The `VolumeSnapshotDataSource` must be set in `--feature-gates=` in the `kube-apiserver`.
 
-1. The [aws-ebs-csi-driver driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) is installed.
+1. The [aws-ebs-csi-driver driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) with alpha features is installed.
+1. If your cluster does not come pre-installed with Snapshot Beta CRDs, install the [Snapshot Beta CRDs](https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd)
 
 ### Usage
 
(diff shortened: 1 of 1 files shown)


robisoh88:
I signed it

examples/kubernetes/snapshot/README.md
```diff
@@ -10,7 +10,8 @@ This driver implements basic volume snapshotting functionality using the [extern
 
 1. The `VolumeSnapshotDataSource` must be set in `--feature-gates=` in the `kube-apiserver`.
 
-1. The [aws-ebs-csi-driver driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) is installed.
+1. The [aws-ebs-csi-driver driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) with alpha features is installed.
+1. If your cluster does not come pre-installed with Snapshot Beta CRDs, install the [Snapshot Beta CRDs](https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd)
```
ayberk, line 14: I think it'd be more useful to link directly to installation part of their README: https://github.com/kubernetes-csi/external-snapshotter#usage

robisoh88: Thanks @ayberk , I agree on this especially should any further changes be made on where their files are stored in the future it would be better to link directly to their README. Please see the new commit.

ayberk:
/ok-to-test
/lgtm
/approve

Thank you!

Issue #622, opened 2020-11-25
Resize Volume fails if capacity is modified in PVC for volume of type io1


jahantech:
/kind bug

**What happened?**
Resized the volume by modifying the capacity in the PVC resource. However the resize failed with following error:
```
Warning  VolumeResizeFailed  26s  volume_expand  error expanding volume "default/test-pod" of plugin "kubernetes.io/aws-ebs": AWS modifyVolume failed for vol-03a0dcd92e5535e4f with InvalidParameterCombination: The parameter iops must be specified for io1 volumes.
```

**What you expected to happen?**
Successful resizing of the volume should have occurred.

**How to reproduce it (as minimally and precisely as possible)?**
- Create a storage class with storage type io1
- Create a persistent volume claim using the above storage class and try to modify the storage

**Anything else we need to know?**:
Possible root cause of the issue is the missing IOPS parameter in the ResizeDisk method: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/6d6e8b76f656789e6356e98ce4f9002c1b333e5c/pkg/cloud/cloud.go#L865
However it is specified in CreateDisk method: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/6d6e8b76f656789e6356e98ce4f9002c1b333e5c/pkg/cloud/cloud.go#L300
**Environment**
- Kubernetes version (use `kubectl version`): `Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.12", GitCommit:"7cd5e9086de8ae25d6a1514d0c87bac67ca4a481", GitTreeState:"clean", BuildDate:"2020-11-12T09:18:55Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.12-eks-31566f", GitCommit:"31566f851673e809d4d667b7235ed87587d37722", GitTreeState:"clean", BuildDate:"2020-10-20T23:25:14Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}`

jahantech:
Workaround was to modify the volume storage and iops via console/api and patch the PV with the correct value. This issue is related to kubernetes.io/aws-ebs provisioner. Closing this issue.

sebradloff:
[...]

PR #626, opened 2020-12-01
Provide "Iops" field when resizing io1 and io2 volumes

zuzzas:

Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/622

Adds `Iops` field set to the current volume's `Iops` field to avoid `InvalidParameterCombination: The parameter iops must be specified for io1 volumes.` error from AWS API.

**What testing is done?**

Manual testing in a Kubernetes cluster.


ayberk:
This is not needed anymore. Issue was fixed on the EBS side.

/close

PR #631, opened 2020-12-02
Allow the driver to perform AWS STS AssumeRole when talking to AWS APIs

solarkennedy:
RFC: I would like to be able to assume into a different role than what the existing instance profile has, and also not use static IAM creds. This new feature allows the driver to assume role based on a `AWS_STS_ROLE` variable, and maintain fresh creds.

I wasn't sure how else to do this without modifying the code. The AWS golang sdk doesn't seem to have the built-in capability without a code change.

Testing done: Compiled and run on environment. Permission errors before, no permission errors after.

If this change is desired (and I'm not doing it wrong), I can add docs to the readme as an alternative way to get creds, update the helm charts for a new variable to be passed in, etc.

### DIFF
--- a/cmd/options/controller_options.go
+++ b/cmd/options/controller_options.go
@@ -33,10 +33,13 @@ type ControllerOptions struct {
 	ExtraVolumeTags map[string]string
 	// ID of the kubernetes cluster.
 	KubernetesClusterID string
+	// AWS Assume Role ARN
+	AssumeRoleArn string
 }
 
 func (s *ControllerOptions) AddFlags(fs *flag.FlagSet) {
 	fs.Var(cliflag.NewMapStringString(&s.ExtraTags), "extra-tags", "Extra tags to attach to each dynamically provisioned resource. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
 	fs.Var(cliflag.NewMapStringString(&s.ExtraVolumeTags), "extra-volume-tags", "DEPRECATED: Please use --extra-tags instead. Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
 	fs.StringVar(&s.KubernetesClusterID, "k8s-tag-cluster-id", "", "ID of the Kubernetes cluster used for tagging provisioned EBS volumes (optional).")
+	fs.StringVar(&s.AssumeRoleArn, "assume-role-arn", "", "IAM ARN for the controller to assume-role into (optional). If unset, uses whatever credentials available to the aws-sdk.")
 }
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -25,10 +25,13 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+
 	dm "github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/devicemanager"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -227,14 +230,24 @@ var _ Cloud = &cloud{}
 
 // NewCloud returns a new instance of AWS cloud
 // It panics if session is invalid
-func NewCloud(region string) (Cloud, error) {
-	return newEC2Cloud(region)
+func NewCloud(region string, assumeRoleArn string) (Cloud, error) {
+	return newEC2Cloud(region, assumeRoleArn)
 }
 
-func newEC2Cloud(region string) (Cloud, error) {
+func newEC2Cloud(region string, assumeRoleArn string) (Cloud, error) {
 	awsConfig := &aws.Config{
 		Region:                        aws.String(region),
 		CredentialsChainVerboseErrors: aws.Bool(true),
+	}
+	sess := session.Must(session.NewSession(awsConfig))
+
+	if assumeRoleArn != "" {
+		if arn.IsARN(assumeRoleArn) {
+			klog.V(5).Infof("Assuming role into ARN %s", assumeRoleArn)
+			awsConfig.Credentials = stscreds.NewCredentials(sess, assumeRoleArn)
+		} else {
+			return nil, fmt.Errorf("Provided ARN: '%s' isn't valid. Make sure it starts with 'arn:' and has the correct number of colons", assumeRoleArn)
+		}
 	}
 
 	endpoint := os.Getenv("AWS_EC2_ENDPOINT")
@@ -245,7 +258,7 @@ func newEC2Cloud(region string) (Cloud, error) {
 	return &cloud{
 		region: region,
 		dm:     dm.NewDeviceManager(),
-		ec2:    ec2.New(session.Must(session.NewSession(awsConfig))),
+		ec2:    ec2.New(sess, awsConfig),
 	}, nil
 }
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -35,6 +35,7 @@ func main() {
 		driver.WithMode(options.DriverMode),
 		driver.WithVolumeAttachLimit(options.NodeOptions.VolumeAttachLimit),
 		driver.WithKubernetesClusterID(options.ControllerOptions.KubernetesClusterID),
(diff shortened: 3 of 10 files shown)


wongma7:
it looks like the go sdk only respects AWS_ACCESS_KEY_ID/AWS_ACCESS_KEY, AWS_SECRET_ACCESS_KEY/AWS_SECRET_KEY, AWS_SESSION_TOKEN

https://github.com/aws/aws-sdk-go/blob/d87d92b4412739c6ef8917979e922fae98b7654c/aws/credentials/env_provider.go

since AWS_STS_ROLE is not a standard environment variable I would prefer an explicit flag like --aws-sts-role or --aws-role-t-assume or something with some nice help text wdyt?

solarkennedy:
Yes, I can try plumbing down `--aws-sts-role` from the cli.

solarkennedy:
Ok, I went with the more explicit `--aws-assume-role`.
At the same time I'm also working on getting CLA approval.

solarkennedy:
Ok @wongma7 , I got my CLA signed. Let me know what I can do next.

pkg/cloud/cloud.go
```diff
@@ -217,15 +218,21 @@ var _ Cloud = &cloud{}
...
+func newEC2Cloud(region string, awsAssumeRole string) (Cloud, error) {
 	awsConfig := &aws.Config{
 		Region:                        aws.String(region),
 		CredentialsChainVerboseErrors: aws.Bool(true),
 	}
+	sess := session.Must(session.NewSession(awsConfig))
+
+	if awsAssumeRole != "" {
+		klog.V(5).Infof("Assuming role into ARN %s", awsAssumeRole)
+		awsConfig.Credentials = stscreds.NewCredentials(sess, awsAssumeRole)
```
wongma7, line 234: I think these creds will expire. Is there a way to use the provider https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/#AssumeRoleProvider to ensure sdk autorefershes them

solarkennedy: Whoops, I misread the docs and thought this was the thing that does that.
Thank you. I'll adjust.

solarkennedy: No, re-reading again, I think this code is correct.

Using `NewCredentials` is how the example code says to do it:
```// Create the credentials from AssumeRoleProvider to assume the role
// referenced by the "myRoleARN" ARN.
creds := stscreds.NewCredentials(sess, "myRoleArn")```
https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/

And you can see in the sdk code, `NewCredentials` uses that provider for you:
https://github.com/aws/aws-sdk-go/blob/v1.36.7/aws/credentials/stscreds/assume_role_provider.go#L257

Also empirically, my `aws-ebs-csi-driver controller` process running on this branch has been up since dec4 and can still attach/detach.

Comment (solarkennedy) [reply pkg/cloud/cloud.go:239]: Bump. I've rebased and squashed on top of master as well.

**Comment by solarkennedy** [comment]
Bump. It has been a month. Can I have a second round of review?

File: cmd/options/controller_options.go [hunk 2]
```diff
@@ -33,10 +33,13 @@ type ControllerOptions struct {
 	ExtraVolumeTags map[string]string
 	// ID of the kubernetes cluster.
 	KubernetesClusterID string
+	// AWS Assume Role ARN
+	AWSAssumeRole string
```
ayberk, line 37: Let's rename this to  `AssumeRoleArn`to make the "arn" requirement more explicit.

pkg/cloud/cloud.go
```diff
@@ -227,15 +228,21 @@ var _ Cloud = &cloud{}
 
 // NewCloud returns a new instance of AWS cloud
 // It panics if session is invalid
-func NewCloud(region string) (Cloud, error) {
-	return newEC2Cloud(region)
+func NewCloud(region string, awsAssumeRole string) (Cloud, error) {
```
ayberk, line 231: Similarly let's rename these "assumeRoleArn".

pkg/cloud/cloud.go
ayberk: I think you're right:

```
... The Credentials value will attempt to refresh the credentials using the provider when Credentials.Get is called, if the cached credentials are expiring. 
```

cmd/options/controller_options.go
```diff
@@ -33,10 +33,13 @@ type ControllerOptions struct {
...
 }
 
 func (s *ControllerOptions) AddFlags(fs *flag.FlagSet) {
 	fs.Var(cliflag.NewMapStringString(&s.ExtraTags), "extra-tags", "Extra tags to attach to each dynamically provisioned resource. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
 	fs.Var(cliflag.NewMapStringString(&s.ExtraVolumeTags), "extra-volume-tags", "DEPRECATED: Please use --extra-tags instead. Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'")
 	fs.StringVar(&s.KubernetesClusterID, "k8s-tag-cluster-id", "", "ID of the Kubernetes cluster used for tagging provisioned EBS volumes (optional).")
+	fs.StringVar(&s.AWSAssumeRole, "aws-assume-role", "", "AWS ARN for the controller to assume-role into (optional).")
```
ayberk, line 44: I think AWS is redundant here but I guess we can keep it. But we should definitely have "arn" in this option, eg, `-aws-assume-role-arn`. We should also briefly explain what happens the operator doesn't provide this value since it's optional.

pkg/cloud/cloud.go
```diff
@@ -227,15 +228,21 @@ var _ Cloud = &cloud{}
...
 
-func newEC2Cloud(region string) (Cloud, error) {
+func newEC2Cloud(region string, awsAssumeRole string) (Cloud, error) {
 	awsConfig := &aws.Config{
 		Region:                        aws.String(region),
 		CredentialsChainVerboseErrors: aws.Bool(true),
 	}
+	sess := session.Must(session.NewSession(awsConfig))
+
+	if awsAssumeRole != "" {
```
ayberk, line 242: What's the behavior if this string is not a valid ARN? I'd say we should verify and if it's not, we should return an error and fail the startup. 

Can we add a test case for that?

ayberk (changes requested):
We definitely should have "arn" in the names and handle the invalid ARN case.

Helm chart, more documentation etc can be done in a follow-up PR but they're definitely most welcome.

cmd/options/controller_options.go
solarkennedy: I put a brief something

pkg/cloud/cloud.go
solarkennedy: Well... it seems to do nothing until it actually tries to exercise it? (I'm surprised)

    cloud.go:243] Assuming role into ARN foo

ayberk: We should definitely verify it then and fail early. Can we add a check here? 

Ref: https://docs.aws.amazon.com/sdk-for-go/api/aws/arn/#IsARN

solarkennedy: Added

ayberk:
Thanks for incorporating the feedback. Appreciate the contribution! This LGTM but I want to wait for at least one more review before I merge this to make sure I'm not missing anything. 

/lgtm

wongma7:
/approve 

thanks again

AndyXiangLi:
Running external test to verify this change:
/test pull-aws-ebs-csi-driver-external-test-latest

ayberk:
@solarkennedy we might also need to add a test case for invalid arn to pass the coverall check.

pkg/cloud/cloud.go
```diff
@@ -227,15 +230,25 @@ var _ Cloud = &cloud{}
...
 		Region:                        aws.String(region),
 		CredentialsChainVerboseErrors: aws.Bool(true),
 	}
+	sess := session.Must(session.NewSession(awsConfig))
+
+	if assumeRoleArn != "" {
+		if arn.IsARN(assumeRoleArn) {
+			klog.V(5).Infof("Assuming role into ARN %s", assumeRoleArn)
+			awsConfig.Credentials = stscreds.NewCredentials(sess, assumeRoleArn)
+		} else {
```
AndyXiangLi, line 248: Can we add unit test for this check?

solarkennedy: added something, this one could be a bad idea because it does actually try to create a ec2 session

solarkennedy:
FYI I've hit my timebox and won't be able to justify working on this code more.
Additionally, we've decided not to use this driver.  If you don't think this functionality will be useful to more people, feel free to close this PR if you don't want to maintain this code.

dntosas:
so until this one is merged, we can't utilize something like `kube2iam` to attach roles to pods, right?

ayberk:
Based on my limited understanding of kube2iam, this shouldn't be a blocker for you to use it. From driver's perspective nothing really changes except you're proxying the requests through the kube2iam container. You can just use the annotations and it should work.

Issue #628, opened 2020-12-02
Support for gp3 volumes


Chili-Man:
**Describe the solution you'd like in detail**
AWS announced the new gp3 ebs class; this driver should support provisioning them. 
https://aws.amazon.com/blogs/aws/new-amazon-ebs-gp3-volume-lets-you-provision-performance-separate-from-capacity-and-offers-20-lower-price/

**Describe alternatives you've considered**
We can work with gp2 for now, but gp3 allows to scale the iops independent of the storage size

**Additional context**

wongma7:
Support has been added to driver version v0.8.0, helm chart version v0.7.0, just released (albeit with some hicccups). 

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG-0.x.md

PR #633, opened 2020-12-03
Add EBS gp3 support

samjo-nyang:
Adding new feature

Support a new volume type gp3 (https://aws.amazon.com/about-aws/whats-new/2020/12/introducing-new-amazon-ebs-general-purpose-volumes-gp3/)

**What testing is done?** 
- `make` succeeded
- `make sanity-test` succeeded
- manual gp3 pvc testing (with my internal EKS cluster) succeeded

implements #628

However, `docker run -v "$PWD":/go -it golang:1.14.1-buster make test` triggers a lot of dependency update. Is this expected and should be in this PR? How can I update aws-sdk only?


samjo-nyang:
I signed it

ayberk:
This is great, thank you!

pkg/cloud/cloud.go
```diff
@@ -252,6 +257,10 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	case VolumeTypeIO1, VolumeTypeIO2:
 		createType = diskOptions.VolumeType
 		iops = capacityGiB * int64(diskOptions.IOPSPerGB)
+	case VolumeTypeGP3:
+		createType = diskOptions.VolumeType
+		iops = capacityGiB * int64(diskOptions.IOPSPerGB)
```
Jan-M, line 262: From my point of view I would like no connection seen between volume size and IOPS. gp3 allows you to spec any any value for IOPS once volume exceeds 32gb. It gives end user 3000IOPS base line, disconnected from the size.

samjo-nyang: I also agree on this comment; should I
1. give another option for gp3 only such as "iopsTotal"?
2. give the "iopsTotal" option and migrate io1 and io2 iopsPerGB to iopsTotal?
3. just simply do max(3000, capacityGB * iopsPerGB)?

ayberk: +1. This is the biggest selling point of gp3. You don't need to set iops while creating a gp3 volume(see [doc](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html)], so you can simply delete this branch and add `VolumeTypeGP3` to the first case block. Line 308 below will set the request.Iops if it's passed by the user.

ayberk (changes requested):
Again, great work. Thank you!

We should probably change the default type to gp3 as it's cheaper and faster. What do you think @wongma7?

wongma7:
/ok-to-test

IMO we can make gp3 the default if we release the change in a new minor version, by default the performance is equal or better, only throughput is a bit more complicated since for gp2 it might involve "burst credits" but I don't think we need to worry about it. Also there are volume type specific service quotas, so for example if a user has increased their gp2 TiB quota, they must remember to increase their gp3 one as well, but again I don't think this is worth worrying about just somethin to note.

pkg/cloud/cloud.go
Jan-M: But the first case uses the odd `iopsPerGB` which someone familiar with backwards compatibility would need to judge changing.

So having the new case allows changing to plain iops and throughput.

pkg/cloud/cloud.go
```diff
@@ -299,6 +308,9 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	if iops > 0 {
 		request.Iops = aws.Int64(iops)
 	}
+	if throughput > 0 {
```
ayberk, line 311: Can you change this to `throughput > 0 && diskOptions.VolumeType == VolumeTypeGP3`? Let's play it safe as throughput is a valid parameter only for gp3.

pkg/cloud/cloud.go
ayberk: You're right! We need to add another field "iops" so the users can directly set the value.

docs/README.md
```diff
@@ -29,13 +29,15 @@ The following CSI gRPC calls are implemented:
...
-| "kmsKeyId"                  |                                   |          | The full ARN of the key to use when encrypting the volume. When not specified, the default KMS key is used |
+| Parameters                  | Values                                 | Default  | Description         |
+|-----------------------------|----------------------------------------|----------|---------------------|
+| "csi.storage.k8s.io/fsType" | xfs, ext2, ext3, ext4                  | ext4     | File system type that will be formatted during volume creation |
+| "type"                      | io1, io2, gp2, gp3, sc1, st1,standard  | gp3      | EBS volume type     |
+| "iopsPerGB"                 |                                        |          | I/O operations per second per GiB. Required when io1 or io2 volume type is specified, and optional when gp3 volume type is specified. If this value multiplied by the size of a requested volume produces a value below the minimum or above the maximum IOPs allowed for the volume type, as documented [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html), AWS will return an error and volume creation will fail |
```
ayberk, line 36: `and optional when gp3 volume type is specified` this doesn't apply anymore.

pkg/cloud/cloud.go
```diff
@@ -138,6 +141,8 @@ type DiskOptions struct {
 	Tags             map[string]string
 	VolumeType       string
 	IOPSPerGB        int
+	IOPSTotal        int
```
ayberk, line 144: I think we should follow the API's naming and call this "iops".

tests/e2e/driver/ebs_csi_driver.go
```diff
@@ -109,6 +109,12 @@ func GetParameters(volumeType string, fsType string, encrypted bool) map[string]
 	if iops := IOPSPerGBForVolumeType(volumeType); iops != "" {
 		parameters["iopsPerGB"] = iops
 	}
+	if iopsTotal := IOPSTotalForVolumeType(volumeType); iopsTotal != "" {
+		parameters["iopsTotal"] = iopsTotal
```
ayberk, line 113: I see other fields are not using them, but can you use the constants you defined above? Ie `IopsTotalKey` and `ThroughputKey `

samjo-nyang:
I think that all feedback are applied and all tests are passed.
- change default volume type to gp3
- add a new option "iops"
- fix readme.md

@ayberk @wongma7 @Jan-M 
Is there anything left to do?

ayberk:
Thank you @samjo-nyang! /lgtm, but let's wait for one more review. 

One minor request, can you squash your commits? Ideally we should have one commit for updating the SDK and another one for gp3 support, but you can simply squash into one.

wongma7:
/lgtm
/approve
thank you, we will make a release shortly!!

PR #638, opened 2020-12-07
Add self to OWNERS

ayberk:
To help with triaging and reviews.


gnufied:
I believe owners is managed via - https://github.com/kubernetes/org you need to open a PR there.

PR #639, opened 2020-12-07
Suffix amazonlinux image with -amazonlinux and push debian image to GitHub

wongma7:
**Is this a bug fix or adding new feature?**fix

**What is this PR about? / Why do we need it?**To match cloudbuild.yaml https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/cloudbuild.yaml and in preparation for the next release v0.8.0, amazonlinux based images will be suffixed with -amazonlinux, like v0.8.0-amazonlinux. And debian based images will become the "default" unsuffixed v0.8.0.

At the same time, I noticed crazy-max/ghaction-docker-buildx@v3 was deprecated https://github.com/crazy-max/ghaction-docker-buildx so I rewrote the action to use docker's official github actions.

**What testing is done?** Not sure if testing is possible? given the extensive rewrite I want to test the new scrijpt but it doesn't seem possible without actually triggering the action. So I will check if the latest tag was successfully pushed after this PR and if not will need a follow-up...

### DIFF
--- a/.github/workflows/container-image.yaml
+++ b/.github/workflows/container-image.yaml
@@ -1,53 +1,79 @@
 name: Container Images
 
-on: push
+on:
+  push:
+
 jobs:
-  build:
+  buildx:
     # this is to prevent the job to run at forked projects
     if: github.repository == 'kubernetes-sigs/aws-ebs-csi-driver'
+    env:
+      IMAGE: aws-ebs-csi-driver
+      DEB_BUILD_TAG: aws-ebs-csi-driver:debian
+      AL2_BUILD_TAG: aws-ebs-csi-driver:amazonlinux
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Set up Docker Buildx
-      id: buildx
-      uses: crazy-max/ghaction-docker-buildx@v3
-      with:
-        buildx-version: latest
-        qemu-version: latest
-    - name: Build container image
-      run: |
-        docker buildx build \
-          -t aws-ebs-csi-driver \
-          --platform=linux/arm64,linux/amd64 \
-          --output="type=image,push=false" . \
-          --target=amazonlinux
-    - name: Push to Github registry
-      run: |
-        USER=$(echo $GITHUB_REPOSITORY | cut -d'/' -f1)
-        BRANCH=$(echo $GITHUB_REF | cut -d'/' -f3)
-        IMAGE=aws-ebs-csi-driver
-        if [ "$BRANCH" = "master" ]; then
-          TAG="latest"
-        else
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image.yaml
```diff
@@ -1,53 +1,69 @@
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Push amazonlinux target to Docker Hub
+        run: |
+          AL2_PUSH_TAG="amazon/$IMAGE:$TAG-amazonlinux"
+          docker tag $AL2_BUILD_TAG $AL2_PUSH_TAG
+          docker push $AL2_PUSH_TAG
```
wongma7, line 69: I realized I changed this from docker buildx build --push=true, so this won't push the arm64 manifest.... need to fix.

.github/workflows/container-image.yaml
```diff
@@ -1,53 +1,79 @@
...
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.REGISTRY_TOKEN }}
+      - name: Push debian target to GitHub
+        run: |
+          DEB_PUSH_TAG="docker.pkg.github.com/$GITHUB_REPOSITORY/$IMAGE:$TAG"
+          docker buildx build \
+            -t $DEB_PUSH_TAG \
```
wongma7, line 55: I don't know if this uses the cache from previous build step ($DEB_BUILD_TAG) . If not, there is no point to $DEB_BUILD_TAG/ $AL2_BUILD_TAG

ayberk: Yeah I'm a little bit confused. What's the benefit of build steps?

wongma7: makes it quicker to debug what failed. if build succeed but push step failed. etc.

wongma7: from local testing the build step image is cached so IMO it makes sense to separate the two steps

wongma7:
I don't have permission to restart travis CI, but it is unrelated and not required for merge so I am manually merging.

wongma7:
The arm build fails, I'll figure it out and send follow up

```
2020-12-08T18:34:29.7913665Z #8 [linux/arm64 debian-base 2/3] RUN clean-install ca-certificates e2fsprog...
2020-12-08T18:34:29.7914574Z #8 9.250 standard_init_linux.go:219: exec user process caused: exec format error
2020-12-08T18:34:29.9412982Z #8 ...
```

Issue #637, opened 2020-12-05
Resource constraints are not enforced on all the containers


tirumerla:
**Is your feature request related to a problem?/Why is this needed**

- When you enforce resource constraints on specific namespace. Deployment of aws-ebs-csi-driver using helm chart throws the error since resources is only specified for [`ebs-plugin`](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/aws-ebs-csi-driver/templates/controller.yaml#L95) container but not for `csi-*` containers in `controller.yaml` and none of the containers in `node.yaml`. 

/feature

**Describe the solution you'd like in detail**

- Add resources block for all the containers in `controller.yaml` and in `node.yaml` reading the resources from values file.
```
{{- with .Values.resources }}
resources: {{ toYaml . | nindent 12 }}
{{- end }}
```

**Describe alternatives you've considered**

- Downloaded & Extracted helm chart -> changed yaml files `controller.yaml` and `node.yaml` to have resource blocks for all the containers -> redeployed with the changes.

**Additional context**
- k8s version: `EKS v1.17.11`
- ebs-csi helm chart: `0.7.1`

Happy to raise PR :)

Thanks

wongma7:
PR welcome!

I would be OK with just templating that same .Values.resources to all containers . Some of the sidecars with their own informers might use a lot of memory depending of # of objects in cluster so it might make sense to have per container values. But probably that'll make the chart too complicatged and we can cross that bridge when we get to it...

tirumerla:
> PR welcome!
> 
> I would be OK with just templating that same .Values.resources to all containers . Some of the sidecars with their own informers might use a lot of memory depending of # of objects in cluster so it might make sense to have per container values. But probably that'll make the chart too complicatged and we can cross that bridge when we get to it...

Thanks @wongma7 this makes sense. Raised PR :)

tirumerla:
Fixed with #640

PR #640, opened 2020-12-08
Apply resource constraints to all sidecar containers

tirumerla:

- Fix to apply resource constraints to all side containers.

- See #637 

**What testing is done?** 

- Updated resource constraints for all the side containers and deployed. Deployment was successful without any resource constraints errors on my EKS cluster.


wongma7:
/ok-to-test
/approve
/lgtm

thanks.

PR #642, opened 2020-12-08
Release v0.8.0

wongma7:
**Is this a bug fix or adding new feature?** new release

gp3 feature alone deserves a release.

I didn't bother to prune the changelog, I think i would prefer if it is exhaustive (i.e. contains ALL PRs) but just surface new features/bug fixes to top and keep everything else under misc.

**What testing is done?** TBH I am not sure how long it will take for the k8s.gcr.io image to becsome available since it is my first time doing it so switching the charts/docs to refer to it might be a bit premature. I can't leave them pointing to docker hub either because I still can't push v0.8.0 debian-based image to amazon docker hub, will only push the v0.8.0-amazonlinux one there.


charts/aws-ebs-csi-driver/templates/statefulset.yaml
```diff
@@ -1,30 +1,30 @@
...
-        {{- include "aws-ebs-csi-driver.labels" . | nindent 8 }}
-    spec:
-      serviceAccountName: ebs-snapshot-controller
-      containers:
-        - name: snapshot-controller
-          image: quay.io/k8scsi/snapshot-controller:v2.1.1
-          args:
-            - --v=5
-            - --leader-election=false
-{{- end }}
```
wongma7, line 30: I fixed line endings here and ran the new-ish `make generate-kustomize`

deploy/kubernetes/base/controller.yaml
```diff
@@ -28,7 +28,7 @@ spec:
         - operator: Exists
       containers:
         - name: ebs-plugin
-          image: amazon/aws-ebs-csi-driver:latest
+          image: k8s.gcr.io/aws-ebs-csi-driver:latest
```
wongma7, line 31: actually this should be k8s.gcr.io/provider-aws/aws-ebs-csi-driver

wongma7:
Also since the chart releaser action doesn't have permissions to update gh-pages index I will manually update it with chart-releaser later

CHANGELOG-0.x.md
```diff
@@ -1,3 +1,38 @@
+# v0.8.0
+
+## Notable changes
+- Images will be built on a Debian base by default. Images built on Amazon Linux will still be available but with the tag suffix `-amazonlinux`.
```
ayberk, line 4: Let's definitely add `gp3 is now default` here.

docs/README.md
```diff
@@ -49,6 +50,7 @@ Following sections are Kubernetes specific. If you are Kubernetes user, use foll
 | AWS EBS CSI Driver \ Kubernetes Version| v1.12 | v1.13 | v1.14 | v1.15 | v1.16 | v1.17 | v1.18 |
```
ayberk, line 50: Time to make this 1.18+? :)

CHANGELOG-0.x.md
wongma7: done

docs/README.md
wongma7: yeah.

i was considering csompressing the rows and columns but I think tehre is some value in acknowleding each version explicitly...to a point. we might be past that point

ayberk:
Let me summon the bot 🧙 
/lgtm

PR #647, opened 2020-12-10
Feature: Add ability to customize node daemonset nodeselector

pliu:
Feature

This allows users to customize the nodeSelector field of the daemonset deploying the ebs-csi-nodes the same way one can customize the nodeSelector of the deployment deploying the controllers. This is needed in the case of a hybrid cluster where some nodes are AWS hosts whereas others are not (the ebs-csi-nodes on the non-AWS hosts continuously crashloop).

**What testing is done?** 
Verified that the Chart renders as expected

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.8.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.7.0
+version: 0.7.1
 kubeVersion: ">=1.14.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
--- a/charts/aws-ebs-csi-driver/templates/node.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node.yaml
@@ -31,6 +31,9 @@ spec:
                 - fargate
       nodeSelector:
         kubernetes.io/os: linux
+        {{- with .Values.node.nodeSelector }}
+{{ toYaml . | indent 8 }}
+        {{- end }}
       hostNetwork: true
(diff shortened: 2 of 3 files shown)


ayberk:
Thank you! Will take a look later.

/ok-to-test

ayberk:
Can you bump the chart version to 0.7.1?

pliu:
@ayberk done :)

ayberk:
/lgtm
/approve

Thank you! 

note to self: we should ideally deploy only to aws nodes by default...

pliu:
> /lgtm
> /approve
> 
> Thank you!
> 
> note to self: we should ideally deploy only to aws nodes by default...

I imagine that would be a bit difficult since you'd need a Kubernetes-level label-equivalent (or just a label), that specifies which nodes are backed by AWS hosts. Currently, we're using a custom label to specify which nodes are backed by AWS hosts, but that label would need to be standardized if we wanted to make this default behavior.

Alternatively one could run it on every node and just be a no-op on non-AWS-host-backed nodes (though the argument could be made that that is unacceptable overhead).

ayberk:
Yeah we need a well-known label, which should be doable especially with the cloud-provider extraction. I'm not a big fan of no-op solution because it can be confusing. We have a workaround with this so it's a nice-to-have for later :)

pliu:
> Yeah we need a well-known label, which should be doable especially with the cloud-provider extraction. I'm not a big fan of no-op solution because it can be confusing. We have a workaround with this so it's a nice-to-have for later :)

Oh nice. I'll look into this cloud provider extraction thing (if you have a link I can start with, that'd be appreciated :)).

ayberk:
Context: https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/20180530-cloud-controller-manager.md

We're building a new version of cloud-provider-aws here, although it's in early stages: https://github.com/kubernetes/cloud-provider-aws#readme

PR #653, opened 2020-12-11
Arm overlay

ayberk:
Maintenance

Driver can't be deployed on arm instances because of non-multiarch sidecar images. This PR updates some of the sidecar images and adds an overlay with multiarch images.

**What testing is done?** 
Manual testing on my cluster + e2e tests.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/clusterrole-attacher.yaml
+++ b/charts/aws-ebs-csi-driver/templates/clusterrole-attacher.yaml
@@ -8,7 +8,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["persistentvolumes"]
-    verbs: ["get", "list", "watch", "update"]
+    verbs: ["get", "list", "watch", "update", "patch"]
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list", "watch"]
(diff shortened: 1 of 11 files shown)


charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -122,7 +122,6 @@ spec:
             - --csi-address=$(ADDRESS)
             - --v=5
             - --leader-election=true
-            - --leader-election-type=leases
```
wongma7, line 125: we don't need this because the default changed to leases right? just checking

wongma7: or they removed the option an dit's leases only?

ayberk: Yeah they removed it and it's leases.

charts/aws-ebs-csi-driver/templates/clusterrole-attacher.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["persistentvolumes"]
-    verbs: ["get", "list", "watch", "update"]
+    verbs: ["get", "list", "watch", "update", "patch"]
```
wongma7, line 11: did you test the helm install /kustomize command? like if I have helm installed a prevoius version them helm install this version, i would assume that helm is smart enough to delete/recreate this object.

Issue #656, opened 2020-12-13
ARM64 aws-ebs-csi-driver images have x86_64 binaries


sipsma:
/kind bug

**What happened?**
The arm64 aws-ebs-csi-driver images worked a few days ago but now all fail to run on arm64 machines. When I pull the images down and inspect them, they appear to all actually by x86 images even though they are labeled arm64. For example:
```
root@ip-10-0-0-243:~# ctr image pull --platform linux/arm64/v8 k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v0.8.0
k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v0.8.0:                                resolved       |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:a8ed0a90b49c16301f3ca934b75db63b8d5c8d68c377a865bce15c9af4a12ff5: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:a5e67acf34d2f09cafca63fd41b6f9898f216ee98dde8ddf235d8edef660baf5:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:4ba180b702c8b29f4d24cb24eeb91a26a5924e797a806bbca1b73d26b5fe8b1e:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:5c3bea1f2f52c6b3cc8be0463f49ba95a2490377d0f848670d1af2ed2d4c6882:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:592701b57e30c44d07074686878f910b68eaa2d4afc75d3c6d53bb58c8807cbc:   done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.3 s                                                                    total:   0.0 B (0.0 B/s)                                         
unpacking linux/arm64/v8 sha256:a8ed0a90b49c16301f3ca934b75db63b8d5c8d68c377a865bce15c9af4a12ff5...
done
root@ip-10-0-0-243:~# mkdir tmp
root@ip-10-0-0-243:~# ctr image mount --platform linux/arm64/v8 k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v0.8.0 tmp
sha256:5bd728051e5a37f591e0acd0c50a51c6871e6953267f5be1c9021f892aca14a0
tmp
root@ip-10-0-0-243:~# file tmp/bin/aws-ebs-csi-driver 
[...]

PR #658, opened 2020-12-15
Use buildx in cloudbuild

wongma7:

cloudbuild is not magic, if arm64 images are desired then buildx must b ecalled

**What testing is done?** 
command works locally, don't have access to cloudbuild to test it in cloudbuild

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -2,18 +2,26 @@ timeout: 1200s
 steps:
   - name: gcr.io/cloud-builders/docker
     args:
+      - buildx
       - build
       - --tag=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver:$_GIT_TAG
       - --tag=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver:latest
-      - --target=debian-base
+      - --platform=linux/arm64,linux/amd64
+      - --output="type=image,push=false"
       - .
+      - --target=debian-base
+      - HOME=/root
   - name: gcr.io/cloud-builders/docker
     args:
+      - buildx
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -2,18 +2,26 @@ timeout: 1200s
 steps:
   - name: gcr.io/cloud-builders/docker
     args:
+      - HOME=/root
```
wongma7, line 5: https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing#simple-build-example
"    # default cloudbuild has HOME=/builder/home and docker buildx is in /root/.docker/cli-plugins/docker-buildx
    # set the home to /root explicitly to if using docker buildx
    # - HOME=/root"

cloudbuild.yaml
```diff
@@ -2,18 +2,26 @@ timeout: 1200s
...
+      - buildx
```
wongma7, line 6: tried to make command look like https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/.github/workflows/container-image.yaml#L25

cloudbuild.yaml
```diff
@@ -2,18 +2,26 @@ timeout: 1200s
 steps:
   - name: gcr.io/cloud-builders/docker
     args:
+      - buildx
       - build
```
ayberk, line 6: do we still need this?

ayberk: oh gotcha `dockerx build...`

wongma7:
cc https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/656

PR #659, opened 2020-12-16
(Try to) fix cloudbuild

wongma7:
**Is this a bug fix or adding new feature?** fix cloudbuild.yaml

**What is this PR about? / Why do we need it?** the attempt to do multiarch cloudbuilds from https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/658 failed horribly with "Step #0: unknown flag: --tag". I think it has something to do with buildx not being available so I am shuffling that stray "HOME=/root" thing that I csopy/pasted from the doc to someplace that makes sense to me. Also sinc ethe github workflow has to enable multiarch using the qemu image then I guess this does too.

**What testing is done?** 
```
docker run -e HOME=/root --entrypoint=bash -it  gcr.io/cloud-builders/docker
root@084b1a2e938b:/# DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build --tag=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver:latest .
[+] Building 0.0s (0/0)
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -2,28 +2,37 @@ timeout: 1200s
 steps:
   - name: gcr.io/cloud-builders/docker
     args:
+      - run
+      - --rm
+      - --privileged
+      - multiarch/qemu-user-static
+      - --reset
+      - -p
+      - "yes"
+  - name: gcr.io/cloud-builders/docker
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
+      - HOME=/root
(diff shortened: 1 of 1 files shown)


wongma7:
Progress, but not there yet.


Step #1: "docker buildx build" requires exactly 1 argument. 

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/aws-ebs-csi-driver-push-images/1339014979134689280

PR #663, opened 2020-12-18
 Bump kubernetes dependencies to 1.19.5 and rest automatically with go mod tidy

wongma7:
**Is this a bug fix or adding new feature?** routine bump

Notably we depend on kubernetes/kubernetes for the testsuites for migraiton testing. I'll send a follow up PR and keep an eye on migration tests.

**What testing is done?** N/A, CI must catch.


wongma7:
I think I'll bump to 1.20

PR #589, opened 2020-10-14
WIP: Rebasing PR #524, Volume metrics

matthiaslee:
rebases #524, (which rebased #421 and fixed #223 and #499)

**Is this a bug fix or adding new feature?**

**What is this PR about? / Why do we need it?**

**What testing is done?**

PR #677, opened 2021-01-04
add volume stats metrics -

### LINKED ISSUE #223: Report Metrics for the driver
Report Metrics for the driver

TBD

**Ref**
* https://github.com/container-storage-interface/spec/issues/236

### LINKED ISSUE #499: Report volume usage
How can I identify the volume usage ?

as of now aws-ebs-csi-driver emits 
`kube_persistentvolume_capacity_bytes`
but it doesn't have anything like 
`kube_persistentvolume_usage_bytes`

Ideally its should support standard metrics as below
```
kubelet_volume_stats_inodes_used
kubelet_volume_stats_used_bytes
kubelet_volume_stats_available_bytes
kubelet_volume_stats_capacity_bytes
kubelet_volume_stats_inodes
kubelet_volume_stats_inodes_free
```

AndyXiangLi:
Fixes #223
Fixes #499 
Implement Volume stats metrics

**What testing is done?** 
Added unit test for the change

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -19,9 +19,12 @@ package driver
 import (
 	"context"
 	"fmt"
+	"golang.org/x/sys/unix"
+	"k8s.io/kubernetes/pkg/volume"
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
@@ -65,6 +68,7 @@ var (
 	nodeCaps = []csi.NodeServiceCapability_RPC_Type{
 		csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME,
 		csi.NodeServiceCapability_RPC_EXPAND_VOLUME,
+		csi.NodeServiceCapability_RPC_GET_VOLUME_STATS,
 	}
 )
 
@@ -344,7 +348,66 @@ func (d *nodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 }
 
 func (d *nodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "NodeGetVolumeStats is not implemented yet")
+	klog.V(4).Infof("NodeGetVolumeStats: called with args %+v", *req)
+	if len(req.VolumeId) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodeGetVolumeStats volume ID was empty")
+	}
+	if len(req.VolumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodeGetVolumeStats volume path was empty")
+	}
+
+	exists, err := d.mounter.ExistsPath(req.VolumePath)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "unknown error when stat on %s: %v", req.VolumePath, err)
+	}
+	if !exists {
+		return nil, status.Errorf(codes.NotFound, "path %s does not exist", req.VolumePath)
+	}
+
+	isBlock, err := d.IsBlockDevice(req.VolumePath)
+
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to determine whether %s is block device: %v", req.VolumePath, err)
+	}
+	if isBlock {
+		bcap, err := d.getBlockSizeBytes(req.VolumePath)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to get block capacity on path %s: %v", req.VolumePath, err)
+		}
+		return &csi.NodeGetVolumeStatsResponse{
+			Usage: []*csi.VolumeUsage{
+				{
+					Unit:  csi.VolumeUsage_BYTES,
+					Total: bcap,
+				},
+			},
+		}, nil
+	}
+
+	metricsProvider := volume.NewMetricsStatFS(req.VolumePath)
+
+	metrics, err := metricsProvider.GetMetrics()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get fs info on path %s: %v", req.VolumePath, err)
+	}
+
+	return &csi.NodeGetVolumeStatsResponse{
+		Usage: []*csi.VolumeUsage{
+			{
+				Unit:      csi.VolumeUsage_BYTES,
+				Available: metrics.Available.AsDec().UnscaledBig().Int64(),
+				Total:     metrics.Capacity.AsDec().UnscaledBig().Int64(),
+				Used:      metrics.Used.AsDec().UnscaledBig().Int64(),
+			},
+			{
+				Unit:      csi.VolumeUsage_INODES,
+				Available: metrics.InodesFree.AsDec().UnscaledBig().Int64(),
+				Total:     metrics.Inodes.AsDec().UnscaledBig().Int64(),
+				Used:      metrics.InodesUsed.AsDec().UnscaledBig().Int64(),
+			},
+		},
+	}, nil
+
 }
 
 func (d *nodeService) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetCapabilitiesRequest) (*csi.NodeGetCapabilitiesResponse, error) {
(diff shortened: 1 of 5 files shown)


pkg/driver/node.go
```diff
@@ -344,7 +348,64 @@ func (d *nodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+
+			{
+				Available: inodesFree,
+				Total:     inodes,
+				Used:      inodesUsed,
+				Unit:      csi.VolumeUsage_INODES,
+			},
+		},
+	}, nil
+
```
gnufied, line 408: Can you post output of this passing csi sanity tests? I assume it did but just to be sure.

pkg/driver/statter.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+type Statter interface {
+	StatFS(path string) (int64, int64, int64, int64, int64, int64, error)
+	IsBlockDevice(string) (bool, error)
+}
+
+var _ Statter = realStatter{}
+
+type realStatter struct {
```
ayberk, line 29: we should name this simply `statter`.

pkg/driver/statter.go
```diff
@@ -0,0 +1,80 @@
...
+
+type fakeStatter struct{}
+
+func NewFakeStatter() fakeStatter {
+	return fakeStatter{}
+}
+
+func (fakeStatter) StatFS(path string) (available, capacity, used, inodesFree, inodes, inodesUsed int64, err error) {
+	// Assume the file exists and give some dummy values back
+	return 1, 1, 1, 1, 1, 1, nil
```
ayberk, line 75: Why don't you return back whatever is passed to this function? It'd be more useful for testing.

pkg/driver/statter.go
```diff
@@ -0,0 +1,80 @@
...
+
+package driver
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/unix"
+)
+
+type Statter interface {
```
ayberk, line 22: Can you add some documentation? It'd be really helpful going forward.

pkg/driver/node.go
```diff
@@ -550,3 +611,17 @@ func hasMountOption(options []string, opt string) bool {
...
+func (d *nodeService) getBlockSizeBytes(devicePath string) (int64, error) {
+	cmd := d.mounter.Command("blockdev", "--getsize64", devicePath)
+	output, err := cmd.Output()
+	if err != nil {
+		return -1, fmt.Errorf("error when getting size of block volume at path %s: output: %s, err: %v", devicePath, string(output), err)
+	}
+	strOut := strings.TrimSpace(string(output))
+	gotSizeBytes, err := strconv.ParseInt(strOut, 10, 64)
+	if err != nil {
+		return -1, fmt.Errorf("failed to parse size %s into int a size", strOut)
```
ayberk, line 624: nit: `failed to parse size %s as int`

pkg/driver/statter.go
```diff
@@ -0,0 +1,80 @@
...
+
+// IsBlock checks if the given path is a block device
+func (realStatter) IsBlockDevice(fullPath string) (bool, error) {
+	var st unix.Stat_t
+	err := unix.Stat(fullPath, &st)
+	if err != nil {
+		return false, err
+	}
+
+	return (st.Mode & unix.S_IFMT) == unix.S_IFBLK, nil
```
ayberk, line 44: I'm going to assume this only works on unix?

ayberk:
Supersedes #589.

wongma7:
fyi, test case failed

 === RUN   TestNodeGetVolumeStats
    node_test.go:1194: Expected no error, got rpc error: code = NotFound desc = path /test does not exist
    panic.go:636: missing call(s) to *mocks.MockMounter.ExistsPath(is equal to /test) /home/prow/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/node_test.go:1180
    panic.go:636: aborting test due to missing call(s)
--- FAIL: TestNodeGetVolumeStats (0.00s)

pkg/driver/node.go
AndyXiangLi: Yeah, sanity test is passed

pkg/driver/statter.go
AndyXiangLi: Yeah, right now it is only working for unix

pkg/driver/sanity_test.go
```diff
@@ -73,6 +76,15 @@ func TestSanity(t *testing.T) {
 	sanity.Test(t, config)
 }
 
+func createDir(targetPath string) (string, error) {
+	if err := os.MkdirAll(targetPath, 0755); err != nil {
+		if !os.IsExist(err) {
```
ayberk, line 81: you can use `IsNotExist` here.

pkg/driver/sanity_test.go
```diff
@@ -309,13 +321,48 @@ func (f *fakeMounter) GetDeviceName(mountPath string) (string, int, error) {
 }
 
 func (f *fakeMounter) MakeFile(pathname string) error {
+	file, err := os.OpenFile(pathname, os.O_CREATE, os.FileMode(0644))
```
ayberk, line 324: I know we talked about this offline, but I want to note it publicly. 

I am not quite sure why we have a different mounter for sanity tests. We should use the code from the driver itself, otherwise we're not really testing the "real" code. 

/cc @wongma7 @gnufied

ayberk (changes requested):
A couple comments, good work!

pkg/driver/sanity_test.go
wongma7: I think it's fine, I'm not too familiar with sanity but seems mocking the mount is common pattern. https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/2014365ea7b3405e024be1d9552aa30d44da4a8b/pkg/mount-manager/fake-safe-mounter.go#L28

I think the sanity test is still testing everything in the driver code surrounding wheerever it actually tries to call "Mount/Statfs".

pkg/driver/statter.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	statfs := &unix.Statfs_t{}
+	err = unix.Statfs(path, statfs)
+	if err != nil {
+		err = fmt.Errorf("failed to get fs info on path %s: %v", path, err)
+		return
+	}
+
+	// Available is blocks available * fragment size
+	available = int64(statfs.Bavail) * int64(statfs.Bsize)
```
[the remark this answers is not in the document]
AndyXiangLi: Yes, they are all system level variables to my understanding, and we should be able to get all the following value in one call: Type   Bsize  Blocks Bfree  Bavail Files  Ffree  Fsid   NamelenFrsize Flags  Spare

pkg/driver/sanity_test.go
```diff
@@ -73,6 +76,15 @@ func TestSanity(t *testing.T) {
 	sanity.Test(t, config)
 }
 
+func createDir(targetPath string) (string, error) {
+	if err := os.MkdirAll(targetPath, 0755); err != nil {
```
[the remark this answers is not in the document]
AndyXiangLi: Actually we need write and exec permit here, minimum 300 will work. But 755 is the default permit sanity_test used to mkdir, so I kept it..
https://github.com/kubernetes-csi/csi-test/blob/054980205e4fc45ab9e940e39bdfd71066d96d87/pkg/sanity/sanity.go#L365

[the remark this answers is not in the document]
ayberk: I wasn't sure why we would need exec on a folder. TIL :) https://unix.stackexchange.com/a/150456

pkg/driver/sanity_test.go
ayberk: Should be fine as long as we don't modify the mount, but if we do that we risk silently failing sanity tests.

@AndyXiangLi can you add a note to the actual mount code? That it's mirrored here and any modification should be applied here too. Not ideal, but I'd feel safer.

AndyXiangLi: @ayberk Sure will do that, and as we talked off line, I think it would be better if we can extract those three functions out of mounter in the future.

pkg/driver/node_test.go
```diff
@@ -1170,32 +1171,102 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+				mockMetadata := mocks.NewMockMetadataService(mockCtl)
+				mockMounter := mocks.NewMockMounter(mockCtl)
+				VolumePath := "./test"
 
-	awsDriver := nodeService{
-		metadata: mockMetadata,
-		mounter:  mockMounter,
-		inFlight: internal.NewInFlight(),
-	}
+				err := os.MkdirAll(VolumePath, 0644)
```
ayberk, line 1213: looks like there are some indentation problems in this file.

AndyXiangLi: fixed

wongma7:
I'll manually merge, coverage decrease is because of getBlockSizeBytes which is not easy to exercise and anyway coveralls is not supposed to be required! ( longstanding TODO to fix this).

/approve

PR #691, opened 2021-01-12
Add a prerequisite to dynamic provisioning

ronenl1:
New documentation

Emphasise that enableVolumeScheduling should be equal to true, in order to provision EBS volumes dynamically.


ayberk:
/ok-to-test
/lgtm
/approve

thank you!

PR #694, opened 2021-01-13
Refactor e2e testing scripts to be more reusable and use them instead…

wongma7:

**What is this PR about? / Why do we need it?** Although I do not like bash, it seems for this case it is the best tool for the job: run kops, helm, docker, awscli, etc. commands.

By comparison, I have found that the tester binary is harder to develop with (need to fork and build it to make changes since for example kops versions and args are hardcoded into it)

Plus, these scripts can be reused in EFS repo, I just have to figure out some subtree magic. 

See also https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/473, I will continue that PR after this.

**What testing is done?** 
In this repo:
```
TEST_PATH=./tests/e2e-migration/... \
EBS_CHECK_MIGRATION=true \
TEST_ID=18512 \
CLEAN=false \
KOPS_STATE_FILE=s3://mattwon \
AWS_REGION=us-west-2 \
AWS_AVAILABILITY_ZONES=us-west-2a \
GINKGO_FOCUS=Dynamic.\*xfs.\*should.store.data \
GINKGO_NODES=1 \
./hack/e2e/run.sh
```
Same scripts reusable in EFS repo:
```
DRIVER_NAME=aws-efs-csi-driver \
TEST_PATH=./test/e2e/... \
TEST_EXTRA_FLAGS='--cluster-name=$CLUSTER_NAME' \
TEST_ID=18512 \
CLEAN=false \
KOPS_STATE_FILE=s3://mattwon \
AWS_REGION=us-west-2 \
AWS_AVAILABILITY_ZONES=us-west-2a \
GINKGO_FOCUS=should.mount.different.paths \
GINKGO_NODES=1 \
./hack/e2e/run.sh
```


hack/e2e/ebs.sh
```diff
@@ -0,0 +1,52 @@
...
+
+function ebs_check_migration() {
+  loudecho "Checking migration"
+  # There should have been no calls to the in-tree driver kubernetes.io/aws-ebs but many calls to ebs.csi.aws.com
+  # Find the controller-manager log and read its metrics to verify
+  NODE=$(kubectl get node -l kubernetes.io/role=master -o json | jq -r ".items[].metadata.name")
+  kubectl port-forward kube-controller-manager-"${NODE}" 10252:10252 -n kube-system &
+
+  # Ensure port forwarding succeeded
+  while true; do
```
ayberk, line 16: Can we change this to give up after a while? Isn't this going to loop forever if forwarding fails?

wongma7: fixed, see commit "Time out if port forward fails"

waiting for tests to ppass, the snapshot ones failed because I didn't install the snapshot CRD, hence commit " Deploy snapshot CRD "

wongma7:
CI tests passed using the refactored script and output looks good

ayberk:
Appreciate the constant improvement! 

/lgtm

PR #695, opened 2021-01-13
NodeExpandVolume no-op for raw block

### LINKED ISSUE #339: Add unit test for NodeExpand
Add unit test for NodeExpand

### LINKED ISSUE #346: Fix raw block volume resizing for EBS
We need to fix raw block volume resizing (`NodeExpandVolume`) to not resize the volume if given `volume_path` in resize request is a block device and not a mounted file system.

AndyXiangLi:
Fixes #346 
Fixes #339 
NodeExpandVolume will check volume type (mount point or block device)and skip resize on block volume.
**What testing is done?** 
Sanity test passed
Added unit tests

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -254,14 +254,47 @@ func (d *nodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if len(volumeID) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
 	}
+	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+	}
 
-	args := []string{"-o", "source", "--noheadings", "--target", req.GetVolumePath()}
+	volumeCapability := req.GetVolumeCapability()
+	// VolumeCapability is optional, if specified, use that as source of truth
+	if volumeCapability != nil {
+		caps := []*csi.VolumeCapability{volumeCapability}
+		if !isValidVolumeCapabilities(caps) {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("VolumeCapability is invalid: %v", volumeCapability))
+		}
+
+		if blk := volumeCapability.GetBlock(); blk != nil {
+			// Noop for Block NodeExpandVolume
+			klog.V(4).Infof("NodeExpandVolume called for %v at %s. Since it is a block device, ignoring...", volumeID, volumePath)
+			return &csi.NodeExpandVolumeResponse{}, nil
+		}
+	} else {
+		// VolumeCapability is nil, check if volumePath point to a block device
+		isBlock, err := d.IsBlockDevice(volumePath)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to determine device path for volumePath [%v]: %v", volumePath, err)
+		}
+		if isBlock {
+			// Skip resizing for Block NodeExpandVolume
+			bcap, err := d.getBlockSizeBytes(volumePath)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to get block capacity on path %s: %v", req.VolumePath, err)
+			}
+			klog.V(4).Infof("NodeExpandVolume called for %v at %s, since given volumePath is a block device, ignoring...", volumeID, volumePath)
+			return &csi.NodeExpandVolumeResponse{CapacityBytes: bcap}, nil
+		}
+	}
+
+	args := []string{"-o", "source", "--noheadings", "--target", volumePath}
 	output, err := d.mounter.Command("findmnt", args...).Output()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/driver/node.go
```diff
@@ -254,14 +254,43 @@ func (d *nodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if len(volumeID) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
 	}
+	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
```
ayberk, line 259: So prior to this PR, we don't have this requirement (at least here). We have to make sure this is not a breaking change. Did we verify that behavior stays the same when volumePath is absent?

pkg/driver/node.go
```diff
@@ -254,14 +254,43 @@ func (d *nodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
+	// VolumeCapability is optional, if specified, use that as source of truth
+	if volumeCapability != nil {
+		caps := []*csi.VolumeCapability{volumeCapability}
+		if !isValidVolumeCapabilities(caps) {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("VolumeCapability is invalid: %v", volumeCapability))
+		}
+
+		if blk := volumeCapability.GetBlock(); blk != nil {
+			// Noop for Block NodeExpandVolume
+			klog.V(4).Infof("NodeExpandVolume succeeded on %v to %s, capability is block so this is a no-op", volumeID, volumePath)
```
ayberk, line 272: Hmm I feel like we should reword this. Maybe something like `"NodeExpandVolume called for %v at %s. Since %v is a block device, ignoring..."`

What do you think?

pkg/driver/node.go
gnufied: volume_path is required, so we should be fine I think.

pkg/driver/node.go
```diff
@@ -254,14 +254,43 @@ func (d *nodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
+	} else {
+		// VolumeCapability is nil, check if volumePath point to a block device
+		isBlock, err := d.IsBlockDevice(volumePath)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to determine device path for volumePath [%v]: %v", volumePath, err)
+		}
+		if isBlock {
+			// Skip resizing for Block NodeExpandVolume
+			klog.V(4).Infof("NodeExpandVolume succeeded on %v to %s, given volumePath is a block device so will not resize volume", volumeID, volumePath)
+			return &csi.NodeExpandVolumeResponse{}, nil
```
gnufied, line 284: Not that kubernetes will use the returned size(at least right now), but is better to return the expanded size if possible. We may have fixes coming up which uses returned size.

gnufied: Also if you do make the change to returned expanded size, we should return expanded block size not the fs size.

pkg/driver/node_test.go
```diff
@@ -1082,6 +1082,92 @@ func TestNodePublishVolume(t *testing.T) {
...
+				VolumeId:   "test-volume-id",
+				VolumeCapability: &csi.VolumeCapability{
+					AccessType: &csi.VolumeCapability_Block{
+						Block: &csi.VolumeCapability_BlockVolume{},
+					},
+					AccessMode: &csi.VolumeCapability_AccessMode{
+						Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+					},
+				},
+			},
```
gnufied, line 1153: Can we use `losetup` and `dd` to check the other case of volume_path being a block device?

pkg/driver/node.go
AndyXiangLi: Yeah, as per https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/vendor/github.com/container-storage-interface/spec/lib/go/csi/csi.pb.go#L4322
I think we should have this check

pkg/driver/node.go
AndyXiangLi: Hi @ayberk What do you think here?

ayberk: It's a great suggestion. We should future-proof wherever we can.

pkg/driver/node_test.go
AndyXiangLi: I tried to use losetup and dd to create loop device when setup testing but with no luck. It failed to find an available loop device when I ran it in linux instance.. This should not be a blocker as I tested the functionality e2e on the cluster. Testing code is [here](https://github.com/AndyXiangLi/aws-ebs-csi-driver/blob/skip-block-resizing-test/pkg/driver/node_test.go#L1098), I will follow up on this later.

ayberk: I'm going to approve to unblock you, but yeah let's follow up.

PR #697, opened 2021-01-14
feat: Add option to provision StorageClasses

gazal-k:
New feature in the helm chart

Most application charts don't usually create `StorageClass`es. They create PVCs.

**What testing is done?** 
Installed the chart with the following custom values:

```yaml
enableVolumeResizing: true
enableVolumeScheduling: true
enableVolumeSnapshot: true

serviceAccount:
  controller:
    name: ebs-csi-controller
    annotations:
      eks.amazonaws.com/role-arn: arn:aws:iam::123:role/cluster-name-kube-system-ebs-csi-controller
  snapshot:
    create: false
    name: ebs-csi-controller
    annotations:
      eks.amazonaws.com/role-arn: arn:aws:iam::123:role/cluster-name-kube-system-ebs-csi-controller

storageClasses:
- name: ebs-sc
  parameters:
    encrypted: "true"
```

Then tried creating a PVC and a pod that used it:

```yaml
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  storageClassName: ebs-sc
  accessModes:
    - "ReadWriteOnce"
  resources:
    requests:
      storage: 5Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: ebs-app
spec:
  containers:
    - name: app
      image: centos
      command: ["/bin/sh"]
      args: ["-c", "while true; do echo $(date -u) >> /data/out; sleep 5; done"]
      volumeMounts:
        - name: persistent-storage
          mountPath: /data
  volumes:
    - name: persistent-storage
      persistentVolumeClaim:
        claimName: ebs-claim
```

Verified the following:

PV was provisioned.
New EBS gp3 volume was created as expected.

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -101,3 +101,13 @@ serviceAccount:
     create: true
     name: ebs-snapshot-controller
     annotations: {}
+
+storageClasses: []
+# Add StorageClass resources like:
+# - name: ebs-sc
+#   # defaults to WaitForFirstConsumer
+#   volumeBindingMode: WaitForFirstConsumer
+#   # defaults to Delete
+#   reclaimPolicy: Retain
+#   parameters:
+#     encrypted: "true"
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "0.9.0"
+appVersion: "0.9.1"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
 version: 0.9.0
--- a/charts/aws-ebs-csi-driver/templates/storageclass.yaml
+++ b/charts/aws-ebs-csi-driver/templates/storageclass.yaml
@@ -1,0 +1,15 @@
+{{- range .Values.storageClasses }}
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: {{ .name }}
+provisioner: ebs.csi.aws.com
+volumeBindingMode: {{ default "WaitForFirstConsumer" .volumeBindingMode }}
+{{- if hasKey . "reclaimPolicy" }}
+reclaimPolicy: {{ .reclaimPolicy }}
+{{- end }}
+{{- with .parameters }}
+parameters:
+{{ toYaml . | indent 2 }}
+{{- end }}
+{{- end }}


gazal-k:
Similar to https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/298

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -98,3 +98,7 @@ serviceAccount:
     create: true
     name: ebs-snapshot-controller
     annotations: {}
+
+storageClasses: []
```
ayberk, line 102: It'd be really nice if we include the options here similar to the efs PR.

gazal-k: You mean like an example? I do have an example storage class entry in comments just like the efs PR. Just that ebs doesn't really need much explicit config. I can provide a more detailed entry I suppose.

gazal-k: I've added reference `parameters` now. Should we also mention other options like `volumeBindingMode` & `reclaimPolicy`?

ayberk: Sorry for the delay here. 

Yeah that would great. Ideally we should note if they're optional and default value if they have one.

ayberk:
/lgtm
/approve

thanks!

gazal-k:
I thought rebasing on master would fix the tests 🤦. So, please approve again @ayberk

ayberk:
Sorry for the back and forth, but we also need to update the chart version. Looks like I missed it when I was reviewing it. I'd wait for the 0.9.0 PR to be merged and than update the chart version to 0.9.1.

gazal-k:
> Sorry for the back and forth, but we also need to update the chart version. Looks like I missed it when I was reviewing it. I'd wait for the 0.9.0 PR to be merged and than update the chart version to 0.9.1.

No worries, that's what PRs are all about anyway :slightly_smiling_face:. I've bumped it to 0.8.2, so this can go in before 0.9.0 if required. Or else, I'll just rebase & bump the version later

ayberk:
/lgtm
/approve

Thank you!!

gazal-k:
not sure what happened here. the chart release seemed to have failed; https://github.com/kubernetes-sigs/aws-ebs-csi-driver/runs/1835265236. It's trying to publish `0.9.0` again

ayberk:
Yeah interesting it's still 0.9.0 on master but in the diff I see it's 0.9.1/

ayberk:
Ah we updated the appVersion not the chart version, that's why.

gazal-k:
> Ah we updated the appVersion not the chart version, that's why.

oops, let me fix that

gazal-k:
@anilkumarpasupuleti @DeVysh @pradeesh-varghese @mfaridk you can start using 0.9.1 of `aws-ebs-csi-driver` from https://kubernetes-sigs.github.io/aws-ebs-csi-driver instead of the custom version which includes this change

PR #698, opened 2021-01-14
modify error message when request volume is in use with other node

### LINKED ISSUE #515: Wrong Error Returned When Volume is Attached
/kind bug

**What happened?**
When a EBS volume is attached to another EC2 instance the `ALREADY_EXISTS` error is returned when it seems like the `FAILED_PRECONDITION` error should be returned instead because the `node_id` and `volume_id` do not match what the request attachment of `volume_id` and `node_id`. 

CSI Spec: https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerpublishvolume-errors

I expected a `FAILED_PRECONDITION` error to be thrown or have this error thrown if the volume was already attached to my node/ec2-instance.

1. Attach a EBS volume to a ec2-instance using any method, easiest would likely be calling 'CreateVolume`, and go through the normal attachment process or by scheduling a job.

2. Call the RPC `ControllerPublishVolume` with the same `volume_id` but different `node_id` than the already attached volume. 

I've located the line that produces this problem here:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/db95482f8963350d70e9932b0936e6794fe76bf2/pkg/cloud/cloud.go#L358

It would be ideal to have this `ALREADY_EXISTS` still work but just also check for the `node_id` that it is attached to before returning that and otherwise return `FAILED_PRECONDITION`.

**Environment**
- Kubernetes version (use `kubectl version`): N/A (Using Nomad v0.11.2)
- Driver version: v0.6.0-dirty

AndyXiangLi:
Fixes #515 
According to https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerpublishvolume-errors
We are returning wrong error code and error message when volume already assigned to different node
TODO: add logic to handle ALREADY_EXISTS scenario 
**What testing is done?** 
Sanity test passed
Added unit tests

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -115,6 +116,9 @@ var (
 	// ErrAlreadyExists is returned when a resource is already existent.
 	ErrAlreadyExists = errors.New("Resource already exists")
 
+	// ErrVolumeInUse is returned when a volume is already attached to an instance.
+	ErrVolumeInUse = errors.New("Request volume is already attached to an instance")
+
 	// ErrMultiSnapshots is returned when multiple snapshots are found
 	// with the same ID
 	ErrMultiSnapshots = errors.New("Multiple snapshots with the same name found")
@@ -133,6 +137,7 @@ type Disk struct {
 	AvailabilityZone string
 	SnapshotID       string
 	OutpostArn       string
+	Attachments      []string
 }
 
 // DiskOptions represents parameters to create an EBS volume
@@ -378,7 +383,7 @@ func (c *cloud) AttachDisk(ctx context.Context, volumeID, nodeID string) (string
 		if err != nil {
 			if awsErr, ok := err.(awserr.Error); ok {
 				if awsErr.Code() == "VolumeInUse" {
-					return "", ErrAlreadyExists
+					return "", ErrVolumeInUse
 				}
 			}
 			return "", fmt.Errorf("could not attach volume %q to node %q: %v", volumeID, nodeID, err)
@@ -396,7 +401,9 @@ func (c *cloud) AttachDisk(ctx context.Context, volumeID, nodeID string) (string
 	// TODO: Double check the attachment to be 100% sure we attached the correct volume at the correct mountpoint
 	// It could happen otherwise that we see the volume attached from a previous/separate AttachVolume call,
 	// which could theoretically be against a different device (or even instance).
-
+	// TODO: Check volume capability matches for ALREADY_EXISTS
+	// This could happen when request volume already attached to request node,
+	// but is incompatible with the specified volume_capability or readonly flag
 	return device.Path, nil
 }
 
@@ -521,6 +528,7 @@ func (c *cloud) GetDiskByID(ctx context.Context, volumeID string) (*Disk, error)
 	}
 
 	volume, err := c.getVolume(ctx, request)
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/cloud/cloud.go
```diff
@@ -115,6 +115,9 @@ var (
 	// ErrAlreadyExists is returned when a resource is already existent.
 	ErrAlreadyExists = errors.New("Resource already exists")
 
+	// ErrAlreadyExists is returned when a resource is already existent.
```
ayberk, line 118: Hehe a copy-paste artifact here I see :)

pkg/cloud/cloud.go
```diff
@@ -115,6 +115,9 @@ var (
 	// ErrAlreadyExists is returned when a resource is already existent.
 	ErrAlreadyExists = errors.New("Resource already exists")
```
ayberk, line 116: Do we not use this anymore?

ayberk: Ok I see the TODO below.

pkg/cloud/cloud.go
```diff
@@ -1044,3 +1052,14 @@ func volumeModificationDone(state string) bool {
 	}
 	return false
 }
+
+func getVolumeAttachmentsList(volume *ec2.Volume) []string {
+	var volumeAttachmentList []string
+	if len(volume.Attachments) > 0 {
```
ayberk, line 1058: Do you need this check?

pkg/driver/controller.go
```diff
@@ -286,19 +286,21 @@ func (d *controllerService) ControllerPublishVolume(ctx context.Context, req *cs
...
 		return nil, status.Errorf(codes.Internal, "Could not get volume with ID %q: %v", volumeID, err)
 	}
 
+	// If given volumeId already assigned to given node, will directly return current device path
 	devicePath, err := d.cloud.AttachDisk(ctx, volumeID, nodeID)
 	if err != nil {
-		if err == cloud.ErrAlreadyExists {
-			return nil, status.Error(codes.AlreadyExists, err.Error())
+		if err == cloud.ErrVolumeInUse {
+			return nil, status.Error(codes.FailedPrecondition, strings.Join(disk.Attachments, ","))
```
ayberk, line 301: I don't think this is what we want. Looking at the API documentation[[1]][[2]], this list will have information about previous attachments as well, but the status will be "detached". So I think what we want to do is filter for "attached". 

[1]: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Volume.html
[2]: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_VolumeAttachment.html

pkg/driver/controller_test.go
```diff
@@ -2697,7 +2697,7 @@ func TestControllerPublishVolume(t *testing.T) {
 					if !ok {
 						t.Fatalf("Could not get error status code from error: %v", srvErr)
 					}
-					if srvErr.Code() != codes.AlreadyExists {
+					if srvErr.Code() != codes.FailedPrecondition {
```
ayberk, line 2700: Can we also check if we're returning the correct instance id in the error message? It's a requirement from the spec.

pkg/cloud/cloud.go
AndyXiangLi: You are right, we actually not need this.

pkg/driver/controller.go
AndyXiangLi: Yeah, this is a good point, will update

pkg/cloud/cloud.go
```diff
@@ -1044,3 +1052,14 @@ func volumeModificationDone(state string) bool {
 	}
 	return false
 }
+
+func getVolumeAttachmentsList(volume *ec2.Volume) []string {
+	var volumeAttachmentList []string
+	for _, attachment := range volume.Attachments {
+		if attachment.State != nil && aws.StringValue(attachment.State) == "attached" {
```
ayberk, line 1059: Can we lowercase `aws.StringValue(attachment.State)` just to be safe?

PR #700, opened 2021-01-18
Allow specifying --volume-attach-limit in the helm chart

keznikl:
Adding support for a new feature

The new flag was added in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/522/files, this makes it easier to use.

**What testing is done?** 
helm template


ayberk:
/ok-to-test

Thanks for your contribution! Can you also update the chart version please?

PR #702, opened 2021-01-20
Refactor inFlight key to add lock per volumeId

### LINKED ISSUE #307: Node service vs idempotency
/kind bug

I think our `inFlight` object isn't providing idempotency for the node service. For example:

1) CO calls `NodeStageVolume` for a 500 GB volume.
2) The driver starts formatting the volume.
3) CO calls calls `NodeStageVolume` **again** for the same volume.
4) The driver returns an error [0] because the first request `NodeStageVolume` call hasn't returned yet (the driver is still formatting the volume).

Since the call is supposed to be idempotent, we shouldn't return an error in step 4.

I believe the best option is to have a per-volume lock, so any operation (stage, publish, resize etc.) on a specific volume will be executed synchronously.

[0] https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/v0.3.0/pkg/driver/node.go#L84-L91

CC @leakingtapan @dkoshkin

### LINKED ISSUE #370: Handle concurrent operation of NodeUnstage with NodeStage and NodeUnpublish with NodePublish
/kind bug

While reviewing the EBS driver code, I found that we don't correctly handle concurrent processing of `NodeUnstage` call if a previous `NodeStage` was in-progress and same thing applies to `NodeUnpublish` call too.

The specs says, if an operation is in-progress for the volume then SP should return error code "10 ABORTED" - https://github.com/container-storage-interface/spec/blob/master/spec.md#timeouts

AndyXiangLi:
Fixes #307 
Fixes #370 
Handle concurrency for node operations (stage/unstage publish/unpublish)
According to https://github.com/container-storage-interface/spec/blob/master/spec.md#concurrency driver's InFlight cache should track in-flight request per volumeId instead of per request to ensure no more than one request per volume at a given time.

### DIFF
--- a/pkg/driver/internal/inflight.go
+++ b/pkg/driver/internal/inflight.go
@@ -17,6 +17,7 @@ limitations under the License.
 package internal
 
 import (
+	"k8s.io/klog"
 	"sync"
 )
 
@@ -29,7 +30,7 @@ type Idempotent interface {
 	String() string
 }
 
-// InFlight is a struct used to manage in flight requests.
+// InFlight is a struct used to manage in flight requests per volumeId.
 type InFlight struct {
 	mux      *sync.Mutex
 	inFlight map[string]bool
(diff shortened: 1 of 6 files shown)


pkg/driver/internal/inflight.go
```diff
@@ -45,26 +46,25 @@ func NewInFlight() *InFlight {
 
 // Insert inserts the entry to the current list of inflight requests.
```
ayberk, line 47: Can we update the comment here?

pkg/driver/internal/inflight.go
```diff
@@ -45,26 +46,25 @@ func NewInFlight() *InFlight {
...
 
 // Delete removes the entry from the inFlight entries map.
 // It doesn't return anything, and will do nothing if the specified key doesn't exist.
-func (db *InFlight) Delete(h Idempotent) {
+func (db *InFlight) Delete(volumeId string) {
 	db.mux.Lock()
 	defer db.mux.Unlock()
 
-	delete(db.inFlight, h.String())
+	delete(db.inFlight, volumeId)
```
ayberk, line 68: so `h.String()` is always equal to `volumeId`?

ayberk:
Mostly LGTM, but ideally we should have @gnufied take a look as well since he opened one of the issues.

pkg/driver/internal/inflight.go
AndyXiangLi: No, previously the key in this cache is the hash of entire request( including volumeId and other fields like targetPath) so the driver has chance to process multiple requests for one volume at same time (Like same volumeId with different targetPath), which should not happen per [spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#concurrency)
That's why I changed the key to volumeId.

ayberk: I guess my question is will the `delete` function work properly with the `volumeId` instead?

ayberk: Oh nvm I'm dumb heh.

AndyXiangLi:
@gnufied How's this looks to you and can we merge this change?

PR #704, opened 2021-01-21
Add support for node existing service accounts

### LINKED ISSUE #703: Support existing Service Account for the node ebs-csi in the Helm chart
We use Pod Security Policies and have Service Accounts authorized to use specific policies. Installing the node Daemonset using the Helm chart does not allow specifying existing Service Account. So, we can't specify the Service Account that has the permissions to create the Pods.

The Helm chart provides an option to specify the name of an existing Service Account for the node service. This is already supported for other components deployed using the chart.

Authorize the `default` Service Account to use the Pod Security Policy for the node ebs-csi. This option is not good for us since other workloads might use that Service Account.

mper0003:
New feature

Fixes #703

**What testing is done?** 
Helm template

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -102,7 +102,9 @@ serviceAccount:
     name: ebs-snapshot-controller
     annotations: {}
   node:
-    name: "" # Name of the service-account to be used
+    create: true
+    name: "ebs-csi-node-sa"
+    annotations: {}
 
 storageClasses: []
 # Add StorageClass resources like:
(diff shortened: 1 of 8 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -101,3 +101,5 @@ serviceAccount:
     create: true
     name: ebs-snapshot-controller
     annotations: {}
+  node:
+    name: "" # Name of the service-account to be used
```
ayberk, line 105: Can we do create, name, annotations fields instead? Although I like the brevity, we should be consistent with the other two.

mper0003: Yes, that's true. I've updated to keep it consistent. Thanks

ayberk:
/lgtm
/approve

Thanks!

PR #705, opened 2021-01-25
add e2e test for volume resizing

### LINKED ISSUE #316: Add e2e test for volume resize
Add e2e test for volume resize

AndyXiangLi:
Fixes #316
Add e2e test for volume resizing

### DIFF
--- a/tests/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go
+++ b/tests/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+   http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testsuites
+
+import (
+	"fmt"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/e2e/driver"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"time"
+
+	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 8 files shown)


tests/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go
```diff
@@ -0,0 +1,94 @@
...
+
+	By("resizing the pvc")
+	updatedPvc, err := client.CoreV1().PersistentVolumeClaims(namespace.Name).Update(pvc)
+	if err != nil {
+		framework.ExpectNoError(err, fmt.Sprintf("fail to resize pvc(%s): %v", pvcName, err))
+	}
+	updatedSize := updatedPvc.Spec.Resources.Requests["storage"]
+
+	By("Wait for resize to complete")
+	time.Sleep(30 * time.Second)
```
ayberk, line 63: How did you come up with this number? Can we do this incrementally? Check every 5 seconds and give up after 1 minute for example.

AndyXiangLi: Yeah that's a good point, it is updated now

tests/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go
```diff
@@ -0,0 +1,92 @@
...
+
+}
+
+// WaitForPvToResize waiting for pvc size to be resized to desired size
+func WaitForPvToResize(c clientset.Interface, ns *v1.Namespace, pvName string, desiredSize resource.Quantity, timeout time.Duration, interval time.Duration) error {
+	By(fmt.Sprintf("Waiting up to %v for pv in namespace %q to be complete", timeout, ns.Name))
+	for start := time.Now(); time.Since(start) < timeout; time.Sleep(interval) {
+		newPv, _ := c.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
+		newPvSize := newPv.Spec.Capacity["storage"]
+		if desiredSize.Equal(newPvSize) {
```
ayberk, line 86: you also return an error if they're not equal, right?

AndyXiangLi: Yeah, it will return error if they are not equal and timeout at line 91

ayberk: Ah I see we can't exit early.

gnufied:
@ayberk @AndyXiangLi  why are we doing this? We already had e2e via testsuites of CSI driver. I don't think we need more e2e (or if anything is missing it should be added to k/k repo).

ayberk:
Can you point me to the e2e tests on the k/k repo? I'll make sure all new tests go there instead.

gnufied:
here we go - https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/volume_expand.go 

We should be running these tests as part of this driver's e2e.

PR #707, opened 2021-01-27
Update sidecars to newer version

### LINKED ISSUE #604: EBS CSI Driver does not work on arm64 based instances
/kind bug

**What happened?**

arm support was added in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/527 for the driver Docker images. However, when deploying the EBS CSI from this repo it references Docker images that are not multiarch. Thus the deployment gets stuck in a crash loop.  So ARM still won't work.

These images:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/963eccfc404f56540f025a93fd4966fa6a73458f/deploy/kubernetes/overlays/stable/kustomization.yaml

As well as snapshotted and resize. The ones from `qauy.io` and not multi arch.

A successful deployment of the EBS CSI Driver on ARM based servers like m6g. 

Try to deploy the EBS CSI on ARM based systems.

**Environment**

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.8-eks-7c9bda", GitCommit:"7c9bda52c425d0d56d7b93f1377a826b4132c05c", GitTreeState:"clean", BuildDate:"2020-08-28T23:04:33Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

- Driver version: 
`v0.7.0`

AndyXiangLi:
Fixes #604 
Continue from PR #596 and update sidecar version in both helm and kustomization 
** What test has been done
Validate sidecar functionality in ARM node
Validate sidecar functionality in AMD node
Kube version: 1.18


charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "0.8.1"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
 version: 0.8.2
```
wongma7, line 5: the version should also be bumped as technicaslly this is a new helm chart with new YAMLs

It's fine if the helm chart `version` goes out of sync with the driver `appVersion` (it already is)

AndyXiangLi: Hi Matthew, should we bump the driver version to 0.9.0?

wongma7: hmm yes that's a good idea, never mind then, let's leave version untouched for now. 

Since there is already demand, we can release driver 0.9.0 + helm chart 0.9.0 later, after this merges.

ayberk: I agree with @wongma7. We should just bump to 0.8.3.

AndyXiangLi: Roger, so I will just update chart version to 0.8.3 in this PR and we can sync the driver version and chart version later(in 0.9.0 release)

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -11,23 +11,23 @@ image:
...
+    repository: k8s.gcr.io/sig-storage/csi-provisioner
+    tag: "v2.0.2"
   attacherImage:
-    repository: quay.io/k8scsi/csi-attacher
-    tag: "v2.2.0"
+    repository: k8s.gcr.io/sig-storage/csi-attacher
+    tag: "v3.0.0"
   snapshotterImage:
-    repository: quay.io/k8scsi/csi-snapshotter
+    repository: k8s.gcr.io/sig-storage/csi-snapshotter
```
ayberk, line 20: We need at least 3.0.0 for this one it seems: https://github.com/kubernetes-csi/external-snapshotter/issues/461

ayberk: @wongma7 maybe we should get rid of the snapshotter before merging this one. Not sure which one is less effort, but this would be a throw-away effort.

AndyXiangLi: Hi Ayberk, just updated snapshotter sidecar and controller to version 3.0.3 and validated they are functioning as expected. Maybe we can retire snapshotter in a separate PR

docs/README.md
```diff
@@ -109,6 +109,8 @@ kubectl create -f https://raw.githubusercontent.com/kubernetes/csi-api/release-1
 ```
 
 #### Deploy driver
+Latest master branch only supports driver deployment on kubernetes version >= 1.17
```
ayberk, line 112: Can we make this more generic? I'm thinking something like "Please see the compatibility matrix above before you deploy the driver". So we don't have to keep updating this part as well.

AndyXiangLi: Updated

ayberk:
/lgtm
/approve

Great work! We're finally officially supporting arm without workarounds :)

ajaykmis:
@ayberk @AndyXiangLi - Thanks for merging this! 

Did someone validate volume snapshotting with this on Arm64?  Do we now use the alpha overlay or stable/arm64 to install the EBS CSI driver. 

- Ajay

AndyXiangLi:
Hi @ajaykmis 
Yes, I validated volume snapshotter is working on Arm64, and I recommend you to use helm to install driver for now. The arm overlay is same as stable overlay after this change and we will remove arm overlay soon to reduce confusion

ajaykmis:
Thanks @AndyXiangLi 

I see this from the doc: 
helm upgrade --install aws-ebs-csi-driver \
    --namespace kube-system \
    --set enableVolumeScheduling=true \
    --set enableVolumeResizing=true \
    --set enableVolumeSnapshot=true \
    aws-ebs-csi-driver/aws-ebs-csi-driver
    
   a couple of questions:
1. what would be this path ${your_chart_path}?  
2. does helm take care of choosing the right overlay? so can I use the same helm command to install EBS driver on both ARM and x86 clusters?

AndyXiangLi:
@ajaykmis 
1. If you have this project on your machine, your_chart_path = ${project_path}/charts/aws-ebs-csi-driver
2. Yes you can run same command on ARM and x86 clusters as all the images we are using are multiarch supported after this change

PR #715, opened 2021-02-03
Increase timout for cloudbuild

ayberk:
Trying to unblock the release. We still need to investigate why builds have been failing since 01/11.


wongma7:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/716

PR #720, opened 2021-02-04
Move cr.yaml out of github workflows

ayberk:
Cleanup

Github considers this file as a workflow, which then always fails because it's a config file. Moving it out so it doesn't trigger a workflow.

**What testing is done?** 
N/A.


wongma7:
lol, thank you

/lgtm
/approve

PR #725, opened 2021-02-05
fix: Update chart version instead of appVersion

gazal-k:
bug fix

amending issue introduced by #697

**What testing is done?**


gazal-k:
@ayberk pls help with this

gazal-k:
Nevermind, saw #724

PR #728, opened 2021-02-05
[aws-ebs-csi-driver] Add the serviceAccountName parameter to the DaemonSet

diranged:
Bug fix

Without this, it is difficult to assign a PodSecurityPolicy to the
DaemonSet pods, which can be required in many more secure clusters where
HostPath and HostNetworking are disabled in the default PSP.


ayberk:
Thank you @diranged. This has been superseded by #704. 

/close

diranged:
:+1:

PR #730, opened 2021-02-06
Allow all fields to be set on StorageClasses

haines:
A new feature

The StorageClasses created by the Helm chart can currently only be configured with `volumeBindingMode`, `reclaimPolicy`, and `parameters`. We have `allowVolumeExpansion: true` set on ours, so we couldn't migrate them to be provisioned by the chart.

This PR allows any field permitted by the API to be specified.

**What testing is done?** 

I've templated the chart with a variety of StorageClasses and validated that the output manifests are correct.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/storageclass.yaml
+++ b/charts/aws-ebs-csi-driver/templates/storageclass.yaml
@@ -4,12 +4,5 @@ apiVersion: storage.k8s.io/v1
 metadata:
   name: {{ .name }}
 provisioner: ebs.csi.aws.com
-volumeBindingMode: {{ default "WaitForFirstConsumer" .volumeBindingMode }}
-{{- if hasKey . "reclaimPolicy" }}
-reclaimPolicy: {{ .reclaimPolicy }}
-{{- end }}
-{{- with .parameters }}
-parameters:
-{{ toYaml . | indent 2 }}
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/templates/storageclass.yaml
```diff
@@ -4,12 +4,5 @@ apiVersion: storage.k8s.io/v1
...
 provisioner: ebs.csi.aws.com
-volumeBindingMode: {{ default "WaitForFirstConsumer" .volumeBindingMode }}
-{{- if hasKey . "reclaimPolicy" }}
-reclaimPolicy: {{ .reclaimPolicy }}
-{{- end }}
-{{- with .parameters }}
-parameters:
-{{ toYaml . | indent 2 }}
-{{- end }}
+{{ omit (dict "volumeBindingMode" "WaitForFirstConsumer" | merge .) "name" | toYaml }}
```
ayberk, line 7: TIL!

ayberk:
/ok-to-test

@haines thanks for your contribution! Could you please update the chart version as well?

haines:
Thanks for reviewing, @ayberk! I've rebased onto master and bumped the chart version.

ayberk:
/lgtm
/approve

Thanks a lot!

PR #731, opened 2021-02-06
Fix daemonset liveness image to multiarch version for ARM kustomization

maxgio92:

I think this is a bug fix.

The ARM-supporting kustomization base still missed the correct (multiarch) liveness sidecar image for the Node service.

The Helm chart already gets the [default](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/v0.9.0/charts/aws-ebs-csi-driver/values.yaml) correctly.

### DIFF
--- a/deploy/kubernetes/base/arm64/node.yaml
+++ b/deploy/kubernetes/base/arm64/node.yaml
@@ -88,7 +88,7 @@ spec:
             - name: registration-dir
               mountPath: /registration
         - name: liveness-probe
-          image: quay.io/k8scsi/livenessprobe:v2.1.0
+          image: k8s.gcr.io/sig-storage/livenessprobe:v2.1.0
           args:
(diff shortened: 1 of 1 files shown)


ayberk:
/approve
/lgtm

Thank you! We deleted the overlay on master but looks like it leaked to the 0.9.0 release.

maxgio92:
Thank you!

maxgio92:
Hi @ayberk maybe it can be tagged v0.9.1?

ayberk:
@maxgio92 Looks like you have a PR to use the branch instead. Do you have a specific need for a tag instead? I'm not keen on having a new release for a kustomization-only change.

maxgio92:
@ayberk I get your point, as the version is related to the driver itself.

The motivation behind it is to use that kustomization by referencing a version's git tag.
Referencing a commit would do the same in the end.

PR #732, opened 2021-02-09
[chart] Allow resources override for node DaemonSet + priorityClassName

dntosas:
CSI controller and node have different needs in a manner of capacity so
in this commit we enable users to define specific resources for the node
component. This will allow users not to reserve not needed resources on
all of their instances as node is a DaemonSet and may not need as much
CPU/Memory as the controller Pods.

Adding a feature of overriding resource for the node

Mitigate overprovisioning capacity for node component

**What testing is done?** 
Tested on our internal clusters

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/node.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node.yaml
@@ -80,8 +80,14 @@ spec:
             timeoutSeconds: 3
             periodSeconds: 10
             failureThreshold: 5
+          {{- if .Values.node.resources }}
+          {{- with .Values.node.resources }}
+          resources: {{ toYaml . | nindent 12 }}
+          {{- end }}
+          {{- else }}
           {{- with .Values.resources }}
           resources: {{ toYaml . | nindent 12 }}
+          {{- end }}
           {{- end }}
         - name: node-driver-registrar
           image: {{ printf "%s:%s" .Values.sidecars.nodeDriverRegistrarImage.repository .Values.sidecars.nodeDriverRegistrarImage.tag }}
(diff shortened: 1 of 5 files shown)


charts/aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -80,7 +81,7 @@ spec:
             timeoutSeconds: 3
             periodSeconds: 10
             failureThreshold: 5
-          {{- with .Values.resources }}
+          {{- with .Values.node.resources }}
```
ayberk, line 84: I am not a big fan of this change. This would be a breaking change and force the operators to update their setup. Can we somehow give this new value priority and if it doesn't exist fallback to `.Values.resources`? We can add a deprecation note for the old field on the values file.

charts/aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -103,7 +104,7 @@ spec:
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
-          {{- with .Values.resources }}
+          {{- with .Values.node.resources }}
```
ayberk, line 107: Similarly here. I understand this makes more sense, but we still have to make sure we don't break existing setups without any warning.

ayberk (changes requested):
/ok-to-test

charts/aws-ebs-csi-driver/templates/node.yaml
dntosas: Good catch ^^

Pushed some changes, adding a condition so the additions would be backwards compatible. Tell me if something like that would work for you :)

ayberk:
Looks great, thanks! It'd be great if you can follow-up with a small CR to add a comment for the resources fields so we have a deprecation warning. 

/lgtm
/approve

Issue #733, opened 2021-02-09
Helm chart: can't annotate storage classes


nicholasmhughes:
/kind bug

**What happened?**
The new ability to create storage classes within the chart is great. However, there isn't a mechanism to annotate the storage classes. This becomes important is you want to tag one as default.

**What you expected to happen?**
Annotations should be handled separately from the rest of the storage class YAML and inserted under metadata.

**How to reproduce it (as minimally and precisely as possible)?**
```
storageClasses:
  - name: ebs-sc
    annotations:
      storageclass.kubernetes.io/is-default-class: "true"
```

**Anything else we need to know?**:
I have the code for this. PR inbound.

**Environment**
- Kubernetes version (use `kubectl version`): v1.18.12
- Driver version: 0.9.0

PR #734, opened 2021-02-09
[chart]  Add storage class annotation and label handling

nicholasmhughes:
Depends on your point of view... :wink:   The issue is categorized as a bug.

Adds storage class annotation and label handling to the chart. See #733 

**What testing is done?**

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -29,6 +29,10 @@ sidecars:
     repository: k8s.gcr.io/sig-storage/csi-node-driver-registrar
     tag: "v2.0.1"
 
+snapshotController:
+  repository: k8s.gcr.io/sig-storage/snapshot-controller
+  tag: "v3.0.3"
+
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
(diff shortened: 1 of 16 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -109,6 +109,9 @@ serviceAccount:
 storageClasses: []
 # Add StorageClass resources like:
 # - name: ebs-sc
+#   # annotation metadata
+#   annotations:
+#     storageclass.kubernetes.io/is-default-class: "true"
```
ayberk, line 114: Did we miss the "labels" field here? It's omitted on the storageclass.yaml.

ayberk (commented):
/ok-to-test

charts/aws-ebs-csi-driver/values.yaml
nicholasmhughes: I handled it in the template. I can't think of a great reason to have a label vs and annotation, so I didn't include it in the sample. I can if required.

ayberk: That'd be nice. Documenting all fields in the example could save operators from reading the template, so I think it's worth doing.

nicholasmhughes: @ayberk I added the label example. Let me know if that looks ok.

nicholasmhughes:
@ayberk is there anything else I need to do for this PR?

ayberk:
@nicholasmhughes If you can squash to a single commit I can merge it. Thanks a lot!

nicholasmhughes:
@ayberk is that good enough?

ayberk:
/lgtm
/approve

Thanks!

nicholasmhughes:
thank you!

PR #744, opened 2021-02-12
Make CreateVolume idempotent

### LINKED ISSUE #165: Race condition in CreateVolume
/kind bug

**What happened?**

When the CO calls `CreateVolume` (through the sidecar container), it passes in a volume name. If a second call to create the volume is issued, we return the same volume created in the first request.

Since there's no way to specify the volume name through the EC2 API, we currently set the volume name as a [tag](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/e014aa97b76420bb2d4a92428242a77766095d23/pkg/cloud/cloud.go#L219-L225). However, this workaround has its own problems. 

Imagine this situation:

* `CreateVolume` is called so a volume named `volume1` is created.
* EC2's `CreateVolumeWithContext` is called and it takes about 15 seconds to effectively create the volume.
* `CreateVolume` is called again, but the call above is not done yet.
* The driver will check if a volume with this name already exists, but it won't find any, so it creates the new volume.
* Now we have 2 volumes tagged with the same name.

During attach, for instance, the driver will notice that there's more than one volume with the same name, so it'll return an error. However, this is just a workaround and a proper solution should ideally be implemented on the EC2 side.

This could be achieved by having a `Name` filed in `ec2.CreateVolumeInput`.

The easiest way is to:

1. Run the CSI driver
2. Use the [csc tool](https://github.com/rexray/gocsi/tree/master/csc) to create 2 volumes:

```
$ csc controller new --endpoint tcp://127.0.0.1:10000 (..) && \
  csi controller new --endpoint tcp://127.0.0.1:10000(...)
```

chrishenzie:
Addresses #165, but the full fix requires an EBS API change.

This PR works around idempotency issues with the `CreateVolume` API. It does so by using the `InFlight` struct to mark a request as "in progress". If an identical `CreateVolumeRequest` is received while another is in progress, we return an error to the caller.

**What testing is done?** 
Unit tests and smoke tests on a k8s cluster on AWS.

This issue can be repeated by the following command (similar to the one in the GitHub issue, however it uses `&` to background the first call):
```
csc controller --endpoint <socket or address> create-volume --cap SINGLE_NODE_WRITER,block <volume name> & \
csc controller --endpoint <socket or address> create-volume --cap SINGLE_NODE_WRITER,block <volume name>
```

When running against master I observed:
```
$ csc controller --endpoint <socket> create-volume --cap SINGLE_NODE_WRITER,block chrishenzie-test & \
> csc controller --endpoint <socket> --cap SINGLE_NODE_WRITER,block chrishenzie-test
[1] 30225
"vol-0e313cca607416c77" 107374182400
"vol-02d575a8de114bbf3" 107374182400
[1]+  Done                    csc controller --endpoint <socket> create-volume --cap SINGLE_NODE_WRITER,block chrishenzie-test
```

This created two volumes with tags `Key:CSIVolumeName, Value:chrishenzie-test` (per the two volume IDs that are returned).

When running against this branch I observed:
```
$ csc controller --endpoint <socket> create-volume --cap SINGLE_NODE_WRITER,block chrishenzie-test & \
> csc controller --endpoint <socket> create-volume --cap SINGLE_NODE_WRITER,block chrishenzie-test
[1] 464
Create volume request for chrishenzie-test is already in progress

Please use -h,--help for more information

"vol-0ee90da47751cc6dd" 107374182400
[1]+  Exit 13                 csc controller --endpoint <socket> create-volume --cap SINGLE_NODE_WRITER,block chrishenzie-test
```

Which created only one volume.

**Open questions**
- At what point do we want to consider requests "identical"?
  - The thing we are depending on to establish uniqueness (at the very least) is the volume name, should we just be using that?
- ~What error code should be returned?~
  - It should be `Aborted` based on the CSI spec

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -27,6 +27,7 @@ import (
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/internal"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -56,6 +57,7 @@ var (
 // controllerService represents the controller service of CSI driver
 type controllerService struct {
 	cloud         cloud.Cloud
+	inFlight      *internal.InFlight
 	driverOptions *DriverOptions
 }
 
@@ -87,6 +89,7 @@ func newControllerService(driverOptions *DriverOptions) controllerService {
 
 	return controllerService{
 		cloud:         cloud,
+		inFlight:      internal.NewInFlight(),
 		driverOptions: driverOptions,
 	}
 }
@@ -200,6 +203,13 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 		return newCreateVolumeResponse(disk), nil
 	}
+
+	// check if a request is already in-flight because the CreateVolume API is not idempotent
+	if ok := d.inFlight.Insert(req); !ok {
+		msg := fmt.Sprintf("Create volume request for %s is already in progress", volName)
+		return nil, status.Error(codes.Aborted, msg)
+	}
+	defer d.inFlight.Delete(req)
 
 	// create a new volume
 	zone := pickAvailabilityZone(req.GetAccessibilityRequirements())
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/mock/gomock"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/internal"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/mocks"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	"google.golang.org/grpc/codes"
@@ -198,6 +199,7 @@ func TestCreateVolume(t *testing.T) {
 
 				awsDriver := controllerService{
 					cloud:         mockCloud,
+					inFlight:      internal.NewInFlight(),
 					driverOptions: &DriverOptions{},
 				}
 
@@ -268,6 +270,7 @@ func TestCreateVolume(t *testing.T) {
 
 				awsDriver := controllerService{
 					cloud:         mockCloud,
+					inFlight:      internal.NewInFlight(),
 					driverOptions: &DriverOptions{},
 				}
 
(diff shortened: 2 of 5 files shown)


chrishenzie:
/assign leakingtapan

@msau42

pkg/driver/controller.go
```diff
@@ -201,6 +204,13 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return newCreateVolumeResponse(disk), nil
 	}
 
+	// check if a request is already in-flight because the CreateVolume API is not idempotent
+	if ok := d.inFlight.Insert(req); !ok {
+		msg := fmt.Sprintf("Create volume request for %s is already in progress", volName)
+		return nil, status.Error(codes.AlreadyExists, msg)
```
msau42, line 210: CSI spec says use "ABORTED" for an operation in progress: https://github.com/container-storage-interface/spec/blame/master/spec.md#L470

pkg/driver/controller.go
```diff
@@ -201,6 +204,13 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return newCreateVolumeResponse(disk), nil
 	}
 
+	// check if a request is already in-flight because the CreateVolume API is not idempotent
+	if ok := d.inFlight.Insert(req); !ok {
```
wongma7, line 208: should we just use the req.Name as the key? I dont think it will ever matter in practice but spec says name is sufficient: https://github.com/container-storage-interface/spec/blob/master/spec.md#controller-service-rpc

wongma7 (approved):

pkg/driver/controller.go
chrishenzie: I think that makes sense, I'll make the change

chrishenzie: One issue with this approach is that this `inFlight` field is shared amongst the entire controller. By limiting the key to just the name we may encounter issues once other operations start depending on this field (assuming they also use only the name as the key).

wongma7: good point, it would need to categorize operations within the inflight map ... lgtm in current state then.

leakingtapan (commented):
It's okay to workaround the issue through inflight check, but I won't mark the original issue as resolved. Since the driver could still crash at the middle of processing the duplicate CreateVolume request, and the restarted driver won't have the original inFlight map as it is only held in memory if nothing changed too much recently. The real fix should be let EBS to take the client token during CreateVolume, which they admitted it's an easy change.

chrishenzie:
@leakingtapan Yes this would still be an issue should the driver crash. Does the real fix you described require an EBS API change that is not currently available? I just want to make sure we're taking the best approach given the current limitations.

leakingtapan:
> Does the real fix you described require an EBS API change that is not currently available?

yep. And the change is to add a new field to the existing request struct which is totally doable.

@wongma7 do you wanna chat with EBS again for their progress on this? As I talked to them earlier, this is ready in the backend, and it just needs to be surfaced up.

@chrishenzie feel free to merge ur change in at the meantime. I just don’t want to close the original issue accidentally

chrishenzie:
I don't think I can LGTM my own PR, would you be able to? I can re-open the issue if it closes or open a separate one that links to it.

msau42:
Instead of "Fixes X", you can use a different keyword (like Addresses) so that github doesn't automatically close the issue.

chrishenzie:
Updated description to use "Addresses" instead of "Fixes".

gnufied:
@AndyXiangLi I think we need a similar issue for `CreateSnapshot` unless we want to allow multiple `CreateSnapshot` calls for same volume.

AndyXiangLi:
@gnufied We definitely don't want multiple CreateSnapshot call in flight for same volume, I will take a look and follow up with my team regarding this! Do we have a ticket created for this?

PR #752, opened 2021-02-18
Set enableVolumeScheduling to true by default in the helm chart

### LINKED ISSUE #727: Helm chart should default enableVolumeScheduling: true
/kind bug

**What happened?**
Deployed the default helm chart and topology aware dynamic provisioning did not work. The kustomize based manifest default to true: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/values/controller.yaml#L4

With the CSI Topology sub-feature at GA status, I would expect this feature gate to default to true.

Thoughts on this?  Is there still a good reason to default the Helm chart to not use the Topology feature?

mtougeron:
Fixes #727 

Changes the default value of `enableVolumeScheduling` from `false` to `true` for the helm chart.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.9.6
+version: 0.9.7
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.9.6
+version: 0.9.7
```
ayberk, line 5: Can we update this to 0.9.8? We recently merged another change.

mtougeron: done

ayberk:
/approve
/lgtm

Thanks!

Issue #569, opened 2020-09-28
Resizing disk partition in EBS from smaller snapshot size


symanovich-a:
/kind bug

**What happened?**
Disk partition isn't resized after creation

**What you expected to happen?**

Disk partition is resized after creation
**How to reproduce it (as minimally and precisely as possible)?**
- Create disk(ebs) snapshot 
- Restore disk from this snapshot but with bigger size.
I see normal disk size from lsblk command, but partition isn't resized. All work good, if i do disk resizing after pvc editing 

**Anything else we need to know?**:

**Environment**
- Kubernetes version 1.17(eks)
- Driver version: 0.6.0

AndyXiangLi:
There is an upstream discussion about this issue https://github.com/container-storage-interface/spec/pull/452
Looks like driver need to resize the fs if detect volume size is larger than current fs when stage volume.
And as per @jsafrane , we can always call resize2fs / xfs_growfs after mount as alternate
@ayberk @leakingtapan what do you think?

PR #753, opened 2021-02-18
Resize filesystem when restore a snapshot to larger size volume

AndyXiangLi:
Fixes: #569 
Before NodePublishVolume, we will check if the device size is matched with the usable filesystem size. And resize the filesystem if there is a mismatch
TODO: Upstream is planning on move the common resize code to kubernetes/mount-util, we should use the resize function from there instead of maintaining our own once the changes are available
**What testing is done?** 
Test done in both ext[x] and xfs filesystem


gnufied:
This is similar to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/595 cc @jsafrane

gnufied:
@AndyXiangLi did you merge it intentionally? I don't think it was reviewed. :-) cc @wongma7

AndyXiangLi:
> @AndyXiangLi did you merge it intentionally? I don't think it was reviewed. :-) cc @wongma7

We talked about this offline and decide to merge it for now. Will update once the upstream is merged
https://github.com/kubernetes/kubernetes/pull/101253

PR #756, opened 2021-02-21
Update livenessprobe image version from 2.1.0 to 2.2.0

### LINKED ISSUE #749: Upgrade livenessprobe to patch memory leak
/kind bug

**What happened?**
livenessprobe v2.1.0 has a memory leak causing it use way more ram than needed. This is patched in [livenessprobe v2.2.0
](https://github.com/kubernetes-csi/livenessprobe/blob/master/CHANGELOG/CHANGELOG-2.2.md#bug-or-regression)
livenessprobe containers usually use ~5MB of ram. 

Let livenessprobe run for a while and watch ram usage.

mowangdk:
Fix #749 

Upgrade livenessprobe to patch memory leak 

**What testing is done?** 
helm install in k8s cluster

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.9.9
+version: 0.9.10
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -21,7 +21,7 @@ sidecars:
     tag: "v3.0.3"
   livenessProbeImage:
     repository: k8s.gcr.io/sig-storage/livenessprobe
-    tag: "v2.1.0"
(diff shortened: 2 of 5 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -21,7 +21,7 @@ sidecars:
     tag: "v3.0.3"
   livenessProbeImage:
     repository: k8s.gcr.io/sig-storage/livenessprobe
-    tag: "v2.1.0"
+    tag: "v2.2.0"
```
ayberk, line 24: Can you update the chart version as well? Otherwise github actions fails.

ayberk (changes requested):

charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.9.8
+version: 0.9.9
```
mowangdk, line 5: @ayberk PTAL

ayberk: I just merged an earlier PR, could you update it to 0.9.10?

mowangdk: ok

ayberk:
/lgtm
/approve

Thanks a lot!!

Issue #766, opened 2021-02-24
Confusing error message when creating "io1" volumes


jsafrane:
/kind bug

**What happened?**
When the storage class requires `io1` volume, but misses `IOPSPerGB` parameter, user gets this event:

```
Warning  ProvisioningFailed    3s               ebs.csi.aws.com_ip-10-0-158-46_2aa28a47-d1e8-4570-9a4d-c4b35ea9f5cf  failed to provision volume with StorageClass "io1-csi": rpc error: code = Internal desc = Could not create volume "pvc-b64c726a-a494-4ce7-bbb1-72822526f160": could not create volume in EC2: InvalidParameterCombination: The parameter iops must be specified for io1 volumes.
           status code: 400, request id: 625e97aa-8238-4a80-92c4-36fde3561c1e
```

`The parameter iops must be specified for io1 volumes.` is very misleading, because it should be `IOPSPerGB`.

PR #767, opened 2021-02-24
Fix error message when IOPSPerGB is missing in io1 volumes

jsafrane:
Explicitly return specific message in this case, otherwise the cloud returns generic `The parameter iops must be specified for io1 volumes`, which mentions `iops`, but the parameter that the user needs to set is `IOPSPerGB`.

Fixes: #766

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -183,6 +183,12 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


wongma7:
indeed 'parameters'  is ambiguous. many THANKS

/lgtm

ayberk:
Don't we need this for io2 as well?

jsafrane:
I did not test it, however, I think that the error message `The parameter iops must be specified for io2 volumes` will be correct.

Issue #621, opened 2020-11-25
Helm chart should manage namespace more consistently


TBBle:
/kind bug

**What happened?**

The Helm chart does not honour the `--namespace` parameter for Helm, all resources hard-code the _kube-system_ namespace, meaning it's possible to install the Helm chart metadata in a different namespace from the actual resources.

**What you expected to happen?**

Either:
* The Helm chart honours `--namespace` and namespaced resources are created in the provided namespace; or
* The Helm chart rendering fails if `--namespace` is not passed as _kube-system_.

**How to reproduce it (as minimally and precisely as possible)?**

```
kubectl create namespace aws-ebs-csi-driver
helm upgrade --install aws-ebs-csi-driver \
    --namespace aws-ebs-csi-driver \
    --set enableVolumeScheduling=true \
    --set enableVolumeResizing=true \
    --set enableVolumeSnapshot=true \
    https://github.com/kubernetes-sigs/aws-ebs-csi-driver/releases/download/v0.7.1/helm-chart.tgz
kubectl --namespace aws-ebs-csi-driver get daemonset ebs-csi-node
```

I would expect to see the AWS EBS CSI driver Daemonset in the output of the last command, or for the helm call to fail.

**Anything else we need to know?**:

I assume that this project currently depends on being installed in _kube-system_, as it is `system-node-critical`. So the obvious resolution to me is to add a check to NOTES.txt that fails rendering if `.Release.Namespace` is not _kube-system_.

Either way, it would be a more-consistent-with-common-practice chart if all the hard-coded _kube-system_ references in the chart were replaced with `.Release.Namespace`.

**Environment**
- Kubernetes version (use `kubectl version`): N/A
- Driver version: Untested, observed in the Helm chart templates.

krmichel:
@ayberk Looks like this didn't get tied to PR #768 and didn't get closed when that was merged.

PR #768, opened 2021-02-24
removed harcoded NAMESPACE from helm chart

alexandrst88:
Override hardcoded `kube-system` namespace

Ability to run helm chart in another namespace, for example `kube-system` could be free out of any mutation or validation webhooks.

Should address https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/621
**What testing is done?** 
Deployed in `example` namespace without issues.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.9.9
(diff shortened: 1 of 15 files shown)


wongma7:
I have read a lot of conflicting info about what the best practice is.

Removing kube-system hardcode seems correct to me though.

https://github.com/helm/helm/issues/5465#issuecomment-723134829

@ayberk

ayberk:
/approve

Looks like this'll break installing via `helm template | kubectl create --namespace foo -f -`[[1]](https://github.com/helm/helm/issues/5465#issuecomment-723182962). We're not advertising installing that way anywhere, so I think it should be fine.

Thanks!

Issue #754, opened 2021-02-19
Volumes may leak during Create operation: Delete/clean volume if waitForVolume fails during Create operation


wongma7:
/kind bug

If AWS API is unavailable (e.g. due to throttling) then waitForVolume could fail even though the volume eventually becomes available and leaks. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L343

Issue also affects in-tree

wongma7:
also, even if we defer deletevolume that could be throttled too so it's not perfect. But I think deletevolume quota and describe* quota are not in the same bucket for accounting purposes.

Could have a leaked vol clean up routine but I don't think that belongs in the driver controller. Anything that does delete is of course risky and routinely scanning for vols to cleanup would exacerbate to any quota problems... BEtter to prevent leaks in first place

wongma7:
Another idea:

CreateVolume can't be passed idempotency token

In lieu of that, we may have to overload tag API and use a tag as a token.

So in CreateVolume, DescribeVolume with filter token=PVCUID. If exists, use it and return early. if DNE, proceed as usual.

PR #769, opened 2021-02-24
Aws client config: increase MaxRetries

josselin-c:
This commit increase the number of retries to a high number (8) so
that the AWS framework retries failing requests until the request
context times-out.

By default, when throttled, the aws client retries a given request
3 times with exponentially increasing delay between retries.
The default number of retries is still low enough that requests are
only going to be retried for a few seconds.
Instead of retrying for a few seconds, we should retry until the
request context times out.

This should reduce the occurence of #754

Mitigation for #754

**What is this PR about? / Why do we need it?*
[waitForVolume](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L804) has a checkTime of 1 minute but current AWS client only retries requests for a few seconds.

This PR is about increasing the number of retries the AWS client performs before giving up so that `waitForVolume` actually waits a full minute or until the request context deadline is reached.

This is important because as explained in #754 a throttled / failed request will result in a leaked volume so we want to reduce the change of it happening.

**What testing is done?** 
`make test`

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -235,6 +235,8 @@ func newEC2Cloud(region string) (Cloud, error) {
 	awsConfig := &aws.Config{
 		Region:                        aws.String(region),
 		CredentialsChainVerboseErrors: aws.Bool(true),
+		// Set MaxRetries to a high value. It will be "ovewritten" if context deadline comes sooner.
+		MaxRetries: aws.Int(8),
 	}
 
 	endpoint := os.Getenv("AWS_EC2_ENDPOINT")


pkg/cloud/cloud.go
```diff
@@ -235,6 +235,8 @@ func newEC2Cloud(region string) (Cloud, error) {
 	awsConfig := &aws.Config{
 		Region:                        aws.String(region),
 		CredentialsChainVerboseErrors: aws.Bool(true),
+		// Set MaxRetries to a high value. It will be "ovewritten" if context deadline comes sooner.
+		MaxRetries: aws.Int(8),
```
wongma7, line 239: for my reference: https://github.com/aws/aws-sdk-go/blob/d87d92b4412739c6ef8917979e922fae98b7654c/aws/client/default_retryer.go

8 make sense to me. I played with value here:  https://play.golang.org/p/sZr9Tgxgt79   SDK jitters the delay every retry but on average 8 retries is the right amount to assuming  a 1 min polling period

wongma7: Since this affects all api calls, i also checked attach detach.

By default, the attacher passes in a context of 15 seconds: https://github.com/kubernetes-csi/external-attacher/blob/3d49e838d8ecc491b0a7ce423f32b7e43aaa51af/cmd/csi-attacher/main.go#L59

So 3 retries was too low even for that.

wongma7: Sorry, I'm mixing up the sdk-level retries and the driver-level retries.

So it's really sdk-level retry * driver-level retry.

Either way, the point is that we need to increase sdk-level retry so that the driver-level retry doesn't prematurely exit due to a throttling error

wongma7:
/approve

w8ing on cla

josselin-c:
cla signed

wongma7:
/lgtm

will manual merge if bot complains about coverlay

wongma7:
FYI @ayberk as I know you are following https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/754 and we are gonna need to discuss how to make the same fix in in-tree cloudprovider and maybe in out-of-tree cloudprovider v2 API that could be consumed by this driver.

PR #771, opened 2021-02-25
delete leaked volume if driver don't know the volume status

### LINKED ISSUE #754: Volumes may leak during Create operation: Delete/clean volume if waitForVolume fails during Create operation

AndyXiangLi:
Fixes #754 
Delete volume if the driver don't know about the volume status after creation
**What testing is done?** 
Could see this log after denying DescribeVolume api call manually
```volume is deleted because it is not in desired state within retry limit```
Added unit test

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -343,6 +343,13 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	}
 
 	if err := c.waitForVolume(ctx, volumeID); err != nil {
+		// To avoid leaking volume, we should delete the volume just created
+		// TODO: Need to figure out how to handle DeleteDisk failed scenario instead of just log the error
+		if _, error := c.DeleteDisk(ctx, volumeID); error != nil {
+			klog.Errorf("%v failed to be deleted, this may cause volume leak", volumeID)
+		} else {
+			klog.V(5).Infof("%v is deleted because it is not in desired state within retry limit", volumeID)
+		}
 		return nil, fmt.Errorf("failed to get an available volume in EC2: %v", err)
 	}
 
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -41,14 +41,15 @@ const (
 
 func TestCreateDisk(t *testing.T) {
 	testCases := []struct {
-		name               string
-		volumeName         string
-		volState           string
-		diskOptions        *DiskOptions
-		expDisk            *Disk
-		expErr             error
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/cloud/cloud_test.go [hunk 1]
```diff
@@ -194,7 +195,8 @@ func TestCreateDisk(t *testing.T) {
 				Tags:             map[string]string{VolumeNameTagKey: "vol-test"},
 				AvailabilityZone: "",
 			},
-			expErr: fmt.Errorf("failed to get an available volume in EC2: timed out waiting for the condition"),
+			cleanUpFailedVolume: true,
+			expErr:              fmt.Errorf("failed to get an available volume in EC2: timed out waiting for the condition"),
```
wongma7, line 199: I think the test case above (DescribeVolume error ) we should also expect to clean up.

(the test case names are a bit confusing since CreateVolume could mean ec2.createvolume or csi.createvolume. If it helps and you dont mind could you disambiguate them in the test names)

pkg/cloud/cloud.go
```diff
@@ -343,6 +343,13 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	}
 
 	if err := c.waitForVolume(ctx, volumeID); err != nil {
+		// To avoid leaking volume, we should delete the volume just created
+		// TODO: Need to figure out how to handle DeleteDisk failed scenario instead of just log the error
+		if _, error := c.DeleteDisk(ctx, volumeID); error != nil {
```
ayberk, line 348: So delete works if the volume is not available when we make the call?

AndyXiangLi: Yeah, I did a test on my account. One concern here is  if the deleteVolume call failed as well then we can not prevent the volume get leaked anyway..

ayberk: Sure, that makes sense.

pkg/cloud/cloud_test.go
ayberk: +1

AndyXiangLi: updated

AndyXiangLi:
Cherry pick upstream fix back to v1.21
https://github.com/kubernetes/kubernetes/pull/100934

Issue #729, opened 2021-02-06
Is it possible to use the standard k8s topology labels?


TBBle:
**Is your feature request related to a problem?/Why is this needed**

Looking into the interactions of this driver with the Cluster Autoscaler, i.e., https://github.com/kubernetes/autoscaler/issues/3230#issuecomment-774387646, I have been unable to identify a good way to support the scale-from-zero case without needing to either change Cluster Autoscaler to always add this driver's topology label, or have each node group's ASG manually specify the topology label `topology.ebs.csi.aws.com/zone`.

/feature

**Describe the solution you'd like in detail**

A parameter to specify the topology label used by the PV dynamic allocation, which in the EKS case, can be set to [`topology.kubernetes.io/zone`](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone) (or `failure-domain.beta.kubernetes.io/zone` on older clusters) and behave correctly.

**Describe alternatives you've considered**

* Hard-coding the `topology.ebs.csi.aws.com/zone` label into Cluster Autoscaler AWS support the same way the standard zone label is hard-coded.
* Manually including the tag `k8s.io/cluster-autoscaler/node-template/label/topology.ebs.csi.aws.com/zone` on each node group definition in eksctl to which the EBS CSI Daemonset will be deployed, i.e. manually matching the local taint/toleration setup.

See https://github.com/kubernetes/autoscaler/issues/3230 for more details on these alternatives.

**Additional context**

I'm not sure what the reason is for using a new label originally, which is why I suggest it be an option, rather than changing the default. I'm guessing in non-EKS deployments on EC2, the standard zone label may not be reliably the AWS Zone, if the AWS Cloud Controller is not in-use?

That said, if the reason is that that the label _must_ be owned by the driver for [spec](https://kubernetes-csi.github.io/docs/topology.html) or behaviour reasons, then this wouldn't be feasible at all, and we'll probably be stuck manually ensuring that the node groups with the tag correctly match the tolerations of the Daemonset.

ayberk:
I really don't know why we didn't use the well-known label. @wongma7 @leakingtapan maybe?

I agree we should just use them instead. Initially we should support both labels though so we can provide a deprecation warning.

leakingtapan:
[...]

PR #773, opened 2021-03-02
Add well-known topology label

ayberk:
Feature

See #729. This is the first part where we start writing the label and also give priority to it when reading topology labels. This change is completely backward-compatible with no change in behavior as far as users are concerned.

**What testing is done?** 
Unit tests
Single AZ e2e (there was one failed case that should be unrelated, probably flake)

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -41,11 +41,14 @@ const (
 
 const (
 	DriverName      = "ebs.csi.aws.com"
-	TopologyKey     = "topology." + DriverName + "/zone"
 	AwsPartitionKey = "topology." + DriverName + "/partition"
 	AwsAccountIDKey = "topology." + DriverName + "/account-id"
 	AwsRegionKey    = "topology." + DriverName + "/region"
 	AwsOutpostIDKey = "topology." + DriverName + "/outpost-id"
+
+	WellKnownTopologyKey = "topology.kubernetes.io/zone"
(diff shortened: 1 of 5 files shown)


pkg/driver/driver.go
```diff
@@ -41,11 +41,14 @@ const (
 
 const (
 	DriverName      = "ebs.csi.aws.com"
-	TopologyKey     = "topology." + DriverName + "/zone"
 	AwsPartitionKey = "topology." + DriverName + "/partition"
 	AwsAccountIDKey = "topology." + DriverName + "/account-id"
 	AwsRegionKey    = "topology." + DriverName + "/region"
```
wongma7, line 46: should we deprecate the region one as well? It seems like we only set it in the outposts case

ayberk: Hmm makes sense yeah. I didn't include it because as you said we only use it internally, but that would be more future-proof.

AndyXiangLi:
LGTM, let's wait for another approval :-)

wongma7:
sorry I think you have to rebase since I moved some stuff to node_linux.go ...

wongma7:
```

pkg/driver/node.go:27:2: imported and not used: "golang.org/x/sys/unix"
pkg/driver/node.go:36:2: volume redeclared as imported package name
	previous declaration at pkg/driver/node.go:28:2 
```

Issue #758, opened 2021-02-22
ebs-csi-controller blocks updating EKS managed node group


imbohyun1:
/kind bug

**What happened?**
While updating managed node group by changing the version of launch template, I observed that if one of nodes from older node group has `ebs-csi-controller`, then it failed to evict pods with `PodEvictionFailure` after it hit the max retries to evict `ebs-csi-controller` pod.

    "Errors": [
      {
        "ErrorCode": "PodEvictionFailure",
        "ErrorMessage": "Reached max retries while trying to evict pods from nodes in node group ng-upgrade",
        "ResourceIds": [
          "ip-192-168-48-49.ap-northeast-2.compute.internal"
        ]
      }
    ]

Looking into the issue, it appeared that kube-scheduler kept rescheduling the ebs-csi-controller pod to the node right after the pod was evicted. While looking at the ebs-csi-controller's manifest (ebs-csi-controller deployment: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/base/controller.yaml), I could see the tolerations setting for the ebs-csi-controller is not preventing from being rescheduled even though the node has been tainted as `eks.amazonaws.com/nodegroup=unschedulable:NoSchedule`.

      tolerations:
        - operator: Exists

According to [Kubernetes document](https://v1-18.docs.kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
), "An empty key with operator Exists matches all keys, values and effects which means this will tolerate everything."

**What you expected to happen?**
I expected that ebs-csi-controller was going to be evicted like other pods to end up with updating managed node group successfully. 
To resolve this, I had to `remove the tolerations` from the ebs-csi-controller or modify tolerations to be applied only to specific effects as below.

      tolerations:
        - operator: Exists
          effect: NoExecute
          tolerationSeconds: 300

**How to reproduce it (as minimally and precisely as possible)?**
1) Create a node group with launch template.
2) Deploy EBS CSI Driver on the cluster by following guidance of [official document](https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi.html).
3) Update managed node group with new version of launch template
4) The node running ebs-csi-controller will not be drained. and It will fail with `PodEvictionFailure`.

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): v1.18
- Driver version: v0.9.0

andrewgeller:
[...]

PR #775, opened 2021-03-04
add toleration time to NoExecute effect

### LINKED ISSUE #588: `tolerationSeconds` for tolerations for controller
At the moment `tolerations` for the controller deployment https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/aws-ebs-csi-driver/templates/controller.yaml looks like this:

```yaml
      tolerations:
        - operator: Exists
```

which effectively allows it run everywhere. Which is fine. The problem is that with this toleration kubernetes won't reschedule the controller if a node where it runs suddenly becomes unavailable.

See https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

> Kubernetes automatically adds a toleration for node.kubernetes.io/not-ready and node.kubernetes.io/unreachable with tolerationSeconds=300, **unless you, or a controller, set those tolerations explicitly.**

(highlight is mine)

So, you run it with default replica count 2, then you get unlucky and both nodes where those replicas run die. Now you end up having no ebs controller whatsoever and it won't fix without manual intervention.

/feature

I think the default `tolerationSeconds: 300` (configurable) should be added as well.

There is no, it's how kubernetes works.

**Additional context**

AndyXiangLi:
Fixes #588 
Distinct two types of tolerations and add default 300s tolerationSeconds to all the NoExecute taints.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.9.13
+version: 0.9.14
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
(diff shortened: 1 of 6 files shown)


wongma7:
I would prefer we stop tolerating NoSchedule. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/758#issuecomment-786792738

AndyXiangLi:
close this PR to address unit test issue, will reopen it later.

PR #776, opened 2021-03-04
add toleration time to NoExecute effect --- Enable in next release

### LINKED ISSUE #588: `tolerationSeconds` for tolerations for controller

AndyXiangLi:
Continue from #775 
Fixes #588
Not tolerate NoSchedule effect and add default 300s tolerationSeconds to all the NoExecute taints.

As we discussed, will default to tolerateAllTaints for now, will disable this in next release.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.9.13
+version: 0.9.14
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -33,6 +33,12 @@ spec:
       tolerations:
         {{- if .Values.tolerateAllTaints }}
         - operator: Exists
+        {{- else }}
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - operator: Exists
+          effect: NoExecute
+          tolerationSeconds: 300
         {{- end }}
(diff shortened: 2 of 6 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -70,7 +70,7 @@ resources:
 
 priorityClassName: ""
 nodeSelector: {}
-tolerateAllTaints: true
```
wongma7, line 73: I do think we should change the default but unsure if it's "safe" by helm standards to do it in a point release.

wdyt @ayberk

ayberk: Hmm this should be mostly fine, but it _might_ still break the existing workflow for some customers with basically no heads-up.

I'd rather change the behavior with a release so we can at least have a note in the release notes.

AndyXiangLi: Agree, How about we keep the current behavior for now, and turn this flag off in the next release(which should be pretty soon)

AndyXiangLi: Updated

PR #782, opened 2021-03-08
Add a note for EKS customers on README

ayberk:
/kind documentation

Customers might (understandably) assume they can use the driver on their EKS clusters. We should make a special note that that's not the case. 

**What testing is done?** 
N/A


wongma7:
This is not true, you can create a PV that says either plugin: csi or plugin: awsElasticBlockStore. If we are talking about migration then it is partly true because a situation where migration is enabled on kubelet but not the control plane (which  would be impossible on EKS) is not supported

ayberk:
You're absolutely right.

/close

PR #786, opened 2021-03-09
Prep for Windows support: Copy pkg/mounter and refactor to use k8s.io/mount-utils

wongma7:
**Is this a bug fix or adding new feature?** new

**What is this PR about? / Why do we need it?** As mentioned in pkg/mounter/README I have copied the csi-proxy implementation of mount.Interface from https://github.com/kubernetes-sigs/azuredisk-csi-driver/tree/master/pkg/mounter . Basically it is a platform independent package where on unix the mounter will be the same SafeFormatAndMount implementation as before and on windows the mounter will be the newer csi-proxy implementation. Since GCE PD also has their own similar implementation, we might want to consider moving it to its own k8s repo, but in meantime copy/pasting the code is the least painful way to consume it IMO. I don't want to depend on azure/gce repos and pull in a mountain of dependencies.

Since pkg/mounter implements the latest mount interface from k8s.io/mount-utils, not k8s.io/utils/mount, I have to do some refactoring to use it.

I also refactored NodeStage tests to be table driven but it's too much effort to refactor them all. The length of the test file and amount of code repetition is becoming too much IMO.

**What testing is done?** 
unit testing locally go test ./pkg/driver/...


pkg/util/util.go
```diff
@@ -59,13 +57,13 @@ func ParseEndpoint(endpoint string) (string, string, error) {
 		return "", "", fmt.Errorf("could not parse endpoint: %v", err)
 	}
 
-	addr := path.Join(u.Host, filepath.FromSlash(u.Path))
```
wongma7, line 62: Tests are failing, this code is pretty confusing.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/util/util_test.go

https://golang.org/pkg/net/url/#URL

`tcp:///127.0.0.1` should mean "listen on path /127.0.0.1"  but we are twisting it to be equivalent to `tcp://127.0.0.1` which should mean "listen on host 127.0.0.1". This seems wrong but I don't care to change it now.

For unix scheme, https://github.com/grpc/grpc/blob/master/doc/naming.md#name-syntax suggests we should support `unix:path` and `unix://absolute-path`. golang url.Parse will parse the first one perfectly but the first segment of the `absolute-path` in the second one will be parsed as the host.

In interest of not breaking anything I will make this change in a future PR so that tests pass and things work on Windows...

wongma7:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/786/pull-aws-ebs-csi-driver-migration-test-latest/1369405539636744192

```

Mar  9 22:15:21.859: INFO: At 2021-03-09 22:15:21 +0000 UTC - event for awsf9sz7: {external-resizer ebs.csi.aws.com } FileSystemResizeRequired: Require file system resize of volume on node
Mar  9 22:15:21.859: INFO: At 2021-03-09 22:15:21 +0000 UTC - event for awsf9sz7: {external-resizer ebs.csi.aws.com } VolumeResizeFailed: Mark PVC "volume-expand-269/awsf9sz7" as file system resize required failed: can't patch status of  PVC volume-expand-269/awsf9sz7 with Operation cannot be fulfilled on persistentvolumeclaims "awsf9sz7": the object has been modified; please apply your changes to the latest version and try again 
```

migration is failing because external-resizer can't mark PVC required? This test doesnt involve the node plugin at all so I don't think it's related to my changes

Also unsure if external-resizer will retry in case of transient patch error like this or if it's a legit bug

/retest

pkg/driver/mount.go
```diff
@@ -19,44 +19,50 @@ package driver
...
+	// Implemented by NodeMounter.SafeFormatAndMount
+	mountutils.Interface
 	FormatAndMount(source string, target string, fstype string, options []string) error
-	GetDeviceName(mountPath string) (string, int, error)
-	MakeFile(pathname string) error
-	MakeDir(pathname string) error
-	ExistsPath(filename string) (bool, error)
+
+	// Implemented by NodeMounter.SafeFormatAndMount.Exec
+	// TODO this won't make sense on Windows with csi-proxy
```
ayberk, line 33: What are the TODOs here? Remove them?

pkg/driver/node_linux.go
```diff
@@ -0,0 +1,109 @@
...
+// findDevicePath finds path of device and verifies its existence
+// if the device is not nvme, return the path directly
+// if the device is nvme, finds and returns the nvme device path eg. /dev/nvme1n1
+func (d *nodeService) findDevicePath(devicePath, volumeID string) (string, error) {
+	exists, err := d.mounter.PathExists(devicePath)
+	if err != nil {
+		return "", err
+	}
+
+	// If the path exists, assume it is not nvme device
```
ayberk, line 41: Can we put a link to the source of this assumption and/or explain why it's safe to do so?

pkg/driver/mount.go
wongma7: i don't know yet what if anything needs to be done on windows until i actually test it which could take a while so I droped a lot of these windows-related TODOS throughout the code.

pkg/driver/node_linux.go
wongma7: i have no clue, keep in mind this is copy/pasted from node.go

wongma7: I think this is it, will link: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html#available-ec2-device-names

ayberk: Thanks. You can do it on the next PR if you like, up to you.

pkg/driver/mount.go
wongma7: so far I've only tested up to the point where it gets the device path (device number on windows) . All of the subsequent execs and mkdirs and mkfiles, i expect to fail, but exactly how to fix that I don't kno

PR #787, opened 2021-03-11
hack/e2e: Support passing helm values as values.yaml and make other similar files optional

wongma7:

**What is this PR about? / Why do we need it?** Minor changes to make the scripts more configurable and usable by EFS driver. https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/354 

It seems like git read tree is the simplest way to consume these scripts from EFS. subtree is more complicated because hack/e2e is a subdirectory of this repo and not its own repo

**What testing is done?** CI of PR https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/354  and CI of this PR.

### DIFF
--- a/hack/e2e/README.md
+++ b/hack/e2e/README.md
@@ -24,3 +24,24 @@ GINKGO_FOCUS=Dynamic.\*xfs.\*should.store.data \
 GINKGO_NODES=1 \
 ./hack/e2e/run.sh
 ```
+
+# git read-tree
+
+Reference: https://stackoverflow.com/questions/23937436/add-subdirectory-of-remote-repo-with-git-subtree
+
+How to consume this directory by read-treeing the ebs repo:
+
+```
+git remote add ebs git@github.com:kubernetes-sigs/aws-ebs-csi-driver.git
+git fetch ebs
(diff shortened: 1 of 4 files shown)


hack/e2e/README.md
```diff
@@ -24,3 +24,24 @@ GINKGO_FOCUS=Dynamic.\*xfs.\*should.store.data \
...
+
+How to consume this directory by read-treeing the ebs repo:
+
+```
+git remote add ebs git@github.com:kubernetes-sigs/aws-ebs-csi-driver.git
+git fetch ebs
+git read-tree --prefix=hack/e2e/ -u ebs/master:hack/e2e
+```
+
+To commit changes and submit them as a PR back to the ebs repo:
```
ayberk, line 40: Can you elaborate on this? Why are we doing this instead of a regular PR?

wongma7: if you've read-treed this director ythen youa re in another repo (efs). so if you want to commit your changes back to theoriginal repo (ebs) you need to do this. does it make sense, not sure what to add here

ayberk: Yeah got it. That's what I kind of guessed but wanted to make sure. lgtm.

Issue #723, opened 2021-02-04
Bump kubernetes dependency to 1.19/1.20


wongma7:
/kind bug

Continue this PR https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/663 

To run the latest tests.

And to pull in https://github.com/kubernetes/kubernetes/pull/98031 once it is released.

wongma7:
fixed

PR #792, opened 2021-03-12
 Bump k8s dependencies to 1.20.4, csi spec to 1.3, external-snapshotter client to v4, go to 1.16

wongma7:
**Is this a bug fix or adding new feature?** fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/723

commits 1-5: go dependency bumps

commits 6-9: YAML image/API version bumps

commits 10-13: e2e test kops version bumps

**What testing is done?**


wongma7:
client api changed, will fix.

 ../../../../pkg/mod/github.com/kubernetes-csi/external-snapshotter/v2@v2.0.1/pkg/client/clientset/versioned/typed/volumesnapshot/v1beta1/volumesnapshot.go:188:5: not enough arguments in call to c.client.Patch(pt).Namespace(c.ns).Resource("volumesnapshots").SubResource(subresources...).Name(name).Body(data).Do
	have ()
	want (context.Context)
../../../../pkg/mod/github.com/kubernetes-csi/external-snapshotter/v2@v2.0.1/pkg/client/clientset/versioned/typed/volumesnapshot/v1beta1/volumesnapshotclass.go:70:5: not enough arguments in call to c.client.Get().Resource("volumesnapshotclasses").Name(name).VersionedParams(&options, scheme.ParameterCodec).Do

wongma7:
pkg/driver/driver.go:128:33: cannot use d (type *Driver) as type csi.ControllerServer in argument to csi.RegisterControllerServer:
	*Driver does not implement csi.ControllerServer (missing ControllerGetVolume method) 

I 'm going to leave it unimplemented

wongma7:
The code added in 1.19 for getting metrics from kcm pod doesn't work for us:

Mar 15 16:56:01.625: INFO: ERROR unable to parse requirement: invalid label value: "kube-controller-manager-ip-172-20-39-70.us-west-2.compute.internal": at key: "name": must be no more than 63 characters

wongma7:
Also, there is no way to disable this metrics grabbing code. We already check KCM metrics ourselves at the end of the test to see if csi or in tree was getting called so it's redundant :/

wongma7:
hmmm I need to fix my logging code:


###
## Printing pod ebs-csi-controller-77fd584dc8-hkg6k ebs-plugin container logs
#
Error from server (Forbidden): Forbidden (user=system:anonymous, verb=get, resource=nodes, subresource=proxy) ( pods/log ebs-csi-controller-77fd584dc8-hkg6k)

wongma7:
actually it's causing test failures as well since some of them check pod logs. 
```
                  s: "error running /usr/local/bin/kubectl --server=https://api-test-cluster-8868-k8s-rjdnu3-1175551248.us-west-2.elb.amazonaws.com --kubeconfig=/root/.kube/config --namespace=volume-3874 exec aws-injector --namespace=volume-3874 -- /bin/sh -c echo 'Hello from aws from namespace volume-3874' > /opt/0/index.html:\nCommand stdout:\n\nstderr:\nerror: unable to upgrade connection: Forbidden (user=system:anonymous, verb=create, resource=nodes, subresource=proxy)\n\nerror:\nexit status 1",
              }, 
```

wongma7:
as of 119, kops kubelet default authorization mode is webook , probably it's cuz of that.

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -18,7 +18,7 @@ sidecars:
     tag: "v3.0.0"
   snapshotterImage:
     repository: k8s.gcr.io/sig-storage/csi-snapshotter
-    tag: "v3.0.3"
+    tag: "v4.0.0"
```
ayberk, line 21: 4.0 requires k8s 1.20+. I don't think we should do this just yet.

https://kubernetes-csi.github.io/docs/external-snapshotter.html

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -31,7 +31,7 @@ sidecars:
 
 snapshotController:
   repository: k8s.gcr.io/sig-storage/snapshot-controller
-  tag: "v3.0.3"
+  tag: "v4.0.0"
```
ayberk, line 34: Same for this one :( https://kubernetes-csi.github.io/docs/snapshot-controller.html

hack/e2e/kops.sh
```diff
@@ -69,3 +78,36 @@ function kops_delete_cluster() {
   loudecho "Deleting cluster ${CLUSTER_NAME}"
   ${KOPS_BIN} delete cluster --name "${CLUSTER_NAME}" --state "${KOPS_STATE_FILE}" --yes
 }
+
+# TODO switch this to python, all this hacking with jq stinks!
```
ayberk, line 82: +1

hack/e2e/kops.sh
```diff
@@ -69,3 +78,36 @@ function kops_delete_cluster() {
...
+function kops_patch_cluster_file() {
+  CLUSTER_FILE=${1}
+  KOPS_PATCH_FILE=${2}
+
+  loudecho "Patching cluster $CLUSTER_NAME with $KOPS_PATCH_FILE"
+
+  # Temporary intermediate files for patching
+  CLUSTER_FILE_0=$CLUSTER_FILE.0
```
ayberk, line 90: What does `$CLUSTER_FILE.0` do here?

ayberk (commented):
👏 🙇

charts/aws-ebs-csi-driver/values.yaml
wongma7: darn, ok

hack/e2e/kops.sh
wongma7: I can't open the file for writing via bash redirect and read it at the same time  to "patch" it hence this convoluted 'intermediate files' idea.

wongma7:
uhh I think I accidentally pushed something to freak out the bot?? checking

wongma7:
something to do with the blacklist /denylist change. https://github.com/kubernetes/test-infra/pull/21687

wongma7:
TooManyLoadBalancers

deploy/kubernetes/base/controller.yaml
```diff
@@ -32,7 +31,7 @@ spec:
           tolerationSeconds: 300
       containers:
         - name: ebs-plugin
-          image: k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v0.10.0
+          image: k8s.gcr.io/provider-aws/aws-ebs-csi-driver:latest
```
ayberk, line 34: Hmm this is kind of dangerous. What if we start building alpha images?

wongma7: people should be using the stable overlay. However I agree best not to even mention latest tag, I'l lcheck where it's coming from (like is it  being propagated from a helm chart?)

wongma7:
replaced by https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/828

Issue #393, opened 2019-10-23
Unify cloud provider between legacy and CSI driver


jsafrane:
[Kubernetes cloud provider for AWS](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/legacy-cloud-providers/aws) is still being actively maintained and bugs are fixed there. It could be worth importing it instead of using a fork in this repo. The code there has been well tested in real world and it would ensure that migration to CSI is as smooth as possible.

On the other hand, the code is old and has very little unit tests so refactoring is hard.

wongma7:
100% agree. For example, https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/389 refers to a fix in legacy cloudprovider. If we use legacy, we could get the fixes for free instead of having to backport every improvement to our "mini" cloudprovider fork in pkg/cloud

leakingtapan:
I like the idea of unifying the two cloud provider between legacy cloud provider and the once being implemented in this CSI driver. I would like more discussion on which one to pick before we can go with importing legacy cloud provider. Several questions:
1. I'm assuming you are proposing to use [legacy provider for AWS](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/legacy-cloud-providers/aws) to replace the [cloud.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go) in this package, right?

2. Both provider defines the interface in slight different way, do we know what are the deltas? Also the cloud.go in the package implements extra features such as snapshot. I don't think it will be a drop in replacement.

3. What's your long term view of the legacy provider? My understanding is that it will be moved to CCM for non EBS related code and EBS related code will be moved into this project. 

I would propose to change this issue topic to **Unify cloud provider between legacy and CSI driver**

jsafrane:
> I'm assuming you are proposing to use legacy provider for AWS to replace the cloud.go in this package, right?

Yes

> Both provider defines the interface in slight different way, do we know what are the deltas? Also the cloud.go in the package implements extra features such as snapshot. I don't think it will be a drop in replacement.

I don't disagree...

> What's your long term view of the legacy provider? My understanding is that it will be moved to CCM for non EBS related code and EBS related code will be moved into this project.

Yes, probably.

> Given above, I would propose to change this issue topic to Unify cloud provider between legacy and CSI driver

[...]

PR #801, opened 2021-03-18
handling describe instances consistency issue

### LINKED ISSUE #389: Handle DescribeInstances eventual consistency
See https://github.com/kubernetes/kubernetes/issues/77389 for more details for the specific race condition.

There is a potential [fix](https://github.com/kubernetes/kubernetes/pull/83567) upstream for in-tree EBS CSI plugin. We should do the same for EBS CSI.

vdhanan:
fixes #389 

the describeInstances API follows an eventual consistency model. the csi driver should handle the fact that it may get inconsistent responses from this API

**What testing is done?** 
manual testing trying to reproduce

### DIFF
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -371,7 +371,7 @@ func TestAttachDisk(t *testing.T) {
 
 			vol := &ec2.Volume{
 				VolumeId:    aws.String(tc.volumeID),
-				Attachments: []*ec2.VolumeAttachment{{State: aws.String("attached")}},
+				Attachments: []*ec2.VolumeAttachment{{Device: aws.String("/dev/xvdba"), InstanceId: aws.String("node-1234"), State: aws.String("attached")}},
 			}
 
 			ctx := context.Background()
@@ -417,6 +417,12 @@ func TestDetachDisk(t *testing.T) {
 			nodeID:   "node-1234",
 			expErr:   fmt.Errorf("DetachVolume generic error"),
 		},
+		{
+			name:     "fail: DetachVolume returned not found error",
+			volumeID: "vol-test-1234",
+			nodeID:   "node-1234",
+			expErr:   fmt.Errorf("DetachVolume not found error"),
+		},
 	}
 
 	for _, tc := range testCases {
@@ -433,7 +439,12 @@ func TestDetachDisk(t *testing.T) {
 			ctx := context.Background()
 			mockEC2.EXPECT().DescribeVolumesWithContext(gomock.Eq(ctx), gomock.Any()).Return(&ec2.DescribeVolumesOutput{Volumes: []*ec2.Volume{vol}}, nil).AnyTimes()
 			mockEC2.EXPECT().DescribeInstancesWithContext(gomock.Eq(ctx), gomock.Any()).Return(newDescribeInstancesOutput(tc.nodeID), nil)
-			mockEC2.EXPECT().DetachVolumeWithContext(gomock.Eq(ctx), gomock.Any()).Return(&ec2.VolumeAttachment{}, tc.expErr)
+			switch tc.name {
+			case "fail: DetachVolume returned not found error":
+				mockEC2.EXPECT().DetachVolumeWithContext(gomock.Eq(ctx), gomock.Any()).Return(nil, awserr.New("InvalidVolume.NotFound", "foo", fmt.Errorf("")))
+			default:
+				mockEC2.EXPECT().DetachVolumeWithContext(gomock.Eq(ctx), gomock.Any()).Return(&ec2.VolumeAttachment{}, tc.expErr)
+			}
 
 			err := c.DetachDisk(ctx, tc.volumeID, tc.nodeID)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/cloud/devicemanager/manager.go
```diff
@@ -119,7 +119,7 @@ func (d *deviceManager) NewDevice(instance *ec2.Instance, volumeID string) (*Dev
 
 	// Check if this volume is already assigned a device on this machine
 	if path := d.getPath(inUse, volumeID); path != "" {
-		return d.newBlockDevice(instance, volumeID, path, true), nil
+		return nil, fmt.Errorf("volume already assigned")
```
AndyXiangLi, line 122: Can we add a warning here?
Something like klog.Warningf("Got assignment call for already-assigned volume: %s@%s, volumeID, instanceID)
And if we can get the volume status for free, it would be better to log the volume status as well (just like in-tree)

AndyXiangLi: One thing related to ControllerPublishVolume idempotent, I think we'd better keep tracking the volume status, per discussion [here](https://github.com/kubernetes/kubernetes/issues/77389), we can only return this error when volume is in detaching status, and if volume is in attached status, we should return the old statement to keep as much idempotent as we can.. LMK you opinion.

pkg/cloud/devicemanager/manager_test.go
```diff
@@ -30,27 +30,39 @@ func TestNewDevice(t *testing.T) {
 		existingDevicePath string
 		existingVolumeID   string
 		volumeID           string
+		status             string
```
AndyXiangLi, line 33: nit: Can we rename this field to volumeStatus?

pkg/cloud/devicemanager/manager.go
```diff
@@ -119,6 +119,10 @@ func (d *deviceManager) NewDevice(instance *ec2.Instance, volumeID string) (*Dev
 
 	// Check if this volume is already assigned a device on this machine
 	if path := d.getPath(inUse, volumeID); path != "" {
+		if status := d.getDeviceStatus(volumeID, instance); status == "detaching" {
```
AndyXiangLi, line 122: nit: Can we rename this field to volumeStatus?

AndyXiangLi: Can you also put a comment here that this fix will only handle the detaching inconsistency, if the status change from detach to attached it will still fail (Just for tracking purpose)

wongma7:
can you describe in detail the exact scenario this is fixing?

If a volume is in detaching state, and we try to attach it back to the same node, what's the problem?

BTW, I have been advocating for us to move to / adopt the in-tree cloudprovider code instead of trying to roll our own and reinvent the wheel.

vdhanan:
> can you describe in detail the exact scenario this is fixing?
> 
> If a volume is in detaching state, and we try to attach it back to the same node, what's the problem?
> 
> BTW, I have been advocating for us to move to / adopt the in-tree cloudprovider code instead of trying to roll our own and reinvent the wheel.

@wongma7 if a pod with a volume attached dies, that volume will be detached, since we don't know where the pod will be recreated. the csi driver checks that the volume has been detached by calling DescribeVolumes. if the pod is recreated on the same node, the csi driver will attempt to reattach the volume. however, if the volume appears in the DescribeInstances call (which is only eventually consistent, meaning it can report stale info) during the attach workflow, the driver assumes it's already assigned and doesn't bother trying to attach it. by erroring out if we see a volume in detaching state, we ensure that the attach workflow will retry, hopefully when DescribeInstances is reporting accurately. (apologies if i used any incorrect terminology here)

AndyXiangLi:
> > can you describe in detail the exact scenario this is fixing?
> > If a volume is in detaching state, and we try to attach it back to the same node, what's the problem?
> > BTW, I have been advocating for us to move to / adopt the in-tree cloudprovider code instead of trying to roll our own and reinvent the wheel.
> 
> @wongma7 if a pod with a volume attached dies, that volume will be detached, since we don't know where the pod will be recreated. the csi driver checks that the volume has been detached by calling DescribeVolumes. if the pod is recreated on the same node, the csi driver will attempt to reattach the volume. however, if the volume appears in the DescribeInstances call (which is only eventually consistent, meaning it can report stale info) during the attach workflow, the driver assumes it's already assigned and doesn't bother trying to attach it. by erroring out if we see a volume in detaching state, we ensure that the attach workflow will retry, hopefully when DescribeInstances is reporting accurately. (apologies if i used any incorrect terminology here)

if the volume is appearing in the instance Attachment list regardless the volume status, we thought this volume is attached to the instance, driver will not try to call attachVolume api and wait for volume to be attached https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L384
This will take long time to timeout (2800s from comment)

wongma7:
What if the volume gets detached and then in our reattach attempt the DescribeInstances is so stale that it returns the volume is still in state `attached`? Then we will proceed and then the volume will get unexpectedly detached no?

Notice in the cloud provider code they check the state in each polling attempt and break out if the volume unexpectedly becomes detached. https://github.com/kubernetes/kubernetes/blob/a55bd631728590045b51a4f65bba31aed1415571/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L2205. This is a better solution.

I would like to see either
a) Full design doc that accounts for all possibility and justification why we need to invent our own solution. We are dealing with race condition that could cause user to wait for ~45 mins or cause volumes to unexpectedly get detached/attached, solution has to be thorough.
or
b) Copy the existing cloud provider solution which has already gone through its own review, been tested over multiple kubernetes versions, etc. Especially if we care about migration compatibility, this is IMO the best option, I just never got around to doing it. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/393

https://github.com/kubernetes/kubernetes/blob/a55bd631728590045b51a4f65bba31aed1415571/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L2178

AndyXiangLi:
> What if the volume gets detached and then in our reattach attempt the DescribeInstances is so stale that it returns the volume is still in state `attached`? Then we will proceed and then the volume will get unexpectedly detached no?
> 
> Notice in the cloud provider code they check the state in each polling attempt and break out if the volume unexpectedly becomes detached. https://github.com/kubernetes/kubernetes/blob/a55bd631728590045b51a4f65bba31aed1415571/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L2205. This is a better solution.
> 
> I would like to see either
> a) Full design doc that accounts for all possibility and justification why we need to invent our own solution. We are dealing with race condition that could cause user to wait for ~45 mins or cause volumes to unexpectedly get detached/attached, solution has to be thorough.
> or
> b) Copy the existing cloud provider solution which has already gone through its own review, been tested over multiple kubernetes versions, etc. Especially if we care about migration compatibility, this is IMO the best option, I just never got around to doing it. #393
> 
> https://github.com/kubernetes/kubernetes/blob/a55bd631728590045b51a4f65bba31aed1415571/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L2178

@vdhanan In your testing, did you see the detached volume back to "attached" status?  I agree to port in-tree waitForAttachmentStatus to CSI as it is more robust

vdhanan:
I ported most of the waitForAttachmentStatus function from in-tree. I think after GA we should just consume the vendor code directly like Matthew mentioned.

AndyXiangLi:
> I ported most of the waitForAttachmentStatus function from in-tree. I think after GA we should just consume the vendor code directly like Matthew mentioned.

Can you add some unit test for the updated function?

wongma7:
/lgtm

thanks, I am really more confident if we just copy the code, I know it's not so glamorous to be doing that but since this issue is so tricky to debug and test (hard to reproduce race condition) I think it's best option!

pkg/cloud/cloud_test.go
```diff
@@ -1230,6 +1230,134 @@ func TestListSnapshots(t *testing.T) {
...
+
+			detachedVol := &ec2.Volume{
+				VolumeId:    aws.String(tc.volumeID),
+				Attachments: []*ec2.VolumeAttachment{{Device: aws.String("/dev/xvdba"), InstanceId: aws.String("1234"), State: aws.String("detached")}},
+			}
+
+			ctx := context.Background()
+
+			switch tc.name {
+			case "success: detached":
```
wongma7, line 1329: (just a style alternative): if u want to avoid depending on test case name you could make these anonymous functions, something like this https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node_test.go#L69

pkg/cloud/cloud_test.go
```diff
@@ -1230,6 +1230,134 @@ func TestListSnapshots(t *testing.T) {
...
+			expectedInstance: "1235",
+			expectedDevice:   "/dev/xvdba",
+			alreadyAssigned:  false,
+			expectError:      true,
+		},
+		{
+			name:             "failure: already assigned but wrong state",
+			volumeID:         "vol-test-1234",
+			expectedState:    volumeAttachedState,
+			expectedInstance: "1235",
```
wongma7, line 1301: should this say 1234? otherwise this test is giving us a false positive?

wongma7: I think it's giving false postiive at the moment cuz, we should be testing the case where: if we set alreadyAttached to true, and DescribeVolumes returns that the volume is detached, we want to error. Correct me if wrong.

vdhanan: yup you're right, it should be 1234

pkg/cloud/cloud_test.go
vdhanan: that's definitely cleaner. i'll use this style next time

AndyXiangLi:
/lgtm
/approve
Thanks!

PR #807, opened 2021-03-22
csi.storage.k8s.io/fstype is case sensitive

jsafrane:
bug

The external-provisioner does not treat `csi.storage.k8s.io/fstype` as case-insensitive. Update the documentation accordingly.

**What testing is done?** 
None :-)


wongma7:
w8 isn't this a bug in external provisioner too then?

/lgtm
/approve

PR #808, opened 2021-03-23
fix deploy stable ecr error kustomization file

ABNER-1:
This is a bug about ecr deploy error kustomization file

If do not fix this error, we can not deployed with this kustomization file.

**What testing is done?** 
I have tested in my eks cluster.


ABNER-1:
tell a joke

ayberk:
/lgtm
/approve

Thanks a lot!

PR #809, opened 2021-03-23
Cap IOPS when calculating from iopsPerGB

jsafrane:
bugfix

Consider AWS IOPS limits when calculating IOPS from `iopsPerGB` to always provision a supported volume. Users should be create a StorageClass "fast" that provides max. IOPS that AWS can support for any volume size they provide, without errors.

Without the PR, users must calculate `iopsPerGB * PVC size` and choose the right PVC size where resulting IOPS does not violate AWS IOPS limits. For example, with iopsPerGB = 10 and PVC size 5 GB, they get error `Volume iops of 50 is too low; minimum is 100.`

Please check the current AWS limits carefully. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html is not exactly clear what's max IOPS per GB for io2 volumes - on one place it's 500:1 and on another 1000:1, 500 seems to be the right ratio.

**What testing is done?** 
Wrote few new unit tests, checking **only** IOPS in `CreateVolumeInput`.
* Tested with 4GB io1 volume with 1, 50 and 500 IOPS per GB, got volume 100, 100 and 200 IOPS.
* Tested with 4GB io2 volume with 1, 50, 500 and 1000 IOPS per GB, got volume 100, 200, 2000 and 2000 IOPS.


jsafrane:
```
/home/prow/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/e2e/dynamic_provisioning.go:535
Expected
    <string>: topology.kubernetes.io/zone
to equal
    <string>: topology.ebs.csi.aws.com/zone
/home/prow/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/e2e/testsuites/testsuites.go:369
```
Did I break it?

ayberk:
That looks related to the recent change I made. I'll take a look today.

pkg/cloud/cloud.go
```diff
@@ -1073,3 +1087,22 @@ func getVolumeAttachmentsList(volume *ec2.Volume) []string {
...
 }
+
+// Calculate actual IOPS for a volume and cap it at supported AWS limits.
+// Using requstedIOPSPerGB allows users to create a "fast" storage class
+// (requstedIOPSPerGB = 50 for io1), which can provide the maximum iops
+// that AWS supports for any requestedCapacityGiB.
+func capIOPS(requestedCapacityGiB int64, requstedIOPSPerGB, minTotalIOPS, maxTotalIOPS, maxIOPSPerGB int64) int64 {
+	iops := requestedCapacityGiB * requstedIOPSPerGB
+
+	if iops < minTotalIOPS {
```
ayberk, line 1098: I'm not a big fan of this change. This might end up in a surprise bill for the customer, so if IOPS is smaller than minimum we should just return and explicit error instead of increasing it.

pkg/cloud/cloud.go
```diff
@@ -1073,3 +1087,22 @@ func getVolumeAttachmentsList(volume *ec2.Volume) []string {
...
+		iops = minTotalIOPS
+	}
+	if iops > maxTotalIOPS {
```
ayberk, line 1101: Can we add a log line for the `> max` cases?

ayberk (changes requested):
cc @AndyXiangLi for verifying EBS limits.

pkg/cloud/cloud.go
wongma7: consider the alternative as described in PR description is forcing the user to inspect the storageclass then edit their PVC to be larger which also incurs a larger bill. Using example in PR description, two scenarios:

1) We accept this PR and always increase iops to the floor: 5GiB at 100iops is 5*.125+100*.0625 = $7.125/month.
2) We force user to set the iops to the floor: 10GiB at 100 iops is  10*.125+100*.0625 = $7.75/month (+ the cost of their time to visit aws docs and pull out their calculator).

This is just one scenario, I don't really want to put the effort into drawing graphs. I think as long as we document this behavior it will be fine. Admins will appreciate they do not have to estimate their users' behavior (if they are on a tight budget they can just set iopspergb to 1 and call it a day) and users do not have to pull out their calculators.

So atm I am inclined to accept the PR.

wongma7 (approved):
I'm fine with this as long as we edit README too

pkg/cloud/cloud.go
ayberk: Here's my reasoning. If a user expects their volume to cost, say, $5 with the values they're setting, it becomes a problem at scale. The difference on paper is only $2, but what if you have 1000 of such volumes across 10 regions? Suddenly your bill is 2 x 10 x 1000 more than you expect. Am I missing something?

We can include the limits in the error message (or even a direct link to the docs) to make it a little bit easier on the user. I understand it's still not a great user experience, but I think it's preferable to having a surprise bill.

AndyXiangLi: To be safe, my idea is we don't make decision for customers, customer can adjust the iopsPerGB or volume size per their requirement as long as we deliver the message clearly (Like Ayberk mentioned, attach the doc link in the error message, etc).

wongma7: The user is mistaken in their $5 calculation if it's predicated on their volume having an impossible GB+IOPS configuration. The choice is not between $7 and $5, it's between $7 (increase both size and iops), $6 (increase only iops), and not running the Pod at all. So really we are debating whether we should let the user know about their mistake or fix it on their behalf. I prefer the latter as long as we document it, let me spell it out.

(Keep in mind on a multi-tenant cluster the person creating PVCs doesn't necessarily have permissions to create StorageClasses. Also the above numbers are imaginary but you get my point)

Say the driver provisioning for StorageClass A with iopspergb x is presented with an impossible GB+IOPS configuration. To remedy the situation, something/somebody must:

1) Increase SIZE and IOPS of volume such that its IOPS is greater than the minimum.
2) Increase IOPS of volume such that its IOPS is greater than the minimum.
3) Decide that fixing the volume to be a valid configuration would be too expensive, so do nothing, and use EFS instead or just not run their workload.

Note option 1) is ALWAYS more expensive than option 2).

If the driver throws an error and refuses to create the volume (like today), option 2 is unavailable. Because for StorageClass A there is no way for them to control iops except by increasing size of volume. So if they want to run their stateful workload they are forced to go with the more expensive option 1

If the driver fixes the volume for the user (if we accept this PR as-is), option 3 is unavailable. I don't think this option would be that popular. I think people would prefer their workload to run than not.

IF this is so contentious, we can make it a flag that defaults to current behavior.

wongma7: BTW  it seems like only bumping iops to min floor is contentious, capping iops at max is a no brainer.

wongma7: also check out this thread, i was having deja vu. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/575 in-tree behaves this way, and I was torn on whether we should behave the same

jsafrane: Well, there is CSI migration on the way and the CSI driver should behave the same as in-tree for migrated volumes. #575 mentions is going to be handled in the translation layer - but how? I don't see any code there and I don't think it's possible.

jsafrane:
BTW, my motivations for this PR are:
* Be in sync with in-tree.
* Better usability. 
  * Admins can prepare "default" storage class (gp2 / gp3?) and something "fast" (fast for *every* volume that users create, regardless of the size).
  * Users should not do any size+iops calculations at all. It's not users who actually create PVCs these days, they use StatefulSets, helm charts or other software.
 
(I don't like `iops` parameter in gp3 volumes because of this - users MUST do the calculations there. The error messages when users get out of the supported bounds are not exactly clear. And a storage class with 10 000 IOPS is fast for some volumes and super slow for others.)

pkg/cloud/cloud.go
jsafrane: Added a log into every branch.

jsafrane:
Updated README.

pkg/cloud/cloud.go
jsafrane: I am open to remove increasing the IOPS, however, but it must come with a very clear error message. The current one is very confusing. With 4 GiB volume and 1 IOPS per GB, I get:

```
could not create volume in EC2: InvalidParameterValue: Volume iops of 4 is too low; minimum is 100.
```

But there is no `iops` setting in the storage class. Only `iopsPerGB`.

jsafrane: talked to @msau42 on slack

What about this:
* Let's cap max IOPS to get within the supported range. I don't see any controversy here.
* For CSI storage classes, **do not** increase iops to the supported range and return error (= as it is today, only with a nicer error message).
* However, *increase the iops for storage classes migrated from in-tree*, to preserve existing behavior.

How to distinguish these two in `CreateVolume`?
  * CSI translation lib can add `alpha.csi.storage.k8s.io/migrated` parameter to CSI StorageClass when it's migrated from in-tree. Note: this translation is in-memory, in the external-provisioner. It's not stored in API server.
  * The driver sees `alpha.csi.storage.k8s.io/migrated` parameter in CreateVolume.
  * When users create a CSI StorageClass with `alpha.csi.storage.k8s.io/migrated` parameter, it's their problem. We will not document the parameter and we reserve the right to remove it in the future. That's why the `alpha` prefix, even though it will be there probably forever :-(

pkg/cloud/cloud.go
```diff
@@ -53,6 +53,17 @@ const (
...
 )
 
+// AWS provisioning limits.
+// Source: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html
+const (
+	io1MinTotalIOPS = 100
+	io1MaxTotalIOPS = 64000
+	io1MaxIOPSPerGB = 50
+	io2MinTotalIOPS = 100
+	io2MaxTotalIOPS = 256000
```
ayberk, line 63: This needs to be 64k, similar to io1. EBS team told us 256k is for block express, which is currently opt-in, so for most customers the limit is still 64k.

pkg/cloud/cloud.go
ayberk: > When users create a CSI StorageClass with alpha.csi.storage.k8s.io/migrated parameter, it's their problem. We will not document the parameter and we reserve the right to remove it in the future. That's why the alpha prefix, even though it will be there probably forever :-(

This also has a nice side-effect of enabling customers to bypass the error by adding this. Feels dirty, but I like it :P

I am all for this approach. It's the best of both worlds: we avoid the surprise bill issue and we don't change the behavior for transitioning volumes.

wongma7: why not make it an explicit parameter then like `ignoreIopsTooLow` (terrible name but ya)? If you introduce this hack then someone will rely on it and complain, doesn't amtter if it says alpha. migration would silently add the parameter, but for new StorageClasses admin will have the choice whether to err on side of ensuring all PVCs get a PV even if overprovisioned v.s. ensuring all PVCs either get an exactly-provisoined PV or no PV at all.

jsafrane: Added parameter `allowAutoIOPSPerGBIncrease` (naming is hard!) + unit tests + documentation, PTAL.

pkg/cloud/cloud.go
jsafrane: Lowered to 64k.

jsafrane:
/test pull-aws-ebs-csi-driver-unit
connection reset during vendor check

pkg/driver/controller.go
```diff
@@ -154,6 +155,8 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 			if err != nil {
 				return nil, status.Errorf(codes.InvalidArgument, "Could not parse invalid iopsPerGB: %v", err)
 			}
+		case AllowAutoIOPSPerGBIncreaseKey:
+			allowIOPSPerGBIncrease = value == "true"
```
ayberk, line 159: Can we make this case insensitive? Since the keys are, it'd be more consistent.

pkg/cloud/cloud.go
```diff
@@ -256,15 +268,25 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
+	case VolumeTypeIO1:
+		createType = diskOptions.VolumeType
+		iops, err = capIOPS(diskOptions.VolumeType, capacityGiB, int64(diskOptions.IOPSPerGB), io1MinTotalIOPS, io1MaxTotalIOPS, io1MaxIOPSPerGB, diskOptions.AllowIOPSPerGBIncrease)
+		if err != nil {
+			return nil, err
+		}
+	case VolumeTypeIO2:
 		createType = diskOptions.VolumeType
-		iops = capacityGiB * int64(diskOptions.IOPSPerGB)
+		iops, err = capIOPS(diskOptions.VolumeType, capacityGiB, int64(diskOptions.IOPSPerGB), io2MinTotalIOPS, io2MaxTotalIOPS, io2MaxIOPSPerGB, diskOptions.AllowIOPSPerGBIncrease)
```
ayberk, line 286: I think this is fine for now, but it might make more sense to create a struct for volume types with each encapsulating their limits, eventually. cc @AndyXiangLi something to keep in mind if we add more types with similar logic.

jsafrane:
Not sure if I missed any comment,IMO it's ready for review. @AndyXiangLi @wongma7 PTAL.

AndyXiangLi:
This looks good to me. Thank you!
/lgtm

jsafrane:
With this merged, I should update CSI translation lib to add `allowAutoIOPSPerGBIncrease` to translated StorageClasses. However, it will pin the library (and the external-provisioner) to the new version of the driver. CSI migration will fail with an older CSI driver version, because it does not know `allowAutoIOPSPerGBIncrease`. Is this allowed? :thinking:

wongma7:
IMO it's fine, we only need to support the combo of external-provisioner+driver we release+test in our helm chart. I guess someone out there may try using new external-provisioner and old driver and run into this issue but we shouldnt have to support it : /

jsafrane:
Filed https://github.com/kubernetes/kubernetes/pull/101082 to update the translation.

PR #814, opened 2021-03-23
Use the old topology key for e2e tests

ayberk:
/bug

e2e tests assumes only one topology key, so checks are hardcoded with the first element of each slice. Due to the recent addition of the well-known label as topology key, these checks might fail based on the ordering of the slice. This PR fixes it.

**What testing is done?** 
Tested on my account and verified it works and does what I expect it to do.

### DIFF
--- a/tests/e2e/testsuites/testsuites.go
+++ b/tests/e2e/testsuites/testsuites.go
@@ -17,9 +17,10 @@ package testsuites
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws"
 	"math/rand"
 	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
 
 	"github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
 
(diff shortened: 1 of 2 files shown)


wongma7:
Can we loop through and check both are present? That is the intention of our code, to provision PVs with both so that they are backwards compatible, we should test it.

wongma7:
I don't mind merging to unblock CI but ideally we should test our driver-specific behavior in our driver-specific test suite otherwise what's the point of it! This kind of thing won't be covered by the external test suite.

wongma7:
/lgtm
/approve

to unblock CI

tests/e2e/testsuites/testsuites.go
```diff
@@ -365,8 +366,25 @@ func (t *TestPersistentVolumeClaim) ValidateProvisionedPersistentVolume() {
...
+			// NodeSelectorTerms: [{[{topology.ebs.csi.aws.com/zone In [us-west-2a]} {topology.kubernetes.io/zone In [us-west-2a]}] []}]
+			// AllowedTopologies: [{[{topology.ebs.csi.aws.com/zone [us-west-2a us-west-2b us-west-2c]}]}]
+			// As you can see tests might fail depending on the ordering of the NodeSelectorTerms. That's why we're doing this "hack".
+			// This is a quick fix to unblock the PRs we have. We really need to improve this. TODO
+
+			keyFound := false
+			for _, v := range t.persistentVolume.Spec.NodeAffinity.Required.NodeSelectorTerms[0].MatchExpressions {
+				if v.Key == "topology.ebs.csi.aws.com/zone" {
+					keyFound = true
+					Expect(v.Key).To(Equal(t.storageClass.AllowedTopologies[0].MatchLabelExpressions[0].Key))
```
wongma7, line 379: This statement is pointless, it's a tautology, if you're in this if statement then this is already guaranteed to be true

wongma7: never mind, I confused storageClass allowedTopoloigies with the PV spec field? I'm a bit lost TBH,whether my SC has the old key or new key, the PV should have both old or new key right?

ayberk: Yeah PV has the both keys, but SC has one (in e2e tests it's always the oldest one, see above tests/e2e/driver/ebs_csi_driver.go). So we have to make sure we're comparing the correct keys. 

For example, this test fails when `t.persistentVolume.Spec.NodeAffinity.Required.NodeSelectorTerms[0].MatchExpressions[0].Key` is not `topology.ebs.csi.aws.com/zone`, which is possible because we also have `topology.kubernetes.io/zone` in the slice.

PR #815, opened 2021-03-23
Track driver deploy time in e2e test pipeline

### LINKED ISSUE #804: [Performance Test] Track driver start time pre-submit
We need to track the driver start time during pre-submit and fail the test if it takes too long to start the driver in the e2e test

**Additional context**
It would be perfect if we can log the start time during e2e test somewhere, so we have a dynamic threshold for each build. If that is not possible, probably start with a hardcoded value.

AndyXiangLi:
Fixes #804  
Add driver start time info during e2e test. so we have better understanding on driver's behavior. 
Set threshold as 20s for now, may adjust as we have more info.

### DIFF
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -25,6 +25,7 @@ source "${BASE_DIR}"/util.sh
 
 DRIVER_NAME=${DRIVER_NAME:-aws-ebs-csi-driver}
 CONTAINER_NAME=${CONTAINER_NAME:-ebs-plugin}
+DRIVER_START_TIME_THRESHOLD_SECONDS=60
 
 TEST_ID=${TEST_ID:-$RANDOM}
 CLUSTER_NAME=test-cluster-${TEST_ID}.k8s.local
@@ -104,11 +105,13 @@ if [[ $? -ne 0 ]]; then
 fi
 
 loudecho "Deploying driver"
+startSec=$(date +'%s')
 "${HELM_BIN}" upgrade --install "${DRIVER_NAME}" \
   --namespace kube-system \
   --set image.repository="${IMAGE_NAME}" \
   --set image.tag="${IMAGE_TAG}" \
   -f "${HELM_VALUES_FILE}" \
+  --wait \
   ./charts/"${DRIVER_NAME}"
 
 if [[ -r "${EBS_SNAPSHOT_CRD}" ]]; then
@@ -116,6 +119,15 @@ if [[ -r "${EBS_SNAPSHOT_CRD}" ]]; then
   kubectl apply -f "$EBS_SNAPSHOT_CRD"
   # TODO deploy snapshot controller too instead of including in helm chart
 fi
+endSec=$(date +'%s')
+secondUsed=$(( (endSec-startSec)/1 ))
+# Set timeout threshold as 20 seconds for now, usually it takes less than 10s to startup
+if [ $secondUsed -gt $DRIVER_START_TIME_THRESHOLD_SECONDS ]; then
+  loudecho "Driver start timeout, took $secondUsed but the threshold is $DRIVER_START_TIME_THRESHOLD_SECONDS. Fail the test."
+  exit 1
+fi
+loudecho "Driver deployment complete, time used: $secondUsed seconds"
+
 
 loudecho "Testing focus ${GINKGO_FOCUS}"
 eval "EXPANDED_TEST_EXTRA_FLAGS=$TEST_EXTRA_FLAGS"


hack/e2e/run.sh
```diff
@@ -116,6 +118,15 @@ if [[ -r "${EBS_SNAPSHOT_CRD}" ]]; then
   kubectl apply -f "$EBS_SNAPSHOT_CRD"
   # TODO deploy snapshot controller too instead of including in helm chart
 fi
+endSec=$(date +'%s')
```
wongma7, line 121: have to wait for the pods to become ready. not so easy in bash to be honest.

AndyXiangLi: Yeah agree, I did some research but no luck to find any tool to track container start up time.
But one thing is if we use helm --wait flag, helm will wait for containersReady condition before exit. IMO It would be good enough as for now.

AndyXiangLi: And open for any suggestion to track this info :-)

wongma7: ooh yeah let's use helm built-in functionality --wait flag, that is cool if they have it. 

Otherwise I was thinking of how to call a python/go script from here and use the kube python/go  client, which would be really ugly...

wongma7: Maybe we can use cloudwatch? The tests run in an AWS-internal account. And if running locally, i don't mind pushing metrics to my own account. We can fail-open in case for whatever reason the metric push fails (transient cloudwatch issue or whatever)

ayberk: That's a good find. Does it wait for everything deployed by the chart to be ready?

hack/e2e/run.sh
```diff
@@ -25,6 +25,7 @@ source "${BASE_DIR}"/util.sh
 
 DRIVER_NAME=${DRIVER_NAME:-aws-ebs-csi-driver}
 CONTAINER_NAME=${CONTAINER_NAME:-ebs-plugin}
+DRIVER_START_TIME_THRESHOLD=25
```
ayberk, line 28: How did we come up with this number? I feel like it should be higher? We can adjust later as we go.

hack/e2e/run.sh
```diff
@@ -104,18 +105,29 @@ if [[ $? -ne 0 ]]; then
...
 if [[ -r "${EBS_SNAPSHOT_CRD}" ]]; then
   loudecho "Deploying snapshot CRD"
   kubectl apply -f "$EBS_SNAPSHOT_CRD"
   # TODO deploy snapshot controller too instead of including in helm chart
 fi
+endSec=$(date +'%s')
+secondUsed=$(( (endSec-startSec)/1 ))
+# Set timeout threshold as 20 seconds for now, usually it takes less than 10s to startup
+if [ $secondUsed -gt $DRIVER_START_TIME_THRESHOLD ]; then
+  loudecho "Driver start timeout, test fail!"
```
ayberk, line 126: You should log here how long it took and what the threshold is, so we can see the gap immediately without reading the code.

ayberk (changes requested):

hack/e2e/run.sh
wongma7: Ideally there is a way to expose the metric publicly though, cloudwatch wont be visible like https://testgrid.k8s.io/provider-aws-efs-csi-driver#e2e-test&width=20 is.

AndyXiangLi: Yes, it will wait for all the containers to be ready

hack/e2e/run.sh
AndyXiangLi: I observed on my cluster that usually takes ~15s so I set this number. But makes sense to increase that a little bit in the initial commit, we can adjust this later.

wongma7:
I think the main bottleneck will be image pulling which we don't really control. I guess if we are trying to measure from ux perspecitve what 'cold start' would look like then it doesn't amtter the details of what is happening under the hood, just the overall #. anyway I'm ok with merging variation of this and seeing how it goes.

hack/e2e/run.sh
ayberk: Yeah looks like it failed on the CI.

AndyXiangLi: I think the delta is the image pull time.. Cold start apparently takes longer than I expected lol

hack/e2e/run.sh
AndyXiangLi: Updated

hack/e2e/run.sh
```diff
@@ -104,18 +105,29 @@ if [[ $? -ne 0 ]]; then
...
 if [[ -r "${EBS_SNAPSHOT_CRD}" ]]; then
   loudecho "Deploying snapshot CRD"
   kubectl apply -f "$EBS_SNAPSHOT_CRD"
   # TODO deploy snapshot controller too instead of including in helm chart
 fi
+endSec=$(date +'%s')
+secondUsed=$(( (endSec-startSec)/1 ))
+# Set timeout threshold as 20 seconds for now, usually it takes less than 10s to startup
+if [ $secondUsed -gt $DRIVER_START_TIME_THRESHOLD ]; then
+  loudecho "Driver start timeout, Cost $secondUsed but the threshold is $DRIVER_START_TIME_THRESHOLD Fail the test."
```
ayberk, line 126: nit: s/Cost/Took

AndyXiangLi: Looks like it took ~30s to start up now, do you think it is ok I change the threshold to 45s? 60s looks a bit too much here

ayberk: I don't see any problem with going to 60s honestly. For now we can use this to make sure we don't introduce a change that'd increase the startup time too much. So let's start with a high number and go from there.

AndyXiangLi: Sounds good! Thank you Ayberk

hack/e2e/run.sh
```diff
@@ -25,6 +25,7 @@ source "${BASE_DIR}"/util.sh
 
 DRIVER_NAME=${DRIVER_NAME:-aws-ebs-csi-driver}
 CONTAINER_NAME=${CONTAINER_NAME:-ebs-plugin}
+DRIVER_START_TIME_THRESHOLD=60
```
ayberk, line 28: one last thing, can you add a comment like `# seconds` here?

AndyXiangLi: Updated, added seconds in the name to make it clear.

hack/e2e/run.sh
```diff
@@ -25,6 +25,7 @@ source "${BASE_DIR}"/util.sh
 
 DRIVER_NAME=${DRIVER_NAME:-aws-ebs-csi-driver}
 CONTAINER_NAME=${CONTAINER_NAME:-ebs-plugin}
+DRIVER_START_TIME_THRESHOLD_SECONDS=60
```
ayberk, line 28: nice :)

PR #820, opened 2021-03-31
release v0.10.0

vdhanan:
release commit

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -6,7 +6,7 @@ replicaCount: 2
 
 image:
   repository: k8s.gcr.io/provider-aws/aws-ebs-csi-driver
-  tag: "v0.9.1"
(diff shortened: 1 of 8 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -70,7 +70,7 @@ resources:
 
 priorityClassName: ""
 nodeSelector: {}
-tolerateAllTaints: true
+tolerateAllTaints: false
```
ayberk, line 73: Why do we have these changes? Release commit should only have the version update.

vdhanan: yeah, in retrospect, this should have been a separate commit. will split it out into two commits next time.

AndyXiangLi:
I will merge this PR first. Like Varun mentioned, next time we should only include release related change.
/approve
/lgtm

PR #822, opened 2021-04-02
Warn users of migrating without draining

partcyborg:
Warn users about migrating nodes without fully draining them of mounted
EBS volumes.

**What testing is done?**


wongma7:
/check-cla

/ok-to-test
/lgtm
/approve

thanks. Definitely need this warning. it seems the best official documentation on migration is still this blog post, but it's a bit dated by now https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/

PR #824, opened 2021-04-07
support volume partition

### LINKED ISSUE #812: Volume partitions are not supported
/kind bug

**What happened?**
For CSI migration we need all in-tree features supported in the CSI driver. In-tree volumes support mounting volume partitions in  `pv.spec.awsElasticBlockStore.partition` flag. Such capability should be present in the CSI driver too.

Enable CSI migration and use following Pod:

```
apiVersion: v1
kind: Pod
metadata:
  name: inline
spec:
  containers:
  - name: inline
    image: centos:7
    args:
    - sleep
    - '99999'
    volumeMounts:
    - mountPath: "/mnt/storage"
      name: inline
  volumes:
  - name: inline
    awsElasticBlockStore:
      fsType: xfs
      volumeID: vol-072b2b73e088c43b5
      partition: 1
```

The CSI translation library makes the partition number available in `volume_attribute: partition`: https://github.com/kubernetes/kubernetes/blob/bbb58fa085ccfc4635aa8544686d960115449d95/staging/src/k8s.io/csi-translation-lib/plugins/aws_ebs.go#L112

AndyXiangLi:
Fixes #812 
Add volume partition support when stage&publish volumes to sync with in-tree driver
**What testing is done?** 
unit test
e2e test

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -121,6 +121,10 @@ func (d *nodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	if !isValidVolumeCapabilities([]*csi.VolumeCapability{volCap}) {
 		return nil, status.Error(codes.InvalidArgument, "Volume capability not supported")
 	}
+	volumeContext := req.GetVolumeContext()
+	if isValidVolumeContext := isValidVolumeContext(volumeContext); !isValidVolumeContext {
+		return nil, status.Error(codes.InvalidArgument, "Volume Attribute is not valid")
+	}
 
 	// If the access type is block, do nothing for stage
 	switch volCap.GetAccessType().(type) {
@@ -158,7 +162,15 @@ func (d *nodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Error(codes.InvalidArgument, "Device path not provided")
 	}
 
-	source, err := d.findDevicePath(devicePath, volumeID)
+	partition := ""
+	if part, ok := volumeContext[VolumeAttributePartition]; ok {
+		if part != "0" {
+			partition = part
+		} else {
+			klog.Warningf("NodeStageVolume: invalid partition config, will ignore. partition = %v", part)
+		}
+	}
+	source, err := d.findDevicePath(devicePath, volumeID, partition)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to find device path %s. %v", devicePath, err)
 	}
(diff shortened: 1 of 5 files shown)


pkg/driver/node.go
```diff
@@ -521,7 +525,13 @@ func (d *nodeService) nodePublishVolumeForBlock(req *csi.NodePublishVolumeReques
 	if !exists {
 		return status.Error(codes.InvalidArgument, "Device path not provided")
 	}
-	source, err := d.findDevicePath(devicePath, volumeID)
+
+	partition := ""
```
wongma7, line 529: partition would not make sense in block context, like block that has been partitioned is not block. we should throw an error imo if we find the attribute

pkg/driver/node_linux.go
```diff
@@ -49,7 +52,15 @@ func (d *nodeService) findDevicePath(devicePath, volumeID string) (string, error
 	// /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol0fab1d5e3f72a5e23
 	nvmeName := "nvme-Amazon_Elastic_Block_Store_" + strings.Replace(volumeID, "-", "", -1)
 
-	return findNvmeVolume(nvmeName)
+	nvmeDevicePath, err := findNvmeVolume(nvmeName)
+	if err != nil {
+		//Return the nvmeDevicePath just for logging purpose
```
wongma7, line 57: the callers never read the returned string though.

pkg/driver/node.go
wongma7: nvm, i think you can still partition a device wthout formatting any of the partitions right ? Not an expert on this obviously

wongma7:
Can we add an e2e test otherwise lgtm

pkg/driver/node.go
wongma7: just checked in-tree and the code is same except in-tree never bothered to set partition of nvme volumes : / with "p#" suffix

AndyXiangLi: in-tree is going to have nvme fix in this PR https://github.com/kubernetes/kubernetes/pull/100500
I'm testing the block scenario and will update here later.

pkg/driver/node_linux.go
AndyXiangLi: u r right... It is logging devicePath, will return empty string then

pkg/driver/node.go
AndyXiangLi: @wongma7 just to confirm we can still partition a device without formatting.

AndyXiangLi:
@wongma7 @jsafrane Update the PR as migration test was failing. Notice that some migration test cases set partition as "0", so explicitly check that scenario and ignore partition 0. Please LMK if you have any idea on this

pkg/driver/node.go
```diff
@@ -521,7 +529,17 @@ func (d *nodeService) nodePublishVolumeForBlock(req *csi.NodePublishVolumeReques
...
+	if part, ok := req.GetVolumeContext()[VolumeAttributePartition]; ok {
+		if part != "0" {
+			partition = part
```
wongma7, line 536: should we validate it? Check that it's an integer? and not something crazy/malicious with weird paths and whatnot.

AndyXiangLi: Yeah, parse to int makes sense to me, is there any special rule for windows partition validation? should that be same?

wongma7: You can leave it same, I'll test it out later.

I think I will make the windows findDevicePath implementation ignore the partition arg. From my understanding, the windows mount code will always mount the first partition so it doesn't matter what partition is set to. https://github.com/kubernetes-csi/csi-proxy/blob/d47b9ecde8f020dd9d5eb2df3e56bf2f10a3f4a3/internal/os/volume/api.go#L87 (similarly, fstype gets ignored in windows world)

wongma7:
/lgtm
thanks

PR #835, opened 2021-04-16
update IAM policy sample and add new driver level tag

### LINKED ISSUE #571: Provide more restrictive example IAM policy, based on tags
When there are multiple EBS volumes in AWS and some of them are used outside of kubernetes cluster, it would be nice to limit somehow access of k8s cluster to only ebs-csi-driver provisioned volumes, to not mock with drives outside of kubernetes world.

/feature

It would be nice to provide example IAM policy, with some tags-based restrictions. I tried to create such policy myself but various things broke. Example (terraform code):
```
  statement {
    actions = [
          "ec2:CreateSnapshot",
          "ec2:CreateTags",
          "ec2:CreateVolume",
          "ec2:DeleteSnapshot",
          "ec2:DeleteTags",
          "ec2:DeleteVolume",
          "ec2:ModifyVolume"
        ]
    resources = ["arn:aws:ec2:*:*:volume/*"]
    # Allow only volumes created by CSI driver - with CSIVolumeName tag
    condition {
      test = "ForAnyValue:StringEquals"
      variable = "aws:TagKeys"
      values = ["CSIVolumeName"]
    }
  }

  statement {
    actions = [
          "ec2:DescribeInstances",
          "ec2:DescribeSnapshots",
          "ec2:DescribeTags",
          "ec2:DescribeVolumes",
          "ec2:DescribeVolumesModifications"
        ]
    resources = ["*"]
  }

  statement {
    actions = [
          "ec2:AttachVolume",
          "ec2:DetachVolume"
        ]
    resources = ["arn:aws:ec2:*:*:volume/*"]

    # Allow attach only volumes with CSIVolumeName tag
    # FIXME: I wasn't able to make this condition working. Probably CSI driver
    # doesn't pass tags in attach/detach API calls
    #
    condition {
      test = "StringLikeIfExists"
      variable = "ec2:ResourceTag/CSIVolumeName"
      values = ["pvc-"]
    }
  }
```

However this tags-based restriction doesn't work and for example delete of volume and attach/detach of the volume fails with not authorized error.

Create policy myself, but I wasn't able to get tag-based restrictions working, for unknown reason.

**Additional context**

AndyXiangLi:
Fixes #571
Update sample IAM policy to scope down the resource restriction. The sample IAM policy is still open enough to cover all driver use cases, dynamic provisioning/static provisioning/snapshot/volume resizing etc. 
It can be scope down further if above use cases are not apply to your cluster at all. (Like only allow driver to attach/detach the volume created by driver if you are not using static provisioning feature)
Add a new tag  `ebs.csi.aws.com/cluster` to identify all the resources created by driver.
**What testing is done?** 
e2e test/ unit test/ manually test

### DIFF
--- a/docs/example-iam-policy.json
+++ b/docs/example-iam-policy.json
@@ -4,23 +4,118 @@
     {
       "Effect": "Allow",
       "Action": [
-        "ec2:AttachVolume",
         "ec2:CreateSnapshot",
-        "ec2:CreateTags",
-        "ec2:CreateVolume",
-        "ec2:DeleteSnapshot",
-        "ec2:DeleteTags",
-        "ec2:DeleteVolume",
+        "ec2:AttachVolume",
+        "ec2:DetachVolume",
+        "ec2:ModifyVolume",
         "ec2:DescribeAvailabilityZones",
         "ec2:DescribeInstances",
         "ec2:DescribeSnapshots",
         "ec2:DescribeTags",
         "ec2:DescribeVolumes",
-        "ec2:DescribeVolumesModifications",
-        "ec2:DetachVolume",
-        "ec2:ModifyVolume"
+        "ec2:DescribeVolumesModifications"
       ],
       "Resource": "*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateTags"
+      ],
+      "Resource": [
+        "arn:aws:ec2:*:*:volume/*",
+        "arn:aws:ec2:*:*:snapshot/*"
+      ],
+      "Condition": {
+        "StringEquals": {
+          "ec2:CreateAction": [
+            "CreateVolume",
+            "CreateSnapshot"
+          ]
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteTags"
+      ],
+      "Resource": [
+        "arn:aws:ec2:*:*:volume/*",
+        "arn:aws:ec2:*:*:snapshot/*"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "aws:RequestTag/ebs.csi.aws.com/cluster": "true"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "aws:RequestTag/CSIVolumeName": "*"
+        }
+      }
+    },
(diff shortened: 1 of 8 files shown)


docs/example-iam-policy.json
```diff
@@ -4,23 +4,118 @@
...
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteVolume"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeName": "*"
```
wongma7, line 80: all existing volumes will have this tag right? meaning this policy is backwards compatible?

AndyXiangLi: Yeah, it will not break the existing flow

wongma7: it does raise the question, why is the new tag necessary then? is it because it's more readable/understandable?

AndyXiangLi: Yeah, we just want to add a more readable tag and make it same for both volume and snapshot.

wongma7: makes sense.

I think we should document somewhere what the tags mean, even if it's in code comments just for maintainers. The current comments are not that useful because they tell us WHAT the tags are, but not what they are FOR.

### CSIVolumeName=foo

added to all volumes, for checking if a given volume was already created so that ControllerPublish/CreateVolume is idempotent.
 
### ebs.csi.aws.com/cluster=true

added to all volumes, for allowing users to use a policy to limit csi driver's permission to just its volumes

### kubernetes.io/cluster/`X`=owned*

added to all volumes if clusterId argument is set to `X`. In in-tree cloudprovider, it's extremely unlikely/impossible for clusterName NOT to be set, so effectively this tag was set on ALL in-tree volumes https://github.com/kubernetes/cloud-provider-aws/blob/5f394ba297bf280ceb3edfc38922630b4bd83f46/pkg/providers/v2/tags.go#L108. In CSI world we have no automatic way to discover the cloud provider's clusterName, so if the user wants their CSI volumes to have these tags like in-tree volumes would, they have to provide clusterId.

### `extrakey`=`extravalue`
if extraTags argument is set

pkg/cloud/cloud.go
```diff
@@ -110,6 +110,8 @@ const (
 	KubernetesTagKeyPrefix = "kubernetes.io"
 	// AWSTagKeyPrefix is the prefix of the key value that is reserved for AWS.
 	AWSTagKeyPrefix = "aws:"
+	//AwsEbsDriverTagKey is the tag to identify if a volume/snapshot is managed by ebs csi driver
```
wongma7, line 113: As i commented above, could you add some more detail about how we use this tag (to enable users to limit the csi driver's permission.

Otherwise, it is difficult to understand why we tag each volume ~4 different ways, as I detailed above

Issue #806, opened 2021-03-19
Emit cloudprovider and throttling metrics


gnufied:
The intree cloudprovider emits cloudprovider API metrics and throttling metrics.

These metrics are important for debugging - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws_metrics.go

wongma7:
cc @ayberk @AndyXiangLi .

We can simply add these metrics.

Or better yet  I would like us to solve https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/393.


there are 3 cloud provider abstractions. our cloud.go, v1/legacy, and v2 and  I want to avoid duplication. https://github.com/kubernetes/cloud-provider-aws/tree/6c10f4b1ba2d0377665cbaa7a2cbe6730591ec64/pkg/providers/v2

i think we should design the v2 interface to replace our cloud.go https://github.com/kubernetes/cloud-provider-aws/tree/6c10f4b1ba2d0377665cbaa7a2cbe6730591ec64/pkg/providers/v2 and then use it as a library.

There is also an argument for using v1 because for migration purposes if we want to ensure 1:1 compatibility and avoid aws related bugs like handlign of eventual consistency then we should use v1.

wongma7:
Also if we use v1 then we emit these metrics for free i think but it's not that hard to add them  so i dont buy it

wongma7:
the throttling in particular is hard to emit unless we copy the retry handler and sdk setup from cloud provider : https://github.com/kubernetes/cloud-provider-aws/blob/5f394ba297bf280ceb3edfc38922630b4bd83f46/pkg/providers/v1/retry_handler.go#L100 .  our custom cloud.go is much simpler.

I am still trying to weigh what is easier, copy/pasting what we need from the old cloud provider, or refactoring to depend on it completely...

wongma7:
[...]

PR #842, opened 2021-04-20
Emit AWS API operation duration/error/throttle metrics

wongma7:
**Is this a bug fix or adding new feature?** fix https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/806

**What is this PR about? / Why do we need it?** This publishes API metrics equivalent to those that the ebs plugin/aws cloud provider embedded in kube-controller-manager publishes today, like describeinstances/describevolume call durations, error counts, and throttle error counts.

I'm not well-versed in how people intend to consume the metrics so this PR is barebones, for now it just exposes them over port 80 using "k8s.io/component-base/metrics/legacyregistry"

DIFFERENCES between my implementation and the cloud provider one:

- I implement a `Complete` handler and refer to request.Operation.Name when emitting metrics. As opposed to wrapping every call and referring to a custom request name.
  - consequence: instead of emitting e.g. cloudprovider_aws_api_request_duration_seconds_bucket{request="describe_instance" , I emit cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances"
- I use SDK's IsErrorThrottle function to decide whether to emit a throttle metric whereas cloudprovider only treats RequestLimitExceeded as throttles.
  - consequence: I don't think it matters for DescribeVolumes/DescribeInstances.
- I did not change our retry logic. Unlike cloudprovider one that has process-wide retry, ours is per-request. So since the quantity of requests will differ so will the metrics, meaning people will have to adjust their alarms?
  - consequence: hard to say for certain without testing under conditions where lots of api calls fail and need to be retried. But generally I feel safer relying on SDK retry logic.

**What testing is done?** 

kubectl port-forward deployment/ebs-csi-controller 8080:80 -n kube-system

excerpt:
```
# HELP cloudprovider_aws_api_request_duration_seconds [ALPHA] Latency of AWS API calls
# TYPE cloudprovider_aws_api_request_duration_seconds histogram
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.005"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.01"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.025"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.05"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.1"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.25"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.5"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="1"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="2.5"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="5"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="10"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="+Inf"} 1
cloudprovider_aws_api_request_duration_seconds_sum{request="DescribeInstances"} 0.204679971
cloudprovider_aws_api_request_duration_seconds_count{request="DescribeInstances"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="0.005"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="0.01"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="0.025"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="0.05"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="0.1"} 3
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="0.25"} 4
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="0.5"} 4
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="1"} 4
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="2.5"} 4
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="5"} 4
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="10"} 4
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeVolumes",le="+Inf"} 4
cloudprovider_aws_api_request_duration_seconds_sum{request="DescribeVolumes"} 0.35412039799999995
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -18,7 +18,11 @@ package main
 
 import (
 	"flag"
+	"net/http"
+
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver"
+	"k8s.io/component-base/metrics/legacyregistry"
 
 	"k8s.io/klog"
 )
@@ -26,6 +30,18 @@ import (
 func main() {
 	fs := flag.NewFlagSet("aws-ebs-csi-driver", flag.ExitOnError)
 	options := GetOptions(fs)
+
+	cloud.RegisterMetrics()
+	if options.ServerOptions.HttpEndpoint != "" {
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", legacyregistry.HandlerWithReset())
+		go func() {
+			err := http.ListenAndServe(options.ServerOptions.HttpEndpoint, mux)
+			if err != nil {
+				klog.Fatalf("failed to listen & serve metrics from %q: %v", options.ServerOptions.HttpEndpoint, err)
+			}
+		}()
+	}
 
 	drv, err := driver.NewDriver(
 		driver.WithEndpoint(options.ServerOptions.Endpoint),
--- a/pkg/cloud/handlers.go
+++ b/pkg/cloud/handlers.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloud
+
+import (
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/request"
+
+	"k8s.io/klog"
+)
+
+// RecordRequestsComplete is added to the Complete chain; called after any request
+func RecordRequestsHandler(r *request.Request) {
+	recordAWSMetric(operationName(r), time.Since(r.Time).Seconds(), r.Error)
+}
+
+// RecordThrottlesAfterRetry is added to the AfterRetry chain; called after any error
+func RecordThrottledRequestsHandler(r *request.Request) {
+	if r.IsErrorThrottle() {
+		recordAWSThrottlesMetric(operationName(r))
+		klog.Warningf("Got RequestLimitExceeded error on AWS request (%s)",
+			describeRequest(r))
+	}
+}
+
+// Return the operation name, for use in log messages and metrics
+func operationName(r *request.Request) string {
+	name := "N/A"
+	if r.Operation != nil {
+		name = r.Operation.Name
+	}
+	return name
+}
+
+// Return a user-friendly string describing the request, for use in log messages
+func describeRequest(r *request.Request) string {
+	service := r.ClientInfo.ServiceName
+	return service + "::" + operationName(r)
+}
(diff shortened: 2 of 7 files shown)


wongma7:
/cc @gnufied 

need your  expertise or testimony from SREs relying on this metric : - )

wongma7:
/cc @AndyXiangLi 

BTW, once we have these metrics, we may need to do some basic comparison / load testing with in-tree driver for API call volume.  In addition to the pod startup type of testing we are doing. I probably can sign up for it, time permitting : D

pkg/cloud/handlers.go
```diff
@@ -0,0 +1,54 @@
...
+	if r.IsErrorThrottle() {
+		recordAWSThrottlesMetric(operationName(r))
+		klog.Warningf("Got RequestLimitExceeded error on AWS request (%s)",
+			describeRequest(r))
+	}
+}
+
+// Return the operation name, for use in log messages and metrics
+func operationName(r *request.Request) string {
+	name := "?"
```
gnufied, line 43: I think convention we use is `N/A` as default name - if we don't know the operation name.

gnufied (commented):
Left some minor comments. It is okay to rename the metrics as long as - we cover it with release notes. I think whenever AWS migration is enabled by default, the release note should capture the renamed metrics.

cmd/main.go
```diff
@@ -27,6 +33,16 @@ func main() {
...
 
+	cloud.RegisterMetrics()
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", legacyregistry.HandlerWithReset())
+	go wait.Until(func() {
+		err := http.ListenAndServe("", mux)
+		if err != nil {
+			klog.Errorf("failed to listen & serve metrics: %v", err)
+		}
+	}, 5*time.Second, wait.NeverStop)
```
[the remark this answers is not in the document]
wongma7: Yeah let me add that, I think if i try to listen on port 80 with hostnetwork true it could conflict with other stuff.

basically the wait.util is to loop forever and retry every 5s in case the server stops for whatever reason (for example, if the port is taken by something else? (i copied this pattern from somewhere else in k8s code base...)

Alternatively we can panic/klog.Fatal and restart the whole process in case the server stops for whatever reason. I think i'll change it to do that, otherwise users will be confused if the pod appears healthy but there r no metrics.

pkg/cloud/handlers.go
wongma7: sounds good I'll change, this part is copied from cloudprovider

gnufied:
also cc @Jiawei0227 and @msau42 who are tracking CSI migration work and metrics migration is a prerequisite for CSI migration.

pkg/cloud/handlers.go
wongma7: fixd

cmd/main.go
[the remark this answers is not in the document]
wongma7: added --http-endpoint (name consistent with sidecars) and klog.Fatal on error.

PR #846, opened 2021-04-22
instance metadata issue fix

### LINKED ISSUE #474: Cannot Run with IAM Service Account and no metadata service
/kind bug

**What happened?**
The ebs-plugin container on the ebs-csi-controller crashes repeatedly while talking to the metadata service:

```
I0325 19:25:18.560010       1 driver.go:62] Driver: ebs.csi.aws.com Version: v0.6.0-dirty
panic: EC2 instance metadata is not available

goroutine 1 [running]:
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.newNodeService(0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/node.go:83 +0x196
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.NewDriver(0xc00016ff70, 0x3, 0x3, 0xc0000a88a0, 0xdcc3a0, 0xc0001edb00)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/driver.go:87 +0x512
main.main()
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/cmd/main.go:31 +0x117
```

When specifying the AWS_REGION variable, and an IAM role service account, the ebs-csi-driver should not need to access the metadata service, and run on its own.

1. Create an IAM role with permissions for the aws-ebs-csi-driver
2. Create an EKS cluster with an OIDC connect provider trust relationship with IAM (https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), especially following the final step of disabling pods from accessing the metadata service to prevent them from assuming the worker instance profile.
3. Deploy aws-ebs-csi-driver using the alpha kustomize overlays, adding the eks.amazonaws.com/role-arn to the service account
4. The ebs-csi-controller pods will start and crash after about 20s

As far as we can tell everything is set up correctly with the role + service account, but the code explicitly tries to instantiate the metadata service, which is firewalled off. Can this be made optional if region is set, and credentials are available via the service account?

**Environment**
EKS v1.14
ebs-csi-driver v0.5.0, v0.6.0-dirty

vdhanan:
fixes #474 

uses downward api and k8s api to populate needed metadata fields

**What testing is done?** 
some unit and manual testing - need to add some more unit tests before merge

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	dm "github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/devicemanager"


charts/aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -65,6 +65,10 @@ spec:
           env:
             - name: CSI_ENDPOINT
               value: unix:/csi/csi.sock
+            - name: CSI_NODE_NAME
+              valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
```
wongma7, line 71: also run `make generate-kustomize` (definitely we need a CI job for this, same as how we have make verify for keeping vendor dir up to date)

pkg/cloud/metadata.go
```diff
@@ -85,13 +91,69 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
...
 	if !svc.Available() {
-		return nil, fmt.Errorf("EC2 instance metadata is not available")
+		klog.Warningf("EC2 instance metadata is not available")
+		nodeName := os.Getenv("CSI_NODE_NAME")
+		if nodeName == "" {
+			return nil, fmt.Errorf("instance metadata is unavailable and CSI_NODE_NAME env var not set")
+		}
+
+		// get node with k8s API
+		node, err := clientset.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
wongma7, line 117: doesn't the node need RBAC permission to do this? we will need to add clusterrole and clusterrolebinding bound to the node service account.

pkg/cloud/metadata.go
```diff
@@ -85,13 +91,69 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
...
+			return nil, err
+		}
+
+		providerID := node.Spec.ProviderID
+		if providerID == "" {
+			return nil, fmt.Errorf("node providerID empty, cannot parse")
+		}
+
+		awsRegionRegex := "(?<=/)([a-z]{2}(-gov)?)-(central|(north|south)?(east|west)?)-[0-9]"
+		awsAvailabilityZoneRegex := "(?<=/)([a-z]{2}(-gov)?)-(central|(north|south)?(east|west)?)-[0-9][a-z]"
```
wongma7, line 128: how much do we have to worry about special cases, will this work on an outpost node for example? see  https://github.com/kubernetes/kubernetes/pull/99130/files#diff-f20dfc917eee752715e89727d403f28f27220826b73bc67ddcc224effe7d1836R270 by @ayberk.

(does sdk provide something so we dont have to write regex?)

pkg/cloud/cloud.go
```diff
@@ -224,7 +224,7 @@ type EC2 interface {
 	DescribeAvailabilityZonesWithContext(ctx aws.Context, input *ec2.DescribeAvailabilityZonesInput, opts ...request.Option) (*ec2.DescribeAvailabilityZonesOutput, error)
 }
 
-type Cloud interface {
+type AWSCloud interface {
```
wongma7, line 227: fair enough. note that i do eventually want to reconcile our cloud implementation with the generic aws cloudprovider implementation https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v2/cloud.go#L63  so "Cloud" is not overly ambiguous, the current naming is probably fine.  https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/806#issuecomment-823535745

wongma7:
/hold

i can see unit tests in progress

charts/aws-ebs-csi-driver/templates/node.yaml
vdhanan: done in new PR #855

pkg/cloud/metadata.go
vdhanan: done in #855

pkg/cloud/cloud.go
vdhanan: my IDE was complaining about Cloud being declared here and in a 3P dependency, but that went away so I don't rename it in PR #855

PR #849, opened 2021-04-23
Fix missing import

wongma7:

**What is this PR about? / Why do we need it?** not sure how https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/842 passed tests at all but an import is missing.

**What testing is done?**


AndyXiangLi:
/lgtm
Thanks!

PR #851, opened 2021-04-23
Add additional logging for outpost arn handling

ayberk:
We recently encountered a case where driver wouldn't use the outpost arn despite metadata being available. I'm adding more logging around it to help troubleshooting the issue in the future.

### DIFF
--- a/pkg/cloud/metadata.go
+++ b/pkg/cloud/metadata.go
@@ -104,7 +104,9 @@ func NewMetadataService(svc EC2Metadata) (MetadataService, error) {
 	// it's guaranteed to be in the form `arn:<partition>:outposts:<region>:<account>:outpost/<outpost-id>`
 	// There's a case to be made here to ignore the error so a failure here wouldn't affect non-outpost calls.
 	if err != nil && !strings.Contains(err.Error(), "404") {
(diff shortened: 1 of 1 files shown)


pkg/cloud/metadata.go
```diff
@@ -104,7 +104,9 @@ func NewMetadataService(svc EC2Metadata) (MetadataService, error) {
 	// it's guaranteed to be in the form `arn:<partition>:outposts:<region>:<account>:outpost/<outpost-id>`
 	// There's a case to be made here to ignore the error so a failure here wouldn't affect non-outpost calls.
 	if err != nil && !strings.Contains(err.Error(), "404") {
-		return nil, fmt.Errorf("something went wrong while getting EC2 outpost arn")
+		return nil, fmt.Errorf("something went wrong while getting EC2 outpost arn: %s", err.Error())
+	} else if err == nil {
+		klog.Infof("Running in an outpost environment with arn: %s", outpostArn)
```
AndyXiangLi, line 109: nit: can you set the log level to 5 for these two log statements. 
Want to make every log explicitly set with a level, so later we can filter the log by their level.

PR #852, opened 2021-04-26
Add eksctl support to e2e scripts

wongma7:
**Is this a bug fix or adding new feature?** testing

**What is this PR about? / Why do we need it?** Would like to test the driver on eks in addition to kops. eksctl command is similar enough to kops that I think the easiest way is to add a short eksctl script (eksctl.sh) like kops.sh

implementation detail differences:
- eksctl doesn't have an equivalent to 'kops export kubecfg' to export kubeconfig and (re) set the current-context to the test cluster. I refactored the scripts such that they export kubeconfig to a file in the test dir instead of .kube/config so that it's easier to keep track of them and to avoid polluting .kube/config
- I haven't figured out yet how to patch eksctl clujster config to add node iam policies, I think first I will add an eks prow job and then work on adding it in a future PR.
- K8S_VERSION must be x.y for eksctl not x.y.z as for kops.
- eksctl doesn't have an equivalent to `kops update --yes`. for kops, first you update the state file in s3 then run update --yes to actually update the cluster resources, eksctl it happens at once, there is no state file?

**What testing is done?** 
tested locally:
```
$ TEST_ID=88888 \
                                              CLEAN=true \
                                              CLUSTER_TYPE=eksctl \
                                              GINKGO_NODES=1 \
                                              AWS_REGION=us-west-2 \
                                              AWS_AVAILABILITY_ZONES=us-west-2a,us-west-2b,us-west-2c \
                                              TEST_PATH=./tests/e2e-kubernetes/... \
                                              GINKGO_FOCUS="asdf" \
                                              K8S_VERSION="1.19" \
                                              ./hack/e2e/run.sh
...
2021-04-26 12:52:35 [ℹ]  will delete stack "eksctl-test-cluster-88888-k8s-local-cluster"
2021-04-26 12:52:35 [✔]  all cluster resources were deleted
###
## OVERALL_TEST_PASSED: 0
#
###
## SUCCESS!
#
```

Will let CI test that i haven't broken the existing kops job. Will add an eks prow job later and then work on fixing it in future PR.

### DIFF
--- a/hack/e2e/eksctl.sh
+++ b/hack/e2e/eksctl.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -uo pipefail
+set -euo pipefail
 
 function eksctl_install() {
   INSTALL_PATH=${1}
--- a/hack/e2e/eksctl.sh
+++ b/hack/e2e/eksctl.sh
@@ -25,13 +25,10 @@ function eksctl_create_cluster() {
 
   CLUSTER_NAME="${CLUSTER_NAME//./-}"
 
-  set +e
-  if ${BIN} get cluster "${CLUSTER_NAME}"; then
-    set -e
+  if eksctl_cluster_exists "${CLUSTER_NAME}" "${BIN}"; then
     loudecho "Upgrading cluster $CLUSTER_NAME with $CLUSTER_FILE"
     ${BIN} upgrade cluster -f "${CLUSTER_FILE}"
   else
-    set -e
     loudecho "Creating cluster $CLUSTER_NAME with $CLUSTER_FILE (dry run)"
     ${BIN} create cluster \
       --managed \
@@ -55,6 +52,19 @@ function eksctl_create_cluster() {
   loudecho "Getting cluster ${CLUSTER_NAME}"
   ${BIN} get cluster "${CLUSTER_NAME}"
   return $?
+}
+
+function eksctl_cluster_exists() {
+  CLUSTER_NAME=${1}
+  BIN=${2}
+  set +e
+  if ${BIN} get cluster "${CLUSTER_NAME}"; then
+    set -e
+    return 0
+  else
+    set -e
+    return 1
+  fi
 }
 
 function eksctl_delete_cluster() {
--- a/hack/e2e/eksctl.sh
+++ b/hack/e2e/eksctl.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+set -uo pipefail
+
+function eksctl_install() {
+  INSTALL_PATH=${1}
+  if [[ ! -e ${INSTALL_PATH}/eksctl ]]; then
+    EKSCTL_DOWNLOAD_URL="https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_$(uname -s)_amd64.tar.gz"
+    curl --silent --location "${EKSCTL_DOWNLOAD_URL}" | tar xz -C "${INSTALL_PATH}"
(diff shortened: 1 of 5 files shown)


hack/e2e/kops.sh
```diff
@@ -57,26 +52,27 @@ function kops_create_cluster() {
     kops_patch_cluster_file "$CLUSTER_FILE" "$KOPS_PATCH_FILE"
 
     loudecho "Creating cluster $CLUSTER_NAME with $CLUSTER_FILE"
-    ${KOPS_BIN} create --state "${KOPS_STATE_FILE}" -f "${CLUSTER_FILE}"
+    ${BIN} create --state "${KOPS_STATE_FILE}" -f "${CLUSTER_FILE}"
+    kops create secret --state "${KOPS_STATE_FILE}" --name "${CLUSTER_NAME}" sshpublickey admin -i "${SSH_KEY_PATH}".pub
```
wongma7, line 56: this is unrelated change, cuz I found out while testing "kops update cluster --state "${KOPS_STATE_FILE}" "${CLUSTER_NAME}" \
    --ssh-public-key="${SSH_KEY_PATH}".pub --yes" is deprecated:

--ssh-public-key string         SSH public key to use (deprecated: use kops create secret instead)

hack/e2e/eksctl.sh
```diff
@@ -0,0 +1,65 @@
...
+  K8S_VERSION=${6}
+  CLUSTER_FILE=${7}
+  KUBECONFIG=${8}
+
+  generate_ssh_key "${SSH_KEY_PATH}"
+
+  CLUSTER_NAME="${CLUSTER_NAME//./-}"
+
+  set +e
+  if ${BIN} get cluster "${CLUSTER_NAME}"; then
```
nckturner, line 29: Refactor into function and `set +e` inside the function?

hack/e2e/eksctl.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+set -uo pipefail
```
nckturner, line 3: `set -e` at top since you are explicitly disabling it down below?

nckturner (commented):
Nice, some small comments.

hack/e2e/run.sh
```diff
@@ -111,23 +141,24 @@ startSec=$(date +'%s')
...
 if [[ -r "${EBS_SNAPSHOT_CRD}" ]]; then
   loudecho "Deploying snapshot CRD"
-  kubectl apply -f "$EBS_SNAPSHOT_CRD"
+  kubectl apply -f "$EBS_SNAPSHOT_CRD" \
+    --kubeconfig "${KUBECONFIG}"
   # TODO deploy snapshot controller too instead of including in helm chart
 fi
 endSec=$(date +'%s')
-secondUsed=$(( (endSec-startSec)/1 ))
+secondUsed=$(((endSec - startSec) / 1))
```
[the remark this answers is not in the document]
wongma7: no clue, my linter just added spaces. I think it's some bash typing thing

hack/e2e/eksctl.sh
wongma7: sure

hack/e2e/eksctl.sh
wongma7: yeah will do I think using set -e  by default and then disabling it as necessary is better than the reverse.

wongma7:
coveralls bugged (did not touc hgo code), manual merge

PR #855, opened 2021-04-27
instance metadata issue fix

### LINKED ISSUE #474: Cannot Run with IAM Service Account and no metadata service

vdhanan:
bug fix (fixes #474)

alternative way to fetch instance metadata

**What testing is done?** 
manual and unit testing

messed up my original PR #846 this is a replacement for that one


pkg/cloud/metadata.go
```diff
@@ -69,13 +75,69 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
...
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	return NewMetadataService(svc, clientset)
```
wongma7, line 88: IMO we should wrap this error in something helpful. like: error getting information from metadata service or node object : err. Looking back at the issue where people reported that the driver couldnt run because of metadata being unavailable, some context would have helped, because otherwise the driver just crashes with weird error.

pkg/cloud/metadata.go
```diff
@@ -69,13 +75,69 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
 func NewMetadata() (MetadataService, error) {
 	sess := session.Must(session.NewSession(&aws.Config{}))
 	svc := ec2metadata.New(sess)
-	return NewMetadataService(svc)
+	// creates the in-cluster config
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, err
```
wongma7, line 81: I think we shouldonly do this if metadata is unavailable.

pkg/cloud/metadata.go
```diff
@@ -69,13 +75,69 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
...
+			return nil, err
+		}
+
+		providerID := node.Spec.ProviderID
+		if providerID == "" {
+			return nil, fmt.Errorf("node providerID empty, cannot parse")
+		}
+
+		awsRegionRegex := "([a-z]{2}(-gov)?)-(central|(north|south)?(east|west)?)-[0-9]"
+		awsAvailabilityZoneRegex := "([a-z]{2}(-gov)?)-(central|(north|south)?(east|west)?)-[0-9][a-z]"
```
wongma7, line 112: @ayberk will this owrk in outpost world?

pkg/cloud/metadata_test.go
```diff
@@ -207,6 +368,31 @@ func TestNewMetadataService(t *testing.T) {
 				if m.GetOutpostArn() != tc.outpostArn {
 					t.Fatalf("GetOutpostArn() failed: expected %v, got %v", tc.outpostArn, m.GetOutpostArn())
 				}
+			} else if !tc.isAvailable {
+				if tc.name == "success: metadata not available, used k8s api" {
+					if err != nil {
+						t.Fatalf("NewMetadataService() failed: expected no error, got %v", err)
+					}
+					if m.GetInstanceID() != "i-abcdefgh123456789" {
```
wongma7, line 376: these could be consts to facilitate hypothetical future refactoring. doesn't mucvh matter though, just nitpicking.

wongma7 (approved):

pkg/cloud/metadata.go
vdhanan: you're right. i'll change lines 80 and 85 to
`if err != nil && !svc.Available() {`

pkg/cloud/metadata.go
vdhanan: can you confirm that this is the idiomatic way to wrap errors?
`func NewMetadata() (MetadataService, error) {
	sess := session.Must(session.NewSession(&aws.Config{}))
	svc := ec2metadata.New(sess)
	// creates the in-cluster config
	config, err := rest.InClusterConfig()
	if err != nil && !svc.Available() {
		return nil, err
	}
	// creates the clientset
	clientset, err := kubernetes.NewForConfig(config)
	if err != nil && !svc.Available() {
		return nil, err
	}
	metadataService, err := NewMetadataService(svc, clientset)
	if err != nil {
		err = fmt.Errorf("error getting information from metadata service or node object: %w", err)
	}
	return metadataService, err
}`

pkg/cloud/metadata.go
vdhanan: i didn't really consider local zones, wavelength, or outposts when i wrote the regular expression. how much does that matter for GA?

wongma7: of those, i think only outpost matters in the sense that we explicitly said we support it. But, on second thought, even if the regex is edited to be more exhaustive to include the outpost case, it will not help them, because outpost users are still reliant on instance metadata being available to retrieve outpost-id..

wongma7: so for now let's just keep it as is I guess

vdhanan: cool cool

wongma7: to be specific, we have no way to find out if we are on an outpost and to retrieve outpostArn in case metadata is disabled.

wongma7:
/lgtm

feel free to squash, i think im going to have to manually merge anyway if coveralls keeps complaining

charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.10.1"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.10.3
+version: 0.10.4
```
wongma7, line 5: we can hold off on a helm release. it won't hurt, but it will also be pointless. Because the chart will include the new RBAC objects, but 0.10.1 of the driver doesnt need them yet.

wongma7:
I'm going to merge, because coveralls is still complaining, and the lates tcommit was just changing helm chart and there's no need to retest.

pkg/cloud/metadata.go
ayberk: What does the `providerId` look like?

There's a case to be made here to allow customers to set outpost-arn themselves so they can use it without the metadata service. I don't think it's a big deal for now but something to consider.

vdhanan: `providerId` looks like `aws:///us-west-2b/i-abcdefgh123456789`

PR #823, opened 2021-04-07
Implement Windows NodePublish/Unpublish

wongma7:
**Is this a bug fix or adding new feature?** /feature

**What is this PR about? / Why do we need it?** In-tree EBS driver supports windows, so too must CSI in case an in-tree EBS volume gets migrated to CSI

I will add build/push job to CI in a separate PR. Since windows images are big, don't want to prematurely upload stuff to registry.

NOTE: this requires a not-yet-released  version of csi-proxy with changes from
- https://github.com/kubernetes-csi/csi-proxy/pull/116
- https://github.com/kubernetes-csi/csi-proxy/pull/126
- https://github.com/kubernetes-csi/csi-proxy/pull/127
I will detail this in README so that my test results are reproducible.

TODO:

- [x] basic/minimal e2e test with migration off
- [x] doc/example

**What testing is done?** 
Windows mount/unmount works, data persists.

I will let CI test that my refactoring has not broken Linux.

```
# build and push image
 docker buildx build --platform windows -f Dockerfile.windows .
```

Manual test:
```
# Create EKS cluster with Windows nodes.

# Checkout this PR.

# Install the driver including the windows daemonset
helm upgrade --install aws-ebs-csi-driver --namespace kube-system ./charts/aws-ebs-csi-driver

# Replace the windows daemonset image with the windows image. It's a TODO to make the image multiarch... # public.ecr.aws/b5w6x5z2/aws-ebs-csi-driver:windows
k edit daemonset ebs-csi-node-windows -n kube-system

kubectl apply -f examples/kubernetes/windows/specs/

k exec -it windows-server-iis-7c5fc8f6c5-t5mk9 -- powershell

PS C:\> New-Item -Path data -Name "testfile1.txt" -ItemType "file" -Value "This 
is a text string."


    Directory: C:\data


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         4/7/2021  12:31 AM             22 testfile1.txt

k delete po windows-server-iis-7c5fc8f6c5-t5mk9

k exec -it windows-server-iis-7c5fc8f6c5-j44qv -- powershell

PS C:\> ls data 


    Directory: C:\data 


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         4/7/2021  12:31 AM             22 testfile1.txt

```

Basic/minimal e2e test:
```
[...]

PR #856, opened 2021-04-28
Cleanup helm chart

### LINKED ISSUE #448: Add ability to change kubelet default path
There is no option in the helm chart values file to define the kubelet mountPath

it would be nice if we can define it if some cluster has some kubelet path other then the default one which is /var/lib/kubelet

### LINKED ISSUE #512: PodDisruptionBudget needed for autoscaling
/kind bug

**What happened?**
The autoscaler can't remove nodes when ebs-csi-controller pods running in the kube-system namespace are lacking a PodDisruptionBudget. See this similar issue:
https://github.com/kubernetes/autoscaler/issues/350 

Could a PodDisruptionBudget be added to the Helm chart? Looks like there's a good example in the nginx-ingress chart:
https://github.com/helm/charts/blob/master/stable/nginx-ingress/templates/controller-poddisruptionbudget.yaml

**Environment**
- Kubernetes version (use `kubectl version`): 1.14.9
- Driver version: 0.5.0

### LINKED ISSUE #672: Controller ends up on fargate nodes, fails to start
/kind bug

**What happened?**

If we first run a pod on a Fargate node and let it finish, it takes a while for the node to go away. If an EBS CSI controller pod is scheduled to start during this time it ends up on the Fargate node, and fails to start with the error: `Pod not supported: SchedulerName is not fargate-scheduler`.

Fargate nodes set taints to prevent non-fargate pods from being scheduled on them. By default, the conroller has tolerations which means the taint is ignored.

I would expect the controller to respect the taint set on the fargate nodes, so that the controller pods are scheduled on nodes where they can actually run.

1. Set up EKS with a Fargate profile matching e.g. the namespace `fargate`
2. Run a pod in the `fargate` namespace
3. Run `kubectl get nodes` to observe that a fargate node has been created
3. Stop the pod
4. Run `kubectl get nodes` to confirm that a node is still running
5. Deploy the EBS CSI controller
6. Observe that the pod fails to start due to the error mentioned above

Seems related to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/591, which is about the DaemonSet and tolerations in general. I think this is slightly different, because even if tolerating everything by default might(?) make sense, tolerating fargate compute nodes definitely does not, because it can't possibly work. Thus, at the very least tolerations should be set for everything _except_ Fargate nodes.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/526 allows tolerations to be configured in the Helm chart. While this is one way to fix the issue, it does not really make sense that the default configuration (and only configuration if we deploy using kustomize?) is broken when fargate nodes are running.

**Environment**
- Kubernetes version (use `kubectl version`): v1.17.12-eks-7684af
- Driver version: 7278cef1d7a1a63a43f733bfe237e1222678e7c3

### LINKED ISSUE #722: AWS EBS CSI Driver Helm chart to inject environment variables
We’d like the AWS EBS CSI Driver Helm chart to have an option to inject environment variables to pods, so that we can set AWS_STS_REGIONAL_ENDPOINTS=regional and use the Helm chart as-is without patching it after the deployment to a private VPC. 

Alternatives considered aren't considered workable for deployments at scale as we want to deploy at scale and keep our configurations as simple as possible.

### LINKED ISSUE #746: Migrate to non deprecated apiVersions on Helm Chart
```
storage.k8s.io/v1beta1 CSIDriver is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 CSIDriver
```

^^

### LINKED ISSUE #758: ebs-csi-controller blocks updating EKS managed node group

### LINKED ISSUE #797: Clean up helm chart + kustomize overlays
/kind bug

- snapshot is not alpha anymore
- the snapshot crd is just randomly in cluster dir (but not present in helm chart? should add it to helm first and then generate the kustomize from helm)
- file naming doesn't match the helm template files

### LINKED ISSUE #887: CSIDriver v1 is not available in 1.17
/kind bug

The chart indicates >=1.17 compatibility. And yet in 1.17 in storage v1 CSIDriver is not available.
Either discontinue support for 1.17 and bump the chart to 1.18 minimum or bust down  resource version to v1beta1

The incoming fix will do the latter.

krmichel:
Fixes #448 
Fixes #512 
Fixes #672 
Fixes #722 
Fixes #746 
Fixes #758 
Fixes #797
Fixes #887 

Cleaning up and adding functionality to the helm chart


krmichel:
@ayberk or @wongma7 Please check out this PR and provide an ok to test

wongma7:
/ok-to-test

Iw ill be busy next 2 weeks but try my best to review before then.

Makefile
```diff
@@ -142,21 +142,22 @@ verify-vendor:
 generate-kustomize: bin/helm
 	cd charts/aws-ebs-csi-driver && ../../bin/helm template kustomize . -s templates/clusterrole-attacher.yaml > ../../deploy/kubernetes/base/clusterrole-attacher.yaml
 	cd charts/aws-ebs-csi-driver && ../../bin/helm template kustomize . -s templates/clusterrole-provisioner.yaml > ../../deploy/kubernetes/base/clusterrole-provisioner.yaml
-	cd charts/aws-ebs-csi-driver && ../../bin/helm template kustomize . -s templates/clusterrolebinding-attacher.yaml > ../../deploy/kubernetes/base/clusterrolebinding-attacher.yaml
-	cd charts/aws-ebs-csi-driver && ../../bin/helm template kustomize . -s templates/clusterrolebinding-provisioner.yaml > ../../deploy/kubernetes/base/clusterrolebinding-provisioner.yaml
+	cd charts/aws-ebs-csi-driver && ../../bin/helm template kustomize . -s templates/clusterrole-resizer.yaml -f ../../deploy/kubernetes/values/resizer.yaml > ../../deploy/kubernetes/overlays/alpha/clusterrole-resizer.yaml
```
wongma7, line 145: should we move resizer out of the alpha overlay along with snapshotter? Then we can delete the alpha overlay

charts/aws-ebs-csi-driver/templates/NOTES.txt
```diff
@@ -1,3 +1,43 @@
 To verify that aws-ebs-csi-driver has started, run:
 
     kubectl get pod -n {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "aws-ebs-csi-driver.name" . }},app.kubernetes.io/instance={{ .Release.Name }}"
+
+
+WARNING: The following values have been deprecated in favor of moving them into the controller or node groups.  They will be removed in a subsequent release.
```
wongma7, line 6: this is great.

charts/aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -160,15 +142,15 @@ spec:
       volumes:
         - name: kubelet-dir
           hostPath:
-            path: /var/lib/kubelet
+            path: {{ .Values.node.kubeletPath }}
```
wongma7, line 145: windows support will be coming soon with a windows-specific daemonset. In which case, I think I'll need to add windows variation of this .Values.node.kubeletPath, unless you see a cleaner solution.

charts/aws-ebs-csi-driver/templates/snapshot-controller.yaml
```diff
@@ -23,40 +23,33 @@ spec:
...
-        {{- else }}
         - key: CriticalAddonsOnly
           operator: Exists
         - operator: Exists
           effect: NoExecute
           tolerationSeconds: 300
-        {{- end }}
         {{- with .Values.tolerations }}
-{{ toYaml . | indent 8 }}
+        {{- toYaml . | nindent 8 }}
```
wongma7, line 40: how can we ensure indenting stays clean as we continue editing these templates? is this just a vanilla yaml linter fixing it up?

charts/aws-ebs-csi-driver/templates/NOTES.txt
```diff
@@ -1,3 +1,43 @@
...
+  enableVolumeResizing:
+  extraCreateMetadata:
+  extraVolumeTags:
+  k8sTagClusterId:
+  nodeSelector:
+  podAnnotations:
+  priorityClassName:
+  region:
+  replicaCount:
+  resources:
```
wongma7, line 38: what about resources vs containerResources, can we deprecate resources in favour of containerResources?

charts/aws-ebs-csi-driver/templates/NOTES.txt
krmichel: Thanks, I though it was a decent way to handle it

charts/aws-ebs-csi-driver/templates/snapshot-controller.yaml
krmichel: I changed this for readability.  Not with a linter or other auto process.  I haven't really found one that formats the helm templates in a very readable way.

wongma7: ok. seems i need to simply do a better job of reviewing then : p

charts/aws-ebs-csi-driver/templates/NOTES.txt
krmichel: We could.  The way it is written currently is that specifying controller.resources will apply to all containers in the deployment, but they can be overridden by setting the individual container resources.  If we don't like that we can decide to get rid of controller.resources on my next PR that makes the breaking changes.

wongma7: okay, it sounds useful then, like i im agine people might want to set some arbitrary nonzero resource.requests for all containers as a starting point and tune for individual containers from there

charts/aws-ebs-csi-driver/templates/node.yaml
krmichel: We could maybe pull it into a helper template or something.  Let me know when that rolls around and we can look at the implications then.  Some of it depends on how different the daemonsets are from a kube perspective.  I assume someone might want to have both daemonsets running in a single cluster.

Makefile
krmichel: I certainly can.  Is it out of alpha?  Moving them is easy enough as long as functionally it is okay to do so.  I assume that also means removing the boolean that turns it on and off?

charts/aws-ebs-csi-driver/templates/node.yaml
wongma7: Yes I plan to install both daemonsets by default so if someone has no windows nodes then the daemonset will simply have 0 pods.

There's a preview in my PR here, the main issue is kubeletPath has to be in windows format like C:\var\lib\kubelet, so simply copying the linux daemonset kubeletPath value won't work : https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/823/files#diff-de20e61e469f4cce4b8334c55256b2b593e91ae9a78fb4c6f0f52a357d695eefR63 Otherwise, I think it would make sense to copy all the other values since the daemonsets otherwise look similar.

Anyway, wanted to give a heads up, we can discuss it in detail in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/823 later, it's still WIP and I'll end up rebasing on this PR nyway

Makefile
wongma7: Yes, it's out of alpha since 1.16, and we recently stopped supporting <=1.16 clusters so now is a good time to remove. ref: https://kubernetes-csi.github.io/docs/volume-expansion.html. 

So yes, we can deprecate enableVolumeResizing and default it true and eventually remove it. To be consistent we should do the same for enableVolumeScheduling and enableVolumeSnapshot. All these flags represent features that were once alpha but have since matured to Beta/GA, they aren't really needed anymore.

that said, I can handle deprecating enableVolumeScheduling enableVolumeSnapshot and enableVolumeResizing in a follow-up PR, don't want to expand the scope of this PR too much.

krmichel: I will move the resizer and scheduler out of alpha, and with that, the enableVolumeResizing and enableVolumeScheduling will no longer be needed so I will just remove them.  I am leaving enableVolumeSnapshot for now but only as a flag on the but only as a gate for the snapshot controller stateful set since the thought right now is to not change how that works.

charts/aws-ebs-csi-driver/templates/_helpers.tpl
```diff
@@ -59,11 +59,24 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+Handle http proxy env vars
+*/}}
+{{- define "aws-ebs-csi-driver.http-proxy" -}}
+- name: HTTP_PROXY
+  value: {{ .Values.proxy.http_proxy | quote }}
+- name: HTTPS_PROXY
+  value: {{ .Values.proxy.http_proxy | quote }}
+- name: NO_PROXY
+  value: {{ .Values.proxy.no_proxy | quote }}
+{{- end -}}
```
AndyXiangLi, line 82: Don't know why verification pass, but we should have an empty line here.

charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: ebs-csi-controller
+      {{- include "aws-ebs-csi-driver.selectorLabels" . | nindent 6 }}
+  {{- if le $replicas 2 }}
+  maxUnavailable: 1
+  {{- else }}
+  minAvailable: 2
+  {{- end }}
```
AndyXiangLi, line 17: Same here, just format

charts/aws-ebs-csi-driver/templates/poddisruptionbudget-snapshot-controller.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: ebs-snapshot-controller
+  labels:
+    {{- include "aws-ebs-csi-driver.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      app: ebs-snapshot-controller
+      {{- include "aws-ebs-csi-driver.selectorLabels" . | nindent 6 }}
+  maxUnavailable: 1
+{{- end }}
```
AndyXiangLi, line 14: same

Makefile
```diff
@@ -140,25 +140,26 @@ verify-vendor:
 
 .PHONY: generate-kustomize
 generate-kustomize: bin/helm
-	cd charts/aws-ebs-csi-driver && ../../bin/helm template kustomize . -n kube-system -s templates/clusterrole-attacher.yaml > ../../deploy/kubernetes/base/clusterrole-attacher.yaml
```
wongma7, line 143: i added -n kube-system, otherwise it seems like helm template uses whatever namespace your current context is :/. (might be rebase issue getting rid of it.)

charts/aws-ebs-csi-driver/templates/_helpers.tpl
wongma7: verify don't cares about yaml/tpl file, only gofmt :/

krmichel: I will add blank lines when I get a chance.

Makefile
krmichel: In this context, the rolebindings are the only thing that cares about the namespace since they need to set the namespace for the service accounts.  The rest of the resources get the namespace they will be created in from the kubectl command.  Unless I am missing the point.  I don't believe that any of the generated /base/ files have any namespace changes in them.  Let me know if there is something I am missing.

wongma7: yeah, it's only the bindings that i'm worried about. I added the arg to guarantee they stay at kube-system because I merged some change that inadvertently changed it and I had to change it back (obviously, I should not have merged it in first place)

krmichel: Okay, let me know if I missed a rolebinding.  I removed the others to make it clear that we aren't forcing a namespace into any of the other manifests.  If you think there is a compelling reason to put them back on the other resources I can but I didn't want others to be confused.

charts/aws-ebs-csi-driver/templates/_helpers.tpl
krmichel: @AndyXiangLi Blank lines added

Makefile
wongma7: makes sense. the right solution is to just review things better, not to force a namespace in the Makefile rule.

wongma7:
/lgtm

/approve

I think the rebase is just the chart version. let's get merged asap to avoid more rebasing pain ..

krmichel:
> /lgtm
> 
> /approve
> 
> I think the rebase is just the chart version. let's get merged asap to avoid more rebasing pain ..

@wongma7 I have rebased and the tests are running.

krmichel:
@wongma7 The tests are passed just needs a /lgtm

wongma7:
/lgtm

thanks

PR #857, opened 2021-04-28
add pod disruption budget for csi controller

### LINKED ISSUE #512: PodDisruptionBudget needed for autoscaling

vdhanan:
fixes #512 

**What testing is done?** 
manually tested with kustomize and helm

### DIFF
--- a/deploy/kubernetes/base/controller.yaml
+++ b/deploy/kubernetes/base/controller.yaml
@@ -77,7 +77,7 @@ spec:
             periodSeconds: 10
             failureThreshold: 5
         - name: csi-provisioner
-          image: k8s.gcr.io/sig-storage/csi-provisioner:v2.0.2
+          image: k8s.gcr.io/sig-storage/csi-provisioner:v2.1.1
           args:
             - --csi-address=$(ADDRESS)
             - --v=5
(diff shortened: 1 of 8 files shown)


deploy/kubernetes/base/controller.yaml
```diff
@@ -91,7 +91,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-attacher
-          image: k8s.gcr.io/sig-storage/csi-attacher:v3.0.0
+          image: k8s.gcr.io/sig-storage/csi-attacher:v3.1.0
```
AndyXiangLi, line 94: Why we need to update the image version?

vdhanan: i didn't update it manually, so i'm assuming one of the `make` recipes bumped the version. @wongma7 do you know why it was updated?

wongma7: yeah probably make generate-kustomize. I double checked and this change makes the helm chart values.yaml match the kustomize, so it's a sound change

Issue #858, opened 2021-04-28
Run chart releaser from branches


wongma7:
**What happened?** This would enable us to branch helm chart releases (e.g. have a different chart for 0.10.x vs 0.9.x).

And it would reduce confusion in the release process. Currently you release the image to registries by pushing a tag on the release branch but you release a corresponding helm chart by pushing something to the master branch.

See efs fix for details: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/333

**What you expected to happen?**

**How to reproduce it (as minimally and precisely as possible)?**

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

PR #859, opened 2021-04-28
Run chart releaser action from release branches only

wongma7:
**Is this a bug fix or adding new feature?** fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/858

"What happened? This would enable us to branch helm chart releases (e.g. have a different chart for 0.10.x vs 0.9.x).

And it would reduce confusion in the release process. Currently you release the image to registries by pushing a tag on the release branch but you release a corresponding helm chart by pushing something to the master branch.

See efs fix for details: kubernetes-sigs/aws-efs-csi-driver#333"

fyi @ayberk @krmichel

**What testing is done?**

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -36,7 +36,7 @@ You also might need to create a release branch, if it doesn't already exist. For
 
 ## Create the release commit in the release branch
 
-### Generate the CHANGELOG
+### Update `CHANGELOG-0.x.md`
 We need to generate the CHANGELOG for the new release by running `./hack/release`. You need to pass previous release tag to generate the changelog.
 
 ```
@@ -45,14 +45,11 @@ python3 release --github-user=ayberk --github-token=$GITHUB_TOKEN note --since <
 
 This will print the CHANGELOG to stdout. You should create a new section for the new version and copy the output there.
 
-### Update the README
+### Update `docs/README.md`
 Search for any references to the previous version on the README, and update them if necessary.
 
-### Update the build and helm deployment files
-1. Update the VERSION variable in the Makefile
-1. Update Helm values
-   - `aws-ebs-csi-driver/values.yaml`
-   - `aws-ebs-csi-driver/Chart.yaml`
+### Update `Makefile`
+Update the VERSION variable in the Makefile
 
 ### Send a release PR to the release branch
 At this point you should have all changes required for the release commit. Verify the changes via `git diff` and send a new PR with the release commit against the release branch. Note that if it doesn't exist, you'll need someone with write privileges to create it for you.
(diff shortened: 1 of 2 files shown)


wongma7:
BTW, this is a double edged sword. because folks browsing master branch might expect the master branch chart to have been released: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/421

So I want others' input and careful consideration :D

docs/RELEASE.md
```diff
@@ -79,15 +79,14 @@ The new tag should trigger a new Github release. Verify that it has run by going
 - Source code (zip)
 - Source code (tar.gz)
 
+## Verify the helm chart release
```
wongma7, line 82: I realize this means that the helm chart release will exist for x minutes before the image to which it is referring will exist in registries.

This is a problem we had in the past that this change would reintroduce.

I think the benefits outweigh it tho.

wongma7: (still thinking outloud)
Actually let me edit the release process to account for this.

ayberk:
Hmm I can see the confusion around not having the latest chart released. My concern is that our chart hasn't been consistent enough to carry all the changes across release branches. Maybe we should do this after @krmichel's changes? When it's more stable?

wongma7:
> Maybe we should do this after @krmichel's changes? When it's more stable?

I'm okay with that.

/hold

krmichel:
I also wonder if there will be confusion about chart versions since they don't generally match the versions of the actual images they use.  Will the helm chart be released only from the most current release branch or any release branch?  Will there be some method other than human caution and observation that will keep the chart versions straight across branches?  Like what if the chart version was 0.27.0 or 1.1.0 on the branch for the 0.11 release of the image.  Would it go up a minor in the release branch for 0.12?  If so, what would the version of the chart be in the 0.11 release branch if a new feature was added to the chart after 0.12 was released.  0.28.0 or 1.2.0 would already be taken and setting it to either 0.27.1/1.1.1 would not accurately represent the change content, and 0.29.0/1.3.0 would falsely imply that the changes from 0.28.0/1.2.0 would be included in that chart version.  It seems like since the chart is a pseudo-autonomous piece of software if you want to release it from branches it should have its own release branches independent of the image releases.  Maybe chart-release-* as a convention.  That would allow for bug fixes to older versions without the complications I was mentioning earlier.  Thoughts?

wongma7:
@krmichel

> Will the helm chart be released only from the most current release branch or any release branch?

Only the most recent release branch, because my main motivation is to protect the helm chart from changes in master branch, not to maintain multiple parallel branches of the helm chart. This thinking goes for the driver as well, because we have never accepted PRs to any branch but master and the most recent release branch. This might change in the future after GA depending on whether we want to maintain multiple minor version tracks in order to support older kubernetes version, but so far we haven't needed to. For example, with the CSIDriver object going from v1beta1 to v1, you are introducing some logic in the chart itself to check the k8s version, so there's no need to maintain different chart tracks for v1beta1 and v1. Anyway, if we assume that we will only ever make changes to the most current release branch, then it mostly avoids the scenarios you described.

> Like what if the chart version was 0.27.0 or 1.1.0 on the branch for the 0.11 release of the image. Would it go up a minor in the release branch for 0.12? If so, what would the version of the chart be in the 0.11 release branch if a new feature was added to the chart after 0.12 was released.

Ideally we would never backport a feature to an old version of the chart, only bugfixes in rare cases, so that would sidestep this scenario. So far, I have been bumping the chart minor version along with the app version. And sometimes we bump the chart minor version independent of the app version. Strictly speaking, as I have mentioned before, I guess we have been abusing semver & helm versioning conventions, but, I really don't know a better way to signal changes to users. At the least, to cut down on confusion, we'll need to start maintaining a chart changelog (someone is working on this in EFS repo). So users will be able to look at chart version, app version, and the changelog for each.

>  It seems like since the chart is a pseudo-autonomous piece of software if you want to release it from branches it should have its own release branches independent of the image releases. Maybe chart-release-* as a convention. 

It might be the case that branching the chart will be the cleanest solution but IMO probably will add more confusion & maintenance burden. The chart still is tied to image releases by `appVersion` like the obvious example is if the chart has a value for toggling a flag that only exists in `appVersion`> x.y. Then it's reasonable to host the chart in the image's release-x.y branch.

wongma7:
/hold cancel

I want to bring this back, https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/995 updated the release process so that we release the kustomize template in release-* branch only after verifying images exist. Similarly, it would make sense to only release the helm chart in release-* branch only after verifying images exist. I think this model has worked OK in efs repo, it requires some additional work from maintainers to cherry-pick helm changes but it is more stable for users

PR #866, opened 2021-04-30
add self as owner

vdhanan:
adding @vdhanan as an owner/maintainer of this repo

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -8,6 +8,7 @@ approvers:
 - wongma7
 - ayberk
 - AndyXiangLi
+- vdhanan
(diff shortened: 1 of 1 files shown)


AndyXiangLi:
I see the invalid use label, are you using the same username you added in the upstream?

vdhanan:
> I see the invalid use label, are you using the same username you added in the upstream?

yeah

vdhanan:
@AndyXiangLi https://github.com/kubernetes/org/blob/main/config/kubernetes-sigs/org.yaml#L561

PR #878, opened 2021-05-10
Fix kustomize RBAC bindings to have namespace kube-system

wongma7:

**What is this PR about? / Why do we need it?** I broke the RBAC. helm used to hardcode kube-system so kustomize templates inherited kube-system. Then we stopped hardcoding it in favour of RElease.namespace so kustomize templates started inheriting default. Unfortunately this breaks things for users who already had kustomize installed in kube-system so let's revert by adding -n kube-system to the helm template command.

**What testing is done?**


AndyXiangLi:
/lgtm
Thanks! Can you also cherry-pick this commit to release v1.0 branch?

TheRealDwright:
This PR broke our aws-ebs-csi-driver deployment which is configured to run in its own namespace. All of the ClusterRoleBinding's and RoleBinding's were shifted to the kube-system namespace causing the driver pods to fail with retrieving resource lock errors.

Issue #876, opened 2021-05-09
Can't start v1 docker image "panic: runtime error: invalid memory address or nil pointer dereference"


lens0021:
/kind bug

**What happened?**

[`amazon/aws-ebs-csi-driver:v1.0.0` and `amazon/aws-ebs-csi-driver:v1.0.0-amazonlinux`](https://hub.docker.com/r/amazon/aws-ebs-csi-driver)  docker image cannot start.

```
$ docker run --rm amazon/aws-ebs-csi-driver:v1.0.0
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12a42e8]

goroutine 1 [running]:
k8s.io/client-go/kubernetes.NewForConfig(0x0, 0x1b52201, 0x4000523c80, 0x0)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/vendor/k8s.io/client-go/kubernetes/clientset.go:395 +0x28
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud.NewMetadata(0x5, 0x0, 0x0, 0x0)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/metadata.go:84 +0xb4
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.newControllerService(0x400007cfa0, 0x400057fb60, 0x0, 0x4000405540, 0x40004afe08)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/controller.go:81 +0x150
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.NewDriver(0x40004aff30, 0x7, 0x7, 0x400057f980, 0x28e3a00, 0x1b522e0)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/driver.go:96 +0x270
main.main()
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/cmd/main.go:46 +0x190
```

```
$ docker run --rm amazon/aws-ebs-csi-driver:v1.0.0-amazonlinux
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12a42e8]

goroutine 1 [running]:
[...]

PR #897, opened 2021-05-20
Only initialize the in-cluster kube client when metadata service is actually unavailable

chrisayoub:

Bug fix

Follow-up PR to this previous PR that fixed instance metadata retrieval when the EC2 metadata service is unavailable:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/855

In our cluster configuration, we run this such that an in-cluster configuration is not available, as we do not give these pods a service account. However, the pods do have access to the EC2 metadata service, as we run the pods with `hostNetwork: true`. When the in-cluster configuration is unavailable, this cause a crash at startup that cannot be avoided, even when setting the `AWS_REGION` environment variable. 

Resolves https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/876

**What testing is done?** 

Unit tests all still pass, and integration tests already exist for this code path. Tested and working in a real cluster when deployed without a serviceAccount.

### DIFF
--- a/pkg/cloud/metadata.go
+++ b/pkg/cloud/metadata.go
@@ -75,15 +75,18 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
 func NewMetadata() (MetadataService, error) {
 	sess := session.Must(session.NewSession(&aws.Config{}))
 	svc := ec2metadata.New(sess)
-	// creates the in-cluster config
-	config, err := rest.InClusterConfig()
-	if err != nil && !svc.Available() {
-		return nil, err
-	}
-	// creates the clientset
-	clientset, err := kubernetes.NewForConfig(config)
-	if err != nil && !svc.Available() {
-		return nil, err
+	var clientset *kubernetes.Clientset
+	if !svc.Available() {
+		// creates the in-cluster config
+		config, err := rest.InClusterConfig()
+		if err != nil {
+			return nil, err
+		}
+		// creates the clientset
+		clientset, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			return nil, err
+		}
 	}
 	metadataService, err := NewMetadataService(svc, clientset)
 	if err != nil {


wongma7:
will let vdhanan do a review and approve

chrisayoub:
Let me know if you need me to try and fix the coverage check, it doesn't seem super trivial to fix based on how the test cases are currently configured, but let me know if you want that fixed. The only thing causing the decrease in coverage is the single `clientset == nil` check.

wongma7:
don't worry about the coverage. I don't think unit tests around this runtime initialization of metadata/client is worth doing

wongma7:
worth doing in this PR *. It requires too much mocking. We'd have to mock instance metadata client, mock/wrap k8s client initialization, etc.

wongma7:
actually the mocking already exists. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/a8c4961cbb336c3e64d38b7ad3b9aabc00398073/pkg/cloud/metadata_test.go

Sorry for the noise, 
I'l ldefer to @vdhanan : do we need a test that if instance metadata is unavailable , we DON'T initialize a k8s client?

vdhanan:
i think it's worth adding a unit test coverage for this. i would just add a boolean `isClientSetNil` field and an error `clientSetNilErr` field to the testCases struct and use those in a new test case to determine whether to initialize the clientset variable and what error to expect

pkg/cloud/metadata_test.go
```diff
@@ -325,11 +327,22 @@ func TestNewMetadataService(t *testing.T) {
...
+			isAvailable:     false,
+			isClientSetNil:  true,
+			clientSetNilErr: fmt.Errorf("instance metadata is unavailable and kubernetes clientset is nil"),
+		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			clientset := fake.NewSimpleClientset(&tc.node)
+			var clientset *fake.Clientset
```
chrisayoub, line 340: What am I doing wrong here? For some reason I can't get this test case to pass

pkg/cloud/metadata.go
```diff
@@ -75,15 +75,20 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
...
+		if err != nil {
+			return nil, err
+		}
+		// creates the clientset
+		clientset, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		clientset = nil
```
vdhanan, line 91: just curious - do you know why we need to explicitly set clientset to nil here? i thought a pointer's zero-value was nil in Go

pkg/cloud/metadata_test.go
vdhanan: can you paste the error?

chrisayoub:
```
=== RUN   TestNewMetadataService/fail:_kube_client_is_nil
W0520 18:25:16.684313   23724 metadata.go:107] EC2 instance metadata is not available
    metadata_test.go:412: NewMetadataService() returned an unexpected error. Expected instance metadata is unavailable and kubernetes clientset is nil, got instance metadata is unavailable and CSI_NODE_NAME env var not set
```

vdhanan: so even though you've set clientset to nil, it's not hitting the if condition you added to `NewMetadataService`, is that right?

pkg/cloud/metadata_test.go
```diff
@@ -394,7 +407,11 @@ func TestNewMetadataService(t *testing.T) {
 					if err == nil {
 						t.Fatalf("NewMetadataService() failed: expected error but got nothing")
 					}
-					if err.Error() != tc.getNodeErr.Error() {
+					if tc.isClientSetNil {
```
wongma7, line 410: can we add an expectedErr field to the test case struct?

The branching is hard to follow, it will be easier if every test case we simply check:

If expectedErr is non empty, did NewMetadataService() return an error?
If so, does the error match what was expected?
If not, test has failed.

I think I originally reviewed this code already and let it pass so I can help refactor .

pkg/cloud/metadata_test.go
wongma7: what dictates whether mockEC2Metadata.Available returns true or not?

wongma7: I think because the pointer got cast into an interface. the interface is not nil even if the implementation is nil.

pkg/cloud/metadata.go
```diff
@@ -75,15 +75,20 @@ func (m *Metadata) GetOutpostArn() arn.ARN {
...
+	if !svc.Available() {
+		// creates the in-cluster config
+		config, err := rest.InClusterConfig()
+		if err != nil {
+			return nil, err
+		}
+		// creates the clientset
+		clientset, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			return nil, err
```
wongma7, line 88: let's just omit the test. in runtime, we will either pass NewMetadataService a function svc or a function clientset. There is no situation where we pass neither. So I don't want to muddy the code just for the sake of passing a test case. I will refactor in a future PR so that it's easier to understand , and write another test, before we release this change.

wongma7: what I'm thinking is, NewMetadataService should take an interface with one function. and we will have two implementations of that interface, one based on ec2 metadata and one based on kubernetes API. If ec2 metadata is unavailable you pass the kubernetes API interface. simple!

pkg/cloud/metadata.go
wongma7: It's not needed, arguably it's good to be explicit / to test our assumption that it is nil but , as I commented above the issue is because the interface can't be nil

pkg/cloud/metadata.go
chrisayoub: I'm actually not sure how to implement what you mean, I'm not super familiar with Go. So, you might want to make your own PR for that approach. I was originally making this PR because I wanted to highlight that the issue existed, but did not previously see the actual issue reported as a bug.

wongma7: yes @chrisayoub I don't want you to work on more than you signed up for, I can do it in a follow up. sorry for the confusion, I'm thinking out loud.

let's merge your PR without the unit test changes (so just the first commit). 

I can take it from there!
- e2e test it
- refactor
- unit test it

wongma7: do a rebase or whatever to drop every commit but the first, force push it, and I'll merge. The coverage check doesn't matter, I can merge it regardless.

chrisayoub: Okay does this look okay now?

vdhanan: does the fix even work, though? i thought from the unit test attempt we learned that the nil clientset is being cast to a non-nil interface, and therefore bypassing the `if clientset == nil` check altogether.

chrisayoub: Ah good point. Do you have a suggested fix we can do for this then, or should it just wait for the follow-up refactor?

vdhanan: if it works, i'm totally cool with merging without unit tests

chrisayoub: I'll test it out in our cluster, I put a hold on this until I do that

vdhanan: i can't really think of anything besides refactoring the function altogether

vdhanan: sounds good. thanks for your help!

chrisayoub: Done, this should be good to go!

wongma7:
/lgtm
/approve

thank you! sorry about all the back and forth regarding whether we need a test or not, sometimes i abuse github comments like my personal notepad : )

PR #902, opened 2021-05-25
Refactor metadata.go to test that k8s client is initialized iff ec2 metadata is unavailable

wongma7:
**Is this a bug fix or adding new feature?** /kind cleanup

**What is this PR about? / Why do we need it?** found in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/897 that it's hard to test whether the k8s client has been initialized or not. in part because NewMetadataService accepts a kubernetes.Interface and you can't check if it's nil.

Refactored NewMetadataService to accept a func that returns a kubernetes.Interface... it's not pretty but should work.

(Also, I removed NewMetadata. Some of this code is hella confusing since we kept the naming of Metadata even though it can be constructed from k8s api now, not just ec2 instance metadata. I stopped short of renaming it, maybe to InstanceInfo or NodeInfo, because I already got  a bit carried away with what is supposed to be a simple refactor, may do it later.)

**What testing is done?** TODO: e2e test on my eks cluster

- [ ] instance metadata enabled (k8s client init should not be attempted, there should be no need to set Node Name env var)
- [ ] instance metadata disabled (k8s client init should be attempted)

### DIFF
--- a/pkg/cloud/metadata_test.go
+++ b/pkg/cloud/metadata_test.go
@@ -29,20 +29,17 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
 	k8s_testing "k8s.io/client-go/testing"
 )
 
-var (
-	stdInstanceID       = "instance-1"
-	stdInstanceType     = "t2.medium"
-	stdRegion           = "instance-1"
-	stdAvailabilityZone = "az-1"
-)
-
 const (
-	nodeName             = "ip-123-45-67-890.us-west-2.compute.internal"
-	nodeObjectInstanceID = "i-abcdefgh123456789"
+	nodeName            = "ip-123-45-67-890.us-west-2.compute.internal"
+	stdInstanceID       = "i-abcdefgh123456789"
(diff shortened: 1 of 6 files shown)


pkg/cloud/metadata_test.go
```diff
@@ -267,143 +218,123 @@ func TestNewMetadataService(t *testing.T) {
...
+						mockEC2Metadata.EXPECT().GetMetadata(OutpostArnEndpoint).Return(tc.getMetadataValue, tc.getMetadataError)
+					} else {
+						mockEC2Metadata.EXPECT().GetMetadata(OutpostArnEndpoint).Return("", fmt.Errorf("404"))
+					}
 				}
-
-				if m.GetInstanceID() != tc.identityDocument.InstanceID {
-					t.Fatalf("GetInstanceID() failed: expected %v, got %v", tc.identityDocument.InstanceID, m.GetInstanceID())
+				if clientsetInitialized == true {
+					t.Errorf("kubernetes client was unexpectedly initialized when metadata is available!")
```
wongma7, line 305: here is the relevant part where we test that clientset initialization func did not get called.

rest of changes in this test file are mainly renaming and simplifying/collapsing some branching

wongma7:
actually instead of manually testing I want https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/907 to test it.

existing kops job will check box 1
new eksctl job will check box 2.

wongma7:
close in favor of https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/907

Issue #904, opened 2021-05-26
Custom Pod Labels in Helm Chart


eytanhanig:
**Is your feature request related to a problem?/Why is this needed**

The Helm chart should allow users to add additional labels to each of the different pods. This is important for a number use cases, including the ability to filter metrics based on custom internal labels.

**Describe the solution you'd like in detail**
The Helm chart should have a `podLables` section for each of the different pod types. This should be similar to how this is currently done in the eks-charts repo. **I have a PR for this queued up and ready to go.**

**Describe alternatives you've considered**
I considered setting this at the top level and having it apply to all pods, however (1) this is more limiting and (2) there is clear intent to migrate away from this.

**Additional context**
Here's a good example of how it's done in in one of the eks-charts: https://github.com/aws/eks-charts/commit/a8723941c12cef7dd8d3d190278a1be153e2cfed

wongma7:
/close

thanks for fix!!

PR #905, opened 2021-05-26
Support custom pod labels in Helm chart

eytanhanig:
Adds the feature requested in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/904

Many organizations use custom labels for a variety of purposes such as grouping metrics and setting affinity rules for other pods. This feature is also present in most of the stable eks-charts.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 1.1.0
+version: 1.1.1
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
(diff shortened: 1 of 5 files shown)


charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 1.1.0
+version: 1.1.1
```
krmichel, line 5: This is adding a feature to the chart so I think the version should be 1.2.0

krmichel (changes requested):

charts/aws-ebs-csi-driver/Chart.yaml
eytanhanig: Good call. Fix pushed.

krmichel (approved):

vdhanan:
thanks for the PR! can you please run `make generate-kustomize` to update the kustomize templates with this change as well?

krmichel:
> thanks for the PR! can you please run `make generate-kustomize` to update the kustomize templates with this change as well?

The values from the values.yaml in the chart are empty dictionaries so unless they get set by some other means they won't cause any changes in the generated kustomize manifests.

vdhanan:
ah, makes sense.
/ok-to-test

vdhanan (approved):

PR #907, opened 2021-05-27
Add test-e2e-external-eks make rule that tests EKS with pod instance metadata disabled. Remove hostNetwork from DaemonSet

wongma7:
**Is this a bug fix or adding new feature?** test

**What is this PR about? / Why do we need it?** test for the case where instance metadata is disabled for pods. EKS specifically recommends this. In this case, only pods with hostNetwork true will have access to instance metadata.

To create the environment, use eksctl --disable-pod-imds.

The expected behavior is that in lieu of instance metadata
- the controller: should get credentials from IAM for SA. This gets configured via eksctl. 
- the node: should get instance info from k8s API. This gets configured by our helm chart. The node should NOT need hostNetwork true anymore because the sole reason that was there was so it could bypass `disable-pod-imds` to touch instance metadata.

TODO for future:
a  test case where instance metadata is totally disabled on instances, not just for pods. hostNetwork is now false for both controller and node, so whether instance metadata is available on host or not should make no difference , but an extra test case won't hurt.

**What testing is done?**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ test-e2e-external:
 .PHONY: test-e2e-external-eks
 test-e2e-external-eks:
 	CLUSTER_TYPE=eksctl \
-	K8S_VERSION="1.19" \
+	K8S_VERSION="1.20" \
 	HELM_VALUES_FILE="./hack/values_eksctl.yaml" \
 	AWS_REGION=us-west-2 \
 	AWS_AVAILABILITY_ZONES=us-west-2a,us-west-2b \
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,17 @@ test-e2e-external:
 	GINKGO_SKIP="\[Disruptive\]|\[Serial\]" \
 	./hack/e2e/run.sh
 
+.PHONY: test-e2e-external-eks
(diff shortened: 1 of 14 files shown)


wongma7:
I0601 23:59:18.405899       1 metadata.go:101] retrieving instance data from ec2 metadata
W0601 23:59:21.746294       1 metadata.go:104] ec2 metadata is not available
I0601 23:59:21.746314       1 metadata.go:112] retrieving instance data from kubernetes api
I0601 23:59:21.747904       1 metadata.go:117] kubernetes api is available

vdhanan:
do we need to bump the helm chart version?

wongma7:
@vdhanan I think these changes should go in the next new `driver + helm chart` release, I don't want to release a helm chart yet.

My reasoning: if we release a new helm chart now, then the driver *should* work even with hostNetwork off and instance metadata off, but it hasn't been tested, the only thing that has been tested is the particular combination of `driver + helm chart` built in this PR.

wongma7:
rebased, conflict was hack/values.yaml got logLevel: 5 added to it (to easily debug tests)

PR #910, opened 2021-05-27
add custom useragent suffix

vdhanan:
new feature

modifies the aws sdk client's useragent string to differentiate the ebs csi driver's requests

**What testing is done?** 
turned on debug logs and confirmed that user agent had new string appended

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -153,6 +153,9 @@ var (
 	VolumeNotBeingModified = fmt.Errorf("volume is not being modified")
 )
 
+// Set during build time via -ldflags
+var driverVersion string
+
 // Disk represents a EBS volume
 type Disk struct {
 	VolumeID         string
@@ -238,6 +241,9 @@ func newEC2Cloud(region string, awsSdkDebugLog bool) (Cloud, error) {
 	if awsSdkDebugLog {
 		awsConfig.WithLogLevel(aws.LogDebugWithRequestErrors)
 	}
+
+	// Set the env var so that the session appends custom user agent string
+	os.Setenv("AWS_EXECUTION_ENV", "aws-ebs-csi-driver-"+driverVersion)
 
 	svc := ec2.New(session.Must(session.NewSession(awsConfig)))
 	svc.Handlers.AfterRetry.PushFrontNamed(request.NamedHandler{
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ VERSION=v1.0.0
 VERSION_AMAZONLINUX=$(VERSION)-amazonlinux
 GIT_COMMIT?=$(shell git rev-parse HEAD)
 BUILD_DATE?=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
-LDFLAGS?="-X ${PKG}/pkg/driver.driverVersion=${VERSION} -X ${PKG}/pkg/driver.gitCommit=${GIT_COMMIT} -X ${PKG}/pkg/driver.buildDate=${BUILD_DATE} -s -w"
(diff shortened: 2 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -321,25 +321,25 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
+		req.Throughput = aws.Int64(throughput)
 	}
 	snapshotID := diskOptions.SnapshotID
 	if len(snapshotID) > 0 {
-		request.SnapshotId = aws.String(snapshotID)
+		req.SnapshotId = aws.String(snapshotID)
 	}
 
-	response, err := c.ec2.CreateVolumeWithContext(ctx, request)
+	response, err := c.ec2.CreateVolumeWithContext(ctx, req, request.WithAppendUserAgent(userAgent()))
```
wongma7, line 342: this shouldn't be necessary, i'm sure there's a way to create the client such that all requests have teh useragent so that you don't have to add the useragent to every single request

charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -73,6 +73,10 @@ spec:
             - --logtostderr
             - --v=5
           env:
+            - name: CSI_DRIVER_VERSION
+              value: {{ .Chart.AppVersion }}
```
wongma7, line 77: it would be better to refer to the driver.driverVersion IMO https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/Makefile#L21

pkg/cloud/cloud.go
wongma7: you can write your own handler and add it tot he chain or leverage the existing handler and add it to the chain https://github.com/aws/aws-sdk-go/blob/v1.38.49/aws/corehandlers/user_agent.go#L22

wongma7: https://github.com/aws/aws-sdk-go/blob/e2d6cb448883e4f4fcc5246650f89bde349041ec/aws/defaults/defaults.go#L77

AddHostExecEnvUserAgentHander is already one of the default handlers

deploy/kubernetes/base/controller.yaml
```diff
@@ -39,6 +39,8 @@ spec:
             - --logtostderr
             - --v=2
           env:
+            - name: CSI_DRIVER_NAME
```
AndyXiangLi, line 42: Why we want to expose the name as env variable? Feel like it would be more accurate if the useragent is fixed..

charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 1.2.0
+version: 1.3.0
```
AndyXiangLi, line 5: correct me if I'm wrong.. it should be 1.0.3 instead of 1.3.0 as the driver version is still on 1.0.x

wongma7: I think, if we avoid exposing env variable, then we dont need to change the chart at all

deploy/kubernetes/base/controller.yaml
wongma7: yeah I agree, better to just keep it fixed/hardcoded

vdhanan: my thinking was that if the driver's name changed for whatever reason, we'd only have to change it in one place. i don't have a strong preference here, so i'll just hardcode it.

charts/aws-ebs-csi-driver/Chart.yaml
vdhanan: the helm chart version is 1.2.0 on master right now. is that a problem?

krmichel: In general, the helm chart version doesn't have to and likely won't match the version of the application.  That is what appVersion is for to specify the version of the application to use if the user doesn't override it.  If the chart keeps getting new features added that work with 1.0.0. of the driver, it will continue to increase in version independent of the driver version.

wongma7:
/lgtm

thanks

PR #912, opened 2021-05-31
remove WellKnownTopologyKey from PV Topology

Elbehery:
bug fix
fixes https://github.com/kubernetes/kubernetes/issues/102248
**What testing is done?** 
modified testing and verified

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -516,8 +516,7 @@ func (d *nodeService) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
 	zone := d.metadata.GetAvailabilityZone()
 
 	segments := map[string]string{
-		TopologyKey:          zone,
-		WellKnownTopologyKey: zone,
+		TopologyKey: zone,
 	}
 
 	outpostArn := d.metadata.GetOutpostArn()
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -655,7 +655,7 @@ func newCreateVolumeResponse(disk *cloud.Disk) *csi.CreateVolumeResponse {
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


Elbehery:
@jsafrane

jsafrane:
lgtm
/assign @wongma7

pkg/driver/node.go
```diff
@@ -516,8 +516,7 @@ func (d *nodeService) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
 	zone := d.metadata.GetAvailabilityZone()
 
 	segments := map[string]string{
-		TopologyKey:          zone,
-		WellKnownTopologyKey: zone,
+		TopologyKey: zone,
```
wongma7, line 519: this part I don't understand, if the volume gets migrated and its TopologyKey gets translated to WellKnownTopologyKey, how will it get scheduled to any node if all nodes only have TopologyKey? the tests passed so clearly it got scheduled

jsafrane: IMO WellKnownTopologyKey labels are still added to nodes by kubelet. And in future, when there is no cloud provider in kubelet, I don't know who adds them... external cloud provider?

jsafrane: Scratch the previous comment. Scheduler / its PV topology code translates in-tree volumes to CSI:

https://github.com/kubernetes/kubernetes/blob/2bbc0e3c54fdf11b0006efe7f65f6a416ea15d32/pkg/controller/volume/scheduling/scheduler_binder.go#L816

jsafrane:
@wongma7, how do you deal with decreased unit test coverage? -0.008%, that's like one line?

wongma7:
@jsafrane yeah coveralls is too finicky, I tried increasing threshold and will remove it if that doesn't work, I'm always manually merging to get around it anyway.

PR #913, opened 2021-05-31
Disable uuid checks on XFS

jsafrane:
Bugfix

By default, XFS does not allow mounting two devices that have the same UUID of the XFS filesystem. As result, it's not possible to mount a volume + its restored snapshot.

Therefore disable UUID check in XFS using a mount option.

**What testing is done?** 
Tested with https://github.com/kubernetes/kubernetes/pull/102538

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -143,12 +143,7 @@ func (d *nodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		fsType = defaultFsType
 	}
 
-	var mountOptions []string
-	for _, f := range mountVolume.MountFlags {
-		if !hasMountOption(mountOptions, f) {
-			mountOptions = append(mountOptions, f)
-		}
-	}
+	mountOptions := collectMountOptions(fsType, mountVolume.MountFlags)
 
 	if ok := d.inFlight.Insert(volumeID); !ok {
 		return nil, status.Errorf(codes.Aborted, VolumeOperationAlreadyExists, volumeID)
@@ -624,6 +619,8 @@ func (d *nodeService) nodePublishVolumeForFileSystem(req *csi.NodePublishVolumeR
 		fsType = defaultFsType
 	}
 
+	mountOptions = collectMountOptions(fsType, mountOptions)
+
 	klog.V(4).Infof("NodePublishVolume: mounting %s at %s with option %s as fstype %s", source, target, mountOptions, fsType)
 	if err := d.mounter.Mount(source, target, fsType, mountOptions); err != nil {
 		if removeErr := os.Remove(target); removeErr != nil {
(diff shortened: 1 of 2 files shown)


jsafrane:
When tested without the second commit (the actual fix in the CSI driver), I can see the newly added test correctly fails when mounting volume + its snapshot on the same node:

```
[It] should concurrently access the volume and restored snapshot from pods on the same node [LinuxOnly]
  /home/prow/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/multivolume.go:337
...
Jun  1 07:58:08.549: INFO: At 2021-06-01 07:53:04 +0000 UTC - event for pod-f465087a-3b4d-4a4a-923a-acac0e6e75e5: {kubelet ip-172-20-45-241.us-west-2.compute.internal} FailedMount: MountVolume.MountDevice failed for volume "pvc-14820d66-fee3-40d6-996e-14eed53ebb45" : rpc error: code = Internal desc = could not format "/dev/xvdbf" and mount it at "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-14820d66-fee3-40d6-996e-14eed53ebb45/globalmount": mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t xfs -o defaults /dev/xvdbf /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-14820d66-fee3-40d6-996e-14eed53ebb45/globalmount
Output: mount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-14820d66-fee3-40d6-996e-14eed53ebb45/globalmount: wrong fs type, bad option, bad superblock on /dev/xvdbf, missing codepage or helper program, or other error.
```
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/913/pull-aws-ebs-csi-driver-external-test/1399630187288596480

There is no system journal, still I think it failed to mount due to uuid conflict. With `-o nouuid` mount option (the second commit), it should succeed (tests are still running).

sgorshkov85:
@jsafrane  Hello, this ticket is critical for our team.  Do you have any ETA for this PR? Sorry, to disturb you, but I need it to schedule my future work.

jsafrane:
Since the e2e test has been merged to Kubernetes (https://github.com/kubernetes/kubernetes/pull/102538), I reworked this PR to update only the driver and add `nouuid` mount option on XFS.

@wongma7 @AndyXiangLi PTAL.

wongma7:
/lgtm
/approve
thanks.
i created an issue so i remember to opt-in to the test when we bump dependencies to 1.22 https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/929

PR #916, opened 2021-06-02
Skip volume expansion if block node

gnufied:
No expansion on the node should be necessary if device is of type raw block


wongma7:
/lgtm
/approve

looks like nodeexpandvolume will no-op anyway, but this is obv. better. thanks

PR #922, opened 2021-06-04
Update driver capabilities

wongma7:

**What is this PR about? / Why do we need it?** to opt-in to (future) e2e tests from kubernetes/kubernetes, capabilities shoudl be accurate. 

**What testing is done?**


wongma7:
currently the test 'should fail to schedule a pod which has topologies that conflict with AllowedTopologies' is being skipped in migration suite because we only create nodes in one zone so there's only one topology to work with. I think we already have non migration test exercising this situation though

PR #925, opened 2021-06-08
Fix podLabels case in Helm chart

eytanhanig:
Bug fix.

This fixes the bug introduced in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/905 where `podLabels` did not have consistent case throughout the Helm chart.

**What testing is done?** 
`helm template` was successfully evaluated when `podLabels` was set for all three pod types.


wongma7:
/lgtm
/approve
/ok-to-test

thanks

Issue #480, opened 2020-04-05
[DOCS] Elaborate on migration instructions


iainbryson:
Currently, the docs say this:

> To turn on the migration, set CSIMigration and CSIMigrationAWS feature gates to true for kube-controller-manager and kubelet.

It would be great to link to (or include inline) an example of this.  i.e. you start with a (presumably EKS) cluster with `StorageClass` with `provisioner: kubernetes.io/aws-ebs`, and you move to the CSI.  Apart from installing the CSI driver, what do you need to do?  Is just `apply`ing `provisioner: kubernetes.io/aws-ebs` on the storage class sufficient? Will it erase the volume? How exactly does one enable the feature gates (i.e. command line examples)?

/triage support

leakingtapan:
Some the of questions are general CSI migration question that might covered in overall csi migration. Have you checked out this blog post: https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/

leakingtapan:
To answer ur questions:

> what do you need to do? Is just applying provisioner: kubernetes.io/aws-ebs on the storage class sufficient? 

Yes. The storage class with `provisioner: kubernetes.io/aws-ebs` should just work. And CSI migration aims at no frontend facing change. 

> Will it erase the volume? 

It will not. CSI driver will start in charge of the volume lifecycle management. and in-tree driver will stay down.

> How exactly does one enable the feature gates (i.e. command line examples)?

Just like how you enable other alpha feature flags. Depends on your cluster provisioning solution. [Here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/feature-gates.yaml#L7) is one example if you are using kops.

cablespaghetti:
I've just been looking at this as precursor for switching to Bottlerocket (which doesn't have in-tree driver support) for our EKS clusters. We of course can't enable alpha feature flags in our EKS clusters.

The details are a bit thin on the ground but this blog post helped a lot: https://medium.com/faun/amazon-eks-upgrade-journey-from-1-16-to-1-17-cb9e88191165

This explains that the CSIMigrationAWS has not been enabled yet in EKS 1.17 and so we have to wait to migrate existing volumes. This is a shame because it basically rules out CSI and Bottlerocket for us. It sounds like we'll have to wait for a future revision of EKS 1.17 to get this.

[...]

PR #932, opened 2021-06-11
Add KubernetesCluster tag to provisioned volumes when cluster-id set

wongma7:

**What is this PR about? / Why do we need it?** users (namely kops) may have policies ocnfigured for their KCM such that KCM can only attach/detach/create/delete volumes with this tag. If user enables  migration, CSI provisions volume, and then user later disables migration, the tag must bep resent so that KCM can attach/detach/create/delete  it. Details:https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/927#issuecomment-858034205

**What testing is done?** 
TODO: I'll do a manual test to sanity check the tag actually shows up. we don't have an automatic tagging e2e test , could be useful (i.e. a test that sets extraTags, clusterID, and checks tags are present in AWS API)
- [x] manual test

### DIFF
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -86,6 +86,11 @@ const (
 	// in-tree volume plugin. Used only when --cluster-id is set.
 	NameTag = "Name"
 
+	// KubernetesClusterTag is tag applied to provisioned EBS volume for backward compatibility with
+	// in-tree volume plugin. Used only when --cluster-id is set.
+	// See https://github.com/kubernetes/cloud-provider-aws/blob/release-1.20/pkg/providers/v1/tags.go#L38-L41.
+	KubernetesClusterTag = "KubernetesCluster"
+
 	// PVCNameTag is tag applied to provisioned EBS volume for backward compatibility
 	// with in-tree volume plugin. Value of the tag is PVC name. It is applied only when
 	// the external provisioner sidecar is started with --extra-create-metadata=true and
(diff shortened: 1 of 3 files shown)


jsafrane:
lgtm. You mention manual tests in TODO and then I see it checked below. Did it succeed?

wongma7:
yep, set k8s-tag-cluster-id to my name and KubernetesCluster tag was added

nckturner:
Its not clear to me if this affects all kops users, or just a subset.  Can we document this somewhere other than code comments so that users know when they need to set `--k8s-tag-cluster-id` (unless its already clear in existing documentation, I did not look very hard).

nckturner:
Its not clear to me if this affects all kops users, or just a subset. Can we document this somewhere other than code comments so that users know when they need to set --k8s-tag-cluster-id (unless its already clear in existing documentation, I did not look very hard).

wongma7:
@nckturner kops needs to set this argument for every CSI installation , or the user has to set it right after installation. IF they don't set it, then volume tags for CSI driver-provisioned volumes won't match tags for in-tree driver-provisioned volumes and if the user has tag-based IAM policies then things will break.  only kops/the user knows what the cluster name is. We dont have any migration-related documentation at all, when we do this will need to be part of it

nckturner:
Sounds good, we can add as a part of migration documentation which there is already an issue for: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/480

PR #933, opened 2021-06-11
Grant EKSCTL_ADMIN_ROLE admin access to eksctl clusters

wongma7:

**What is this PR about? / Why do we need it?** this is part of the effort to get eksctl test CI in every PR.

To get Kubernetes API access to an EKS cluster you need to either be a) cluster creator or b) added to the aws-auth ConfigMap. Of course if you do not have Kubernetes API access in the first place then you can't edit the configmap, so the cluster creator has to do some bootstrapping. This PR enables the cluster creator to bootstrap one additional EKSCTL_ADMIN_ROLE with admin access to the cluster.

By contrast, with kops, if you have AWS API access to the kops cluster config in S3, then you can give yourself Kubernetes API access.

So why do we need one extra admin besides the cluster creator in the first place...I don't want to divulge too much pointless implementation detail about how the CI works but basically, the scripts are being run by an IAM entity in this account: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/aws/sig-aws-credentials.yaml#L17. The cluster creator will be that IAM entity. In case we need to access EKS clusters' Kubernetes API, we don't want to reuse that entity's creds. Better to at cluster-creation time grant a different, dedicated role the permission to access Kubernetes API via EKSCTL_ADMIN_ROLE.

**What testing is done?** will run eks test

### DIFF
--- a/hack/e2e/kops.sh
+++ b/hack/e2e/kops.sh
@@ -22,12 +22,14 @@ function kops_create_cluster() {
   CLUSTER_NAME=${2}
   BIN=${3}
   ZONES=${4}
-  INSTANCE_TYPE=${5}
-  K8S_VERSION=${6}
-  CLUSTER_FILE=${7}
-  KUBECONFIG=${8}
-  KOPS_PATCH_FILE=${9}
-  KOPS_STATE_FILE=${10}
+  NODE_COUNT=${5}
+  INSTANCE_TYPE=${6}
+  K8S_VERSION=${7}
+  CLUSTER_FILE=${8}
+  KUBECONFIG=${9}
+  KOPS_PATCH_FILE=${10}
+  KOPS_PATCH_NODE_FILE=${11}
+  KOPS_STATE_FILE=${12}
 
   generate_ssh_key "${SSH_KEY_PATH}"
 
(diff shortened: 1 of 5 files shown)


wongma7:
`Error: arn: invalid prefix `

wongma7:
2nd commit contains changes backported from fsx edition of scripts https://github.com/kubernetes-sigs/aws-fsx-csi-driver/commits/master/hack/e2e

hack/e2e/kops.sh
```diff
@@ -88,36 +93,63 @@ function kops_delete_cluster() {
   ${BIN} delete cluster --name "${CLUSTER_NAME}" --state "${KOPS_STATE_FILE}" --yes
 }
 
-# TODO switch this to python or work exclusively with yaml, all this
-# hacking with jq stinks!
+# TODO switch this to python, work exclusively with yaml, use kops toolbox
+# template/kops set?, all this hacking with jq stinks!
 function kops_patch_cluster_file() {
-  CLUSTER_FILE=${1}    # input must be json
+  CLUSTER_FILE=${1}    # input must be yaml
```
nckturner, line 99: This is a function I'd like to see done in something other than bash.  Good candidate for moving reusable parts of this to the test-infra repository?  (Not PR blocking, but something we can consider for the future).

hack/e2e/kops.sh
```diff
@@ -88,36 +93,63 @@ function kops_delete_cluster() {
...
   KOPS_PATCH_FILE=${2} # input must be yaml
+  KIND=${3}            # must be either Cluster or InstanceGroup
+  ROLE=${4}            # must be either Master or Node
 
   loudecho "Patching cluster $CLUSTER_NAME with $KOPS_PATCH_FILE"
 
-  # Temporary intermediate files for patching
+  # Temporary intermediate files for patching, don't mutate CLUSTER_FILE until
```
nckturner, line 106: Does this create the cluster.yaml file before the test?  Is this something we can use `KOPS_FEATURE_FLAGS="SpecOverrideFlag"` for?  Also, are these changes in kops.sh necessary for this PR?

wongma7: cluster.yaml gets created by the dry run create somewhere else. if it already exists, the pathcing won't hapepn.

No these changes aren't necessary, I just wanted to piggyback the changes from fsx repo

hack/e2e/kops.sh
wongma7: yes, see the comment. 

I know I have gone back and forth on this a lot, I decided we should use Go. You can see here https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/927/files#diff-1174eb3051f9c1a9e1f65c760347ef47446d5b93c712925b446b29ff591184e8R36 I'm trying it out and it's fine.

hack/e2e/kops.sh
wongma7: to be clear the pr desription is only relevant for first commit.

commit 2 and 3 are snuck in

nckturner: Ok, might want to break it up or update the PR description.

PR #937, opened 2021-06-15
Default extra-create-metadata true so that volumes get created with pvc/pv tags

wongma7:
**Is this a bug fix or adding new feature?** this flag has been in external-provisioner for a long time now, we should default it true so that volumes get created with pvc/pv tags kubernetes.io/created-for  https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/9f4ce44bd3c1d454177bd97b12170b6f95bfafde/pkg/driver/constants.go#L89

**What is this PR about? / Why do we need it?** people want tags https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/180

**What testing is done?**

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -46,7 +46,7 @@ enableVolumeSnapshot: false
 
 # Moving to values under controller
 affinity: {}
-extraCreateMetadata: false
+extraCreateMetadata: true
 extraVolumeTags: {}
 k8sTagClusterId:
 nodeSelector: {}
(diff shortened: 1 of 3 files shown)


wongma7:
/cc @jsafrane 

do you see any issue with enabling this flag on by default? imo it needs to be on by default for migration feature parity with in tree driver so that the same tags get added to volumes. I think that was your intention when you added the tagging functionality but not sure.  https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/9f4ce44bd3c1d454177bd97b12170b6f95bfafde/pkg/driver/constants.go#L89 thanks.

jsafrane:
IMO, it's good to have it on by default.
/lgtm

Issue #942, opened 2021-06-17
Snapshotter sidecar is deployed by Helm chart even if enableVolumeSnapshot == false


judgeaxl:
/kind bug

**What happened?**

When disabling snapshots, the ebs-snapshot-controller is no longer deployed, but the snapshot sidecar in the ebs-csi-controller is still deployed, causing repeated errors in the logs if the snapshot CRDs have not been deployed.

Example:
```
E0617 04:39:20.714430       1 reflector.go:127] github.com/kubernetes-csi/external-snapshotter/client/v3/informers/externalversions/factory.go:117: Failed to watch *v1beta1.VolumeSnapshotClass: failed to list *v1beta1.VolumeSnapshotClass: the server could not find the requested resource (get volumesnapshotclasses.snapshot.storage.k8s.io)
```

**What you expected to happen?**

In addition to not deploying the snapshot controller itself, it should not deploy the snapshot sidecar for the ebs-csi-controller pod, or at least disable it so that I don't get errors about a feature I don't need.

**How to reproduce it (as minimally and precisely as possible)?**

On a 1.19 or 1.20 cluster without the `SnapshotVolumeContent` and `SnapshotVolumeClass` CRDs, deploy the EBS-CSI Helm chart 1.2.0-1.2.2 with `enableVolumeSnapshot = false` and tail the ebs-csi-controller pod logs.

**Anything else we need to know?**:

I believe this worked in an earlier version of the chart, but I'm afraid I don't remember in what version.

From what I can tell, the snapshot controller stateful set checks the `enableVolumeSnapshot` flag, but the ebs-csi-controller does not check it for the csi-snapshotter sidecar container.

**Environment**
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #959, opened 2021-07-01
Release v1.1.1 and chart v1.2.4

wongma7:
**Is this a bug fix or adding new feature?**

**What is this PR about? / Why do we need it?** cherrypick of https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/958 onto master branch for housekeeping, doc update, andh elm chart release.

**What testing is done?**

Issue #635, opened 2020-12-04
Don't install snapshot controller


wongma7:
/kind bug

**What happened?** Snapshot controller is installed by helm.

**What you expected to happen?** Snapshot controller should not be installed, the driver should reuse an existing installation of snapshot controller, since the controller might be reused by other drivers. Or helm could somehow check if the controller already exists. Trying to install it again can cause conflict issues.

**How to reproduce it (as minimally and precisely as possible)?** do helm install

**Anything else we need to know?**: see also https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/601

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

wongma7:
I think the issue is we haven't ujpdated the yamls ever since snapshot controller gto split into a sidecar and separate controller. It's still a bug because we are out of date for not doing so!!

ayberk:
This confused me when I was updating the sidecars for the arm overlay. Unless someone else picks it up by then, I'll do this as part of updating sidecar images.

wongma7:
Our chart's handling of snapshotter controller and CRDs is causing a lot of headaches for both users and maintainers (i have spent way too much time thinking and writing comments about it -_-). I intend to remove the CRDs and controller and simply list it as a prerequisite to the chart. As of 1.20 kops will install the controller and CRDs by default. Other distros may do the same. Users whose distros don't will have to install themselves, we'll link them to https://github.com/kubernetes-csi/external-snapshotter

PR #960, opened 2021-07-01
[helm-chart] csi-snapshotter in ebs-csi-controller now checks for enableVolumeSnapshot before including it in containers

missingcharacter:
bug fix

Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/942

**What testing is done?** 
locally, yes

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -189,6 +189,7 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+{{- if or .Values.enableVolumeSnapshot (.Capabilities.APIVersions.Has "snapshot.storage.k8s.io/v1") }}
         - name: csi-snapshotter
           image: {{ printf "%s:%s" .Values.sidecars.snapshotterImage.repository .Values.sidecars.snapshotterImage.tag }}
           args:
@@ -210,6 +211,7 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+{{- end }}
         - name: csi-resizer
           image: {{ printf "%s:%s" .Values.sidecars.resizerImage.repository .Values.sidecars.resizerImage.tag }}
           imagePullPolicy: Always


wongma7:
This doesn't fully fix the issue because we might have users who already have the controller and CRDs installed, so they set enableVolumeSnapshot false. In that case they DO want the sidecar still running. 

TBH I'm leaning heavily toward not installing the snapshot controller and CRDS at all : ( https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/635#issuecomment-872534979, let me know your thoughts.

missingcharacter:
@wongma7 

>  I intend to remove the CRDs and controller and simply list it as a prerequisite to the chart

This makes total sense, I think it is best

wongma7:
Yeah my full reasoning is in the comment, @krmichel if we want to stop installing the CRDs and snapshot controller, and instead point users to docs to how to install it (leveraging the help text that prints right after an install) do we need to bump the helm chart major version?

missingcharacter:
@wongma7 
> do we need to bump the helm chart major version?
Yes, please 🙇🏽‍♂️

wongma7:
OK. I'm actually prepping helm chart 1.2.4 with your previous PR right now. So that will leave the 1.x.x series of charts in an okay-ish state (if you set enableVolumeSnapshot true it will install everything you need, if it's false then you might run into #942)

So, after 1.2.4 is released we can take some time to make any other breaking changes we want.

(EFS repo made a 2.0.0 release recently as well , so I am not worried at all about making a new major version too fast or whatever.)

missingcharacter:
@wongma7 is it OK if I just close this PR?

missingcharacter:
Actually, I think I can add a `capabilities` check to install the sidecar if `enableVolumeSnapshot` is true or the snapshot CRDs are present, it is not perfect but it would leave `1.2.4` in a better state

krmichel:
> Yes, please 🙇🏽‍♂️ (I know you are not asking me)

I fully agree.  I already have the second half of the major chart update that needs to do things like getting rid of the old not well-organized values and bump the helm version requirement so I will try to get those done and in PR so we can roll them into 2.0 as well.

missingcharacter:
Ok, this PR should leave `1.2.4` in better shape to then remove CRDs and snapshot controller off it

wongma7:
/lgtm

I'll rebase on this / include it in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/959

PR #965, opened 2021-07-05
Several breaking changes to the helm chart

krmichel:
**Is this a bug fix or adding new feature?**
Both and includes a breaking change

Fixes #635 
Fixes #847 
Fixes #964 

**What is this PR about? / Why do we need it?**
Removes deprecated helm values
Removes support for helm v2
Removes installation of the [Snapshot controller](https://github.com/kubernetes-csi/external-snapshotter)

PR #969, opened 2021-07-08
Better control over log verbosity.

michalschott:
This is chart update.

This PR will allow to have more control over log verbosity.

**What testing is done?** 
Chart has been templated and applied against local kind cluster.


charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -238,6 +238,7 @@ spec:
           image: {{ printf "%s:%s" .Values.sidecars.livenessProbeImage.repository .Values.sidecars.livenessProbeImage.tag }}
           args:
             - --csi-address=/csi/csi.sock
+            - --v={{ .Values.controller.logLevel }}
```
krmichel, line 241: I don't think that the liveness probe supports a `--v` flag

charts/aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -132,6 +132,7 @@ spec:
           image: {{ printf "%s:%s" .Values.sidecars.livenessProbeImage.repository .Values.sidecars.livenessProbeImage.tag }}
           args:
             - --csi-address=/csi/csi.sock
+            - --v={{ .Values.node.logLevel }}
```
krmichel, line 135: I don't think that the liveness probe supports a `--v` flag

charts/aws-ebs-csi-driver/templates/node-windows.yaml
```diff
@@ -140,6 +140,7 @@ spec:
           image: {{ printf "%s:%s" .Values.sidecars.livenessProbeImage.repository .Values.sidecars.livenessProbeImage.tag }}
           args:
             - --csi-address=unix:/csi/csi.sock
+            - --v={{ .Values.nodeWindows.logLevel }}
```
krmichel, line 143: I don't think that the liveness probe supports a `--v` flag

charts/aws-ebs-csi-driver/templates/node-windows.yaml
```diff
@@ -58,7 +58,7 @@ spec:
             - --volume-attach-limit={{ .Values.volumeAttachLimit }}
             {{- end }}
             - --logtostderr
-            - --v=5
+            - --v={{ .Values.nodeWindows.logLevel }}
```
krmichel, line 61: The windows node currently uses the `node.*` values so this should be changed to match

charts/aws-ebs-csi-driver/templates/node-windows.yaml
```diff
@@ -108,7 +108,7 @@ spec:
           args:
             - --csi-address=$(ADDRESS)
             - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
-            - --v=5
+            - --v={{ .Values.nodeWindows.logLevel }}
```
krmichel, line 111: The windows node currently uses the `node.*` values so this should be changed to match

charts/aws-ebs-csi-driver/templates/snapshot-controller.yaml
```diff
@@ -57,12 +57,12 @@ spec:
             {{- include "aws-ebs-csi-driver.http-proxy" . | nindent 12 }}
           {{- end }}
           args:
-            - --v=5
+            - --v={{ .Values.snapshotController.logLevel }}
```
krmichel, line 60: You should remove this since the snapshot controller will soon no longer be installed via this chart

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -129,6 +129,9 @@ controller:
 # The "maximum number of attachable volumes" per node
 volumeAttachLimit:
 
+snapshotController:
```
krmichel, line 132: You should remove this since the snapshot controller will soon no longer be installed via this chart.  Also, `snapshotController` already exists as a value.

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -148,6 +151,9 @@ node:
     liveness: {}
   volumeAttachLimit:
 
+nodeWindows:
```
krmichel, line 154: The windows node currently uses the node.* values so this should be removed

krmichel (changes requested):
I am currently working on PR #965 that will reorganize some of the sidecar-related values.  I believe that these additional log level values would fit better in the new structure.  What do you think about one of these options?

1. I incorporate these changes into that PR
2. You wait until after those changes to make your PR so the values can be put in the better structure

krmichel:
I went ahead and made changes on PR #965 if you want to check them out.

wongma7:
/ok-to-test

I'm OK with merging this change as part of #965 , will simplify rebasing pain for all.

charts/aws-ebs-csi-driver/templates/controller.yaml
michalschott:
```
➜  aws-ebs-csi-driver git:(master) docker run --rm -it k8s.gcr.io/sig-storage/livenessprobe:v2.2.0 -h                                                           <aws:saml>
Usage of /livenessprobe:
...
  -v value
    	number for the log level verbosity
...
```

michalschott:
Actioned all suggestions from above.

michalschott:
Yay, looks like this one is no longer required. Thanks!

PR #970, opened 2021-07-08
Feature/allow add debug args

mkkatica:
Adding a new feature.

I was trying to debug a deployment of the EBS CSI and needed the `--aws-sdk-debug-log` flag set to true. I could not do it from the chart, so I had to keep editing the deployment in-place, and my CI process kept redeploying the helm chart. Being able to set this value from the chart, and other aribitrary commandline flags would make it easier for power users to debug.

**What testing is done?** 
I've templated the chart a few times with different values, and deployed it to a cluster without issue.

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -60,7 +60,7 @@ topologySpreadConstraints: []
 
 controller:
   # If arbitrary args like "--aws-sdk-debug-log=true" need to be passed, use this value
-  # additionalArgs: []
+  additionalArgs: []
   affinity: {}
   # True if enable volume scheduling for dynamic volume provisioning
   env:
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -59,6 +59,8 @@ tolerations: []
 topologySpreadConstraints: []
 
 controller:
+  # If arbitrary args like "--aws-sdk-debug-log=true" need to be passed, use this value
+  # additionalArgs: []
```
krmichel, line 63: Don't comment this line out

krmichel (changes requested):

krmichel:
@mkkatica you will need to sign the CLA before this can be tested/merged

mkkatica:
> @mkkatica you will need to sign the CLA before this can be tested/merged

I went through our internal process to get added to our Company CLA. That is complete and the status should be reflected in the license check.

krmichel:
> I went through our internal process to get added to our Company CLA. That is complete and the status should be reflected in the license check.

Yep, looks good.  Now you just need a kubernetes-sigs member to flag it as okay to test.

krmichel (approved):

wongma7:
/ok-to-test
/lgtm
/approve

thanks

PR #973, opened 2021-07-15
Chart values

tlcowling:

**What testing is done?**


tlcowling:
Made this in error, apologies for the spamming!

PR #982, opened 2021-07-21
utilize latest go sdk to ensure createVolume idempotency

### LINKED ISSUE #165: Race condition in CreateVolume

AndyXiangLi:
Fixes #165 

EBS team has updated the createVolume api to accept ClientToken in the createVolume request. doc [here](https://github.com/aws/aws-sdk-go/releases/tag/v1.40.4)
With ClientToken, EBS will be able to ensure the createVolume Idempotency at Api level.
 I think it is ok to keep inflight cache as it is, it ensure the driver level idempotency, and this change is more on the api level.
**What testing is done?** 
unit test
sanity test
Manually tested described in the original re-produce step and observed only one volume is created
```
csc controller create-volume -e <endpoint> --cap 1,1 xiang-test --req-bytes 1024000 & \
> csc controller create-volume -e <endpoint> --cap 1,1 xiang-test --req-bytes 1024000
[1] 2801280
Create volume request for xiang-test is already in progress

Please use -h,--help for more information
"vol-0c8323ca0f9ab1a6d"	1073741824

### CODE REVIEW

File: pkg/driver/controller.go [hunk 1]
```diff
@@ -115,19 +115,6 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	defer d.inFlight.Delete(volName)
 
-	disk, err := d.cloud.GetDiskByName(ctx, volName, volSizeBytes)
```
AndyXiangLi, line 118: With the volumeName in the ClientToken, we no longer need describeVolume to check if the volume has already been created or not. But we will also lose some visibility to customer, if the two createVolume requests have same clientToken(volumeName) but different parameters (size or snapshotId), ebs api will return general idempotency violation exception. (instead of saying diff size or diff snapshotId)

Open to any idea, if you want me to add it back

vdhanan: i'm good with removing this code

pkg/cloud/cloud_test.go
```diff
@@ -203,6 +203,30 @@ func TestCreateDisk(t *testing.T) {
...
 			expCreateVolumeErr:   fmt.Errorf("CreateVolume generic error"),
 		},
+		{
+			name:       "fail: ec2.CreateVolume returned snapshot not found error",
+			volumeName: "vol-test-name-error",
+			diskOptions: &DiskOptions{
+				CapacityBytes:    util.GiBToBytes(1),
+				Tags:             map[string]string{VolumeNameTagKey: "vol-test", AwsEbsDriverTagKey: "true"},
+				AvailabilityZone: expZone,
+			},
```
vdhanan, line 213: i'm confused by this test case. don't you need a snapshot ID in the diskOptions to get a Snapshot not found error?

AndyXiangLi: It's testing [cloud.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L350) so we don't need to actually provide a valid snapshotId, this test case just expect go sdk to return snapshot not found fault code to the client and see if client can handle the error properly.

vdhanan: i see. we're just checking that the error gets handled. makes sense.

vdhanan:
let's remove the vendor directory soon haha. these PRs are unnecessarily big

AndyXiangLi:
> let's remove the vendor directory soon haha. these PRs are unnecessarily big

Yeah, agree

Issue #985, opened 2021-07-22
[chart] Add updateStrategy to chart for daemonsets


stevehipwell:
**Is your feature request related to a problem?/Why is this needed**
There is no way to set the `updateStrategy` for the daemonsets which means that a node upgrade can only happen as fast as this daemonset can restart. For example the aws-vpc-cni already has this [functionality](https://github.com/aws/eks-charts/blob/a46497dc49804fd0cb985cf8ce8e7ba8745819fc/stable/aws-vpc-cni/values.yaml#L103).

**Describe the solution you'd like in detail**
Add `updateStrategy` to the values for the node and use it in the node templates.

**Describe alternatives you've considered**
n/a

**Additional context**
I'm happy to open a PR to add this in.

PR #988, opened 2021-07-23
[chart] Node update strategy & auto driver image tag

stevehipwell:
This PR adds the ability to configure the node update strategy and uses the chart app version for the image tag.

Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/985.

With the default update strategy the more nodes added to the cluster the longer it takes to upgrade, this PR makes this customisable and sets the default to the 10% unavailable strategy used by the aws-vpc-cni and the EKS kube-proxy daemonsets.

**What testing is done?** 
Helm template output validated.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "1.2.0"
+appVersion: 1.2.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.0.3
+version: 2.0.4
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
@@ -16,3 +16,9 @@ maintainers:
     email: chengpan@amazon.com
   - name: krmichel
     url: https://github.com/krmichel
+annotations:
+  artifacthub.io/changes: |
+    - kind: changed
+      description: Use chart app version as default image tag
+    - kind: changed
+      description: Add updateStrategy to daemonsets
--- a/charts/aws-ebs-csi-driver/templates/node-windows.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node-windows.yaml
@@ -10,6 +10,8 @@ spec:
     matchLabels:
       app: ebs-csi-node
       {{- include "aws-ebs-csi-driver.selectorLabels" . | nindent 6 }}
+  updateStrategy:
+    {{ toYaml .Values.node.updateStrategy | nindent 4 }}
   template:
     metadata:
       labels:
@@ -54,7 +56,7 @@ spec:
         {{- end }}
       containers:
         - name: ebs-plugin
-          image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+          image: {{ printf "%s:%s" .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
           args:
             - node
             - --endpoint=$(CSI_ENDPOINT)
--- a/charts/aws-ebs-csi-driver/CHANGELOG.md
+++ b/charts/aws-ebs-csi-driver/CHANGELOG.md
@@ -1,6 +1,25 @@
 # Helm chart
 
-# v2.0.0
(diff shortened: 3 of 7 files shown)


krmichel (approved):

stevehipwell:
@vdhanan any chance this could be merged?

vdhanan:
thanks for the PR, i'll review it today

stevehipwell:
@vdhanan I've rebased to capture the last chart release and updated the changelog again for the last chart release. Is there any chance of a review today, I need the update strategy feature ASAP.

charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "1.1.3"
+appVersion: 1.1.3
```
krmichel, line 2: Is there a reason to add the `v` elsewhere instead of just setting this to "v1.1.3"

krmichel (changes requested):

charts/aws-ebs-csi-driver/Chart.yaml
stevehipwell: Although the spec says that `appVersion` needn't be a valid SemVer version, there are tools that either don't work correctly or error if it's not. The comment about quotes being recommended is also misleading as they're only relevant if you have a version that is a different data type.

stevehipwell: Additionally, for many reasons which I won't bother listing, sticking with a valid SemVer 2 version here will cause the least potential for disruption as tools and technologies move on.

stevehipwell:
@krmichel @vdhanan I've rebased and updated the changelog again.

charts/aws-ebs-csi-driver/Chart.yaml
krmichel: Fair enough

krmichel (approved):

krmichel:
@stevehipwell I am not an owner, so I can't trigger a merge.  @vdhanan can you get this merged for Steve?

charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -16,3 +16,9 @@ maintainers:
     email: chengpan@amazon.com
   - name: krmichel
     url: https://github.com/krmichel
+annotations:
+  artifacthub.io/changes: |
+    - kind: changed
+      description: Use chart app version as default image tag
+    - kind: changed
+      description: Add updateStrategy to daemonsets
```
vdhanan, line 24: i'm curious what purpose these annotations serve when we already have a helm chart changelog

vdhanan:
sorry for the delayed review, and thanks for the PR -- could you please run `make generate-kustomize` so that the updateStrategy is available for kustomize users as well? otherwise, LGTM. i'll get this merged immediately once that change is made

charts/aws-ebs-csi-driver/templates/node-windows.yaml
```diff
@@ -54,7 +56,7 @@ spec:
         {{- end }}
       containers:
         - name: ebs-plugin
-          image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+          image: {{ printf "%s:%s" .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
```
AndyXiangLi, line 59: Will .Values.image.tag be empty after change?
Can we use the v{Chart.appversion} in values.yaml continue use the tag field in the template?

charts/aws-ebs-csi-driver/Chart.yaml
krmichel: They will show up on artifact.io when the issue to publish there gets completed

charts/aws-ebs-csi-driver/templates/node-windows.yaml
krmichel: If image.tag is not set it will fall back to the value in appVersion.  When running a helm install/upgrade you can either use --values to pass a yaml file of values that could contain image.tag to provide a version or use --set image.tag to set the version to use.

https://helm.sh/docs/helm/helm_install/

stevehipwell:
Thanks @vdhanan, I've re-generated the Kustomize maifests.

PR #991, opened 2021-07-24
Release v1.1.3

wongma7:

**What is this PR about? / Why do we need it?** bug in our internal build means the ECR tag is pointing to the debian image, fix.

**What testing is done?** 
N/A


wongma7:
admin merging, we don't have a retest not required thingie....

PR #992, opened 2021-07-24
Release 1.1.3

wongma7:

cherry-picking helm/doc changes from release-1.1 branch to master.

we released 1.1.2 with base image changes for cve, and then 1.1.3 right after cuz of an issue with the ECR build.

/hold
**What testing is done?**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN make
 FROM amazonlinux:2 AS amazonlinux
 RUN yum update -y
 RUN yum install ca-certificates e2fsprogs xfsprogs util-linux -y
+RUN yum clean all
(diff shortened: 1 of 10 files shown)


wongma7:
/hold cancel

nvm.

Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN make
 FROM amazonlinux:2 AS amazonlinux
```
SaranBalaji90, line 20: As discussed offline using https://hub.docker.com/r/amd64/amazonlinux/ reduced image size significantly (from 750MB to 250MB). We should move to that instead of using amazonlinux:2

RCarretta:
> we released 1.1.2 with base image changes for cve, and then 1.1.3 right after cuz of an issue with the ECR build.

Hey, could you check https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/993 ?

PR #1006, opened 2021-07-31
Cleanup and refactor helm chart

AtkinsChang:

base on #1013 

- update RBAC from upstream
- unify style (ex. `default "foo" .Value.foo` vs `.Value.foo | default "foo"`)
- unify `socket-dir` mount path

#### Controller
- add options:
  - `controller.podSecurityContext`
  - `controller.securityContext`
- use `system-cluster-critical` as default value of `priorityClassName`
- prevent from deploying csi-snapshotter related resources if not supported

#### Node
- add options:
  - `node.podSecurityContext`
  - `node.securityContext`
  - `node.additionalArgs` (this option is already in controller)
- set `ImagePullPolicy` according to `image.pullPolicy`

#### Sidecar
- add `sidecars.<name>.image.pullPolicy` option for images

@krmichel is it ok to rename any resource under `deploy/kubernetes`?


AtkinsChang:
I signed it

krmichel:
I am not very familiar with kustomize but I believe the files can be renamed as long as the `kustomization.yaml` is updated correctly to reflect the changes.  The makefile for this project has a target called `generate-kustomize` that generates the files in the `deploy/kubernetes/base` directory and should be run when changes to the chart have been made.  If the kustomize files are going to be renamed then the make target needs to be updated to reflect that.  This project would like the kustomize files to match names with the helm templates that create them.  I don't know that this chart is big enough to warrant splitting the templates into sub-directories, but am also not opposed to it.  @wongma7 do you have any thoughts on organizing the templates into sub-directories by function?  If we are going to do that I would like to see the moving/organizing of the templates done as its own PR and to ensure that the files all show as moved in git.  The combination of moving files around, renaming a lot of the resources, and the new editions is, in my opinion, too much to deal with in a single PR.  I also don't know a whole lot about the driver itself. @wongma7, @ayberk, @AndyXiangLi, or @vdhanan do any of you know if the RBAC should be synced as indicated with the comments in the changes like `this should be sync from upstream`?  Instead of truncating the name and fullname named templates at 40 characters, which is based on the length of fairly arbitrary strings, I think it would be better to have a suffix name helper that combined the name/fullname with a provided suffix and then truncates based on a specific suffix.  This could either be done by adding the suffix and then truncating or by doing some math based on suffix length and truncating the name before adding the suffix.  Since it seems like the suffix is the more descriptive part, in this case, doing the latter is probably the better option.  I just don't want to truncate information when we don't need to.

Please, unless one of the owners says that they don't want the templates reorganized, split this into two PRs.  One for the template rename/reorg and another for the resource renaming/adding additional functionality/image version updates.

AtkinsChang:
@krmichel 
maybe I should refine my question more accurately:
should the kustomize related files in `deploy/kubernetes/base` keep the compatibility? if not then just update the `Makefile`, if yes then can be handled via `{{- if eq .Release.Name "kustomize" }}` 

> do any of you know if the RBAC should be synced as indicated with the comments in the changes like this should be sync from upstream? 

Some RBAC are used by the [external-attacher](https://github.com/kubernetes-csi/external-attacher), [external-provisioner](https://github.com/kubernetes-csi/external-provisioner), [external-resizer](https://github.com/kubernetes-csi/external-resizer), [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter).
For example, kubernetes-csi/external-attacher#146  update the deprecated `csinodeinfos` but [clusterrole-attacher.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/2c9ad2b89b7b4355fd5ee28769d7cda489d25307/charts/aws-ebs-csi-driver/templates/clusterrole-attacher.yaml#L15-L16) is not update. It is lucky because some people update [clusterrole-provisioner.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/clusterrole-provisioner.yaml#L27-L29) so the `external-attacher` sidecar JUST WORKS.
I think we should leave these untouched and keep it sync when the sidecar version is updated.
Put the part related to this plugin to another `ClusterRole` (for example `nodes` API introduce in #907)

I known this is a big PR and will split this. But I'd like to hear which to do and which not so I open a draft PR. Thanks for advice.

charts/aws-ebs-csi-driver/templates/node/daemonset-windows.yaml
```diff
@@ -1,28 +1,29 @@
...
-        {{- if .Values.node.podLabels }}
-        {{- toYaml .Values.node.podLabels | nindent 8 }}
-        {{- end }}
+        {{- include "aws-ebs-csi-driver.node-windows.labels" . | nindent 8 }}
+      {{- with .Values.node.podLabels }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- with .Values.node.podAnnotations }}
       annotations:
-        {{- toYaml . | nindent 8 }}
```
krmichel, line 25: Indents like this are to make the template more readable and show where the result will be.  I would leave them in.

charts/aws-ebs-csi-driver/templates/node/daemonset-windows.yaml
```diff
@@ -37,53 +38,63 @@ spec:
...
         {{- with .Values.node.nodeSelector }}
-        {{- toYaml . | nindent 8 }}
+          {{- toYaml . | nindent 8 }}
         {{- end }}
-      serviceAccountName: {{ .Values.node.serviceAccount.name }}
-      priorityClassName: {{ .Values.node.priorityClassName | default "system-node-critical" }}
+      serviceAccountName: {{ include "aws-ebs-csi-driver.node.serviceAccountName" . }}
+      priorityClassName: {{ default "system-node-critical" .Values.node.priorityClassName }}
       tolerations:
-        {{- if .Values.node.tolerateAllTaints }}
```
krmichel, line 45: Again for readability, I would leave these indents.

charts/aws-ebs-csi-driver/templates/node/daemonset-windows.yaml
```diff
@@ -37,53 +38,63 @@ spec:
...
+          {{- with .Values.node.logLevel }}
+            - --v={{ . }}
+          {{- end }}
+          {{- with .Values.node.additionalArgs }}
+            - {{ toYaml . | nindent 12 }}
+          {{- end }}
           env:
             - name: CSI_ENDPOINT
-              value: unix:/csi/csi.sock
+              value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
```
krmichel, line 81: Are these paths different on different nodes?  Should this be configurable?

charts/aws-ebs-csi-driver/templates/node/daemonset-windows.yaml
```diff
@@ -37,53 +38,63 @@ spec:
...
-            {{- end }}
+          {{- end }}
           volumeMounts:
             - name: kubelet-dir
               mountPath: C:\var\lib\kubelet
               mountPropagation: "None"
-            - name: plugin-dir
-              mountPath: C:\csi
+            - name: socket-dir
+              mountPath: C:\var\lib\csi\sockets\puuginproxy
```
krmichel, line 97: `puuginproxy` is that a typo?

krmichel (commented):
The majority of what you are adding as new options has no default values and since you are using `with` blocks I don't think you will need any if to prevent them from doing anything undesirable in the kustomize manifests.  You will need to make sure that the templates that you are adding API checks to have `--api-versions 'snapshot.storage.k8s.io/v1'` added in the makefile to ensure the manifests are generated correctly.  Let me know if you need anything else before splitting this PR.

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -13,40 +13,46 @@ sidecars:
...
+      tag: v2.3.0
+      pullPolicy: IfNotPresent
     resources: {}
   resizer:
     env: []
     image:
       repository: k8s.gcr.io/sig-storage/csi-resizer
-      tag: "v1.0.0"
+      tag: v1.2.0
+      pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
AtkinsChang: I think `Always` is wrong value with fixed image version, and taking the risk of Docker Hub API rate limiting

charts/aws-ebs-csi-driver/templates/node/daemonset-windows.yaml
AtkinsChang: This can be set to any path, but this and the volume mount path should be same.

charts/aws-ebs-csi-driver/values.yaml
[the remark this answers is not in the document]
krmichel: I don't believe that `k8s.gcr.io` is docker hub.  I think it is google cloud.  Also, I believe, if there is already a local copy of the image and the digest for the image hasn't changed it won't pull but instead reuse the local cached images since it hasn't changed.

charts/aws-ebs-csi-driver/templates/node/daemonset-windows.yaml
krmichel: What I mean is that on one windows system could this be a different path than on another?  Should the user be able to specify/override this value?

krmichel:
I would split the PR before making any more updates.

charts/aws-ebs-csi-driver/values.yaml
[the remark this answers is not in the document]
AtkinsChang: Ok if you think it is reasonable to have `csi-resizer` using `Always` as default image pull policy and all other containers using `IfNotPresent` then I will do it

AtkinsChang:
#1013 is what you talk about

krmichel:
> #1013 is what you talk about

Sort of.  I was wanting to split the renaming of the resources inside of the templates from moving/renaming the templates files themselves.  Doing them together, based on git figuring out what is a rename based on content similarity, causes some of the files to appear to be deleted and then new files added since the resource renaming, inside the templates, causes them to no longer be similar enough to be a rename.  The first PR would just be moving/renaming files and then fixing the paths in the Makefile to ensure git tracks all the renames.

PR #1008, opened 2021-08-03
[chart] Add controller strategy

stevehipwell:
This PR adds the ability to configure the controller update strategy.

With the default update strategy and tightly mapped zones the surge pods cause issues.

What testing is done?
Helm template output validated.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.2.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.0.4
+version: 2.1.0
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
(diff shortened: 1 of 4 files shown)


charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.2.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.0.4
+version: 2.0.5
```
krmichel, line 5: This PR is adding new features to the chart.  The version should be 2.1.0

krmichel (changes requested):

charts/aws-ebs-csi-driver/Chart.yaml
stevehipwell: I wasn't sure, a lot of PRs I've been doing recently have only be accepted as patches, will fix.

stevehipwell: Done.

krmichel (approved):

stevehipwell:
@vdhanan could you take a look at this?

stevehipwell:
@vdhanan tests are all passing, could you or @jsafrane take a look at getting this merged?

vdhanan:
thanks!

PR #1015, opened 2021-08-06
Update helm chart alongside kustomize, after images have been pushed, for consistency

wongma7:
**Is this a bug fix or adding new feature?** docs

**What is this PR about? / Why do we need it?** If we are going to do two commits to the release branch per release,

1. to trigger the image build
2. to update the kustomize yaml to refer to the image

it makes sense to me that helm yaml updates should go into step 2, not 1.

**What testing is done?**

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -106,3 +106,7 @@ Send a PR to merge both the release and post-release commits to the main branch.
 ## Verify the helm chart release
 
 Visit the [Releases](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/releases) pages to verify we have a new helm chart release.
+
+## Update AWS EKS documentation
+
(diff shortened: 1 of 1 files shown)


docs/RELEASE.md
```diff
@@ -36,7 +36,7 @@ You also might need to create a release branch, if it doesn't already exist. For
 
 ## Create the release commit in the release branch
 
-### Generate the CHANGELOG
```
AndyXiangLi, line 39: Should we rename the file or create separate changelog file? It is not 0.x anymore.

wongma7: youre right, I'll make a CHANGELOG-1.x in separate PR as rename would break link

AndyXiangLi:
/lgtm
Thanks!

PR #1016, opened 2021-08-09
Fix github build failing because of wrong docker hub registry name 

wongma7:

build is failing. There is no "library" in the registry name https://github.com/kubernetes-sigs/aws-ebs-csi-driver/runs/3256645072

**What testing is done?** locally can pull from docker.io/amazon but not docker.io/library/amazon, the former is the correct one.


AndyXiangLi:
/lgtm
Thanks!

PR #1019, opened 2021-08-10
Node Publish Mount Idempotent

### LINKED ISSUE #955: Make NodePublish mount  idempotent
/kind bug

**What happened?** as of kubernetes 1.20, kubelet doesn't check if the target path exists or is mounted for us https://github.com/kubernetes/kubernetes/pull/88759 . path existence should be idempotent i.e. if path already exists, NodeStage/NodePublish should not error. mount should be idempotent i.e. if the thing that's supposed to be mounted at the path is already there then  NodeStage/NodePublish should not error

NodeStage:
- [X] path existence idempotent https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/951bafe01a39ecf3575ec786fbb6da5ba2bc4889/pkg/driver/node.go#L175
- [X] mount idempotent https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/951bafe01a39ecf3575ec786fbb6da5ba2bc4889/pkg/driver/node.go#L203

NodeVolume:
- [X] path existence idempotent https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/951bafe01a39ecf3575ec786fbb6da5ba2bc4889/pkg/driver/mount_linux.go#L56
- [ ] mount idempotent https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/951bafe01a39ecf3575ec786fbb6da5ba2bc4889/pkg/driver/node.go#L627

**Environment**
- Driver version:

nirmalaagash:
Issue Fix #955  - Node publish mount idempotent

The PR adds mount idempotent to the CSI driver which provides multiple requests of 'NodePublishVolume' with the same request parameters would not throw error. We need it for the idempotency in mount.

**What testing is done?** 
Unit testing.
Manual testing is done using [csc](https://github.com/rexray/gocsi/tree/master/csc) tool. Below are the screenshots which shows the NodePublishVolume call of same request made multiple times. At first, it publishes the volume at the target and later request provides the information that the target is already mounted.

Screen Shot 2021-08-10 at 11 55 29 AM
Screen Shot 2021-08-10 at 11 55 07 AM

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -620,10 +620,39 @@ func (d *nodeService) nodePublishVolumeForFileSystem(req *csi.NodePublishVolumeR
 	mountOptions = collectMountOptions(fsType, mountOptions)
 
 	klog.V(4).Infof("NodePublishVolume: mounting %s at %s with option %s as fstype %s", source, target, mountOptions, fsType)
-	if err := d.mounter.Mount(source, target, fsType, mountOptions); err != nil {
-		if removeErr := os.Remove(target); removeErr != nil {
-			return status.Errorf(codes.Internal, "Could not remove mount target %q: %v", target, err)
+	/*
+		Checking if it's a mount point. There are three cases,
+		1. true, err when the directory does not exist or corrupted.
+		2. false, nil when the path is already mounted with a device.
+		3. true, nil when the path is not mounted with any device.
+	*/
+	notMnt, err := d.mounter.IsLikelyNotMountPoint(target)
+	if err != nil && !os.IsNotExist(err){
+		//Checking if the path exists and error is related to Corrupted Mount, in that case, the system could unmount and mount.
+		_, pathErr := mountutils.PathExists(target)
+		if pathErr != nil && mountutils.IsCorruptedMnt(pathErr) {
+			klog.V(4).Infof("Target path %q is a corrupted directory",target)
+		}else{
+			return status.Errorf(codes.Internal, "Could not mount %q at %q: %v", source, target, err)
 		}
+	}
+
+	if !notMnt {
+		//Return error when any of the directory inside is not readable which means that a device could be mounted.
+		_, err = os.ReadDir(target)
+		if err != nil {
+			klog.V(4).Infof("Error occurred at reading directory %q. Trying to Unmount.",target)
+			//Reading the directory failed and trying to unmount and remount
+			if mntErr := d.mounter.Unmount(target); mntErr != nil {
+				return status.Errorf(codes.Internal, "Unable to unmount the target %q : %v", target, err)
+			}
+		} else {
+			klog.V(4).Infof("Target path %q is already mounted",target)
+			return nil
+		}
+	}
+
+	if err := d.mounter.Mount(source, target, fsType, mountOptions); err != nil {
 		return status.Errorf(codes.Internal, "Could not mount %q at %q: %v", source, target, err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/driver/node.go
```diff
@@ -620,13 +620,41 @@ func (d *nodeService) nodePublishVolumeForFileSystem(req *csi.NodePublishVolumeR
...
+		1. true, err when the directory does not exist or corrupted.
+		2. false, nil when the path is already mounted with a device.
+		3. true, nil when the path is not mounted with any device.
+	*/
+	notMnt, err := d.mounter.IsLikelyNotMountPoint(target)
+	if err != nil && !os.IsNotExist(err) {
+		//Checking if the path exists and error is related to Corrupted Mount, in that case, the system could unmount and mount.
+		_, pathErr := mountutils.PathExists(target)
+		if pathErr != nil && mountutils.IsCorruptedMnt(pathErr) {
+			klog.V(4).Infof("Target path %q is a corrupted directory", target)
```
vdhanan, line 634: are we not returning an error if target path is a corrupted directory?

pkg/driver/node.go
```diff
@@ -620,13 +620,41 @@ func (d *nodeService) nodePublishVolumeForFileSystem(req *csi.NodePublishVolumeR
...
+		} else {
+			return status.Errorf(codes.Internal, "Could not mount %q at %q: %v", source, target, err)
 		}
-		return status.Errorf(codes.Internal, "Could not mount %q at %q: %v", source, target, err)
 	}
 
+	if !notMnt {
+		//Return error when any of the directory inside is not readable which means that a device could be mounted.
```
vdhanan, line 641: could you clarify what this comment means? not sure i'm following

vdhanan:
are all the new code paths (i.e. the if/else conditions) covered by the existing unit tests? i feel like we probably want to add a couple new ones, right?

pkg/driver/node.go
nirmalaagash: Chances are that the device mounted or the mount point is corrupted because of improper unmounting. Hence in the next phase we try to read the directory and unmount it.

pkg/driver/node.go
nirmalaagash: if a device is already mounted in the target directory, the directories inside the directory would not be accessible until the device is unmounted from the target directory. Hence in that case, it would return an error.

bertinatto (commented):
Thank you for working on this, @nirmalaagash.

Does your fix also work block volumes? It seems we're only covering filesystem volumes.

nirmalaagash:
> Thank you for working on this, @nirmalaagash.
> 
> Does your fix also work block volumes? It seems we're only covering filesystem volumes.

@bertinatto Yes. This PR only covers the filesystem volumes.

pkg/driver/node.go
```diff
@@ -620,13 +620,41 @@ func (d *nodeService) nodePublishVolumeForFileSystem(req *csi.NodePublishVolumeR
...
-	if err := d.mounter.Mount(source, target, fsType, mountOptions); err != nil {
-		if removeErr := os.Remove(target); removeErr != nil {
-			return status.Errorf(codes.Internal, "Could not remove mount target %q: %v", target, err)
+	/*
+		Checking if it's a mount point. There are three cases,
+		1. true, err when the directory does not exist or corrupted.
+		2. false, nil when the path is already mounted with a device.
+		3. true, nil when the path is not mounted with any device.
+	*/
+	notMnt, err := d.mounter.IsLikelyNotMountPoint(target)
```
bertinatto, line 629: An easy way to cover block devices as well is to move the check right below the in flight check.

You may want to take a look at the solutions both [GCP PD](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/a26ef36a9e6bf92cc32ee93bb72b2a985d61e3fe/pkg/gce-pd-csi-driver/node.go#L117) and Azure Disk implemented.

nirmalaagash: @bertinatto Thanks. New commit covering block devices is pushed.

nirmalaagash:
> are all the new code paths (i.e. the if/else conditions) covered by the existing unit tests? i feel like we probably want to add a couple new ones, right?

 [Commit](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1019/commits/ac4a32e41b989445145bd4a18d3a60f132636cf3) include Unit Test Cases covering the conditions implemented. Thanks.

PR #1035, opened 2021-08-31
Fix spacing in RELEASE.md

wongma7:

**What testing is done?**


wongma7:
not sure why github is running buildx job in this pr, it should only run on pushes to release/master branches?

wongma7:
overriding merge because it's just a docs change

PR #1041, opened 2021-09-02
hash volume name to get client token

### LINKED ISSUE #1038: could not create volume in EC2: InvalidParameterValue: Client token must be less than or equal to 64 characters
/kind bug
/kind regression

**What happened?**
After upgrade from aws-ebs-csi-driver@v1.1.1 to aws-ebs-csi-driver@v1.2.0, PVC creation fails with

```
could not create volume in EC2: InvalidParameterValue: Client token must be less than or equal to 64 characters
```

The PR introduces the regression seems to be https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/982 (cc @AndyXiangLi).

PVC creation to work with aws-ebs-csi-driver@v1.2.0.

1. Create a sample StatefulSet and PVC

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: claim
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 4Gi
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: app
  labels:
    app: app
spec:
  serviceName: app
  replicas: 1
  selector:
    matchLabels:
      app: app
  template:
    metadata:
      labels:
        app: app
    spec:
      containers:
      - name: app
        image: centos
        command: ["/bin/sh"]
        args: ["-c", "while true; do echo $(date -u) >> /data/out.txt; sleep 5; done"]
        volumeMounts:
        - name: persistent-storage
          mountPath: /data
      volumes:
      - name: persistent-storage
        persistentVolumeClaim:
          claimName: claim
```

2. Make sure that the PV fails to be provisioned

```
$ k describe pvc claim

Events:
  Type     Reason                Age              From                                                                                      Message
  ----     ------                ----             ----                                                                                      -------

  Warning  ProvisioningFailed    0s               ebs.csi.aws.com_csi-driver-controller-f9b9fdf-gkprq_0e0ed09e-055c-4516-80e2-e66122f7da60  failed to provision volume with StorageClass "default": rpc error: code = Internal desc = Could not create volume "pv-shoot--foooooo--rxt19rv5mw-6f4e647a-6b9f-46e8-8477-a0e6b546d257": could not create volume in EC2: InvalidParameterValue: Client token must be less than or equal to 64 characters
           status code: 400, request id: <omitted>
```

**Environment**
- Kubernetes version (use `kubectl version`): v1.18.20
- Driver version: v1.2.0
- External provisioner - v2.1.2

vdhanan:
fixes #1038 

hash the volume name using sha-256 in case the name is longer than 64 characters.

**What testing is done?** 
manually tested that a volume could be created with new token

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -18,6 +18,8 @@ package cloud
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"os"
@@ -320,9 +322,12 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		}
 	}
 
+	// We hash the volume name to generate a unique token that is less than or equal to 64 characters
+	clientToken := sha256.Sum256([]byte(volumeName))
+
 	requestInput := &ec2.CreateVolumeInput{
 		AvailabilityZone:  aws.String(zone),
-		ClientToken:       aws.String(volumeName),
+		ClientToken:       aws.String(hex.EncodeToString(clientToken[:])),
 		Size:              aws.Int64(capacityGiB),
 		VolumeType:        aws.String(createType),
 		TagSpecifications: []*ec2.TagSpecification{&tagSpec},


pkg/cloud/cloud.go
```diff
@@ -320,9 +322,12 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		}
 	}
 
+	// We hash the volume name to generate a unique token that is less than or equal to 64 characters
+	clientToken := sha256.Sum256([]byte(volumeName))
```
AndyXiangLi, line 326: I think encoding is enough here, why do we need extra hashing?

vdhanan: in this step, we are hashing the `volumeName`, resulting in a byte slice of length 32 (256 bits). in the next step, we hex encode the slice and convert it to a string, ensuring it is always length 64. i think this is the safest way to guarantee we never exceed 64 characters in the token.

AndyXiangLi:
/lgtm
thanks

PR #1043, opened 2021-09-02
release 1.2.1

vdhanan:
patch

fixes mount idempotency bug

**What testing is done?**


james-callahan:
You missed updating the version in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/v1.2.1/deploy/kubernetes/overlays/stable/kustomization.yaml#L7

PR #1053, opened 2021-09-10
Add changelog for the latest patch releases

ialidzhikov:
/kind documentation


ialidzhikov:
It seems that the same changes are already done with a72f36b566e3d77792f175a87116c9ad588d6713 and 3224d3ded539d6669ba21a04e20e6815e02a4c7f.

Issue #1048, opened 2021-09-08
[release-1.1]/[release-1.2] Prow GCR build is slow


wongma7:
/kind bug

**What happened?**

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/aws-ebs-csi-driver-push-images timeouts

For comparison, the github action builds take about 15 minutes.

Possible mitigations, they r not mutually exclusive
1. Stop building the debian variant, only build amazonlinux variant. This was planned for 1.3.x but we can "backport" this change to 1.1.x and 1.2.x branches too. It should halve the build time.
2. Replace GCR prow build with public ECR GitHub actions build. This way we will only need to maintain one build pipeline, github actions. I think we can't simply stop pushing to GCR* without first offering an alternative registry like public ECR that allows unauthenticated pulls (because it is useful for testing or for those who don't want to register/authenticate just to pull an image ) and that doesn't throttle (unlike Docker Hub), so if we want to stop pushing to GCR via Prow we will need to push to public ECR via GitHub actions . cc https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1011

**What you expected to happen?**

**How to reproduce it (as minimally and precisely as possible)?**

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

wongma7:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/aws-ebs-csi-driver-push-images/1435723992483762176

#21 [linux/arm64 builder 4/4] RUN make
#21 DONE 3424.7s 

takes over an hour on prow but ~10m on github.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/actions/runs/1215072462


#14 [linux/arm64 builder 4/4] RUN make
#14 DONE 564.3s 

Either way it is slow I think because we are building for arm using arm virtualization when we could be building for arm without virtualization (GOARCH)

wongma7:
Test started today at 10:40 AM passed after 1h4m6s. (more info) 

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/aws-ebs-csi-driver-push-images/1436021635202158592

PR #1062, opened 2021-09-15
[release-1.2] Build and push multi-arch/os (amazon and windows, no debian) image manifest via Make rules

wongma7:

**What is this PR about? / Why do we need it?** cherry-pick of https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/957 on to release-1.2 branch

The main motivation is to mitigate the slow builds in this branch https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1048 .  Building amazonlinux + windows manifest using this Makefile is faster than building amazonlinux + debian manifest using the old Makefile I suspect because the old one wastes a lot of time doing multi-platform binary builds while virtualizing the platforms whereas this one leverages --platform=$BUILDPLATFORM and GOOS to multi-platform binaries on x86.

**What testing is done?** I will check after this merges if the release-1.2 branch prow build finishes within an hour or not.


nckturner (approved):
I think getting rid of debian makes sense -- also is there anything holding us back from using a minimal/distroless image?

wongma7:
@nckturner we rely on stuff like mkfs.ext4 and such and so  we just install those using yum + amazonlinux base image. something like this https://github.com/aws/eks-distro-build-tooling where the minimum binaries/libs are copied from an amazonlinux base image in to a distroless image is possible but requires some effort.

PR #1069, opened 2021-09-22
Allow default fstype to be overriden via values.yaml

jcrsilva:

New feature.

Default `fstype` is hardcoded as ext4, users may wish to override it to use other filesystems instead of specifying them explicitly in the StorageClass resource.

**What testing is done?** 

Helm templating + test deploy on k8s cluster.

Change is backwards compatible, as the default is still `ext4`

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.3.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.2.1
+version: 2.3.0
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -68,6 +68,9 @@ controller:
   # If arbitrary args like "--aws-sdk-debug-log=true" need to be passed, use this value
   additionalArgs: []
   affinity: {}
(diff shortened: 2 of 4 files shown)


charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.3.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.2.1
+version: 2.2.2
```
krmichel, line 5: With a new feature please bump the version a minor.
```suggestion
version: 2.3.0
```

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -126,6 +126,9 @@ controller:
   #    topologyKey: kubernetes.io/hostname
   #    whenUnsatisfiable: ScheduleAnyway
   topologySpreadConstraints: []
+  # The default filesystem type of the volume to provision when fstype is unspecified in the StorageClass.
+  # If the default is not set and fstype is unset in the StorageClass, then no fstype will be set
+  defaultFsType: ext4
```
krmichel, line 131: Please move this to the correct alphabetical place in the list of values under controller.

charts/aws-ebs-csi-driver/Chart.yaml
jcrsilva: Done

charts/aws-ebs-csi-driver/values.yaml
jcrsilva: Done

jcrsilva:
Updated `Changelog.md`

krmichel (approved):

PR #1074, opened 2021-09-29
Refactor pkg/cloud/metadata.go into pkg/cloud/metadata_*.go files

wongma7:
**Is this a bug fix or adding new feature?** refactor

**What is this PR about? / Why do we need it?** I am planning to consume or copy pkg/cloud/metadata_* code from here to the EFS repo so I wanted to split up metadata.go and cloud_interface.go to make it easier. this is justa refactor for readability and easier consumption it should have no consequences on functionality 

**What testing is done?** make mockgen succeeded, will let CI test the rest


pkg/cloud/metadata_k8s.go
```diff
@@ -0,0 +1,83 @@
...
+	"os"
+	"regexp"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+type KubernetesAPIClient func() (kubernetes.Interface, error)
```
nckturner, line 15: nit: What's the reason for the metadata_ prefix for all files?  This seems to just be a Kubernetes client wrapper, maybe client.go or kubernetes.go makes for sense.

nckturner: Otherwise lgtm, will merge if you prefer that name.

wongma7: the main purpose is for  `KubernetesAPIInstanceInfo` to return a Metadata, i.e. how to retrieve Metadata/InstanceInfo from k8s versus from ec2. (and later, one to retieve from ECS for fargate environment). the function you are looking at is just wrapper stuff for testing, KubernetesAPIInstanceInfo is the important one.

nckturner: Yeah that makes sense and I see the reasoning for the similar file names, especially given they all are different mechanisms for finding the same information.

PR #1075, opened 2021-09-29
Added more volumeAttachLimit values based on instanceTypes

nirmalaagash:
Update in Volume Attach Limit - Fix

VolumeAttachLimit is updated. Different instances have different attachment limit and it is updated in this PR.
We need it because previously only two common values were used irrespective of the type of instance. Few instances allow only 19 or 11 volumeAttachLimit, but the current repo accepts limits upto 39 or 24. Hence it should be updated accordingly.

Reference: 
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances

**What testing is done?** 
Unit testing is done.
Checked it using csc tool. Below are the screenshots. When the CO calls the NodeGetInfo, the volumeAttachLimit is called on the node and the volumeAttachLimit is calculated based on the other attachments(ENI and BlockDevices attached) to the node.
Screen Shot 2021-09-29 at 2 07 05 PM
The result 37 because the node already has a ENI attached to it and two EBS root volume. The queried instance is t2.large.
Screen Shot 2021-09-29 at 2 06 54 PM

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
@@ -833,6 +834,57 @@ func (c *cloud) getVolume(ctx context.Context, request *ec2.DescribeVolumesInput
 	}
 
 	return volumes[0], nil
+}
+
+func (c *cloud) GetVolumeAttachLimit(ctx context.Context, nodeID string) (int, error) {
+	instance, err := c.getInstance(ctx, nodeID)
+	if err != nil {
+		return 0, err
+	}
+
+	instanceType := string(*instance.InstanceType)
+	networkInterfacesAttached := len(instance.NetworkInterfaces)
+	blockDevicesAttached := len(instance.BlockDeviceMappings)
+
+	nitroInstances := []string{"A1", "C5", "C5a", "C5ad", "C5d", "C5n", "C6g", "C6gd", "C6gn", "D3", "D3en", "G4", "I3en", "Inf1",
+		"M5", "M5a", "M5ad", "M5d", "M5dn", "M5n", "M5zn", "M6g", "M6gd", "M6i", "p3dn", "P4",
+		"R5", "R5a", "R5ad", "R5b", "R5d", "R5dn", "R5n", "R6g", "R6gd", "T3", "T3a", "T4g", "X2gd", "z1d"}
+
+	//Static values are taken from the https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#nitro-system-volume-limits
+	if instanceType == "d3.8xlarge" || instanceType == "d3en.12xlarge" {
+		return (3 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if instanceType == "inf1.xlarge" || instanceType == "inf1.2xlarge" {
+		return (26 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if instanceType == "inf1.6xlarge" {
+		return (23 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if instanceType == "inf1.24xlarge" {
+		return (11 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if instanceType == "mac1.metal" {
+		return (16 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if ok, _ := regexp.MatchString("^u-.*.metal$", instanceType); ok {
+		return (19 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if ok, _ := regexp.MatchString("^u-.*$", instanceType); ok {
+		return (27 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if ok, _ := regexp.MatchString("^.*.metal$", instanceType); ok {
+		return (31 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if ok := contains(nitroInstances, strings.ToLower(instanceType)); ok {
+		return (28 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else {
+		//For the Linux Instances, the number of attachments supported is 40. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#linux-specific-volume-limits
+		return (40 - networkInterfacesAttached - blockDevicesAttached), nil
+	}
+
+}
+
+func contains(arr []string, target string) bool {
+	for _, prefix := range arr {
+		prefix = strings.ToLower(prefix)
+		if strings.HasPrefix(target, prefix) {
+			return true
+		}
+	}
+
+	return false
 }
 
 func (c *cloud) getInstance(ctx context.Context, nodeID string) (*ec2.Instance, error) {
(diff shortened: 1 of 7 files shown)


pkg/cloud/cloud.go
```diff
@@ -835,6 +836,57 @@ func (c *cloud) getVolume(ctx context.Context, request *ec2.DescribeVolumesInput
...
 
+func (c *cloud) GetVolumeAttachLimit(ctx context.Context, nodeID string) (int, error) {
+	instance, err := c.getInstance(ctx, nodeID)
+	if err != nil {
+		return 0, err
+	}
+
+	instanceType := string(*instance.InstanceType)
+	networkInterfacesAttached := len(instance.NetworkInterfaces)
+	blockDevicesAttached := len(instance.BlockDeviceMappings)
```
wongma7, line 847: I believe (might need to verify) that nodegetinfo gets called once at driver startup when it registers with kubelet and then never again afterward. If this `blockDevicesAttached`  counts volumes attached by the driver itself, it could result inthe limit unexpectedly lowering in case the driver is restarted (e.g. if I upgrade it). For example this would cut the limit effectively in half because if the instance type has limit 10 , then it attaches 5 volumes, then I restart the driver (let's say I'm upgrading it) then this function will return 10-5 = 5 so it will refuse to attach more.

pkg/cloud/cloud.go
```diff
@@ -835,6 +836,57 @@ func (c *cloud) getVolume(ctx context.Context, request *ec2.DescribeVolumesInput
 	return volumes[0], nil
 }
 
+func (c *cloud) GetVolumeAttachLimit(ctx context.Context, nodeID string) (int, error) {
+	instance, err := c.getInstance(ctx, nodeID)
```
wongma7, line 840: this means every node needs describeinstances permission right? I don't think we explicitly grant these in our e2e test cases so i'm not sure how the tests are passing (I will check) . In any case it would mean additional requirement for node policy and api call load so needs to be documented . If the same info is available in instance metadata it should attempt to retrieve from there first

pkg/driver/node.go
```diff
@@ -678,16 +683,22 @@ func (d *nodeService) nodePublishVolumeForFileSystem(req *csi.NodePublishVolumeR
...
-	ebsNitroInstanceTypeRegex := "^[cmr]5.*|t3|z1d"
+
 	instanceType := d.metadata.GetInstanceType()
-	if ok, _ := regexp.MatchString(ebsNitroInstanceTypeRegex, instanceType); ok {
-		return defaultMaxEBSNitroVolumes
+	instanceID := d.metadata.GetInstanceID()
+
+	limit, err := d.cloud.GetVolumeAttachLimit(ctx, instanceID)
+	if err != nil {
+		//CO decides on the number of volume of the same type to be published.
```
wongma7, line 696: can you check if the e2e tests are currently doing this? by adding a log line here.

wongma7: I think they are doing this because I don't see the "The volumeAttachLimit for the instance type %q is %d" line being printed in the logs https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/1075/pull-aws-ebs-csi-driver-external-test-eks/1443331300164374528 and the verbosity is set to 5 https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/values.yaml#L4

pkg/cloud/cloud.go
```diff
@@ -835,6 +836,57 @@ func (c *cloud) getVolume(ctx context.Context, request *ec2.DescribeVolumesInput
...
+	} else if ok, _ := regexp.MatchString("^u-.*.metal$", instanceType); ok {
+		return (19 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if ok, _ := regexp.MatchString("^u-.*$", instanceType); ok {
+		return (27 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if ok, _ := regexp.MatchString("^.*.metal$", instanceType); ok {
+		return (31 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else if ok := contains(nitroInstances, strings.ToLower(instanceType)); ok {
+		return (28 - networkInterfacesAttached - blockDevicesAttached), nil
+	} else {
+		//For the Linux Instances, the number of attachments supported is 40. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#linux-specific-volume-limits
```
gnufied, line 873: @wongma7 is there a way to get this via APIs from AWS? it would be nice even if we have an internal ticket to implement this in future. :-)

wongma7: no idea, @nirmalaagash would know more than me actually. I imagine this info could be returned by https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTypes.html some day, they have MaximumNetworkInterfaces which is similar to what we want ...

nirmalaagash: @gnufied Right now, we don't have APIs to get this.

pkg/cloud/cloud.go
nirmalaagash: Metadata does not give enough information about the Network interfaces and the volumes attached. Hence Node needs describeInstances permission to look into the number of attachments.

pkg/cloud/cloud.go
nirmalaagash: Yes, you are right. NodeGetInfo is called once at driver startup. Like you said, if the driver is restarted, then it would reduce the number of attachments to its current number. Would you rather suggest a different approach to it?

wongma7: we would need a way to distinguish volumes attached by csi, which should not affect the limit, from volumes that are not, which should affect the limit, and I don't think describeinstances has enough info for that. I think we would need to call describeVolumes once with all the volumes and filter out the csi-specific tags? But that is another API call and I am not sure we should be making an API call in the first place.

pkg/cloud/cloud.go
wongma7: Then I think there needs to be a fallback in case the permission is not granted. Even if  we return the default limit without accounting for existing # of ENI / volumes, it is better than returning 0 jsut because the customer has not granted permission. Asking them for this permission is a breaking change and there are security reasons they will not want their nodes to be able to hit the aws api

wongma7: what about block-device-mapping/ebsN ? "The virtual devices associated with any Amazon EBS volumes. Amazon EBS volumes are only available in metadata if they were present at launch time or when the instance was last started. The N indicates the index of the Amazon EBS volume (such as ebs1 or ebs2). "

or network/interfaces/macs/mac/device-number for ENIS? https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html

ialidzhikov:
@nirmalaagash, can you rebase this PR?

stevehipwell:
It'd be great to get this rebased and reviewed as we're currently seeing the issue this resolves.

stevehipwell:
@nirmalaagash has this PR been made irrelevant by #1175 or is the dynamic lookup functionality in this PR still desirable?

rdpsin:
Closing this since it is already fixed in #1175.

PR #1078, opened 2021-10-05
Move mocks to parent package to avoid import cycle

wongma7:

I would like to move the cloud mocks to pkg/cloud from pkg/driver because they belong in pkg/cloud but that results in import cycle because mock_cloud.go depends on pkg/cloud e.g. for cloud.DiskOptions. If I recall correctly they are in pkg/driver as a hack to avoid this import cycle.
```
# github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud
package github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud (test)
	imports github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/mocks
	imports github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud: import cycle not allowed in test
```
It seems simplest to just get rid of the mocks package and let mocks live in same package as what they are mocking.

(This is part of some effort to reorganize pkg/cloud to be easily copyable/consumable from EFS repo, see also "Refactor pkg/cloud/metadata.go into pkg/cloud/metadata_*.go files" https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1074)
**What testing is done?**


nckturner:
I don't see any problem with this, other than a desire to keeping testing related resources separate from application logic.  Another note would be that this package seems general enough to consider moving it to cloud-provider-aws and importing it from there to both driver repositories.

PR #1106, opened 2021-11-04
Support for custom labels

nlamirault:

Add custom labels to metadata for all resources

**What testing is done?**

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/_helpers.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_helpers.tpl
@@ -41,7 +41,11 @@ helm.sh/chart: {{ include "aws-ebs-csi-driver.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
+app.kubernetes.io/component: csi-driver
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+{{- if .Values.customLabels }}
+{{ toYaml .Values.customLabels }}
 {{- end }}
 {{- end -}}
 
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -8,6 +8,10 @@ image:
   tag: ""
   pullPolicy: IfNotPresent
 
+# -- Custom labels to add into metadata
+customLabels: #{}
```
krmichel, line 12: I think custom labels should be empty for the default values for the chart

charts/aws-ebs-csi-driver/templates/_helpers.tpl
```diff
@@ -41,8 +41,12 @@ helm.sh/chart: {{ include "aws-ebs-csi-driver.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
+app.kubernetes.io/component: csi-driver
```
krmichel, line 44: I am not familiar enough with csi, in general, to know if `csi-driver` is a sufficient component name or if it should be `ebs-csi-driver`.  @wongma7 @ayberk thoughts?

krmichel (changes requested):

charts/aws-ebs-csi-driver/values.yaml
nlamirault: done

charts/aws-ebs-csi-driver/templates/_helpers.tpl
wongma7: I think it's OK, the example from https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ is `database` which is very generic. https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.21/hostpath/csi-hostpath-driverinfo.yaml#L9 also is setting it so lgtm

PR #1109, opened 2021-11-10
Update to go 1.17

bertinatto:
Since Kubernetes is now built with go 1.17 [1], do the same here.

[1] https://github.com/kubernetes/kubernetes/pull/106209

CC @jsafrane @wongma7


go.mod
```diff
@@ -26,6 +24,69 @@ require (
...
+	k8s.io/csi-translation-lib v0.21.0 // indirect
+	k8s.io/kube-openapi v0.0.0-20210305001622-591a79e4bda7 // indirect
+	k8s.io/kubectl v0.0.0 // indirect
+	k8s.io/kubelet v0.0.0 // indirect
+	k8s.io/legacy-cloud-providers v0.0.0 // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.15 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.1.0 // indirect
+	sigs.k8s.io/yaml v1.2.0 // indirect
+)
+
```
bertinatto, line 89: xref: https://golang.org/doc/go1.17#go-command
```
Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies. 
```

PR #1082, opened 2021-10-07
Search for nvme device path even if non-nvme exists

wongma7:
**Is this a bug fix or adding new feature?** /bug

**What is this PR about? / Why do we need it?** (intended to) fix https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1076 and maybe https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1027.

NodeStage is not idempotent for nvme volumes. This is most apparent in kubelet 1.20 because it is guaranteed to call NodeStage again for the same volume when it restarts. Then you will see the errors in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1076 because NodeStage compares the path /dev/xvbda to the path /dev/nvme1n1, when the intention is to compare /dev/nvme1n1 with /dev/nvme1n1 and return NodeStage as no-op success.

The fix is to ensure that the path we compare is the "canonical" one, so even if we see that /dev/xvbda exists we still need to search for the "canonical" nvme representation of the device because that is what will show up in the mount table. Then we are comparing what we find with what is in the mount table and it should match.

**What testing is done?** 
### e2e In CI:

I updated the test to create nitro c5 instances so this code will get exercised

### e2e on my cluster:

built and pushed an image to my registry
```
$ OUTPUT_TYPE=registry OS=linux TAG=oct7-1082-1 REGISTRY=x.dkr.ecr.us-west-2.amazonaws.com make image
```
restarted kubelet to repro the issue.
```
$ journalctl -ru kubelet -l | less
[...]

PR #1141, opened 2021-12-16
Fix canonical nvme device resolution in more cases

PaulFurtado:
Bug fix further improving on https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1082

- It allows the `/dev/xvdXX` symlinks to continue to function, but properly identifies them as nvme devices based on the `/dev/nvme` prefix
- No matter what kind of symlinking is setup, it will always find the canonical block device because all symlinks are always resolved now. This means that if some crazy non-nitro system had `/dev/xvdb` symlinked to `/dev/sdb`, even that would work correctly.
- nvme devices resolved via the `/dev/xvdXX` symlinks will now correctly handle appending the partition string
- The `/dev/disk/by-id` fallback continues to work correctly
- It fixes logging for `node_linux.go`: this file was using klog v2 while the rest of the repo was using v1 so its logging was going nowhere

**What testing is done?** 
- Tested on:
  - a kubelet with just the `/dev/xvdXX` udev rules for nvme devices
  - a kubelet with just the `/dev/disk/by-id` udev rules for nvme devices
  - a kubelet with both sets of udev rules present
- On each of those, I confirmed that:
  - pods with PVCs can be created and deleted successfully
  - when kubelet in 1.21 is restarted, the CSI driver correctly recognizes that the volumes are already mounted and the operation is idempotent

### DIFF
--- a/pkg/driver/node_linux.go
+++ b/pkg/driver/node_linux.go
@@ -27,8 +27,20 @@ import (
 	"strings"
 
 	"golang.org/x/sys/unix"
-	"k8s.io/klog/v2"
+	"k8s.io/klog"
 )
+
+func (d *nodeService) appendPartition(devicePath, partition string) string {
+	if partition == "" {
+		return devicePath
+	}
+
+	if strings.HasPrefix(devicePath, "/dev/nvme") {
+		return devicePath + nvmeDiskPartitionSuffix + partition
+	}
+
+	return devicePath + diskPartitionSuffix + partition
+}
 
 // findDevicePath finds path of device and verifies its existence
 // if the device is not nvme, return the path directly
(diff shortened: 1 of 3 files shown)


wongma7:
thanks, this makes sense to me at a glance. from what i can tell resolving symlinks is exactly what linux means "canonicalizing" a path is before putting entry in mount table https://github.com/util-linux/util-linux/blob/441f9b9303d015f1777aec7168807d58feacca31/libmount/src/cache.c https://github.com/util-linux/util-linux/blob/441f9b9303d015f1777aec7168807d58feacca31/lib/canonicalize.c#L121 so it follows that if we are trying to canonicalize device path to compare it with what's in mount table https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/5d782870e3ee2e56e36202362d7b65d54ce8f866/pkg/driver/node.go#L200-L203 we should follow symlinks all the way, not make any assumptions about the depth of symlinks or such which could break at any time with a EC2 new instance type etc.

pkg/driver/node_linux.go
```diff
@@ -48,12 +60,26 @@ func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (st
 	}
 
 	if exists {
```
wongma7, line 62: should we still try to fallback to the findNvmeVolume method in case there is an error in this block (if lstat or evalsymlinks somehow fails)? Is there any situation where we receive a devicePath that exists but that could not be lstatted or evalsymlink'd? If not then it is not necessary to fallback and this PR is good as-is

PaulFurtado: @wongma7 I cannot think of a situation where that could be possible.

PaulFurtado:
Also worth noting: we have been running with this patch in our clusters internally and it has eliminated all of the EBS attach/detach issues we've had since upgrading kubernetes from 1.19 to 1.21.

wongma7:
/lgtm
/approve

thanks

PR #1145, opened 2021-12-16
Fix the wrong app and helm chart version

gtxu:

**What testing is done?**

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: 1.6.0
+appVersion: 1.5.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.6.0
+version: 2.6.1
 kubeVersion: ">=1.17.0-0"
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: 1.6.0
+appVersion: 1.5.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.6.0
+version: 2.5.1
```
krmichel, line 5: You need to increase the helm chart version to 2.6.1.  Decreasing it will still have the latest version be broken.

krmichel (changes requested):

charts/aws-ebs-csi-driver/Chart.yaml
gtxu: Working on it

krmichel (approved):

AndyXiangLi:
Can you fix the commits issue? There should be only one commit in this PR.

AndyXiangLi:
/lgtm
Thanks!

phantasm66:
Thanks guys!

PR #1149, opened 2021-12-22
Updated deprecated exec command in READMEs

mathisve:
This is a slight fix in the `examples/kubernetes` READMEs

In some of the documentation, you are told to execute a command the format of which is deprecated. This gives you the following warning:
`kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.`

With my changes, it no longer uses the deprecated way of executing commands in pod.

**What testing is done?** 

As you can see by the following snippet, it no longer shows the **deprecated** warning when using the double dashes in the exec command.
```
$ kubectl run --image=nginx my-app
pod/my-app created
$ kubectl exec -it my-app echo "test"
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
test
$ kubectl exec -it my-app -- echo "test"
test
```

Signed-off-by: Mathis Van Eetvelde <mathis.vaneetvelde@protonmail.com>

### DIFF
--- a/examples/kubernetes/block-volume/README.md
+++ b/examples/kubernetes/block-volume/README.md
@@ -25,7 +25,7 @@ app    1/1     Running   0          16m
 Verify the device node is mounted inside the container:
 
 ```sh
-$ kubectl exec -ti app -- ls -al /dev/xvda
+$ kubectl exec -it app -- ls -al /dev/xvda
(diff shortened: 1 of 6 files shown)


torredil:
Closing as this has already been addressed.

PR #1157, opened 2022-01-20
Allow controller to use metadata service to fetch default AZ when not specified

hntrmrrs:

New feature (specify default AZ behaviour) along with some bugfixes and clarified log messages

If `zone` is unspecified in `VolumeOptions`, the default behaviour chooses a random availability zone. In practice, it is always the first availability zone listed (almost always(?) `<region>a`). This PR allows the controller operator to specify a flag which allows the default availability zone to be retrieved from the metadata service (as the region is).

**What testing is done?** 

Unit testing and some end-to-end testing in a staging environment on AWS.


hntrmrrs:
I'm having some trouble with the new LFX platform for the CLA, but hopefully it will get resolved quickly as I've signed the agreement on behalf of my organisation.

hntrmrrs:
I signed it

PR #1169, opened 2022-02-16
Update cluster role configuration for external snapshotter

gtxu:
Remove un-necessary access to secrets by the external-snapshotter-role

External-snapshoter does not require access to secrets for its functionality, keep the access to all secret values may cause other security issues.


AndyXiangLi:
Can you add comment there and saying why it is not needed now?

gtxu:
> Can you add comment there and saying why it is not needed now?
Done.

AndyXiangLi:
/approve
/lgtm
Thanks!

PR #1171, opened 2022-02-17
Update helm charts for a secrect exposure

gtxu:
Removed external-snapshot-role gain u-necessary access to secrets

### DIFF
--- a/CHANGELOG-0.x.md
+++ b/CHANGELOG-0.x.md
@@ -1,6 +1,6 @@
 # v1.5.1
 ## Notable changes
-* Address CVE ALAS-2021-1552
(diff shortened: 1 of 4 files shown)


AndyXiangLi:
/approve
/lgtm
thanks

stevehipwell:
@AndyXiangLi if access to secrets might still be required shouldn't this removal have been a conditional switch rather than requiring manual changes to the yaml?

gtxu:
This will not require customers to modify it. Leaving this comment for reference if the our snapshot future feature that requires secret access. i.e. customers use other credentials to encrypt snapshot volume. Currently, the driver don't use any secret, so do the snapshotter. So we comment this part.

stevehipwell:
Thanks @gtxu I think the comment confused me reading as a user, but reading as a maintainer it makes sense.

Issue #1139, opened 2021-12-10
NodeGetInfo: MaxVolumesPerNode is wrong for a lot of nitro instance types


ialidzhikov:
/sig storage
/kind bug
/area provider/aws
/priority critical-urgent

**What happened?**
The current handling for volume limits is copied from the in-tree plugin:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/f298d57e290a73f355fd96b96c3bc34385877012/pkg/driver/node.go#L691-L702

The problems I see with the current handling:
- The nitro instance types regex `^[cmr]5.*|t3|z1d` is not good enough to cover all nitro instance types.
From the [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html):
```
The following instances are built on the Nitro System:

Virtualized: A1, C5, C5a, C5ad, C5d, C5n, C6g, C6gd, C6gn, C6i, D3, D3en, DL1, G4, G4ad, G5, G5g, I3en, Im4gn, Inf1, Is4gen, M5, M5a, M5ad, M5d, M5dn, M5n, M5zn, M6a, M6g, M6gd, M6i, p3dn.24xlarge, P4, R5, R5a, R5ad, R5b, R5d, R5dn, R5n, R6g, R6gd, R6i, T3, T3a, T4g, high memory (u-*), VT1, X2gd, and z1d

Bare metal: a1.metal, c5.metal, c5d.metal, c5n.metal, c6g.metal, c6gd.metal, i3.metal, i3en.metal, m5.metal, m5d.metal, m5dn.metal, m5n.metal, m5zn.metal, m6g.metal, m6gd.metal, mac1.metal, r5.metal, r5b.metal, r5d.metal, r5dn.metal, r5n.metal, r6g.metal, r6gd.metal, r6i.metal, u-6tb1.metal, u-9tb1.metal, u-12tb1.metal, u-18tb1.metal, u-24tb1.metal, x2gd.metal, and z1d.metal
```

I will leave to the reader to discover how many of these instance types are not covered by the current regex.

- aws-ebs-csi-driver does not consider attached network interfaces or instance store volumes:
From [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#nitro-system-volume-limits):
```
[...]

PR #1175, opened 2022-02-25
Fix nodeService.getVolumesLimit() adding more instance types

rdpsin:
Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1139

This commit fixes the `getVolumesLimit` function.

Previously, the volume limit for Nitro instances was based off of a
regex. This regex is insufficient for the wide variety of Nitro
instances that AWS supports today.

Nitro instances share attachments with ENIs and NVMe instance
stores.

The number of attached ENIs is queried through IMDS, against the
`network/interfaces/macs` endpoint.

The number of already attached block devices is also queried through
IMDS, against the `block-device-mapping` endpoint.

Currently, IMDS does not provide a way to query NVMe instance store
volumes. So, all instances that support NVMe instance stores are listed
in `pkg/cloud/volume_limits.go`

This commit also added more tests to cover various scenarios.

**What testing is done?** 

Start a cluster with `m6i.8xlarge` worker node.

CSINode before the PR: 

```
  spec:
    drivers:
    - allocatable:
        count: 39
      name: ebs.csi.aws.com
```

CSINode after the PR (with 2 attached ENIs and 1 root block device):

```
spec:
    drivers:
    - allocatable:
        count: 25
      name: ebs.csi.aws.com
```

Checked it against various other instance types too.

### DIFF
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -1,0 +1,283 @@
+package cloud
+
+import (
+	"regexp"
+	"strings"
+)
+
+/// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances
+const (
+	highMemoryMetalInstancesMaxVolumes   = 19
+	highMemoryVirtualInstancesMaxVolumes = 27
+	baremetalMaxVolumes                  = 31
+	nonNitroMaxAttachments               = 39
+	nitroMaxAttachments                  = 28
+)
+
+/// It is possible to have an instance family where the virtualized instances are Nitro
+/// and metal instances are not
+var nonNitroInstances = map[string]struct{}{
+	"c6i.metal": {},
+	"g5g.metal": {},
+}
+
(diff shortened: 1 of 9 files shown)


rdpsin:
PTAL @wongma7

pkg/cloud/volume_limits.go
```diff
@@ -0,0 +1,282 @@
...
+	if v, ok := nvmeInstanceStoreVolumes[it]; ok {
+		return v
+	}
+	return 0
+}
+
+/// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes
+/// IMDS does not provide NVMe instance store data; we'll just list all instances here
+var nvmeInstanceStoreVolumes = map[string]int{
+	"c5ad.large":     1,
```
AndyXiangLi, line 117: Can we get these config from describeInstanceType API instead of hard code the value?

rdpsin: Yeah, it's an ugly solution, but I didn't want to make another API call just to get that information. It's possible we can do it just once at the start, but we would also need to update the policy if we want to do that.

AntonioDiTuri:
On what version was this fix added? I cannot see the issue #1175 in the changelog.md file

PR #1187, opened 2022-03-15
Update to kops 1.23.0

wongma7:
**Is this a bug fix or adding new feature?** routine update

I am not updating K8S_VERSION yet (so test clusters will still be 1.20.8 which is outdated) but just want to make sure updating kops by itself works.

**What testing is done?**  CI will exercise

### DIFF
--- a/hack/e2e/kops.sh
+++ b/hack/e2e/kops.sh
@@ -10,11 +10,16 @@ source "${BASE_DIR}"/util.sh
 function kops_install() {
   INSTALL_PATH=${1}
   KOPS_VERSION=${2}
-  if [[ ! -e ${INSTALL_PATH}/kops ]]; then
-    KOPS_DOWNLOAD_URL=https://github.com/kubernetes/kops/releases/download/v${KOPS_VERSION}/kops-${OS_ARCH}
-    curl -L -X GET "${KOPS_DOWNLOAD_URL}" -o "${INSTALL_PATH}"/kops
-    chmod +x "${INSTALL_PATH}"/kops
+  if [[ -e "${INSTALL_PATH}"/kops ]]; then
+    INSTALLED_KOPS_VERSION=$("${INSTALL_PATH}"/kops version)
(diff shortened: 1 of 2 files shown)


olemarkus:
We'll run into an issue where kOps unconditionally installs the CSI driver. We should look into how the addon can be sufficiently configured for the testing to keep working.

/remove-lifecycle stale

ConnorJC3:
You can disable installing the driver with kops, just merge this into the `hack/kops-patch.yaml`:

```yaml
spec:
  cloudConfig:
    awsEBSCSIDriver:
      enabled: false
```

Ref in kops doc: https://kops.sigs.k8s.io/addons/#snapshot-controller

olemarkus:
This also disables the feature flags that should be set when using the CSI driver. Besides, it is not really given that just disabling the CSI driver like that will work. Hence the suggestion to ensure the pipeline can configure kops sufficiently (this is also how AWS CCM is tested).

ConnorJC3:
The [existing kops-patch.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/kops-patch.yaml) already has the needed feature flags from before kops even supported installing the driver, so no problems there.

I looked into how the CCM does it but they seem to do it effectively the same way: https://github.com/kubernetes/cloud-provider-aws/blob/b9c8d82db9edc3be45f82f8d5320d006c62aa50c/hack/e2e/run.sh#L138 - rather than using a file they manually specify parameters for `kops create cluster`, but the principle is the same.

olemarkus:
True. The feature flags should be taken care of. 

Yeah so CCM does this `--override=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}` rather than using helm. So kOps is aware of CCM.

In k8s 1.25, one can't disable EBS CSI Driver (given that the KEP goes through) so that's my main worry with using the helm approach. But for k8s < 1.25 this should work.

Issue #1184, opened 2022-03-08
Upgrade PDB to `policy/v1`


zerkms:
**Is your feature request related to a problem?/Why is this needed**

/feature

```
policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+
```

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/de1c6569f054c844cdc25bf45b59dc83dc0e437b/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml

**Describe the solution you'd like in detail**

**Describe alternatives you've considered**

**Additional context**

PR #1196, opened 2022-04-01
Upgrade PodDisruptionBudget api version for kubernetes 1.21+

wangshu3000:
This is an API upgrade PR.

This is a PR to upgrade PDB api version for kubernetes 1.21+
The policy/v1beta1 api version will not be supported after kubernetes 1.25+. So we need to upgrade this version.
Use policy/v1 for k8s 1.21+, to avoid displaying any warning of deprecated api version.

This PR is related to below issue:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1184

**What testing is done?** 
Tested on different kubernetes version in minikube.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
@@ -1,4 +1,8 @@
+{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
 kind: PodDisruptionBudget
 metadata:
   name: ebs-csi-controller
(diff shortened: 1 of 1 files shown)


gtxu:
Hi @wangshu3000 would you please update your CLA?

wangshu3000:
> Hi @wangshu3000 Would you please update your CLA? Thanks！

@gtxu  Just updated CLAs. Thanks.

zerkms:
The suggested implementation is not portable, instead comparisons with api versions should be used:

```
{{ if $.Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" -}}
```

wangshu3000:
i check k8s version, because of new version of helm, if people use helm template to generate yaml offline. The APIVersion.Has can not return the accuate api version, this happend in other repo previously, if it's ok for this repo  then let me update it, thanks.

zerkms:
It's the opposite: `helm template` DOES NOT have reliable version available, but you may have capabilities mocked: https://helm.sh/docs/helm/helm_template/

wangshu3000:
> It's the opposite: `helm template` DOES NOT have reliable version available, but you may have capabilities mocked: https://helm.sh/docs/helm/helm_template/
> 
> Also for example look at
> 
> https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/705b545c8818a8f0699a3315838d8f39d5633e1a/charts/aws-ebs-csi-driver/templates/controller.yaml#L196

Aha. You are right, i remember it wrongly.  Just updated, thanks for pointing it out.

Issue #1182, opened 2022-03-03
Chart fails to build if a tag value in controller.extraVolumeTags contains a ":"


Kaezon:
/kind bug

**What happened?**
Trying to build the Chart with the following values file with this values file:
```
controller:
  extraVolumeTags:
    my-tag: "me: myself and I"
    my-other-tag: "something else"
```

Presents the following error:
```
INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].args[2]): invalid type for io.k8s.api.core.v1.Container.args: got "map", expected "string"
```

**What you expected to happen?**
I expect the chart to build, and for the controller to be configured with the extra volume tags.

**How to reproduce it (as minimally and precisely as possible)?**
Add any extra tags which are set with a value containing ":"

**Anything else we need to know?**:
It seems like the problem is in the processing of extraVolumeTags here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/de1c6569f054c844cdc25bf45b59dc83dc0e437b/charts/aws-ebs-csi-driver/templates/_helpers.tpl#L71

The arg string is unquoted, so the ":" ends up turning the string into a map after templating.

Changing the line to `{{- printf "- \"--extra-tags=%s\"" (join "," $result.pairs) -}}` fixed the problem in my experiments.
I'm planning on making a PR with the fix, but I wanted to open this to feedback first.

**Environment**
- Kubernetes version (use `kubectl version`):
  - Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:48:33Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23+", GitVersion:"v1.23.3-2+d441060727c463", GitCommit:"d441060727c4632b67d09c9118a36a8590308676", GitTreeState:"clean", BuildDate:"2022-01-26T21:57:05Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"linux/amd64"}
- Driver version:
  - master/v1.5.1

Kaezon:
Fix was merged

PR #1198, opened 2022-04-05
1182 quote extra tags arg

Kaezon:
Bug fix for #1182

This fix add quotes around the extra-tags argument in order to prevent special characters such as ":" from breaking the manifest YAML after template rendering.

**What testing is done?**
Manual test with "helm template" to validate render output, as well as a live deployment to make sure a deployment would run with the offending tag value.


Kaezon:
Trying to get the CLA taken care of ASAP

torredil:
Thank you for your contribution :)
/lgtm
/hold

cc @gtxu @rdpsin

torredil (approved):

PR #1205, opened 2022-04-11
Ensure image OCI compliance

torredil:
- Bug fix

- The `os.version` is being incorrectly set which makes the image not OCI compliant and thus incompatible with containerd. The `os.version` on Windows should be a 4-part string similar to 10.0.14393.1066. Reference: https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/config.go#L100
- The current annotate overrides base image values, unnecessarily.

**What testing is done?** 
- Built images and verified description is correct with `docker manifest inspect`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,8 @@ all: all-image-docker
 # Builds all linux and windows images and pushes them
 all-push: all-image-registry push-manifest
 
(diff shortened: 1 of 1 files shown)


AndyXiangLi:
/approve
/lgtm
Thanks!

PR #1206, opened 2022-04-12
Add namespace for namespaced resources in helm chart

hanlins:
It's a bug fix.

Some namespaced resources has namespace
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/0fbd9f793c2da68abcdce991e045c0d72675e0e8/charts/aws-ebs-csi-driver/templates/clusterrolebinding-attacher.yaml#L11
where other doesn't
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/0fbd9f793c2da68abcdce991e045c0d72675e0e8/charts/aws-ebs-csi-driver/templates/node.yaml#L2-L6

Those resources without namespace are created in the default namespace where other resources are created in release namespace. This caused confusions when the `default` namespace is missing where helm says the resources is supposed to be deployed in a different namespace. This is what I'm seeing from helm outputs:
```
[ root@debug-55595fdf68-lxhfp:/etc/nginx ]$ HELM_DRIVER=configmap helm list -A
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /etc/nginx/config2-2
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /etc/nginx/config2-2
NAME          	NAMESPACE   	REVISION	UPDATED                                	STATUS  	CHART                     	APP VERSION
aws-pca-issuer	cert-manager	1       	2022-04-07 01:25:16.129127964 +0000 UTC	deployed	aws-privateca-issuer-1.2.0	1.2.0
cert-manager  	cert-manager	1       	2022-04-07 01:25:12.755634499 +0000 UTC	deployed	cert-manager-v1.6.1       	v1.6.1
csi           	kube-system 	1       	2022-04-07 01:25:07.452054908 +0000 UTC	failed  	aws-ebs-csi-driver-2.4.0  	1.4.0
proxy         	tidb-admin  	1       	2022-04-07 01:25:20.476145199 +0000 UTC	deployed	cluster-proxy-0.1.0       	1.2.3
tidb-operator 	tidb-admin  	1       	2022-04-07 01:25:18.871475022 +0000 UTC	deployed	tidb-operator-v1-canary   	v1-canary
```
However the resources are deployed in a different namespace:
```
[ root@debug-55595fdf68-lxhfp:/etc/nginx ]$ kubectl get deploy,ds -A | grep csi
infra          deployment.apps/ebs-csi-controller                    2/2     2            2           48m
infra         daemonset.apps/ebs-csi-node   2         2         2       2            2           kubernetes.io/os=linux   48m
```
This is confusing when I initially haven't create `infra` namespace. The chart failed during install due to the missing of namespace where helm shows the namespace is supposed to be `kube-system` which already exists.

**What testing is done?** 
Manual test is done. Now the resources are all created correctly in `kube-system` namespace in my test environment.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -3,6 +3,7 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: ebs-csi-controller
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "aws-ebs-csi-driver.labels" . | nindent 4 }}
 spec:
--- a/charts/aws-ebs-csi-driver/templates/node-windows.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node-windows.yaml
@@ -3,6 +3,7 @@ kind: DaemonSet
 apiVersion: apps/v1
 metadata:
   name: ebs-csi-node-windows
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "aws-ebs-csi-driver.labels" . | nindent 4 }}
 spec:
--- a/charts/aws-ebs-csi-driver/templates/node.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node.yaml
@@ -3,6 +3,7 @@ kind: DaemonSet
 apiVersion: apps/v1
 metadata:
   name: ebs-csi-node
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "aws-ebs-csi-driver.labels" . | nindent 4 }}
 spec:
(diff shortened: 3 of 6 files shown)


stevehipwell:
I'm not a maintainer here, but as a previously outspoken critic of adding this pattern to the charts I maintain due to it not being "idiomatic", I've realised the error of my ways and am now pushing to make it the idiom. As soon as your chart requires knowledge of namespaces outside the resource metadata, for example RBAC, the namespace needs to be set in some resource specs. If the namespace isn't being set in the resource metadata explicitly there is the potential for ambiguity and if you are using `helm template` to generate manifests it can actually break.

/approve

gtxu:
Thanks @hanlins ! 
/lgtm
/approve

PR #1210, opened 2022-04-13
Update golang.org/x/crypto for CVE-2022-27191

jsafrane:
`golang.org/x/crypto/ssh` server can crash in certain conditions. This CSI driver most probably does not run any ssh server, but update the package to a safe version to be 100% sure and prevent dumb dep scanning tools from raising an issue. Using today's golang.org master, they don't use tags.

Updated using these commands, they updated few additional libs:

```
go get -u golang.org/x/crypto
go mod tidy
go mod vendor
```


jsafrane:
Added a second commit bumping go image version to 1.17.

gtxu:
Hi @jsafrane , thank you for the contribution. This PR need rebase, we are awaiting it to merge.

PR #1212, opened 2022-04-15
Driver v1.5.3 post-release

torredil:

**What testing is done?**

### DIFF
--- a/deploy/kubernetes/base/controller.yaml
+++ b/deploy/kubernetes/base/controller.yaml
@@ -31,7 +31,7 @@ spec:
           tolerationSeconds: 300
       containers:
         - name: ebs-plugin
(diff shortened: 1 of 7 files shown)


deploy/kubernetes/base/controller.yaml
```diff
@@ -31,7 +31,7 @@ spec:
           tolerationSeconds: 300
       containers:
         - name: ebs-plugin
-          image: k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v1.5.1
+          image: k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v1.5.3
```
AndyXiangLi, line 34: Shall we use ecr for kustomize as well?

PR #1231, opened 2022-05-09
Replace Windows 2004(EOL) with ltsc2019

torredil:

- Stop building for Windows 2004 which has reached EOL for all editions.
- Build ltsc2019.

**What testing is done?**

- Deployed CSI driver with helm chart using new ltsc2019 image.


rdpsin:
Have we verified functionality with the new image? I don't expect there to be a problem, but don't want any surprises.

Issue #1235, opened 2022-05-12
IOPS parameters is not picked up by driver without defining volume type


651juan:
/kind bug

**What happened?**
When not defining the type parameter explicitly, the IOPS parameter is not sent when a volume is created. 

As you can see the IOPS is not present in the request from the driver
**What you expected to happen?**

I would expect it to have the IOPS parameter explicitly defined with the type's default value. 
**How to reproduce it (as minimally and precisely as possible)?**
I am using the aws-ebs-csi-driver helm chart with the following values: 
```
storageClasses:
- name: high-iops-ebs-sc
  parameters:
    iops: "6000"
```
The storage class is created with the correct parameters: 
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    meta.helm.sh/release-name: aws-ebs-csi-driver
    meta.helm.sh/release-namespace: trader
  creationTimestamp: "2022-05-11T16:49:53Z"
  labels:
    app.kubernetes.io/managed-by: Helm
  name: high-iops-ebs-sc
  resourceVersion: "24843"
  uid: 05e5b0ff-b001-4c5c-b335-6b15dff5bb5c
parameters:
  iops: "6000"
provisioner: ebs.csi.aws.com
```
The EBS volumes created through the driver are not created with this IOPS parameter. Looking at CloudTrail the follow request is sent: 
```
    "requestParameters": {
        "size": "10",
        "zone": "eu-central-1a",
        "volumeType": "gp3",
        "encrypted": false,
}
```
When I add the type to the helm chart values parameters as follows: 
```
storageClasses:
- name: high-iops-ebs-sc
  parameters:
    type: "gp3"
    iops: "6000"
```
The request in cloud trail is correct with the correct IOPS.
**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): 1.21
- Helm Chart version: 2.6.4

651juan:
#783 seems to be similar but was never addressed.

PR #1236, opened 2022-05-13
Fix IOPS parameter bug when no volume type is defined

torredil:

- Bug Fix: #1235 

- The IOPS parameter, defined in the StorageClass.parameters, is not picked up by the driver when the volume type is not defined (i.e. defaults to gp3). 

### Steps to Reproduce

- Create a StorageClass without a volume type but with IOPS

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: test
parameters:
  iops: "6000"
provisioner: ebs.csi.aws.com
```

- Provision a volume with this StorageClass.

```
k apply -f claim.yaml
```

- You'll see that the IOPS values is not honored in the volume created. 

```
 % aws ec2 describe-volumes --volume-id vol-085b747b34dd7f330 --query 'Volumes[0].[Iops, VolumeType]'
[
    3000,
    "gp3"
]
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -299,6 +299,8 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		throughput = int64(diskOptions.Throughput)
 	case "":
 		createType = DefaultVolumeType
+		iops = int64(diskOptions.IOPS)
+		throughput = int64(diskOptions.Throughput)
 	default:
 		return nil, fmt.Errorf("invalid AWS VolumeType %q", diskOptions.VolumeType)
 	}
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -299,6 +299,8 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		throughput = int64(diskOptions.Throughput)
 	case "":
```
rdpsin, line 300: This can be merged with `case VolumeTypeGP3`

gtxu: Here we have `DefaultVolumeType` for "" (instead of `gp3`), and we can change the default volume type at L114. I don't think it can be merged to `case VolumeTypeGP3`. And the `default:` here handles all the invalid inputs.

PR #1260, opened 2022-06-02
Bump up sidecar versions

gtxu:

bump up the following sidecars:
external-attacher v3.1.0 -> v3.4.0
external-resizer v1.1.0 -> v1.4.0
external-provisioner v2.2.2 -> v3.1.0
node-driver-registrar v2.3.0 -> v2.5.1
livenessprobe v2.4.0 -> v2.5.0

re-generate kustomize profile

**What testing is done?** 
Unit test and functionality test

### DIFF
--- a/deploy/kubernetes/base/controller.yaml
+++ b/deploy/kubernetes/base/controller.yaml
@@ -81,7 +81,7 @@ spec:
             periodSeconds: 10
             failureThreshold: 5
         - name: csi-provisioner
-          image: k8s.gcr.io/sig-storage/csi-provisioner:v2.2.2
+          image: k8s.gcr.io/sig-storage/csi-provisioner:v3.1.0
           imagePullPolicy: IfNotPresent
           args:
             - --csi-address=$(ADDRESS)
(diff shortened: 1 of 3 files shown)


Hokwang:
please bump up helm chart version

deploy/kubernetes/base/controller.yaml
```diff
@@ -5,6 +5,7 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: ebs-csi-controller
+  namespace: default
```
torredil, line 8: What is the reason for explicitly adding this? By default, all resources are created in the default namespace, unless a namespace is specified.

torredil:
@Hokwang Bumping up helm chart in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1262

deploy/kubernetes/base/controller.yaml
gtxu: Yes, this is generated by template `namespace: {{ .Release.Namespace }}` in helm template. It should be exclude to kustomize profile. WIP address it.

PR #1276, opened 2022-06-15
Node driver registrar socket path fix

### LINKED ISSUE #1274: kubeletPath needs to be universally used
/kind bug

**What happened?**
/var/lib/kubelet is used when kubeletPath is set to something else.

For the kubeletPath to be consistently used everywhere

Install aws-ebs-csi-driver in k0s.

k0s uses /var/lib/k0s/kubelet instead of /var/lib/kubelet

**Environment**
1.23.6

- Driver version:
1.6.2

makhov:

Fixes #1274.

The current implementation doesn't respect `.Values.node.kubeletPath` and doesn't work in case kubelet uses a non-default directory.

Related to https://github.com/k0sproject/k0s/issues/1842

**What testing is done?** 

Manual testing

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/node.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node.yaml
@@ -84,7 +84,7 @@ spec:
             {{- end }}
           volumeMounts:
             - name: kubelet-dir
-              mountPath: /var/lib/kubelet
+              mountPath: {{ .Values.node.kubeletPath }}
               mountPropagation: "Bidirectional"
             - name: plugin-dir
               mountPath: /csi
--- a/charts/aws-ebs-csi-driver/templates/node.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node.yaml
@@ -117,7 +117,7 @@ spec:
             - name: ADDRESS
               value: /csi/csi.sock
             - name: DRIVER_REG_SOCK_PATH
-              value: /var/lib/kubelet/plugins/ebs.csi.aws.com/csi.sock
+              value: {{ printf "%s/plugins/ebs.csi.aws.com/csi.sock" (trimSuffix "/" .Values.node.kubeletPath) }}
             {{- if .Values.proxy.http_proxy }}
             {{- include "aws-ebs-csi-driver.http-proxy" . | nindent 12 }}
             {{- end }}


charts/aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -117,7 +117,7 @@ spec:
             - name: ADDRESS
               value: /csi/csi.sock
             - name: DRIVER_REG_SOCK_PATH
-              value: /var/lib/kubelet/plugins/ebs.csi.aws.com/csi.sock
+              value: {{ printf "%s/plugins/ebs.csi.aws.com/csi.sock" (trimSuffix "/" .Values.node.kubeletPath) }}
```
torredil, line 120: Hi @makhov, thank you for your contribution! 
To address #1274, can you please include this change in [charts/aws-ebs-csi-driver/templates/node.yaml](charts/aws-ebs-csi-driver/templates/node.yaml) as well? Specifically, in [volumeMounts](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b05ed2d4fc67e7986cbd026c448b349911908af4/charts/aws-ebs-csi-driver/templates/node.yaml#L87) and [volumes](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/430194761f903a21097ea01d0a9ca21b34d99ed6/deploy/kubernetes/base/node.yaml#L110).

torredil (changes requested):

charts/aws-ebs-csi-driver/templates/node.yaml
makhov: Volumes and mounts are correct. `DRIVER_REG_SOCK_PATH` should contain the path on the host, not on the container.

> `--kubelet-registration-path`: This is the path to the CSI driver socket on the host node that kubelet will use to issue CSI operations

https://github.com/kubernetes-csi/node-driver-registrar#required-arguments

torredil: That fix is not to deal with the registry socket - because, as you pointed out, that path is relative to the host. The issue is that the kubelet attempts to mount volumes inside its own directory. Thus, with this fix the driver will be able to start up but will not mount volumes correctly because the kubelet will ask for volume mounts at (for example) `/var/lib/k0s/kubelet/volumes/...`, but the driver will have `/var/lib/kubelet` mounted instead of `/var/lib/k0s/kubelet`. Both changes are needed to properly support a modified kubeletPath.

makhov: Ok, now I see the point: kubelet sends a request to the driver containing the real path on the host node, so we need to mount it into the pod with the proper path.

Thanks! I fixed the volumeMount, volume [was already fixed](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/69e98ed251c0b12cb0f3c2108680a3d5bed74948/charts/aws-ebs-csi-driver/templates/node.yaml#L157).

Issue #1248, opened 2022-05-23
csi-resizer getting OOMKilled


bwmetcalf:
/kind bug

**What happened?**
We are using version v1.6.0-eksbuild.1 of the aws-ebs-csi-driver addon.  We have one environment where the csi-resizer container is using substantially more memory and getting OOMKilled.  Since the memory limits setting cannot be controlled with the addon, we have no way of increasing it.  We can edit the deployment live, but it gets reverted by EKS.  However, the real issue is that this container is using so much memory:
```
ebs-csi-controller-748586d5cc-9fbr5            csi-resizer                    13m          148Mi
```
Our other environments looks like
```
ebs-csi-controller-5458b96c8d-bmgmf             csi-resizer                    1m           21Mi
```
The memory settings for csi-resizer are
```
    Limits:
      cpu:     100m
      memory:  128Mi
    Requests:
      cpu:     10m
      memory:  40Mi
```
which is baked into the chart.

**What you expected to happen?**
csi-resizer should not be using so much memory.

**How to reproduce it (as minimally and precisely as possible)?**
This is unclear, but would welcome pointers on how to troubleshoot.

**Anything else we need to know?**:
This is occurring even though we aren't resizing and volumes.

**Environment**
- Kubernetes version (use `kubectl version`): 1.22
- Driver version: v1.6.0-eksbuild.1

gtxu:
Hi @bwmetcalf, the usage exceeded 128Mi is abnormal, can you specify the "other environments" and the environment that the issue occurs? Plus can you provide more details on the issue container by retrieve the log (`kubectl logs ebs-csi-controller-748586d5cc-9fbr5 -c csi-resizer -n kube-system`), if it convenient to you, thanks!

bwmetcalf:
Thanks @gtxu.  Here are the logs from one of the killed pods
```
$ kubectl logs ebs-csi-controller-6588dc7455-hzgdw -c csi-resizer -n kube-system
I0523 17:14:44.388430       1 main.go:83] Version : v
I0523 17:14:44.487283       1 connection.go:153] Connecting to unix:///var/lib/csi/sockets/pluginproxy/csi.sock
I0523 17:14:44.487729       1 common.go:111] Probing CSI driver for readiness
I0523 17:14:44.489037       1 main.go:128] CSI driver name: "ebs.csi.aws.com"
[...]

PR #1280, opened 2022-06-20
Set handle-volume-inuse-error to false

stijndehaes:
If online resizing is supported this should
be set to false. Otherwise, the resizer
will watch pods and use a lot more memory.

This fixes: #1248

It set's the check to see if the volume is in use to false. This should only be turned on
when the volume supports live resizing. I believe EBS volumes support resizing when in use.

**What testing is done?** 
This setting is used in other csi implementations

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -224,6 +224,7 @@ spec:
           args:
             - --csi-address=$(ADDRESS)
             - --v={{ .Values.sidecars.resizer.logLevel }}
+            - --handle-volume-inuse-error=false
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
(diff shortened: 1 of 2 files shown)


torredil (changes requested):
Hi, thank you for your contribution! Can you please also include this change in [deploy/kubernetes/base/controller.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b05ed2d4fc67e7986cbd026c448b349911908af4/deploy/kubernetes/base/controller.yaml#L127)? thanks.

stijndehaes:
> Hi, thank you for your contribution! Can you please also include this change in [deploy/kubernetes/base/controller.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b05ed2d4fc67e7986cbd026c448b349911908af4/deploy/kubernetes/base/controller.yaml#L127)? thanks.

Yeah no problem! I added it to the PR

torredil:
Thanks!
/lgtm
/approve
cc @gtxu @rdpsin

torredil (approved):

stijndehaes:
@torredil Hey I saw master branch failing because the helm chart has the same version. Did I have to update the helm charts version in my PR? I can make a second PR to update the version if that would help :)

torredil:
@stijndehaes No worries! We are currently working on releasing `v1.7.0` and will bump up the Helm chart version as part of the release process. Thanks :)

PR #1282, opened 2022-06-20
Prepare for release v1.7.0

torredil:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,4 +23,4 @@ jobs:
             ## CHANGELOG
             See [CHANGELOG](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG-0.x.md) for full list of changes
           draft: false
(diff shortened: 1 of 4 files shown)


torredil:
cc: @wmesard

torredil:
Sorry for the noise. 
/retest-required

.github/workflows/release.yaml
```diff
@@ -23,4 +23,4 @@ jobs:
             ## CHANGELOG
             See [CHANGELOG](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG-0.x.md) for full list of changes
           draft: false
-          prerelease: false
+          prerelease: true
```
olemarkus, line 26: Curious about this change. It seems unusual to flag a semver stable tag as a prerelease.

olemarkus: I think the quick fix of this would be to have the job set `draft: true` rather than `prerelease: true`.

PR #1319, opened 2022-07-25
Validate fs type before mounting

torredil:

- Validate file system type before mounting.

**What testing is done?** 
- `make test`

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/driver/node.go
```diff
@@ -56,7 +57,12 @@ const (
 )
 
 var (
-	ValidFSTypes = []string{FSTypeExt2, FSTypeExt3, FSTypeExt4, FSTypeXfs}
+	ValidFSTypes = map[string]struct{}{
```
rdpsin, line 60: Should we also add NTFS to this list?

torredil: We definitely should :)

Issue #1267, opened 2022-06-08
default kmsKeyID is not used when value is null


amilanoski:
/kind bug

**What happened?**
default kmskeyid is not used even though CMK is set in EC2 Dashboard >EBS Encryption 

**What you expected to happen?**

For aws-csi-driver to use the default CMK kms key when set in the aws console.

**How to reproduce it (as minimally and precisely as possible)?**
values.yaml 
```
storageClasses:
# Add StorageClass resources like:
- name: ebs-sc
  # annotation metadata
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  # label metadata
  # labels:
  #   my-label-is: supercool
  # defaults to WaitForFirstConsumer
  volumeBindingMode: WaitForFirstConsumer
  # defaults to Delete
  reclaimPolicy: Delete
  parameters:
    encrypted: "true"
  allowVolumeExpansion: true
```

**Anything else we need to know?**:
kms key is set to the aws generated kms key rather then the CMK we have created.

only work around is to set the kmskeyid to the arn we require.  Either the docs need to be updated or there is a bug.  I am not sure.  Hoping for some guidance and answers. Thanks in advance. 

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:  helm chart version: 2.6.3

torredil:
This is not a bug and currently expected behavior. To clarify: Amazon EBS automatically creates a unique AWS managed key in each Region where you store AWS resources. This [KMS key](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html#:~:text=Amazon%20EBS%20automatically%20creates%20a%20unique%20AWS%20managed%20key%20in%20each%20Region%20where%20you%20store%20AWS%20resources.%20This%20KMS%20key%20has%20the%20alias%20alias/aws/ebs.%20By%20default%2C%20Amazon%20EBS%20uses%20this%20KMS%20key%20for%20encryption.) has the alias alias `/aws/ebs`. By default, Amazon EBS uses this KMS key for encryption. 

If you wish to use a different kms key, you **must** specify it in the StorageClass. If the encrypted parameter is set to "true" but no key is specified, the AWS generated kms key will be used by default.

amilanoski:
Thanks for confirming @torredil and updating the docs. I have tested some different paths with default EBS CMK configured in the console.

option 1 - **values.yaml  - encryption:true**
```
storageClasses:
[...]

PR #1329, opened 2022-07-29
Update kOps to `v1.23.0` + Update `parameters.md`

torredil:
- Update kOps to `v1.23.0` + disable default installation of CSI driver in kOps to prevent testing conflicts
- Update `parameters.md` to clarify #1267
- Disable `Pre-provisioned PV` tests in CSI migration.

**What testing is done?** 
- CI

closes #1187


torredil:
Sorry for the noise, sniped by flaky CI and fixing co-author info in commit msg.
PTAL @wongma7

wongma7:
/lgtm

I think the pull-aws-ebs-csi-driver-migration-test failures/flakes are because of bug in the tests, we might consider skipping all 'pre-provisioned' test cases since 'dynamic' test cases are basically a superset of those.
`Jul 29 17:38:56.144: INFO: At 2022-07-29 17:33:45 +0000 UTC - event for exec-volume-test-preprovisionedpv-wg7c: {attachdetach-controller } FailedAttachVolume: AttachVolume.Attach failed for volume "aws-444x6" : rpc error: code = Internal desc = Could not attach volume "vol-01c6fe3ef04e529d3" to node "i-03497c820bd31d419": could not attach volume "vol-01c6fe3ef04e529d3" to node "i-03497c820bd31d419": InvalidVolume.ZoneMismatch: The volume 'vol-01c6fe3ef04e529d3' is not in the same availability zone as instance 'i-03497c820bd31d419' `

for the non-migration test suites, CSI topology ensures this problem never happens. But for the pull-aws-ebs-csi-driver-migration-test suite, the pre-provisioned tests naively create the PV without any zone/topology label, it should add the label :
https://github.com/kubernetes/kubernetes/pull/108696 

What I dont understand is how we are the only ones hitting this failure , it seems like it should also affect GCE PD test cases , anyway i would selfishly like https://github.com/kubernetes/kubernetes/pull/108696  to merge anyway.

hack/kops-patch.yaml
```diff
@@ -172,3 +172,6 @@ spec:
       ExpandCSIVolumes: "true"
       CSIInlineVolume: "true"
```
wongma7, line 173: BTW when bumping K8S_VERSION we can remove CSIDriverRegistry, CSINodeInfo, and CSIBlockVolume, all of these are GA since 1.18 so no need to explicitly set true

torredil: ack, will do 👍

docs/parameters.md
```diff
@@ -6,10 +6,10 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
 | Parameters                  | Values                                 | Default  | Description         |
 |-----------------------------|----------------------------------------|----------|---------------------|
 | "csi.storage.k8s.io/fstype" | xfs, ext2, ext3, ext4                  | ext4     | File system type that will be formatted during volume creation. This parameter is case sensitive! |
-| "type"                      | io1, io2, gp2, gp3, sc1, st1,standard  | gp3*     | EBS volume type     |
+| "type"                      | io1, io2, gp2, gp3, sc1, st1,standard  | gp3*     | EBS volume type.     |
```
rdpsin, line 9: FYI, the * is missing from the new version of this doc. 

```
'gp3' is not supported on Outposts. Outpost customer will need to choose another volume type.
```

torredil:
Most recent CI was failing due to overriding the `GINKGO_SKIP` value in [run.sh](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b05ed2d4fc67e7986cbd026c448b349911908af4/hack/e2e/run.sh#L73). Latest commit accounts for this by also skipping `"\[Disruptive\]"` tests.

hack/e2e/run.sh
```diff
@@ -52,7 +52,7 @@ IMAGE_TAG=${IMAGE_TAG:-${TEST_ID}}
 # eksctl: mustn't include patch version (e.g. 1.19)
 K8S_VERSION=${K8S_VERSION:-1.20.8}
```
ConnorJC3, line 53: This PR already implements several CI-related changes, maybe just bump this now?

hack/kops-patch.yaml
```diff
@@ -172,3 +172,6 @@ spec:
...
     anonymousAuth: false
+  cloudConfig:
+    awsEBSCSIDriver:
+      enabled: false
```
ConnorJC3, line 177: nit: removed newline at end of file

ConnorJC3 (approved):
/lgtm some minor nitpicks

hack/e2e/run.sh
ConnorJC3: On second thought, this would require splitting the version for EKS/kops created clusters, probably better as its own PR

hack/kops-patch.yaml
ConnorJC3: :+1:

PR #1330, opened 2022-07-29
Pass GOPROXY to image builder

wongma7:

**What is this PR about? / Why do we need it?** for networks where proxy.golang.org is blocked, go mod download fails.

**What testing is done?**  ~~testing now, now go mod download seems to be working but it is taking a while.~~
edit: it worked. took 15+ minutes but I assume it will be cached :pray: '
```
10 [builder 4/6] RUN go mod download
#10 DONE 1052.4s
```


torredil:
Thanks for this :) 

/lgtm
/hold

torredil (approved):

PR #1334, opened 2022-08-03
Only run helm action when Chart.yaml modified

ConnorJC3:

"Bug fix"

The helm release action gets run every time the `charts/` directory is modified. However, if the helm version hasn't been bumped (either by accident or because the new changes intentionally shouldn't be released) then the commit has a sad red X when merged into master:

image

This PR fixes that by only running the helm release action when `Chart.yaml` is modified (which very likely is a version bump).

**What testing is done?** 

Manually tested on test branch of forked repo

### DIFF
--- a/.github/workflows/helm-chart-release.yaml
+++ b/.github/workflows/helm-chart-release.yaml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
     paths:
-      - "charts/**"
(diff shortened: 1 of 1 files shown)


torredil:
Thanks for this.

/ok-to-test
/lgtm
/hold

PR #1340, opened 2022-08-12
Change the example policy to have tighter permissions.

huguesalary:
Draft PR until some conversation happens. Specifically, with this PR, the example policy won't work without the end user looking at it and changing `CLUSTER_ID` to their cluster id/name.

## Description 

The example policy is too lax.

This policy presents the following issues:

* It allows the IAM Role to delete any EBS volume tagged with "CSIVolumeName" OR "ebs.csi.aws.com/cluster" OR "kubernetes.io/cluster/*"
* It allows the IAM Role to delete any EBS snapshot tagged with "CSIVolumeSnapshotName" OR "ebs.csi.aws.com/cluster"

This means the EBS CSI IAM Role can delete any volume or snapshot from any other cluster.

This commit changes this policy and replaces the OR conditions by AND conditions, as well as restricting kubernetes.io/cluster/* to kubernetes.io/cluster/EKS_CLUSTER_ID.

This means the EBS CSI IAM Role can only delete volume or snapshot it owns.

### DIFF
--- a/docs/example-iam-policy-strict.json
+++ b/docs/example-iam-policy-strict.json
@@ -1,0 +1,126 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:AttachVolume",
+        "ec2:DetachVolume",
+        "ec2:ModifyVolume",
+        "ec2:DescribeAvailabilityZones",
+        "ec2:DescribeInstances",
+        "ec2:DescribeSnapshots",
+        "ec2:DescribeTags",
+        "ec2:DescribeVolumes",
+        "ec2:DescribeVolumesModifications"
+      ],
+      "Resource": "*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateTags"
+      ],
+      "Resource": [
+        "arn:aws:ec2:*:*:volume/*",
+        "arn:aws:ec2:*:*:snapshot/*"
+      ],
+      "Condition": {
+        "StringEquals": {
+          "ec2:CreateAction": [
+            "CreateVolume",
+            "CreateSnapshot"
+          ]
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteTags"
+      ],
+      "Resource": [
+        "arn:aws:ec2:*:*:volume/*"
+      ],
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeName": "*",
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
+          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteTags"
+      ],
+      "Resource": [
+        "arn:aws:ec2:*:*:snapshot/*"
+      ],
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeSnapshotName": "*",
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
+          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "aws:RequestTag/CSIVolumeName": "*",
+          "aws:RequestTag/ebs.csi.aws.com/cluster": "true",
+          "aws:RequestTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteVolume"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeName": "*",
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
+          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateSnapshot"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeName": "*",
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
+          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteSnapshot"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeSnapshotName": "*",
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
+          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    }
+  ]
+}
--- a/docs/install.md
+++ b/docs/install.md
@@ -9,8 +9,9 @@
 * Important: If you intend to use the Volume Snapshot feature, the [Kubernetes Volume Snapshot CRDs](https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd) must be installed **before** the EBS CSI driver. For installation instructions, see [CSI Snapshotter Usage](https://github.com/kubernetes-csi/external-snapshotter#usage).
 
 ## Installation
-### Set up driver permissions
-The driver requires IAM permissions to talk to Amazon EBS to manage the volume on user's behalf. [The example policy here](./example-iam-policy.json) defines these permissions. AWS maintains a managed policy, available at ARN `arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy`. 
+
+### Set up driver permission
+The driver requires IAM permissions to talk to Amazon EBS to manage the volume on user's behalf. [The example policy here](./example-iam-policy.json) defines these permissions. A [stricter example policy](./example-iam-policy-strict.json) is also provided, but needs to be modified before use (specifically, replace `<<CLUSTER_ID>>` with the ID of your cluster). AWS maintains a managed policy, available at ARN `arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy`.
 
 Note: Add the below statement to the example policy if you want to encrypt the EBS drives. 
 ```


torredil:
Hey, thanks for this. I agree that the example policy is too lax, but I am concerned this change could inconvenience users since the policy will no longer work without making modifications.

As @bryantbiggs mentioned in https://github.com/terraform-aws-modules/terraform-aws-iam/issues/266, some add-ons have a "general" policy and a "restricted" policy. How about instead of replacing this policy, we create a new one: `example-iam-policy-restricted.json`?

huguesalary:
Sounds good to me; I'll modify the PR

huguesalary:
I created a new policy called `example-iam-policy-strict.json` and added a very short callout to it in the documentation.

torredil:
Thanks!

/ok-to-test
/lgtm
/hold

cc: @gtxu

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,127 @@
...
+        "ec2:DeleteTags"
+      ],
+      "Resource": [
+        "arn:aws:ec2:*:*:volume/*"
+      ],
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeName": "*",
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
+          "ec2:ResourceTag/kubernetes.io/cluster/CLUSTER_ID": "owned"
```
bryantbiggs, line 50: should we make this a bit more clear that users need to specify the cluster ID

```suggestion
          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
```

docs/install.md
```diff
@@ -10,7 +10,7 @@
 
 ## Installation
 #### Set up driver permission
-The driver requires IAM permission to talk to Amazon EBS to manage the volume on user's behalf. [The example policy here](./example-iam-policy.json) defines these permissions. There are several methods to grant the driver IAM permission:
+The driver requires IAM permission to talk to Amazon EBS to manage the volume on user's behalf. [The example policy here](./example-iam-policy.json) defines these permissions. A [stricter example policy](./example-iam-policy-strict.json) is also provided, but needs to be modified before use. There are several methods to grant the driver IAM permission:
```
bryantbiggs, line 13: Perhaps a note here as well that `<<CLUSTER_ID>>` needs to be updated to set the cluster ID value in the policy

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,127 @@
...
+        "ec2:DetachVolume",
+        "ec2:ModifyVolume",
+        "ec2:DescribeAvailabilityZones",
+        "ec2:DescribeInstances",
+        "ec2:DescribeSnapshots",
+        "ec2:DescribeTags",
+        "ec2:DescribeVolumes",
+        "ec2:DescribeVolumesModifications"
+      ],
+      "Resource": "*"
```
0x9090, line 18: Is there a way to tighten this, or limit it to only the k8s accounts? This would give a lot of ec2 permissions to anyone who used this policy

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,127 @@
...
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteTags"
+      ],
+      "Resource": [
+        "arn:aws:ec2:*:*:volume/*"
+      ],
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/CSIVolumeName": "*",
```
0x9090, line 48: Can we set up a naming scheme that we can tighten this one too?

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,127 @@
...
+          "aws:RequestTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteVolume"
+      ],
+      "Resource": "*",
```
0x9090, line 89: Deletions are a bit more sketchy than just creating volumes. Can we make a naming scheme for them as a resources, and specify it here?

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,127 @@
...
+          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteSnapshot"
+      ],
+      "Resource": "*",
```
0x9090, line 117: A resource naming scheme we can limit to would be beneficial, but maybe the conditionals are enough to tighten it down to only the k8s ec2 snapshots

0x9090 (changes requested):

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,127 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateSnapshot",
```
mitchellhuang, line 7: Remove this, `ec2:CreateSnapshot` is action is now defined on line 101.

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,127 @@
...
+          "ec2:ResourceTag/CSIVolumeName": "*",
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
+          "ec2:ResourceTag/kubernetes.io/cluster/<<CLUSTER_ID>>": "owned"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateSnapshot"
```
mitchellhuang, line 101: Dupe here.

docs/example-iam-policy-strict.json
huguesalary: I think the `CSIVolumeName` is not predictable, that's why it's set to `*`. Line 50 is what makes this tighter, by ensuring the volume whose tags are being deleted does actually belong to the cluster.

docs/example-iam-policy-strict.json
huguesalary: The condition line 90 is enforcing the volume being deleted belong to the cluster deleting it

docs/example-iam-policy-strict.json
huguesalary: Same as other comments

docs/example-iam-policy-strict.json
mitchellhuang: For the Describe* actions, you cannot, per AWS [docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-policy-structure.html#UsingWithEC2_Actions):
> Currently, the Amazon EC2 Describe* API actions do not support resource-level permissions. For more information about resource-level permissions for Amazon EC2, see [IAM policies for Amazon EC2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-policies-for-amazon-ec2.html).

docs/example-iam-policy-strict.json
```diff
@@ -0,0 +1,126 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:AttachVolume",
+        "ec2:DetachVolume",
+        "ec2:ModifyVolume",
```
mitchellhuang, line 9: For these actions can we scope them further to only allow this on the volumes provisioned by the CSI driver?

huguesalary: I _think_ I tried that originally, but ran into issues; I don't remember what exactly. If someone can confirm scoping this further won't cause any issue, I'm happy to do it

mitchellhuang: Here is a cleaned up policy to do this. I also removed the broad `ec2:DescribeTags` action because there is no reference to that API call anywhere in the codebase. I wonder why it's there in the first place?

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "ec2:DescribeInstances",
        "ec2:DescribeSnapshots",
        "ec2:DescribeVolumes",
        "ec2:DescribeVolumesModifications"
      ],
      "Resource": "*"
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:CreateTags"
      ],
      "Resource": [
        "arn:aws:ec2:*:*:volume/*",
        "arn:aws:ec2:*:*:snapshot/*"
      ],
      "Condition": {
        "StringEquals": {
          "ec2:CreateAction": [
            "CreateVolume",
            "CreateSnapshot"
          ]
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:DeleteTags"
      ],
      "Resource": [
        "arn:aws:ec2:*:*:volume/*"
      ],
      "Condition": {
        "StringLike": {
          "ec2:ResourceTag/CSIVolumeName": "*",
          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
          "ec2:ResourceTag/kubernetes.io/cluster/<%= cluster_name %>": "owned"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:DeleteTags"
      ],
      "Resource": [
        "arn:aws:ec2:*:*:snapshot/*"
      ],
      "Condition": {
        "StringLike": {
          "ec2:ResourceTag/CSIVolumeSnapshotName": "*",
          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
          "ec2:ResourceTag/kubernetes.io/cluster/<%= cluster_name %>": "owned"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:CreateVolume"
      ],
      "Resource": "*",
      "Condition": {
        "StringLike": {
          "aws:RequestTag/CSIVolumeName": "*",
          "aws:RequestTag/ebs.csi.aws.com/cluster": "true",
          "aws:RequestTag/kubernetes.io/cluster/<%= cluster_name %>": "owned"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:AttachVolume",
        "ec2:DetachVolume"
      ],
      "Resource": [
        "arn:aws:ec2:*:*:volume/*"
      ],
      "Condition": {
        "StringLike": {
          "ec2:ResourceTag/CSIVolumeName": "*",
          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
          "ec2:ResourceTag/kubernetes.io/cluster/<%= cluster_name %>": "owned"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:AttachVolume",
        "ec2:DetachVolume"
      ],
      "Resource": [
        "arn:aws:ec2:*:*:instance/*"
      ],
      "Condition": {
        "StringLike": {
          "ec2:ResourceTag/kubernetes.io/cluster/<%= cluster_name %>": "owned"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:ModifyVolume",
        "ec2:DeleteVolume"
      ],
      "Resource": "*",
      "Condition": {
        "StringLike": {
          "ec2:ResourceTag/CSIVolumeName": "*",
          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
          "ec2:ResourceTag/kubernetes.io/cluster/<%= cluster_name %>": "owned"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:CreateSnapshot",
        "ec2:DeleteSnapshot"
      ],
      "Resource": "*",
      "Condition": {
        "StringLike": {
          "ec2:ResourceTag/CSIVolumeSnapshotName": "*",
          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true",
          "ec2:ResourceTag/kubernetes.io/cluster/<%= cluster_name %>": "owned"
        }
      }
    }
  ]
}

```

huguesalary: Have you tested this? 

I don't think `CreateSnapshot` and `DeleteSnapshot` can be bundled together. The condition for `CreateSnapshot` would be `"aws:RequestTag...`. 

However, I remember running into issues trying to make this policy as clean as possible. Changes that seemed innocuous actually weren't.

mitchellhuang: FWIW I did a basic test with this policy and it appears to work.

```
kubectl apply -f aws-ebs-csi-driver/examples/kubernetes/dynamic-provisioning/manifests
kubectl delete -f aws-ebs-csi-driver/examples/kubernetes/dynamic-provisioning/manifests
```

The trick is you need to scope the condition on the `instance` side and `volume` side in separate statements. If you do it on `*`, the instances will not have the additional tags that the volumes have, denying the driver permission to attach/detatch. Found it from this blog post [here](https://www.iampulse.com/resources/policies/attach-or-detach-volumes-to-an-ec2-instance-based-on-tags).

mitchellhuang: Oh for the snapshots I did not. Good catch.

huguesalary: Ok, so you tested this last policy and everything works? If so, I'll commit that

mitchellhuang: I did not test snapshots yet. Are you able to test? Otherwise I can try to do it tomorrow.

mitchellhuang: Ok, I tested this final [policy](https://gist.github.com/mitchellhuang/ed141207493745ea22ac02f27396df16) with snapshots and it appears to work. Ended up just doing `*` resources for `CreateSnapshot` like you had.

huguesalary:
@torredil any chance we could get this merged? Thank you!

0x9090 (approved):

kevarr:
Does this policy allow for deleting volumes created by the in-tree storage plugin?

PR #1341, opened 2022-08-15
Enable CI tests in K8s `1.25`

torredil:

- Upgrade `k8s.io/kubernetes` to `1.25`.
- Upgrade `github.com/onsi/ginkgo` to `v2`.
- Use `kubetest2` for `e2e-kubernetes` and use the upstream binary for this test. 
- Split `K8S_VERSION` in tests between `kOps` and `eksctl` so that we can test them independently.
- Removed migration tests due to in-tree driver removal.
- Remove `CSIDriverRegistry`, `CSINodeInfo`, and `CSIBlockVolume` from `hack/kops-patch.yaml` as the feature gates have been GA since K8s `v1.18`.


ConnorJC3:
Needs eksctl version bump to https://github.com/weaveworks/eksctl/releases/tag/v0.109.0-rc.0 for k8s 1.23 support.

olemarkus:
@torredil I wonder if it is worth implementing a feature in kops where you tell kops the addon is enabled, but it's selfmanaged. That way kops will behave as if it installed the addon, but won't deploy the manifest. We do this e.g for cert-manager, which a lot of people prefer to self-manage.

This will make it a lot easier to migrate to 1.24, where CSI driver is mandatory and cannot be disabled.

ConnorJC3:
kops patch needs merge with this to disable security feature of preventing nodes from accessing IMDSv2:

```yaml
spec:
  instanceMetadata:
    httpPutResponseHopLimit: 3
```

torredil:
@olemarkus Can you clarify this point: "CSI driver is mandatory and cannot be disabled". My understanding was that even in kOps v1.24, users would still be able to disable the CSI driver as currently done in [kops-patch.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/c2eb9660508ea90761142eac8b29236c37e3f955/hack/kops-patch.yaml#L177).
```  
cloudConfig:
    awsEBSCSIDriver:
      enabled: false
 ```
If this is not the case, then I think the solution you are proposing would be worth implementing.

**Comment by olemarkus** [comment]
> @olemarkus Can you clarify this point: "CSI driver is mandatory and cannot be disabled". My understanding was that even in kOps v1.24, users would still be able to disable the CSI driver as currently done in [kops-patch.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/c2eb9660508ea90761142eac8b29236c37e3f955/hack/kops-patch.yaml#L177).
> 
> ```
> cloudConfig:
>     awsEBSCSIDriver:
>       enabled: false
> ```
> 
> If this is not the case, then I think the solution you are proposing would be worth implementing.

To be precise, kOps 1.24 with K8s 1.24 will enable external cloud-controller-manager, which in turn requires CSI. So the above snippet will fail with "external cloud controller manager requires CSI driver to be enabled" or something to that effect.

**Comment by ConnorJC3** [comment]
@olemarkus thoughts on just disabling the external CCM in CI as a temporary fix? I can't think of any reason it would be needed in CI.

Also, I think I might have seen someone mention this elsewhere on another issue, but I wonder if as a long term fix it would be better to add a kops option to override the EBS CSI Driver container repo/tag. Then, the CI could just deploy the driver via kops which would be easier and probably more accurately represent real-world customer installations.

**Comment by torredil** [comment]
Adding this here for context: As of kOps 1.22, new clusters running Kubernetes 1.22 on AWS will restrict Pod access to the instance metadata service. This means that Pods will also be prevented from directly assuming instance roles. See IAM Roles for ServiceAccounts and instance metadata documentation. Before this, all Pods running on your cluster have access to the instance IAM role. Consider enabling the protection mentioned above and use IRSA for your own workloads. See https://kops.sigs.k8s.io/iam_roles/#adding-additional-policies

**Comment by olemarkus** [comment]
> @olemarkus thoughts on just disabling the external CCM in CI as a temporary fix? I can't think of any reason it would be needed in CI.

kOps is moving away from in-tree CCM. From kOps perspective it doesn't make much sense to add any option to disable it.

> 
> Also, I think I might have seen someone mention this elsewhere on another issue, but I wonder if as a long term fix it would be better to add a kops option to override the EBS CSI Driver container repo/tag. Then, the CI could just deploy the driver via kops which would be easier and probably more accurately represent real-world customer installations.

Yes. That is also an option. There are quite a few images involved though, so a question of how much configurability you need. E.g do you want to test all the sidecars?

FWIW, we run the CSI driver e2e test against all our configurations now. So e.g the 1.23 results can be seen here: https://testgrid.k8s.io/kops-k8s-1.23#kops-aws-k8s-1-23 (tests containing the string "External Storage")

File: hack/kops-patch.yaml [hunk 1]
```diff
@@ -175,3 +165,5 @@ spec:
   cloudConfig:
     awsEBSCSIDriver:
       enabled: false
+    instanceMetadata:
```
olemarkus, line 168: Instead of doing this one, you can also set node affinity to run on the control plane.

ConnorJC3: That defeats the purpose of the multi-az tests though, right?

olemarkus: Only the controller needs metadata afaik. And it runs leader elected anyway.

torredil: @olemarkus This isn't very well documented but my understanding is that currently the node service relies on the metadata service to derive the max number of attachable volumes, instance id, and AZ for dynamic provisioning.

olemarkus: Both kops and EKS runs with hop limit 1 and this works... Perhaps the controller fetches this information when the daemonset cannot?

ConnorJC3: Node can pull metadata from k8s, it only bombs out if both ec2 and k8s metadata aren't available. The reason that the controller needs metadata here is because we're using it to pull an instance IAM role.

kops can also setup IRSA instead of instance roles but requires a publicly-readable S3 bucket to do so, so it's easier to just enable IMDSv2 than jump through the hoops to get the test-infra folks to setup a bucket and modify the test scripts to use it.

ConnorJC3: I think I'm in favor of my original solution here because it feels less janky and changing the affinity would slightly lower the test coverage (as the default affinity wouldn't be tested), but I'm not particularly committed either way and I'll let you all figure out what's best here.

olemarkus: The S3 state store you are already using probably supports IRSA already. But agree IRSA is overkill.

My suggestion to run controller on the control plane is that it has a hop limit of 3, not 1. So you don't need to change the hop limit of worker nodes if you do so.

olemarkus:
@MurphyPuppy See e.g https://testgrid.k8s.io/kops-k8s-1.23#kops-aws-k8s-1-23 mentioned above. The k8s community already tests this driver on multiple versions and configurations.

olemarkus:
Some great work here!

Makefile
```diff
@@ -193,13 +187,13 @@ test-e2e-external:
 	EBS_INSTALL_SNAPSHOT="true" \
 	TEST_PATH=./tests/e2e-kubernetes/... \
 	GINKGO_FOCUS="External.Storage" \
-	GINKGO_SKIP="\[Disruptive\]|\[Serial\]" \
+	GINKGO_SKIP="\[Disruptive\]|\[Serial\]|provisioning should mount multiple PV pointing to the same storage on the same node" \
```
ConnorJC3, line 190: Maybe leave a comment explaining this?

hack/e2e/run.sh
```diff
@@ -189,30 +197,37 @@ fi
...
-    OVERALL_TEST_PASSED=0
-  else
-    loudecho "One of test or migration failed"
-    OVERALL_TEST_PASSED=1
-  fi
+if [[ $TEST_PATH == "./tests/e2e-kubernetes/..." ]]; then
+  pushd $(pwd)/tests/e2e-kubernetes
+  packageVersion=$(echo $(cut -d '.' -f 1,2 <<< $KOPS_VERSION))
+
+  kubetest2 noop \
```
ConnorJC3, line 205: don't we need `set -x` and `set +e` stuff here not to crash out on test failure (see next diff)

hack/update-gomock
```diff
@@ -16,7 +16,8 @@
...
 
-./bin/mockgen -package cloud -destination=./pkg/cloud/mock_ec2.go -source pkg/cloud/ec2_interface.go
-./bin/mockgen -package cloud -destination=./pkg/cloud/mock_cloud.go -source pkg/cloud/cloud_interface.go
-./bin/mockgen -package cloud -destination=./pkg/cloud/mock_metadata.go -source pkg/cloud/metadata_interface.go
-./bin/mockgen -package driver -destination=./pkg/driver/mock_mount.go -source pkg/driver/mount.go
+mockgen -package cloud -destination=./pkg/cloud/mock_ec2.go -source pkg/cloud/ec2_interface.go
+mockgen -package cloud -destination=./pkg/cloud/mock_cloud.go -source pkg/cloud/cloud_interface.go
+mockgen -package cloud -destination=./pkg/cloud/mock_metadata.go -source pkg/cloud/metadata_interface.go
+mockgen -package driver -destination=./pkg/driver/mock_mount.go -source pkg/driver/mount.go
+sed -i '/type MockMounter struct {/a \\tmount_utils.Interface' pkg/driver/mock_mount.go
```
ConnorJC3, line 23: Add explanation comment

tests/e2e-kubernetes/manifests.yaml
```diff
@@ -5,7 +5,7 @@ ShortName: ebs
 StorageClass:
   FromFile: storageclass.yaml
 SnapshotClass:
-  FromFile: snapshotclass.yaml
+  FromName: true
```
ConnorJC3, line 8: Not sure I agree with this, would love to see other's thoughts

tests/e2e/suite_test.go
```diff
@@ -61,13 +55,12 @@ func TestE2E(t *testing.T) {
 	// Run tests through the Ginkgo runner with output to console + JUnit for Jenkins
 	var r []Reporter
 	if framework.TestContext.ReportDir != "" {
+		// Create the directory if it doesn't already exists
+		// NOTE: junit report can be simply created by executing your tests with the new --junit-report flags instead.
```
ConnorJC3, line 59: "can be simply created by executing your tests with the new --junit-report flags instead" -> "can be created with new --junit-report flag"

hack/e2e/run.sh
```diff
@@ -189,30 +197,37 @@ fi
...
-  loudecho "MIGRATION_PASSED: ${MIGRATION_PASSED}"
-  if [ "${TEST_PASSED}" == 0 ] && [ "${MIGRATION_PASSED}" == 0 ]; then
-    loudecho "Both test and migration passed"
-    OVERALL_TEST_PASSED=0
-  else
-    loudecho "One of test or migration failed"
-    OVERALL_TEST_PASSED=1
-  fi
+if [[ $TEST_PATH == "./tests/e2e-kubernetes/..." ]]; then
+  pushd $(pwd)/tests/e2e-kubernetes
```
ConnorJC3, line 202: use `${PWD}` like below

ConnorJC3 (changes requested):

hack/e2e/run.sh
torredil: Good eyes, added this 👍

tests/e2e/suite_test.go
torredil: Done.

hack/e2e/run.sh
torredil: Done.

tests/e2e-kubernetes/manifests.yaml
torredil: `FromFile` would be more appropriate if we had additional parameters in the snapshotclass object. One benefit of this change is that we no longer have to maintain our own snapshotclass for testing.

https://github.com/kubernetes/kubernetes/blob/5556d3336b986dcd7ec7de85f731425be91273b4/test/e2e/storage/external/external.go#L99

olemarkus: We've had issues using this one because it _copies_ the given object per test and when the original object is also the default, you end up with multiple defaults resulting in failures. I tried to fix this in k/k for StorageClass, would assume its the same for SnapshotClass. https://github.com/kubernetes/kubernetes/pull/111730

torredil: @olemarkus [FromExistingClassName](https://github.com/kubernetes/kubernetes/blob/5556d3336b986dcd7ec7de85f731425be91273b4/test/e2e/storage/external/external.go#L103) behaves as you describe, but we specify [FromName]() here which I believe behaves differently. Either way, I can revert this change if you / other reviewers feel strongly about it.

ConnorJC3: In that case, I think it's fine as-is

wongma7:
/lgtm
thx

torredil:
Good to go
thanks team 👍

PR #1344, opened 2022-08-16
Example policy to support encrypted drives

challapradyumna:
Doc update

The example policy should include some sensible defaults for encrypted drives to work.

**What testing is done?** 
Used the policy to create drives

### DIFF
--- a/docs/example-iam-policy.json
+++ b/docs/example-iam-policy.json
@@ -140,15 +140,6 @@
           "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true"
         }
       }
-    },
-    {
-      "Effect": "Allow",
-      "Action": [
-          "kms:Decrypt",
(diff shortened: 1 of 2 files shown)


torredil:
Thanks for raising this up. There's been a few issues related to IAM policies recently which indicates we should improve the documentation in this area. There are no benefits to giving the driver permissions it doesn't need, so a better long-term solution would be having a baseline IAM policy which contains the minimum permissions the driver requires to properly function, and then separately document additional permissions that can be added for other functionality such as snapshots and encrypted drives. The idea is that driver users shouldn't have to guess which policies they need to get something to work or provide more access than needed.

challapradyumna:
Instead of updating the sample policy. I've added a note to the install doc itself with the necessary details.

torredil (approved):

PR #1353, opened 2022-08-17
Set inital release to draft status

gtxu:
Signed-off-by: Gengtao Xu <gengtaox@amazon.com>

Big fix

Once release tag it post a new version in release section while the images are not ready. To avoid confusion we set the initial release as draft release and mark it as official release once the all processes are done.

**What testing is done?**


torredil:
We can close this one since we need this change in the release branch, otherwise we need to cherry-pick it there, 
added change to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1351

PR #1356, opened 2022-08-22
Builds should not set/get GOPROXY in the environment

wmesard:
Go settings are managed by `go env`. This is the way.

Also, simplify the build instructions. `GO111MODULE` does not need to
be set in the build environment.

Signed-off-by: Wayne Mesard <wsm@amazon.com>

**What testing is done?**


torredil (approved):
/lgtm
/hold

PR #1358, opened 2022-08-23
c6i.metal and g5g.metal are nitro instances

wmesard:
Signed-off-by: Wayne Mesard <wsm@amazon.com>

Bug

An earlier version of the [Available instance types list](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances) was unclear on whether `c6i.metal` and `g5g.metal`. They are. And the docs now say so unambiguously. This PR updates the code to match the doc (and reality).

**What testing is done?** 

`make && make test`

### DIFF
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -14,13 +14,6 @@ const (
 	nitroMaxAttachments                  = 28
 )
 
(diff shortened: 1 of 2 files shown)


torredil (approved):
/lgtm
/approve
/hold

Issue #1357, opened 2022-08-23
Error on  ebs-csi-controller pod


nicodeur:
/kind bug

Hi,

**What happened?**
Container `ebs-plugin` failed to run on `ebs-csi-controller` pod.
See error bellow : 
```
I0823 14:45:34.396446       1 metadata.go:85] retrieving instance data from ec2 metadata
W0823 14:45:40.688951       1 metadata.go:88] ec2 metadata is not available
I0823 14:45:40.688972       1 metadata.go:96] retrieving instance data from kubernetes api
I0823 14:45:40.689867       1 metadata.go:101] kubernetes api is available
panic: did not find aws instance ID in node providerID string

goroutine 1 [running]:
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.newNodeService(0xc000088820)
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/node.go:95 +0x269
github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver.NewDriver({0xc0003a3f30, 0x8, 0x40c6be})
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/driver.go:98 +0x286
main.main()
	/go/src/github.com/kubernetes-sigs/aws-ebs-csi-driver/cmd/main.go:46 +0x365
```
ebs-csi-controller run on a fargate node, strange ?
<img width="1301" alt="image" src="https://user-images.githubusercontent.com/6707556/186190111-d296b8e2-12f1-4151-998e-b95313c728a9.png">

**What you expected to happen?**

`ebs-csi-controller` run without error, and we can use `pvc` on a pod.

**How to reproduce it (as minimally and precisely as possible)?**

Installed [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) follow [aws documentation] https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html on EKS

**Anything else we need to know?**:
Our eks use ec2 nodes and fargate nodes. We want to use Volumes only on ec2 node because it doesn't work on fargate.

**Environment**

- Kubernetes version (use `kubectl version`): 1.23.9
```
[...]

PR #1359, opened 2022-08-23
Add warning message when region is unavailable on the controller

ConnorJC3:

Adds clearer warning message so customers with cases such as #1357 understand they should manually supply the region.

**What testing is done?** 

N/A

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -81,6 +81,7 @@ func newControllerService(driverOptions *DriverOptions) controllerService {
 		klog.V(5).Infof("[Debug] Retrieving region from metadata service")
 		metadata, err := NewMetadataFunc(cloud.DefaultEC2MetadataClient, cloud.DefaultKubernetesAPIClient, region)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/driver/controller.go
```diff
@@ -81,6 +81,7 @@ func newControllerService(driverOptions *DriverOptions) controllerService {
 		klog.V(5).Infof("[Debug] Retrieving region from metadata service")
 		metadata, err := NewMetadataFunc(cloud.DefaultEC2MetadataClient, cloud.DefaultKubernetesAPIClient, region)
 		if err != nil {
+			klog.Errorf("Could not determine region from any metadata service. The region can be manually supplied via the AWS_REGION enviornment variable (controller.region parameter in the helm chart).")
```
ConnorJC3, line 84: I'm somewhat on the fence about whether to include information about the helm chart. Technically, it is CO-specific information, but it undeniable that most customers in this situation will be using the helm chart. Open to thoughts about this.

torredil: Would prefer CO agnostic error logging.

torredil (approved):
/lgtm

PR #1360, opened 2022-08-23
Add controller nodeAffinity to prefer EC2 over Fargate

torredil:
Signed-off-by: torredil <torredil@amazon.com>

- Prefer to schedule controller pods on EC2 over Fargate. `preferredDuringSchedulingIgnoredDuringExecution`: The scheduler tries to find a node that meets the rule. If a matching node is not available, the scheduler still schedules the Pod.


olemarkus:
This looks pretty nice!

ConnorJC3 (approved):
/lgtm

Issue #1365, opened 2022-08-25
Statically provisioned  PV/PVC don't set group ownership when fsGroup is defined


jbehrends:
/kind bug

**What happened?**
When following the "[static-provisioning](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/static-provisioning/manifests)" example/method of using the aws-ebs-csi-driver, setting securityContext.fsGroup in your pod or deployment does NOT configure group ownership of the EBS volume mount.  Setting fsGroup when using the "dynamic-provisioning" example DOES set the group ownership of the EBS volume correctly.  This is an issue because our deployments use non-root uid/gid's and as a result the process running in the container can't write to the EBS volume mount because it's owned by root:root.

**What you expected to happen?**
Setting "securityContext.fsGroup" to a specific GID value should result in the volume mount and the files contained in the mount  having the same GID applied to them.

**How to reproduce it (as minimally and precisely as possible)?**
1. Create an EBS volume in the AWS console.
2. Create a PV using the example provided [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/examples/kubernetes/static-provisioning/manifests/pv.yaml)
3. Create a PVC using the example provided [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/examples/kubernetes/static-provisioning/manifests/claim.yaml)
4. Use the example provided [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/examples/kubernetes/static-provisioning/manifests/pod.yaml) and add "securityContext.fsGroup" and set a value of 1000.  Then create the pod.

The resulting pod will have a volume mount with ownership of root:root and it SHOULD be root:1000

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): EKS 1.21
- Driver version: 1.11.2

ConnorJC3:
Kubelet is responsible for managing `fsGroup` permissions (there is an optional feature in k8s 1.22+ for the driver to manage it, but we don't opt in), The driver does not interfere in any way or otherwise set permissions on volumes. This bug should be [reported to k/k](https://github.com/kubernetes/kubernetes/).

joshbranham:
[...]

PR #1178, opened 2022-02-28
EBS MultiAttach

sr-auto:
**Is this a bug fix or adding new feature?**
New Feature: EBS Multi Attach

**What is this PR about? / Why do we need it?**
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/449

**What testing is done?**

PR #1377, opened 2022-09-08
Change fsGroupPolicy to File

ConnorJC3:
See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1365#issuecomment-1240827211

This _might_ need to be revisited in the future with multi-attach volumes, but for now is 100% safe.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-ebs-csi-driver/templates/csidriver.yaml
@@ -7,3 +7,4 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: false
+  fsGroupPolicy: File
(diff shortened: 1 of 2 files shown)


gnufied:
> This might need to be revisited in the future with multi-attach volumes, but for now is 100% safe.

Is the driver going to support read-write-many multi-attach volumes? It seems like standard filesystems like ext4, XFS does not have necessary fencing to multi-attach same volume in write mode.

ConnorJC3:
EBS has multiple writer support for multi-attach volumes, but I've heard it's very wonky and has several limitations. I haven't tested but I assume they at least partially deal with the filesystem issue on their end?

There's a long-open PR for multi-atrach support that would allow multiple writers (#1178), but it needs serious reworking and probably won't be merged anytime soon.

z0rc:
This changes immutable field in CSIDriver resource on existing deployments. It was undefined previously and therefore kubernetes sets it to `ReadWriteOnceWithFSType` by default.

How this change is expected to be applied by end user?

gnufied:
Hmm. not a helm person, but is there a way it could delete and recreate the resource for immutable fields? We do that in our operators when something immutable has to be changed inside an object. :(

PR #1380, opened 2022-09-12
Add enableMetrics configuration

torredil:
- This PR adds a Helm chart configuration which facilitates scraping AWS metrics when `enableMetrics` is set to true and the Prometheus Operator is installed, by deploying:

  1.  A [Service](https://kubernetes.io/docs/concepts/services-networking/service/) object that exposes the EBS CSI Driver's controller metric port through a `ClusterIP`. 
  
  2. A [ServiceMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#:~:text=Alertmanager-,ServiceMonitor,-See%20the%20Alerting) object which updates the Prometheus scrape configuration and allows scraping metrics from the endpoint defined. For more information, see the manifest `metrics.yaml`.

- `httpEndpoint` is deprecated.

**What testing is done?** 
- See `docs/metrics.md`.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -79,6 +79,9 @@ spec:
             {{- with .Values.controller.k8sTagClusterId }}
             - --k8s-tag-cluster-id={{ . }}
             {{- end }}
+            {{- if and (.Values.controller.enableMetrics) (not .Values.controller.httpEndpoint) }}
+            - --http-endpoint=0.0.0.0:3301
+            {{- end}}
             {{- with .Values.controller.httpEndpoint }}
             - --http-endpoint={{ . }}
             {{- end }}
@@ -133,6 +136,11 @@ spec:
             - name: healthz
               containerPort: 9808
               protocol: TCP
+            {{- if .Values.controller.enableMetrics }}
+            - name: metrics
+              containerPort: 3301
+              protocol: TCP
+            {{- end}}
           livenessProbe:
             httpGet:
               path: /healthz
--- a/charts/aws-ebs-csi-driver/templates/metrics.yaml
+++ b/charts/aws-ebs-csi-driver/templates/metrics.yaml
@@ -1,0 +1,40 @@
+{{- if .Values.controller.enableMetrics -}}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: ebs-csi-controller
+  namespace: kube-system
+  labels:
+    app: ebs-csi-controller
+spec:
+  selector:
+    app: ebs-csi-controller
+  ports:
+    - name: metrics
+      port: 3301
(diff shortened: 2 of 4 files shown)


charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -79,6 +79,9 @@ spec:
             {{- with .Values.controller.k8sTagClusterId }}
             - --k8s-tag-cluster-id={{ . }}
             {{- end }}
+            {{- if .Values.controller.enableMetrics }}
```
ConnorJC3, line 82: This should probably explicitly check the user hasn't manually specified `.Values.controller.httpEndpoint`, or else there will be a conflict

torredil: Good catch 👍  

```  
 httpEndpoint: "localhost:8080"
 enableMetrics: true
```

with

```
{{- if and (.Values.controller.enableMetrics) (not .Values.controller.httpEndpoint) }}
```

output:

```
    Args:
      controller
      --endpoint=$(CSI_ENDPOINT)
      --http-endpoint=localhost:8080
      --logtostderr
      --v=2
```

--- 
no http endpoint specified for good measure:

```
 httpEndpoint: 
 enableMetrics: true
```

with
```
{{- if and (.Values.controller.enableMetrics) (not .Values.controller.httpEndpoint) }}
```

output:
```
    Args:
      controller
      --endpoint=$(CSI_ENDPOINT)
      --http-endpoint=0.0.0.0:3301
      --logtostderr
      --v=2
```

ConnorJC3 (approved):
current version lgtm

wmesard (approved):

charts/aws-ebs-csi-driver/templates/metrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+---
+{{- if (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") -}}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: ebs-csi-controller
+  namespace: kube-system
+  labels:
+    app: ebs-csi-controller
+    release: prometheus
```
stevo-f3, line 27: It would be great if these labels were configurable / overridable without needing to use helm's kustomize post-renderers support.

In my setup `release` label should have a different value. Maybe `app` label can remain fixed, but `release: prometheus` one should ideally be part of e.g. `controller.serviceMonitor.additionalLabels`, to allow for overrides. Then also enable switch can be moved under `controller.serviceMonitor.enabled`

WDYT @torredil ?

torredil: @stevo-f3 Thanks for the feedback, included this improvement in #1419.

PR #1381, opened 2022-09-12
Allow all taint for toleration csi-node by default

gtxu:
Signed-off-by: Gengtao Xu <gengtaox@amazon.com>

This PR removed the previous policy of csi-node taint toleration and make allowing all taint by default. For csi-node, unless customized configuration, we should allow all nodes taint for PV availability.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -36,8 +36,8 @@ kubectl create secret generic aws-secret \
     --from-literal "access_key=${AWS_SECRET_ACCESS_KEY}"
 ```
 
-#### Config node toleration settings
-By default, driver tolerates taint `CriticalAddonsOnly` and has `tolerationSeconds` configured as `300`, to deploy the driver on all nodes, please set Helm `Value.node.tolerateAllTaints` to true before deployment
+#### Configure driver toleration settings
+By default, the driver controller tolerates taint `CriticalAddonsOnly` and has `tolerationSeconds` configured as `300`; and the driver node tolerates all taints. If you don't want to deploy the driver node on all nodes, please set Helm `Value.node.tolerateAllTaints` to false before deployment. Add policies to `Value.node.tolerations` to configure customized toleration for nodes.
 
 #### Deploy driver
 Please see the compatibility matrix before you deploy the driver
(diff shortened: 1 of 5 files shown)


torredil (commented):
This would be a breaking change. Let's just set `tolerateAllTaints: true` in [values.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/3900759a2950deb6fac051605497936e2de8d3fc/charts/aws-ebs-csi-driver/values.yaml#L206) instead and update the [docs](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/3900759a2950deb6fac051605497936e2de8d3fc/charts/aws-ebs-csi-driver/values.yaml#L206).

ConnorJC3:
Agree with @torredil : just changing the default is better here. This change prevents helm users from customizing the tolerations when they don't want to tolerate all taints.

docs/install.md
```diff
@@ -36,8 +36,8 @@ kubectl create secret generic aws-secret \
     --from-literal "access_key=${AWS_SECRET_ACCESS_KEY}"
 ```
 
-#### Config node toleration settings
-By default, driver tolerates taint `CriticalAddonsOnly` and has `tolerationSeconds` configured as `300`, to deploy the driver on all nodes, please set Helm `Value.node.tolerateAllTaints` to true before deployment
+#### Config driver toleration settings
```
ConnorJC3, line 39: Nit: "Config" -> "Configure" (might as well fix this now)

docs/install.md
```diff
@@ -36,8 +36,8 @@ kubectl create secret generic aws-secret \
...
+By default, the driver controller tolerates taint `CriticalAddonsOnly` and has `tolerationSeconds` configured as `300`; the driver node tolerates all taint. If you don't want to deploy the driver node on all nodes, please set Helm `Value.node.tolerateAllTaints` to false before deployment, you can add policy to `Value.node.tolerations` to configure customized toleration for nodes.
```
ConnorJC3, line 40: Grammar Nit: "all taint" -> "all taints"

ConnorJC3, line 40: Grammar Nit: "deployment, you can add policy" -> "deployment. Add policies"

ConnorJC3, line 40: Grammar Nit: "; the" -> "; and the"

ConnorJC3 (commented):
All the code changes lgtm, spotted a few spots the docs grammar could be improved

ConnorJC3 (approved):
lgtm 👍

torredil (approved):
/lgtm

PR #1401, opened 2022-09-27
Update workflows

torredil:
- Add workflow to scan container images (CSI driver + sidecars) when changes are made in `'deploy/kubernetes/overlays/stable/gcr/kustomization.yaml'` and daily.
- Add workflow to scan on repo mode which helps identify potential vulnerabilities that might get introduced with new PRs.
- Remove `codeQL` workflow.


.github/workflows/trivy-sidecars.yaml
```diff
@@ -0,0 +1,49 @@
+name: Trivy Scanner - Container Images
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'deploy/kubernetes/overlays/stable/gcr/kustomization.yaml'
+  schedule:
+   # Run “At minute 0 past every 24th hour.”
```
ConnorJC3, line 9: Remove this comment

.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Run Trivy vulnerability scanner in repo mode
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: 'fs'
+          ignore-unfixed: true
+          format: 'sarif'
+          output: 'trivy-ebs-csi-results.sarif'
```
ConnorJC3, line 19: Probably just call this `results.sarif` to be consistent with the other workflow?

.github/workflows/trivy-sidecars.yaml
```diff
@@ -0,0 +1,49 @@
+name: Trivy Scanner - Container Images
```
ConnorJC3, line 1: I think this file should be renamed to `trivy-containers.yaml` or similar now that we're not just scanning the sidecars

.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,25 @@
+name: Trivy Scanner - EBS CSI Driver
```
ConnorJC3, line 1: Change name to "Trivy Scanner"

ConnorJC3 (changes requested):

.github/workflows/trivy.yaml
torredil: Done.

PR #1409, opened 2022-09-30
Add support for io2 Block Express volumes

ConnorJC3:


ConnorJC3:
/hold for reviews

Me and @torredil manually tested, this is good for review.

torredil:
```
$ kubectl describe node i-0a8088f85b6d3bedb          
                             
Name:               i-0a8088f85b6d3bedb
Roles:              node
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=r5b.large
                    node.kubernetes.io/instance-type=r5b.large
```

```
$ kubectl get pvc           
                           
NAME        STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
ebs-claim   Bound    pvc-862780bc-9b3b-4abb-8c3f-6f3c303a64d2   2000Gi     RWO            ebs-sc         2m51s
```

```
$ cat storageclass.yaml   
                                                           
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
parameters:
  type: io2
  iops: "100000"
provisioner: ebs.csi.aws.com
```

```
$ cat claim.yaml      
                                                            
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 2000Gi
```

image


/lgtm
/hold

torredil:
cc: @rdpsin

rdpsin:
/lgtm

Can we also confirm that setting IOPS greater than 64K on non-bx instances will return an error? (Tbf, I don't see why not, but would still like to confirm).

rdpsin:
FYI, it'd be useful to add some details in the commit message for posterity.

rdpsin:
Thanks.

/lgtm

ConnorJC3:
@rdpsin there's no way to know at the time of creation what type of instance the volume will be attached to. so it will be created, but fail to attach to instances that do not support block express.

Unfortunately the error message is somewhat poor (`Could not attach volume "vol-006100676decdeaa0" to node "i-01365b727b32e9f74": attachment of disk "vol-006100676decdeaa0" failed, expected device to be attached but was detaching`), but I believe that is something that has to be improved on the EBS side.

rdpsin:
I see. This can potentially lead to leaked resources. As long as we are explicit in our docs, I guess it should be fine. Can you add this?

docs/parameters.md
```diff
@@ -3,25 +3,27 @@ There are several optional parameters that may be passed into `CreateVolumeReque
...
+| "encrypted"                  | true, false                                        | false   | Whether the volume should be encrypted or not. Valid values are "true" or "false".                                                                                                                                                                                                                                                                                                             |
+| "blockExpress"               | true, false                                        | false   | Enables the creation of [io2 Block Express volumes](https://aws.amazon.com/ebs/provisioned-iops/#Introducing_io2_Block_Express) by increasing the limit for io2 volumes to 256000 for io volumes.                                                                                                                                                                                              |
```
torredil, line 15: nit: "by increasing the limit for io2 volumes to 256000 for io volumes." -> "by increasing the IOPS limit for io2 volumes to 256,000." Would be good to also add a note here specifying this parameter is only supported for certain instance types.

torredil (commented):
Discussed offline and decided against instance type validation using [metadata](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/562c0a236d8b61c2f5900325a554bf7bb7cddf8c/pkg/cloud/metadata.go#L57) during volume mounting to:

1. Avoid introducing new dependency on metadata. Long-term, we want to move to a world where metadata is an entirely optional dependency and not a requirement on both the controller & node.
2. Avoid being fragile to changes by requiring a CSI driver update when/if EBS supports block express on other instance types. It would be best to rely on the EBS api as source of truth. The downside here is that setting IOPS > 64K on non-bx supported instances results in `failed, expected device to be attached but was detaching` which is not very helpful. 

Manually tested, these changes lgtm.

docs/parameters.md
ConnorJC3: Reworded, also added a note about instance types.

pkg/driver/controller.go
```diff
@@ -189,6 +196,10 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}
 
+	if volumeType != cloud.VolumeTypeIO2 && blockExpress {
```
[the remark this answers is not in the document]
ConnorJC3: Fixed

pkg/cloud/cloud.go
```diff
@@ -63,15 +63,16 @@ const (
...
-	io2MaxIOPSPerGB = 500
-	gp3MaxTotalIOPS = 16000
-	gp3MinTotalIOPS = 3000
-	gp3MaxIOPSPerGB = 500
+	io1MinTotalIOPS   = 100
+	io1MaxTotalIOPS   = 64000
+	io1MaxIOPSPerGB   = 50
+	io2MinTotalIOPS   = 100
+	io2MaxTotalIOPS   = 64000
+	io2BxMaxTotalIOPS = 256000
```
rdpsin, line 71: np: Can this be `io2BlockExpressMaxTotalIOPS`.

pkg/driver/controller.go
```diff
@@ -174,6 +175,12 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		case PVNameKey:
 			volumeTags[PVNameTag] = value
 			tProps.PVName = value
+		case BlockExpressKey:
+			if value == "true" {
+				blockExpress = true
+			} else {
+				blockExpress = false
+			}
```
rdpsin, line 183: np: Don't really need the `else` clause.

ConnorJC3: Fair enough, changed this for `isEncrypted` to be consistent

torredil (approved):
/unhold

PR #1418, opened 2022-10-06
Add parameter to force enable snapshotter sidecar

### LINKED ISSUE #1416: Controller's csi-snapshotter container enablement from values file
I want to be able to enable controller's csi-snapshotter container from values file not only to depend on api-versions from the server. This is needed in case to test with helm template command which is done locally not on the server-side.

/feature

A csiSNapshotter property can be added in the values file and checked in controller template with additional or statement for local tests with helm template.

Screenshot 2022-10-05 at 17 36 26
Screenshot 2022-10-05 at 17 35 44

**Additional context**
There is an issue with integrating the helm chart and using snapshot functionality in argo cd because argo is using helm template command with --api-versions flag but the api-versions that external-snapshotter is adding (snapshot.storage.k8s.io/v1) are not seen - they are added after argo deployment. That way the csi-snapshotter container can't be deployed.

ConnorJC3:

New feature

Fixes #1416 

**What testing is done?** 

Manually tested with `helm template`

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -258,7 +258,7 @@ spec:
           securityContext:  
             {{- toYaml . | nindent 12 }}
           {{- end }}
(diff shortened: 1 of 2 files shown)


torredil (approved):
/lgtm
/hold

PR #1419, opened 2022-10-06
Add serviceMonitor.labels parameter

### LINKED ISSUE #1420: [helm] Don't check capabilities when `enableMetrics: true`
/kind bug

**What happened?**
ServiceMonitor resource is guarded behind the `.Capabilities.APIVersions.Has` check. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1380/files#diff-2412b20598542fa98f5a07543074554a7ca2bf9fa414b35e7cbc83a429f3c142R19

This check is problematic, as in order to resolve it `helm template` should be executed with `--validate` flag and connect to cluster that has CRDs installed.

I would strongly prefer to have `.Capabilities.APIVersions.Has` check dropped or replaced with user settable value, like `serviceMonitor: true`

Try to install this chart with `enableMetrics: true` and cluster that *doesn't* have prometheus-operator CRDs installed. I would strongly prefer to see helm error, that ServiceMonitor is unknown resource kind, instead of having it skipped and chart installed.

`.Capabilities.APIVersions.Has` is good to deal with kubernetes native resources that have multiple versions available, like `policy/v1` and `policy/v1beta1`. In most cases this function can be successfully resolved without need to connect to cluster. But it isn't suitable to check CRDs availability, where installation failure is desirable outcome instead of having resource silently omitted.

**Environment**
- Kubernetes version (use `kubectl version`): 1.23
- Driver version: v1.12.0

CC @torredil

torredil:

- Allow modifying `labels` for the `ServiceMonitor` object. 
- Parameterize the hardcoded namespace in `metrics.yaml`.
- Address https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1380#discussion_r988904466
- closes  #1420

**What testing is done?** 
- Rendered Helm chart.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/metrics.yaml
+++ b/charts/aws-ebs-csi-driver/templates/metrics.yaml
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: ebs-csi-controller
-  namespace: kube-system
+  namespace: {{ .Release.Namespace }}
   labels:
     app: ebs-csi-controller
 spec:
@@ -16,22 +16,24 @@ spec:
       targetPort: 3301
   type: ClusterIP
 ---
-{{- if (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") -}}
+{{- if or .Values.controller.serviceMonitor.forceEnable (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: ebs-csi-controller
-  namespace: kube-system
+  namespace: {{ .Release.Namespace }}
   labels:
     app: ebs-csi-controller
-    release: prometheus
+    {{- if .Values.controller.serviceMonitor.labels }}
+    {{- toYaml .Values.controller.serviceMonitor.labels | nindent 4 }}
+    {{- end }}
 spec:
   selector:
     matchLabels:
       app: ebs-csi-controller
   namespaceSelector:
     matchNames:
-      - kube-system
+      - {{ .Release.Namespace }}
   endpoints:
     - targetPort: 3301
       path: /metrics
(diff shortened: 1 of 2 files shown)


torredil:
Addressing #1420 in this PR.

- `.Values.controller.serviceMonitor.forceEnable = true` & CRDs not installed.

```
$ helm template ./ | grep -A 20 "monitoring.coreos.com/v1"

apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: ebs-csi-controller
  namespace: default
  labels:
    app: ebs-csi-controller
    release: prometheus
spec:
  selector:
    matchLabels:
      app: ebs-csi-controller
  namespaceSelector:
    matchNames:
      - default
  endpoints:
    - targetPort: 3301
      path: /metrics
      interval: 15s
```

- `.Values.controller.serviceMonitor.forceEnable = false` & CRDs not installed.
```
$ helm template ./ | grep -A 20 "monitoring.coreos.com/v1"
```

charts/aws-ebs-csi-driver/templates/metrics.yaml
```diff
@@ -16,22 +16,24 @@ spec:
       targetPort: 3301
   type: ClusterIP
 ---
-{{- if (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") -}}
+{{- if or (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") (.Values.controller.serviceMonitor.forceEnable) }}
```
ConnorJC3, line 19: Nit: parenthesis not needed around `forceEnable` value

ConnorJC3 (changes requested):

charts/aws-ebs-csi-driver/templates/metrics.yaml
torredil: Done.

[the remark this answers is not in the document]
torredil: Done.

ConnorJC3 (approved):
/lgtm

PR #1427, opened 2022-10-13
refactor: Move the default controller tolerations in the helm chart values

Linutux42:
Give more flexibility to the helm chart users for the csi-controller tolerations.
The default values remains unchanged and the default behavior is also unchanged.
However, one with different needs regarding those tolerations can modify them at will.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -41,11 +41,6 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       tolerations:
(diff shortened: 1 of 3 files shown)


ConnorJC3:
This seems to have caused helm to reorder the tolerations for some reason ¯\\\_(ツ)\_/¯

@Linutux42 can you run `make generate-kustomize` and commit the result, thanks.

torredil:
/lgtm
Thank you for the contribution 👍 
let's squash the commits before merging.

Linutux42:
@torredil I've squashed my commits into a single one as requested :)

PR #1436, opened 2022-10-28
helm: Expose volumes and volumeMounts for the ebs-csi-controller deployment

cnmcavoy:
Feature. 

Allows users who wish to integrate the controller with IRSA authentication manually (i.e not using the https://github.com/aws/amazon-eks-pod-identity-webhook) to configure the controller with the correct mounts for authentication. For example:
```
    volumes:
    - name: aws-token
      projected:
        sources:
        - serviceAccountToken:
            audience: "sts.amazonaws.com"
            expirationSeconds: 86400
            path: token
    volumeMounts:
    - mountPath: "/var/run/secrets/eks.amazonaws.com/serviceaccount/"
      name: aws-token
  ```

**What testing is done?** 
 Helm template verification.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -54,7 +54,7 @@ spec:
         {{- $constraints | toYaml | nindent 8 }}
       {{- end }}
       {{- with .Values.controller.securityContext }}
-      securityContext:  
+      securityContext:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       {{- with .Values.controller.initContainers }}
@@ -131,6 +131,9 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
+          {{- with .Values.controller.volumeMounts }}
+          {{- toYaml . | nindent 12 }}
+          {{- end }}
           ports:
             - name: healthz
               containerPort: 9808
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: charts/aws-ebs-csi-driver/templates/controller.yaml [hunk 1]
```diff
@@ -203,12 +206,15 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
+          {{- with .Values.controller.volumeMounts }}
+          {{- toYaml . | nindent 12 }}
+          {{- end }}
```
ConnorJC3, line 211: Is there a reason the sidecar containers also need access to these mounts? If not, could you remove them. If yes, could you separate out a config value per-sidecar.

charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -345,3 +363,6 @@ spec:
       volumes:
         - name: socket-dir
           emptyDir: {}
+        {{- with .Values.controller.volumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
```
ConnorJC3, line 368: nit: add missing newline to end of file

charts/aws-ebs-csi-driver/templates/controller.yaml
cnmcavoy: Wasn't necessary/

PR #1450, opened 2022-11-16
Update example policy

torredil:

-  Remove invalid policy conditions.
-  Introduce a new policy condition that allows for deleting volumes created by the in-tree storage plugin.

### DIFF
--- a/docs/example-iam-policy.json
+++ b/docs/example-iam-policy.json
@@ -72,18 +72,6 @@
     {
       "Effect": "Allow",
       "Action": [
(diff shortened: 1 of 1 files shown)


kevarr:
@torredil Related documentation has drifted a little bit here [install.md#migrating-from-in-tree-ebs-plugin](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md#migrating-from-in-tree-ebs-plugin)

PR #1452, opened 2022-11-21
Support specifying block size for filesystem format

ConnorJC3:

New feature

Adds support to specify the block size during filesystem format. Currently not supported on windows (or for NTFS volumes at all), that can change once the csi-proxy is replaced with privileged Windows containers.

**What testing is done?** 

Manual/New unit tests/CI


rdpsin:
Different filesystems have different valid block sizes. E.g., `ext4` will only accept 1K, 2K or 4K block sizes. Should the driver check for that?

rdpsin (changes requested):

docs/parameters.md
```diff
@@ -14,6 +14,7 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
...
 | "kmsKeyId"                   |                                                    |         | The full ARN of the key to use when encrypting the volume. If not specified, AWS will use the default KMS key for the region the volume is in. This will be an auto-generated key called `/aws/ebs` if not changed.                                                                                                                                                                            |
+| "blockSize"                  |                                                    |         | The block size to use when formatting the underlying filesystem. Only supported on linux nodes and with fstype `ext2`, `ext3`, `ext4`, or `xfs`.                                    |
```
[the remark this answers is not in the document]
ConnorJC3: For now, I'll leave it as is - we can easily add the new option and deprecate the old if/when this change happens. We should consider bringing it up at the next SIG Storage meeting, but this will likely be a long path before standardization.

pkg/driver/node.go
```diff
@@ -111,6 +112,17 @@ func newNodeService(driverOptions *DriverOptions) nodeService {
 func (d *nodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
 	klog.V(4).Infof("NodeStageVolume: called with args %+v", *req)
 
+	context := req.GetVolumeContext()
+	blockSize, ok := context[BlockSizeKey]
+	if ok {
+		// This check is already performed on the controller side
+		// However, because it is potentially security-sensitive, we redo it here to be safe
```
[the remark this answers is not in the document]
ConnorJC3: Shortly after this, the block size is directly inserted into `formatOptions` and passed to `mount-utils`. An attacker that could theoretically control `blockSize` here could use that on linux to insert arbitrary `mkfs` parameters (a potential security threat). On Windows, it's a complete takeover: formatOptions is concatenated into the string.

In theory, this value is checked in the controller, but I'm not 100% confident there's no way to induce a value here without going through the controller (on the contrary, I'm pretty confident it is possible by directly manipulating `CSIVolume` objects).

Checking that it looks like an integer here is practically free and makes 100% sure no malicious values ever get passed to `formatOptions`

pkg/driver/controller.go
```diff
@@ -178,6 +179,12 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 			if value == "true" {
 				blockExpress = true
 			}
+		case BlockSizeKey:
+			_, err = strconv.Atoi(value)
+			if err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Could not parse blockSize: %v", err)
```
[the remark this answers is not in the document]
ConnorJC3: Fixed

pkg/driver/controller.go
```diff
@@ -187,6 +194,26 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}
 
+	if len(blockSize) > 0 {
+		for _, volCap := range req.GetVolumeCapabilities() {
+			switch volCap.GetAccessType().(type) {
+			case *csi.VolumeCapability_Block:
+				return nil, status.Error(codes.InvalidArgument, "Cannot use block size with block volume")
```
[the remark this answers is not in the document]
ConnorJC3: I will test this and report back, but I believe it is just silently ignored. That feels really wrong here though, and I strongly believe we should raise an error - the user requested we set the block size in a situation where we definitely cannot.

docs/parameters.md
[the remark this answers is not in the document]
rdpsin: We should open an issue in the CSI spec repo to gauge interest.

torredil:
> Different filesystems have different valid block sizes. E.g., ext4 will only accept 1K, 2K or 4K block sizes. Should the driver check for that?

`mkfs` rounds up to a valid block size.

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
parameters:
  blocksize: "3535"
  csi.storage.k8s.io/fstype: ext4
  
$ kubectl describe pv

Name:              pvc-bd9fc0a3-6481-4811-8ad4-b2022b8241c8
Labels:            <none>
Annotations:       pv.kubernetes.io/provisioned-by: ebs.csi.aws.com
Finalizers:        [kubernetes.io/pv-protection external-attacher/ebs-csi-aws-com]
StorageClass:      ebs-sc
Status:            Bound
Claim:             default/ebs-claim
Reclaim Policy:    Delete
Access Modes:      RWO
VolumeMode:        Filesystem
Capacity:          4Gi
Node Affinity:
  Required Terms:
    Term 0:        topology.ebs.csi.aws.com/zone in [us-east-1d]
Message:
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            ebs.csi.aws.com
    FSType:            ext4
    VolumeHandle:      vol-0d6ec6acb3bbdbdd7
    ReadOnly:          false
    VolumeAttributes:      blocksize=3535
                           storage.kubernetes.io/csiProvisionerIdentity=1669217011030-8081-ebs.csi.aws.com
Events:                <none>

$ kubectl logs ebs-csi-node-qtqpt -n kube-system -c ebs-plugin

I1123 19:07:54.622264       1 mount_linux.go:528] Disk "/dev/nvme1n1" appears to be unformatted, attempting to format as type: "ext4" with options: [-b 3535 -F -m0 /dev/nvme1n1]
I1123 19:07:55.111667       1 mount_linux.go:538] Disk successfully formatted (mkfs): ext4 - /dev/nvme1n1 /var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/20ca089c5b265bd40055b47b3b882793681fd817a59484c5f1bf70f8bd5a6523/globalmount 

$ root@app:/# tune2fs -l /dev/nvme0n1p1 |grep '^Block size:'
Block size:               4096
```

pkg/driver/controller.go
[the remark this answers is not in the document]
ConnorJC3: Update: it does silently fail, but this case is different, because (at least on k8s) the `external-attacher` _always_ injects the default fstype, even on block volumes, so we are incapable of detecting presence/absence.

pkg/driver/controller.go
```diff
@@ -58,6 +58,17 @@ var (
 
 const isManagedByDriver = "true"
 
+// BlockSizeFSTypes contains the filesystems that a custom block size is supported on
+var (
+	BlockSizeFSTypes = map[string]struct{}{
```
wmesard, line 63: Why is this exported? Suggest `s/BlockSizeFSTypes/blockSizeFSTypes/`.

But really since this list is exactly `ValidFSTypes`, why is it being added at all?

pkg/driver/controller_test.go
```diff
@@ -147,7 +147,9 @@ func TestCreateVolume(t *testing.T) {
 	stdVolCap := []*csi.VolumeCapability{
 		{
 			AccessType: &csi.VolumeCapability_Mount{
-				Mount: &csi.VolumeCapability_MountVolume{},
+				Mount: &csi.VolumeCapability_MountVolume{
+					FsType: "ext4",
+				},
```
wmesard, line 152: Suggest _not_ touching this variable, since you are implicitly modifying three dozen existing unit tests. If the new block size code can't handle `defaultFsType`, I think that's a bug that should be fixed, not worked around in unit tests.

But it is worthwhile to test both cases here (when fsType is both specified and not).

wmesard (changes requested):

pkg/driver/controller.go
ConnorJC3: oops, artifact of copy-pasting `ValidFSTypes`, fixed

pkg/driver/controller_test.go
ConnorJC3: The problem here is that the default fsType is de-facto set by the external-attacher: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/86f165e433eb7a7aff14a56540d9dd695dc1a756/charts/aws-ebs-csi-driver/templates/controller.yaml#L195

This means that if the `fsType` received for `CreateVolume` is blank I have no idea what `fsType` I'll be getting later when the volume is attached, and by then it will be too late.

No other code in `CreateVolume` touches the fstype so that's why I didn't bother making a separate test setup for it, but I can if you think it's relevant.

pkg/driver/controller.go
wmesard: > But really since this list is exactly `ValidFSTypes`, why is it being added at all?

Care to comment on this? My concern is that this is a bug-in-waiting. Some day, someone is going to add a new entry to  `ValidFSTypes` and forget to add it here. So the benefit it's providing now should outweigh the cost of that future bug.

ConnorJC3: Oh, that's actually a typo on my part, this list should _not_ contain NTFS, as we don't (currently) support custom block sizes on that filesystem. To be more general: the difference is that `ValidFSTypes` contains _all_ possible fstypes, and this array contains the fstypes that we support custom block sizes for.

wmesard: In that case, I suggest making it a negative list instead of a positive one. That is, we can safely assume that we can specify a block size for a file system unless it is in a list of known defectives.

Rationale:
* The worst-case outcome of a "supported" list is the future bug I described above. This list will have to be updated whenever a new file system type is added.
* The worst-case outcome of a "not supported" list is that some customer will try to set the block size and it will either be ignored or will cause the FS creation to fail. This list will likely never be touched again.

I think the second outcome is the lesser of two evils.

pkg/driver/controller_test.go
wmesard: > The problem here is that the default fsType is de-facto set by the external-attacher

Then I'm not sure how I feel about the `defaultFsType` [code](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L49) in _pkg/driver/node.go_. Confused, I guess. In any case, as long as the default supports block size configuration, everything should just work, right? I don't love the idea of baking that assumption into a unit test. But as assumptions go, that one seems pretty safe.

> No other code in CreateVolume touches the fstype **today**

FTFY 😄. Let's write unit tests that make the fewest possible assumptions about the code they're testing.

wmesard:
> Different filesystems have different valid block sizes. E.g., `ext4` will only accept 1K, 2K or 4K block sizes. Should the driver check for that?

That seems like a violation of [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). Do these file systems have a stable interface to query their supported block sizes? Are we going to update the CSI driver when the FS or the underlying hardware adds support for new block sizes? How will we know if software and hardware on the node supports the new block size?

Also, that list only applies to x86. ext4 on ARM supports larger block sizes *today*.

Let components do its own input validation. As a caller, just make sure you are handling and reporting the errors correctly.

pkg/driver/node.go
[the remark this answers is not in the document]
wmesard: It's not our interface, so I'm not sure how we can validate the input. Is it a number of bytes? Number of pages? Does the underlying command allow a "k" suffix? What about "Ki"? Does it allow negative values? `mkfs.ext4` does, `mkfs.xfs` does not.

Unless you're saying that we should *make* it our interface, validate the input, and take on the job of converting it to whatever format is appropriate for the underlying file system. I don't think we should do that.

pkg/driver/controller_test.go
ConnorJC3: > Then I'm not sure how I feel about the defaultFsType [code](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L49) in pkg/driver/node.go.

That code will handle the case if the volume is attached with a blank fstype. With kubernetes, that won't ever actually happen (because, as I mentioned above, the external attacher has its own default fstype logic), but it is possible with other COs.

We can move to assuming the default fstype will support a custom block size. Today, that's a relatively safe option, as it is highly unlikely anybody is using NTFS as their default filesystem, and that is the only filesystem for which custom block sizes is not supported.

pkg/driver/controller.go
ConnorJC3: Moved to exclusion-based list, and now check in both the controller and node (to account for when the controller allows through an invalid fstype)

pkg/driver/controller_test.go
ConnorJC3: Because of the discussion above, the controller code will now allow blank fstypes (and thus the case when the default fstype is invalid will fail on mount), so this change is no longer necessary. 

I agree we should test a variety of capabilities here, but that's a relatively big change and out of scope for this PR - I will raise a PR at a later date to improve this testing.

ConnorJC3:
/remove-hold

I believe I've addressed all the existing feedback, this PR is ready for a fresh round of review and otherwise ready to merge.

pkg/driver/node.go
```diff
@@ -155,6 +145,24 @@ func (d *nodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Errorf(codes.InvalidArgument, "NodeStageVolume: invalid fstype %s", fsType)
 	}
 
+	context := req.GetVolumeContext()
+	blockSize, ok := context[BlockSizeKey]
+	if ok {
+		// This check is already performed on the controller side
+		// However, because it is potentially security-sensitive, we redo it here to be safe
+		_, err := strconv.Atoi(blockSize)
```
wmesard, line 153: But really, neither place should be doing this, because it is not your interface. mkfs.xfs [allows](https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/mkfs/xfs_mkfs.c#n1426) suffixes (e.g. "4k"), while this code will reject them. No big deal I guess. The user will figure out what's happening and work around it. But what about some future hypothetical file system that takes `"small"`, `"medium"`, and `"large"` instead of a number?

wmesard (approved):

pkg/driver/node.go
ConnorJC3: I can try to make the validation smarter, but I'm of the opinion we should just leave it dumb. As explained above, we cannot accept arbitrary values here, because it leads to an RCE (and even when it doesn't, a significant security risk). We'd have to rebuild the validation for every filesystem ourselves, and getting it wrong here means exposing our users/customers to security holes.

All existing filesystems I'm aware of accept this value as an integer. It's theoretically possible, but I strongly doubt any future filesystem will change that.

torredil (approved):

Issue #1390, opened 2022-09-16
helm fsGroupPolicy patch fails with immutable field


bcha:
/kind bug

**What happened?**

Upgrading with helm..
```sh
-     helm.sh/chart: aws-ebs-csi-driver-2.10.1
-     app.kubernetes.io/version: "1.11.2"
+     helm.sh/chart: aws-ebs-csi-driver-2.11.0
+     app.kubernetes.io/version: "1.11.3"
```
...results in `Error: UPGRADE FAILED: cannot patch "ebs.csi.aws.com" with kind CSIDriver: CSIDriver.storage.k8s.io "ebs.csi.aws.com" is invalid: spec.fsGroupPolicy: Invalid value: "File": field is immutable`

**What you expected to happen?**
Smooth upgrade.

**How to reproduce it (as minimally and precisely as possible)?**
Upgrade from helm chart 2.10.1 to 2.11.0

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): 1.21
- Driver version: 1.11.3

z0rc:
Caused by https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1377

ConnorJC3:
Hi all, sorry we didn't catch this!

Firstly, if you want to upgrade to helm chart 1.11.0, you can do so by pre-deleting the `CSIDriver` object, for example with `kubectl delete csidriver ebs.csi.aws.com`. This won't affect existing volumes, but it will prevent the driver from responding to future requests (similar to if the driver container had crashed), so follow it with an upgrade.

If you can't delete the `CSIDriver` object for any reason, I'm pushing an additional parameter `useOldCSIDriver` that can be set to `true` to use the old object. This will allow an upgrade from an old version without needing to recreate the `CSIDriver` object. You can track that change in #1391. That PR also adds a note to the helm chart changelog about this issue.

ConnorJC3:
cc @olemarkus I know you were going to update the kops version of the addon soon, you'll either need custom logic to replace the `CSIDriver` or opt to use the old version on existing clusters.

torredil:
If you run into this while upgrading please read @ConnorJC3's comment above and take a look at the Helm chart [release notes](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/CHANGELOG.md), thanks.

PR #1454, opened 2022-11-21
Add test-helm-chart target to Makefile to test Helm chart upgrades

torredil:
- Add test-helm-chart target to Makefile to test Helm chart upgrades
- Prevent this class of issue: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1390

**What testing is done?** 
- Manual
- CI

1. Pull changes locally.
2. Run `CLUSTER_TYPE=kops KOPS_STATE_FILE="s3://devcluster-state-e2e" make test-helm-chart`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,6 +204,14 @@ test-e2e-external-eks:
 	GINKGO_SKIP="\[Disruptive\]|\[Serial\]" \
 	./hack/e2e/run.sh
 
+.PHONY: test-helm-chart
+test-helm-chart:
+	AWS_REGION=us-west-2 \
+	AWS_AVAILABILITY_ZONES=us-west-2a,us-west-2b,us-west-2c \
+	EBS_INSTALL_SNAPSHOT="true" \
+	HELM_CT_TEST="true" \
+	./hack/e2e/run.sh
+
 .PHONY: verify-vendor
 test: verify-vendor
 verify: verify-vendor
--- a/charts/aws-ebs-csi-driver/templates/tests/helm-tester.yaml
+++ b/charts/aws-ebs-csi-driver/templates/tests/helm-tester.yaml
@@ -1,0 +1,202 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: helm-sa
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: test-role
+rules:
+  - apiGroups: [ "" ]
+    resources:
+      - events
+      - nodes
+      - pods
+      - replicationcontrollers
+      - serviceaccounts
+      - configmaps
+      - persistentvolumes
+      - persistentvolumeclaims
+    verbs: [ "list" ]
+  - apiGroups: [ "" ]
+    resources:
+      - services
+      - nodes
+      - nodes/proxy
+      - persistentvolumes
+      - persistentvolumeclaims
+      - pods
+      - pods/log
+    verbs: [ "get" ]
+  - apiGroups: [ "" ]
+    resources:
+      - namespaces
+      - persistentvolumes
+      - persistentvolumeclaims
+      - pods
+      - pods/exec
+    verbs: [ "create" ]
+  - apiGroups: [ "" ]
+    resources:
+      - namespaces
+      - persistentvolumes
+      - persistentvolumeclaims
+      - pods
+    verbs: [ "delete" ]
+  - apiGroups: [ "" ]
+    resources:
+      - persistentvolumeclaims
+    verbs: [ "update" ]
+  - apiGroups: [ "" ]
+    resources:
+      - pods/ephemeralcontainers
+    verbs: [ "patch" ]
+  - apiGroups: [ "" ]
+    resources:
+      - serviceaccounts
+      - configmaps
+    verbs: [ "watch" ]
+  - apiGroups: [ "apps" ]
+    resources:
+      - replicasets
+      - daemonsets
+    verbs: [ "list" ]
+  - apiGroups: [ "storage.k8s.io" ]
+    resources:
+      - storageclasses
+    verbs: [ "create" ]
+  - apiGroups: [ "storage.k8s.io" ]
+    resources:
+      - storageclasses
+      - csinodes
+    verbs: [ "get" ]
+  - apiGroups: [ "storage.k8s.io" ]
+    resources:
+      - storageclasses
+    verbs: [ "delete" ]
+  - apiGroups: [ "snapshot.storage.k8s.io" ]
+    resources:
+      - volumesnapshots
+      - volumesnapshotclasses
+      - volumesnapshotcontents
+    verbs: [ "create" ]
+  - apiGroups: [ "snapshot.storage.k8s.io" ]
+    resources:
+      - volumesnapshots
+      - volumesnapshotclasses
+      - volumesnapshotcontents
+    verbs: [ "get" ]
+  - apiGroups: [ "snapshot.storage.k8s.io" ]
+    resources:
+      - volumesnapshotcontents
+    verbs: [ "update" ]
+  - apiGroups: [ "snapshot.storage.k8s.io" ]
+    resources:
+      - volumesnapshots
+      - volumesnapshotclasses
+      - volumesnapshotcontents
+    verbs: [ "delete" ]
+  - apiGroups: [ "authorization.k8s.io" ]
+    resources:
+      - clusterroles
+    verbs: [ "list" ]
+  - apiGroups: [ "authorization.k8s.io" ]
+    resources:
+      - subjectaccessreviews
+    verbs: [ "create" ]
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -204,6 +204,15 @@ test-e2e-external-eks:
 	GINKGO_SKIP="\[Disruptive\]|\[Serial\]" \
 	./hack/e2e/run.sh
 
+.PHONY: test-helm-chart
+test-helm-chart:
+	AWS_REGION=us-west-2 \
+	AWS_AVAILABILITY_ZONES=us-west-2a,us-west-2b,us-west-2c \
+	EBS_INSTALL_SNAPSHOT="true" \
```
ConnorJC3, line 211: Is this necessary? Does this test snapshots in any way?

torredil: It's necessary. We run the `External.Storage` test suite with kubetest2 [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/662cbf898e74ad016f4b1d76cf5e9654131b70aa/charts/aws-ebs-csi-driver/templates/tests/helm-tester.yaml#L192) which includes the snapshot tests, if we don't install the snapshot controller and CRDs those tests will fail.

Technically, we could opt out of testing snapshot functionality here with `GINKGO_SKIP` but I couldn't think of a good reason for specifically skipping these tests.

charts/chart_schema.yaml
```diff
@@ -0,0 +1,37 @@
...
+  repository: str(required=False)
+  condition: str(required=False)
+  tags: list(str(), required=False)
+  enabled: bool(required=False)
+  import-values: any(list(str()), list(include('import-value')), required=False)
+  alias: str(required=False)
+---
+import-value:
+  child: str()
+  parent: str()
```
ConnorJC3, line 37: What is the point of this file? Can we document that? Is there an advantage over setting `validate-chart-schmea: false`?

charts/config.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - charts/aws-ebs-csi-driver
+helm-extra-args: "--timeout 1000s"
+upgrade: true
+skip-missing-values: true
+namespace: kube-system
+release-label: release
+kubectl-timeout: 1000s
+skip-clean-up: false
+print-logs: true
+debug: true
```
ConnorJC3, line 19: Is it not possible to put this file in the test directory? Can it be named something other than `config.yaml`?

ConnorJC3: Also, `debug: true`?

charts/lintconf.yaml
```diff
@@ -0,0 +1,42 @@
...
+    indent-sequences: whatever      # - list indentation will handle both indentation and without
+    check-multi-line-strings: false
+  key-duplicates: enable
+  line-length: disable              # Lines can be any length
+  new-line-at-end-of-file: enable
+  new-lines:
+    type: unix
+  trailing-spaces: enable
+  truthy:
+    level: warning
```
ConnorJC3, line 42: Same questions as `chart_schema.yaml` and `config.yaml`

hack/e2e/chart-testing.sh
```diff
@@ -0,0 +1,17 @@
...
+  INSTALL_PATH=${1}
+  CHART_TESTING_VERSION=${2}
+  if [[ ! -e ${INSTALL_PATH}/chart-testing ]]; then
+    CHART_TESTING_DOWNLOAD_URL="https://github.com/helm/chart-testing/releases/download/v${CHART_TESTING_VERSION}/chart-testing_${CHART_TESTING_VERSION}_linux_amd64.tar.gz"
+    curl --silent --location "${CHART_TESTING_DOWNLOAD_URL}" | tar xz -C "${INSTALL_PATH}"
+    chmod +x "${INSTALL_PATH}"/ct
+  fi
+
+  export PATH=/root/.local/bin:$PATH
+  python3 -m pip install --upgrade pip
```
ConnorJC3, line 15: What is the point of self-upgrading pip?

hack/e2e/chart-testing.sh
```diff
@@ -0,0 +1,17 @@
...
+  python3 -m pip install --user yamllint yamale
```
ConnorJC3, line 16: Do these need to end up on the PATH? My testing with the kubekins image has pip loudly complaining the installation directory is not.

hack/e2e/run.sh
```diff
@@ -76,6 +77,8 @@ TEST_EXTRA_FLAGS=${TEST_EXTRA_FLAGS:-}
 EBS_INSTALL_SNAPSHOT=${EBS_INSTALL_SNAPSHOT:-"false"}
 EBS_INSTALL_SNAPSHOT_VERSION=${EBS_INSTALL_SNAPSHOT_VERSION:-"v6.1.0"}
 
+HELM_TEST=${HELM_TEST:-"false"}
```
ConnorJC3, line 80: I feel like `HELM_TEST` is ambiguous - maybe we should consider something like `HELM_CT_TEST`?

hack/e2e/run.sh
```diff
@@ -263,6 +284,8 @@ else
   loudecho "Not cleaning"
 fi
 
+OVERALL_TEST_PASSED="${TEST_PASSED}"
```
ConnorJC3, line 287: Why is this now here?

Makefile
```diff
@@ -204,6 +204,15 @@ test-e2e-external-eks:
...
+	HELM_EXTRA_FLAGS='--set=controller.k8sTagClusterId=$$CLUSTER_NAME' \
```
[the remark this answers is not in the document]
torredil: Done.

charts/chart_schema.yaml
torredil: Removed this file and set `validate-chart-schmema: false`.

charts/config.yaml
torredil: Done.

hack/e2e/chart-testing.sh
torredil: Due to errors being logged; these have been fixed in the latest kubekins image so this line has now been removed.

hack/e2e/run.sh
torredil: Otherwise we run into `OVERALL_TEST_PASSED: unbound variable` because the variable will not be set if `"${HELM_CT_TEST}" == true`.

torredil:
Attaching a run-through with the latest kubekins image here with this revision:
[log.txt](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/files/10282343/log.txt)

tests/ct-lintconfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+    indent-sequences: whatever      # - list indentation will handle both indentation and without
+    check-multi-line-strings: false
+  key-duplicates: enable
+  line-length: disable              # Lines can be any length
+  new-line-at-end-of-file: enable
+  new-lines:
+    type: unix
+  trailing-spaces: enable
+  truthy:
+    level: warning
```
ConnorJC3, line 42: For this PR, can we consider deleting this and just turning `validate-yaml` off in the config (which I assume is what this is used for?)

This config already has basically all the checks disabled, in my opinion it would be best to follow up later with a more sane config and a revamped chart.

charts/aws-ebs-csi-driver/templates/tests/helm-tester.yaml
```diff
@@ -0,0 +1,201 @@
+apiVersion: v1
```
ConnorJC3, line 1: Place a new line with `---` on it before this one

charts/aws-ebs-csi-driver/templates/tests/helm-tester.yaml
```diff
@@ -0,0 +1,201 @@
...
+kind: Pod
+metadata:
+  name: helm-test
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: helm-test
+      image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20220624-1a63fdd9f2-master
```
ConnorJC3, line 181: Not relevant to this PR, but in the future we should consider getting a bot to auto update this similar to how it's done on test-infra.

hack/e2e/run.sh
ConnorJC3: That doesn't answer the question, shouldn't it be right after line 238?

ConnorJC3 (changes requested):

ConnorJC3:
Commit `8eed2185d3c5f0086f4e3726535ae8c1222cf229` (latest as of commenting) passes with latest kubekins image:

```
Deleted cluster: "test-cluster-15561.k8s.local"
###
## OVERALL_TEST_PASSED: 0
#
###
## SUCCESS!
#
root@483133e2920e:/workspace/aws-ebs-csi-driver# git rev-parse HEAD
8eed2185d3c5f0086f4e3726535ae8c1222cf229
```

/lgtm

ConnorJC3 (approved):

PR #1461, opened 2022-12-19
Chart Cleanup

ConnorJC3:
Add some sensible defaults to the chart (that are already used in the EKS-managed version of the driver), and also add `image.containerRegistry` parameter to allow prefixing all repositories (we will set this to `public.ecr.aws` in the upcoming 1.15.0 driver release when we switch the default to the EKS-D sidecars).

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -3,6 +3,8 @@
 # Declare variables to be passed into your templates.
 
 image:
+  # Container registry to be used, will prefix all repositories (including sidecars)
+  containerRegistry: ""
   repository: public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver
   # Overrides the image tag whose default is v{{ .Chart.AppVersion }}
   tag: ""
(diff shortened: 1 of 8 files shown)


rdpsin:
While #1454 is still pending, can you post the results of manually testing/linting the helm chart upgrade?

ConnorJC3:
Current version passes #1454, snippet from end of testing log:

```
Deleted cluster: "test-cluster-29719.k8s.local"

###
## OVERALL_TEST_PASSED: 0
#
###
## SUCCESS!
# 
```

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -164,12 +177,21 @@ controller:
...
   #   maxUnavailable: 1
-  resources: {}
+  resources:
+    requests:
+      cpu: 10m
+      memory: 40Mi
+    limits:
+      cpu: 100m
+      memory: 256Mi
   # We usually recommend not to specify default resources and to leave this as a conscious
```
torredil, line 195: Now that we specify default resource limits, we should update this comment.

torredil (approved):

PR #1465, opened 2022-12-27
Update README.md

torredil:
- Update compatibility information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The [Amazon Elastic Block Store](https://aws.amazon.com/ebs/) Container Storage 
 
 ## Releases
 
-The EBS CSI Driver publishes monthly releases and bi-weekly Helm chart releases. Security vulnerabilities will be addressed as soon as possible.
+The EBS CSI Driver publishes monthly releases. Unscheduled releases may be published for patches to security vulnerabilities and other fixes deemed urgent.
 
 The EBS CSI Driver follows [semantic versioning](https://semver.org/). The version will be bumped following the rules below:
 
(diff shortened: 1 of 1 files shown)


rdpsin:
Can you update the commit description and message to more accurately reflect what's going on in the PR? (`Update README.md` seems very vague).

rob-howie-depop:
You missed a reference to the "compatibility matrix" here:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md#deploy-driver

I just spent an hour reading through AWS pages and this repo, before realising that someone might have deleted it!

torredil:
@rob-howie-depop Thanks for catching that! https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1484

chrisRedwine:
I'm curious why this was removed. The current [docs](https://github.com/kubernetes-sigs/aws-ebs-csi-driver?tab=readme-ov-file#compatibility) just state that the AWS EBS CSI Driver is compatible with all versions supported by K8s and/or EKS - but surely, for example, `0.1.0` wouldn't work with K8s `1.30` - so it's unclear to me what this actually means.

ConnorJC3:
@chrisRedwine we only test/maintain compatibility for supported versions (i.e. as of writing, the most recent two minor versions). As such, our README accurately reflects this: https://github.com/kubernetes-sigs/aws-ebs-csi-driver?tab=readme-ov-file#compatibility

Older versions may work but we don't test their compatibility and thus don't publish information about it. Additionally, such old versions are not supported and very likely are missing bugfixes, dependency updates to address CVEs, and other similar important changes, so we would strongly recommend sticking with the latest stable version if you are installing/upgrading the EBS CSI Driver.

PR #1479, opened 2023-01-13
Migrate Trivy workflow to grab images from values.yaml

ConnorJC3:


.github/workflows/trivy-containers.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
         with:
           # Dynamically build the matrix of images to scan using `deploy/kubernetes/overlays/stable/gcr/kustomization.yaml`
```
torredil, line 24: This comment is no longer true.

torredil (approved):
/lgtm

PR #1483, opened 2023-01-16
Prevent WaitForAttachmentState from being piled up

### LINKED ISSUE #1482: csi-attacher falls into unrecoverable state on request spike
/kind bug

**What happened?**
In my cluster, there are many EBS volumes. Previously, I rolled all nodes out, and then csi-attacher went bad state. `AttachVolume`/`DetachVolume` calls failed for 3+ hours. I inspected why it was happen, and that was because ebs-csi-driver won't cancel to process the request even when Kubernetes times out. Therefore, `ControllerPublishVolume` and `ControllerUnpublishVolume` requests has piled up and made vicious cycle (resulting in very high RPS to DescribeVolume).
I understand the behavior is to complete the request anyway so the driver could immediately return for next the request. However, the failing calls were `WaitForAttachmentState`, which is not related to operation cancellation. I want to submit a PR which handles `context canceled` error to prevent `WaitForAttachmentState` calls from being stacked.

WaitForAttachmentState should not be piled up.

Spike request `ControllerPublishVolume`/`ControllerUnpublishVolume` to make `DescribeVolume` calls fail

**Environment**
- Kubernetes version (use `kubectl version`): 1.23
- Driver version: 1.1.0

qbx2:
Close #1482

bug fix

We should cancel `WaitForAttachmentState` to prevent it from being piled up. It is not related to operation completion, so this PR won't affect [the behavior](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/3ce83e588de90bb4eac34af17122b0f0e56aa2fa/docs/design.md?plain=1#L47).

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -606,6 +606,8 @@ func (c *cloud) WaitForAttachmentState(ctx context.Context, volumeID, expectedSt
 					klog.Warningf("Waiting for volume %q to be attached but the volume does not exist", volumeID)
 					return false, err
 				}
+			} else if errors.Is(err, context.Canceled) {
+				return true, nil
 			}
 
 			klog.Warningf("Ignoring error from describe volume for volume %q; will retry: %q", volumeID, err)


qbx2:
@torredil Ping. Could you please review?

torredil:
This issue has been addressed by https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1621.

/close

PR #1485, opened 2023-01-17
feature: remove depricated kubernetes api

RiegerSebastian:
The PodDisruptionBudget api "policy/v1" is deprecated since v1.21 and will be removed with v1.25.
This API is available since v1.21


ConnorJC3:
/close

As I explained above, this change is incompatible with some versions of k8s our users use. If you have a more specific suggestion or need support, feel free to file an issue.

PR #1494, opened 2023-01-24
Helm allowed topologies

nichoio:

This is a small fix for Helm chart.

This PR is just a small addition to the Helm chart. The `storageClass` template was missing the `allowedTopologies` field. `allowedTopologies` is a feature relevant for EBS storage classes because it allows choice of a specific availability zone (AZ) when provisioning `PersistentVolumes` (PV). Selecting a specific AZ can prevent edge cases where a PV was once provisioned in some AZ and is later supposed to be reused by some other EC2 instance which cannot provisioned to same AZ (certain EC2 instance types are not available in all AZs).

**What testing is done?** 

Deployed to an EKS v1.23.14 cluster for testing. I ran `make generate-kustomize` but kustomize manifests are not affected because per default, Helm chart does not render any `storageClass` manifests.

### DIFF
--- a/charts/aws-ebs-csi-driver/Chart.yaml
+++ b/charts/aws-ebs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.15.0
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 2.16.0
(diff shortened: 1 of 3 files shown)


nichoio:
I'll close this and recreate properly so that commit ID matches with my Github user ID. Rookie mistake.

PR #1495, opened 2023-01-24
Volume Health Monitoring

ConnorJC3:

Feature

Implements https://kubernetes.io/docs/concepts/storage/volume-health-monitoring/

**What testing is done?** 

Manually tested, as well as significant additional tests in-driver


ConnorJC3:
/hold

Known caveats:

Using non EKS-D sidecar until their team can build the external health monitor sidecar as well.

We should consider what the appropriate timing is for health checks, the default is once per minute per volume.

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -89,6 +89,17 @@ sidecars:
     securityContext:
       readOnlyRootFilesystem: true
       allowPrivilegeEscalation: false
+  healthMonitor:
```
torredil, line 92: We should consider allowing users to specify their own monitoring interval via the `monitor-interval` flag: https://github.com/kubernetes-csi/external-health-monitor/blob/f11edf7690317fbad79ac0506a166eef3d78bf18/cmd/csi-external-health-monitor-controller/main.go#L55

docs/volume-health-monitoring.md
```diff
@@ -0,0 +1,30 @@
...
+
+The controller must be passed the CLI flag `--volume-health-monitoring=true` (exposed via the Helm parameter `controller.volumeHealthMonitoring`).
+
+Additionally, the controller needs to run with the [External Health Monitor sidecar](https://github.com/kubernetes-csi/external-health-monitor) (this will be performed automatically when using Helm with `controller.volumeHealthMonitoring`).
+
+## Operation
+
+The EBS CSI Driver will report any volumes with `impaired` status as abnormal. This will record a `VolumeConditionAbnormal` event on the `PersistenVolumeClaim` associated with the abnormal volume. This event will include the most recent message from the AWS API about the volume's status.
+
+At this time, no other action will be taken on abnormal volumes. The cluster administrator and/or users should monitor `PersistentVolumeClaim`s for `VolumeConditionAbnormal` and take manual steps to rectify broken volumes.
```
torredil, line 30: nit: broken -> impaired

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -113,6 +124,10 @@ controller:
   additionalArgs: []
   sdkDebugLog: false
   loggingFormat: text
+  # EXPERIMENTAL
```
torredil, line 127: 👍

torredil (commented):
Largely lgtm besides a few suggestions and really minor nits.

Let's add a test case for the `volume-health-monitoring` flag in `TestControllerOptions`: 
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b926e1ee4551e60b44c9b5a887409a8feaf91d86/cmd/options/controller_options_test.go#L25

```
		{
			name:  "lookup volume-health-monitoring",
			flag:  "volume-health-monitoring",
			found: true,
		},
```

docs/volume-health-monitoring.md
```diff
@@ -0,0 +1,30 @@
+# Volume Health Monitoring
+
+The EBS CSI Driver contains experiment support for [Volume Health Monitoring](https://kubernetes.io/docs/concepts/storage/volume-health-monitoring/). Because Volume Health Monitoring is an alpha feature, this support is subject to change as the feature evolves.
```
rdpsin, line 3: np: s/experiment/experimental/g

docs/volume-health-monitoring.md
```diff
@@ -0,0 +1,30 @@
...
+
+The EBS CSI Driver must be given permission to access the [`DescribeVolumeStatus` EC2 API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVolumeStatus.html) for the volumes in the cluster. This example snippet can be used in an IAM policy to grant access to `DescribeVolumeStatus` on all volumes:
+
+```json
+{
+  "Effect": "Allow",
+  "Action": [
+    "ec2:DescribeVolumeStatus"
+  ],
+  "Resource": "*"
```
rdpsin, line 17: How about restricting to just volumes created by the CSI driver?

rdpsin (approved):

charts/aws-ebs-csi-driver/values.yaml
ConnorJC3: Leaving this open until we decide what the default for `monitor-interval` will be.

docs/volume-health-monitoring.md
ConnorJC3: I think that's probably better left to #1340. Right now, our default example IAM policy gives the driver global permissions for several APIs like `DescribeVolumes`

ConnorJC3:
Fixed most of the review comments, ready for re-review on anything marked resolved.

Thinking about also using this PR to implement `ListVolumes` to optimize API calls.

charts/aws-ebs-csi-driver/values.yaml
wmesard: Every configuration option represents a failure of the software to do the right thing automatically. Every configuration option needs to be documented and protected by unit tests, thereby increasing the cognitive load of user and developer alike. Sometimes they are necessary, but only as a last resort.

docs/volume-health-monitoring.md
wmesard: If you're going to add to the entropy that #1340 has to clean up, at least leave a note (if not a new rev) over there.

pkg/cloud/cloud.go
```diff
@@ -166,6 +166,9 @@ var (
 
 	// VolumeNotBeingModified is returned if volume being described is not being modified
 	VolumeNotBeingModified = fmt.Errorf("volume is not being modified")
+
+	// ErrInternal is an error that is returned when "impossible" events occur.
+	ErrInternal = errors.New("Internal error (this should be impossible!)")
```
wmesard, line 171: As this error is used today, I agree that it should be impossible. Honestly, I wouldn't have bothered to write that check in the fist place, since it's bordering on `if 1 + 1 != 2` territory.

That said, error messages should be specific or actionable. "`(this should be impossible!)`" is neither. Suggest deleting the parenthetical.

ConnorJC3: Changed to `(submit a bug report)`, which is actionable

ConnorJC3:
Closing this for now because it's stale and will likely be reworked

PR #1503, opened 2023-02-02
Update installation docs

torredil:

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -6,13 +6,11 @@
 
 * If you are using a self managed cluster, ensure the flag `--allow-privileged=true` for `kube-apiserver`.
 
-* Important: If you intend to use the csi-snapshotter functionality, the [CSI Snapshotter](https://github.com/kubernetes-csi/external-snapshotter). **must be installed before** the EBS CSI driver.
-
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -6,12 +6,10 @@
 
 * If you are using a self managed cluster, ensure the flag `--allow-privileged=true` for `kube-apiserver`.
 
-* Important: If you intend to use the csi-snapshotter functionality, the [CSI Snapshotter](https://github.com/kubernetes-csi/external-snapshotter). **must be installed before** the EBS CSI driver.
-
-* If you are managing EBS volumes using static provisioning, get yourself familiar with [EBS volume](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AmazonEBS.html).
+* Important: If you intend to use the csi-snapshotter functionality, the [CSI Snapshotter](https://github.com/kubernetes-csi/external-snapshotter) **must be installed before** the EBS CSI driver.
```
rdpsin, line 9: How about adding installation instructions here? They're just a couple of lines, right?

ConnorJC3: I think it would be better just to link directly to their installation guide: https://github.com/kubernetes-csi/external-snapshotter#usage

Putting the instructions here is just asking for trouble when they get changed and we forget to update our docs.

torredil: Directly linked to the installation instructions and clarified that the CRDs specifically must be installed before the driver. Let me know if we can refine this further.

PR #1504, opened 2023-02-07
Pin k8s.io/dynamic-resource-allocation to v0.26.0

bertinatto:

This is a bugfix intended to fix the following error:

    ```
    $ go list -mod readonly -m all
    go: k8s.io/dynamic-resource-allocation@v0.0.0: invalid version: unknown revision v0.0.0
    ```

This patch pins `k8s.io/dynamic-resource-allocation` to `v0.26.0`, like the other kube dependencies.

**What testing is done?** 

```$ go list -mod readonly -m all```


torredil (approved):
Thanks

/lgtm
/retest

PR #1508, opened 2023-02-13
Don't trim device prefix for existing volumes in DeviceManager

ConnorJC3:
If a volume is (already) mounted at /dev/sdX, DM will trim the prefix and claim the volume should be mounted at /dev/xvdX - this causes the driver to get stuck in an infinite loop waiting for the volume to be attached to the wrong location. This bug commonly is triggered by manually mounting a volume using the AWS console, because it uses the prefix /dev/sdX by default.


ConnorJC3:
Superceded by #1518

PR #1510, opened 2023-02-13
Fix --version output

ConnorJC3:
Fix version handling code to output version info, which was inadvertantly removed in ad25982


torredil (approved):
/lgtm

PR #1512, opened 2023-02-14
The conditional API version is not necessary since the driver is requ…

vbrinza:
…ired starting from 1.23 which already has the policy/v1

Syntax fix
The conditional API version is not necessary since the driver is required starting from 1.23 which already has the policy/v1
**What testing is done?** 
N/A


ConnorJC3:
/close

Hi @vbrinza, we still have users on older versions of Kubernetes that only support `policy/v1beta1`. Just because the EBS CSI Driver is required on EKS 1.23 and above does not mean that all users are running 1.23 yet! For further information please see this similar PR: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1485

If you are running into an issue with the conditional API please reach out for support via a GitHub issue or to AWS directly, thanks!

PR #1518, opened 2023-03-02
Stop treating prefixes as magic in DeviceManager

ConnorJC3:
DM treats the prefixes "/dev/xvd" and "/dev/sd" as magic. This leads to a bug where volumes with a prefix starting with "/dev/sd" (for example, "/dev/sdf") will get incorrectly mangled into a "/dev/xvd" form (i.e. converting "/dev/sdf" to "/dev/xvdf"). This will cause a volume that is already mounted and prefixed with "/dev/sd" (for example, a volume manually mounted via the AWS console) to get stuck in an infinite loop as the node will believe it is mounted at the wrong path.

With this change, instead of storing only the end of the path in the inflight and existing volumes map, we instead store the entire device path. The previously leaky abstraction from the allocator to store only the end of the path is contained to the allocator, which has a new parameter for the desired prefix.

This commit also vastly increases the range of paths the allocator can generate from 52 simultaneously mounted volumes to 650.

### DIFF
--- a/pkg/cloud/devicemanager/allocator.go
+++ b/pkg/cloud/devicemanager/allocator.go
@@ -22,8 +22,7 @@ import (
 
 // ExistingNames is a map of assigned device names. Presence of a key with a device
 // name in the map means that the device is allocated. Value is irrelevant and
-// can be used for anything that NameAllocator user wants.  Only the relevant
-// part of device name should be in the map, e.g. "ba" for "/dev/xvdba".
+// can be used for anything that NameAllocator user wants.
 type ExistingNames map[string]string
 
 // On AWS, we should assign new (not yet used) device names to attached volumes.
@@ -36,9 +35,8 @@ type ExistingNames map[string]string
 // device name reuse.
 type NameAllocator interface {
 	// GetNext returns a free device name or error when there is no free device
-	// name. Only the device name is returned, e.g. "ba" for "/dev/xvdba".
-	// It's up to the called to add appropriate "/dev/sd" or "/dev/xvd" prefix.
-	GetNext(existingNames ExistingNames) (name string, err error)
+	// name. The prefix (such as "/dev/xvd" or "/dev/sd") is passed as a parameter.
+	GetNext(existingNames ExistingNames, prefix string) (name string, err error)
 }
 
 type nameAllocator struct{}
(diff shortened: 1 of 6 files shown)


pkg/cloud/devicemanager/allocator_test.go
```diff
@@ -53,11 +53,11 @@ func TestNameAllocatorError(t *testing.T) {
 	allocator := nameAllocator{}
 	existingNames := map[string]string{}
 
-	for i := 0; i < 52; i++ {
-		name, _ := allocator.GetNext(existingNames)
+	for i := 0; i < 25*26; i++ {
```
rdpsin, line 56: Where did this number come from? Can we store it as a descriptive constant?

pkg/cloud/devicemanager/allocator.go
```diff
@@ -22,8 +22,7 @@ import (
 
 // ExistingNames is a map of assigned device names. Presence of a key with a device
 // name in the map means that the device is allocated. Value is irrelevant and
-// can be used for anything that NameAllocator user wants.  Only the relevant
-// part of device name should be in the map, e.g. "ba" for "/dev/xvdba".
+// can be used for anything that NameAllocator user wants.
 type ExistingNames map[string]string
```
rdpsin, line 26: If we don't care about the value, it should be `struct{}`.

rdpsin:
Please update the commit title to be more descriptive of the actual problem. 'Stop treating prefixes as magic' isn't very informative.

pkg/cloud/devicemanager/allocator_test.go
ConnorJC3: It's the max number of volumes that the allocator can allocate. It can't really be a constant (because the allocator uses nested for loops), but I'll add a comment making it more obvious.

pkg/cloud/devicemanager/allocator.go
ConnorJC3: Although we don't care about the value there _is_ a value: the caller passes in a `map[string]string`, which is not compatible with a `map[string]struct{}`.

torredil (approved):

ConnorJC3:
/remove-hold

Minor fix, but someone else should re-review before this is merged.

ConnorJC3:
/retest
Flake?

Issue #1535, opened 2023-03-15
Helm tests fail 


camaeel:
/kind bug

**What happened?**
I did `helm template | kubectl apply -f -` 
Test pod failed. 

Log from test pod contains a lot of: 
```
ebs-csi-driver-test kubetest2 {"msg":"Test Suite starting","completed":0,"skipped":0,"failed":0}
ebs-csi-driver-test kubetest2 Running Suite: Kubernetes e2e suite - /workspace/_rundir/e2e-kubernetes
ebs-csi-driver-test kubetest2 =======================================================================
ebs-csi-driver-test kubetest2 Random Seed: 1678883296 - will randomize all specs
ebs-csi-driver-test kubetest2
ebs-csi-driver-test kubetest2 Will run 240 of 7332 specs
ebs-csi-driver-test kubetest2 Running in parallel across 25 processes
ebs-csi-driver-test kubetest2 ------------------------------
ebs-csi-driver-test kubetest2 [SynchronizedBeforeSuite] [FAILED] [60.066 seconds]
ebs-csi-driver-test kubetest2 [SynchronizedBeforeSuite]
ebs-csi-driver-test kubetest2 test/e2e/e2e.go:76
ebs-csi-driver-test kubetest2
ebs-csi-driver-test kubetest2   Begin Captured StdOut/StdErr Output >>
ebs-csi-driver-test kubetest2     {"msg":"Test Suite starting","completed":0,"skipped":0,"failed":0}
ebs-csi-driver-test kubetest2   << End Captured StdOut/StdErr Output
ebs-csi-driver-test kubetest2
ebs-csi-driver-test kubetest2   Begin Captured GinkgoWriter Output >>
ebs-csi-driver-test kubetest2     [SynchronizedBeforeSuite] TOP-LEVEL
ebs-csi-driver-test kubetest2       test/e2e/e2e.go:76
ebs-csi-driver-test kubetest2     Mar 15 12:28:22.551: INFO: >>> kubeConfig: /root/.kube/config
ebs-csi-driver-test kubetest2     Mar 15 12:28:22.561: INFO: Waiting up to 30m0s for all (but 0) nodes to be schedulable
[...]

PR #1536, opened 2023-03-15
Disable tests in helm chart

camaeel:
Bug #1535 

Ability to disable running helm test hooks

**What testing is done?** 
Helm template


camaeel:
helm template --skip-tests does the trick

PR #1542, opened 2023-03-22
Add non-negative check on getVolumeLimit()

### LINKED ISSUE #1361: Invalid attach limit value -1 cannot be added to CSINode object for "ebs.csi.aws.com"
/kind bug

**What happened?**
We upgraded the ebs-csi-driver from version 1.4.0 to 1.10.0. We have an existing alert that tracks when more than 5 pods are stuck in ContainerCreating. This alert fired and upon running `kubectl describe` on the affected pods, we found the following events:
```
Events:
  Type     Reason              Age                 From                     Message
  ----     ------              ----                ----                     -------
  Normal   Scheduled           19m                 default-scheduler        Successfully assigned namespace_foo/pod_foo to ip-200-200-200-200.ap-southeast-1.compute.internal
  Warning  FailedMount         17m                 kubelet                  Unable to attach or mount volumes: unmounted volumes=[pgdata], unattached volumes=[user-volume root-volume pgconf-volume recover-volume pgdata primary-volume backup pgwal-volume dshm backrestrepo-volume]: timed out waiting for the condition
  
.... More FailedMount events

  Warning  FailedAttachVolume  48s (x10 over 19m)  attachdetach-controller  AttachVolume.Attach failed for volume "pvc-19d6d86d-ef49-43ad-a55b-14bd95969f9a" : rpc error: code = Internal desc = Could not attach volume "vol-03d1ad62aa5d7da79" to node "i-0b0a9b426ebfcd260": attachment of disk "vol-03d1ad62aa5d7da79" failed, expected device to be attached but was attaching
```

When we looked into the node, we found that there were a total of 31 pods with 1 volume each scheduled on the node (m5a.xlarge). 5 of those pods, the most recently scheduled on the node, were pending.

When investigating we found the following symptoms:
1. We had 2 affected nodes - both nodes had pods scheduled where their volumes had issues attaching and mounting. Both nodes had `CSINode` objects that looked like the following, and is missing an `allocatable` field which informs the scheduler how many volumes the node can accept:
```
kubectl get csinode/ip-200-200-200-200.ap-southeast-1.compute.internal -o yaml
apiVersion: storage.k8s.io/v1
kind: CSINode
metadata:
  ...
spec:
  drivers:
  - name: ebs.csi.aws.com
    nodeID: i-0b0a9b426ebfcd260
    topologyKeys:
```
2. Upon looking at kubelet logs (kubelet calls the `NodeGetInfo` function on the csi plugin when the plugin registers itself [here](https://github.com/kubernetes/kubernetes/blob/64ed9145452d2d1d324d2437566f1ea1ce76f226/pkg/volume/csi/csi_plugin.go#L135)), we see this error: `nodeinfomanager.go:566] Invalid attach limit value -1 cannot be added to CSINode object for "ebs.csi.aws.com"` (logged [here](https://github.com/kubernetes/kubernetes/blob/64ed9145452d2d1d324d2437566f1ea1ce76f226/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L561)). This indicates that the attach limit, which is returned by nodeService.getVolumesLimit() is -1, which isn't correct - getVolumesLimit should return the `max_volumes_per_node` ([spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetinfo)) , which should never be negative.

The scheduler keeps track of the currently attached volumes on a node. When scheduling a pod with a volume that is not already attached to the node, it takes the advertised volume limit of the node (which is returned by NodeGetInfo) and subtracts the currently attached volumes. If the result is a positive number, the scheduler schedules the pod on the node. If there is no limit (because the number returned by `NodeGetInfo` is negative), the scheduler always returns true ([code](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L112)).

We have tracked down the issue to this change: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1175 . This change starts returning the # of volumes that can be additionally attached to the node (it queries the number of volumes already attached) instead of the maximum number of volumes the instance can handle.

nodeService.volumeLimits should return the maximum number of volumes the instance can handle. Kubernetes scheduler should not schedule pods with volumes on nodes that are at their volume limit.

1. Install aws-ebs-csi-driver v1.4.0.
2. Create a node with a volume limit (ex. m5a.xlarge).
3. Schedule as many pods as required to fill the node to its volume limit (ex: 25 pods with volumes on a m5a.xlarge)
4. Upgrade the aws-ebs-csi-driver to 1.10.0 (this restarts every ebs-csi-driver pod running on each node and kubelet will call `NodeGetInfo`). 
5. Schedule more pods with volumes on the same node - they should get scheduled and the pods will be stuck in ContainerCreating

**Environment**
- Kubernetes version (use `kubectl version`): 1.21.7 and 1.21.9
- Driver version: 1.10

gtxu:
Fix #1361 
When volume attach limit not correct calculated, it may return negative value which is not allowed in the [csi/spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetinfo)
The `-1` return in the issue reflect to [metadata_ec2.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d889329395deee4ed8fcccbd6bf4b204ac9d7f7d/pkg/cloud/metadata_ec2.go#L63), when all available volume attached, the ENI turns to be actually ENI +1 for Nitro instances.

**Reproduce the issue**
On a cluster with aws-ebs-csi-driver installed,
1. create a nitro instance node
2. schedule pods with pvc to the node till reach/or exceed the attach limit
3. restart aws-ebs-csi-driver
4. check kubectl log, find the same error message as the issue #1361 

**What testing is done?** 
Unit test added for maximum volume/eni allocated cases
CI test passed


pkg/driver/node.go
```diff
@@ -729,6 +729,7 @@ func (d *nodeService) getVolumesLimit() int64 {
 		return d.driverOptions.volumeAttachLimit
 	}
 
+	//snow device
```
ConnorJC3, line 732: Nit: remove this comment

pkg/driver/node.go
```diff
@@ -743,10 +744,15 @@ func (d *nodeService) getVolumesLimit() int64 {
 	if isNitro {
 		enis := d.metadata.GetNumAttachedENIs()
 		nvmeInstanceStoreVolumes := cloud.GetNVMeInstanceStoreVolumesForInstanceType(instanceType)
-		availableAttachments = availableAttachments - enis - blockVolumes - nvmeInstanceStoreVolumes
+		availableAttachments = availableAttachments - enis - nvmeInstanceStoreVolumes
+	}
+
+	if blockVolumes < availableAttachments {
```
ConnorJC3, line 750: This seems excessively complicated compared to just doing `if availableAttachments < 0`, is there a reason we're not just doing that?

torredil (commented):
- Suggest providing a minimal reproducible example of https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1361.
- Suggest adding details to **What testing is done?** section.

ConnorJC3:
/lgtm

:+1:

ConnorJC3:
/remove-hold
All good

hanyuel (approved):

torredil (approved):

PR #1546, opened 2023-03-27
Grab snow device path via last character rather than trimming

ConnorJC3:

Bug fix

Due to #1518, volumes now start with /dev/xvda rather than /dev/xvdb. Because of this, the hardcoded path in the snow code stopped working, breaking mounting on snow devices. This change instead creates the snow device path based on the last character of the generated path, rather than trying to do it via trimming, which is unreliable.

Snow devices are currently limited to 10 volumes attached which makes this 100% safe, but this will need to be revisited if snow every adds suport for >26 volumes.

**What testing is done?** 

Manual (with snow customer), added additional unit test case

### DIFF
--- a/pkg/driver/node_linux.go
+++ b/pkg/driver/node_linux.go
@@ -101,7 +101,11 @@ func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (st
 
 	if util.IsSBE(d.metadata.GetRegion()) {
 		klog.V(5).InfoS("[Debug] Falling back to snow volume lookup", "devicePath", devicePath)
-		canonicalDevicePath = "/dev/vd" + strings.TrimPrefix(devicePath, "/dev/xvdb")
+		// Snow completely ignores the requested device path and mounts volumes starting at /dev/vda .. /dev/vdb .. etc
+		// Morph the device path to the snow form by chopping off the last letter and prefixing with /dev/vd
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

pkg/driver/node_linux.go
```diff
@@ -101,7 +101,7 @@ func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (st
 
 	if util.IsSBE(d.metadata.GetRegion()) {
 		klog.V(5).InfoS("[Debug] Falling back to snow volume lookup", "devicePath", devicePath)
-		canonicalDevicePath = "/dev/vd" + strings.TrimPrefix(devicePath, "/dev/xvdb")
+		canonicalDevicePath = "/dev/vd" + devicePath[len(devicePath)-1:]
```
rdpsin, line 104: np: Can you add a comment here explaining why you're doing this?

hanyuel: For my own understanding: the 26-volume limit doesn't exist for the current implementation, correct?

ConnorJC3:
/retest

Is this the new test breakage :(

ConnorJC3:
/retest

Prow was actually broken, hopefully fixed now: https://github.com/kubernetes/test-infra/pull/29167

Issue #1545, opened 2023-03-27
Setting tag in VolumeSnapshotClass doesn't work as expected


hanyuel:
/kind bug

**What happened?**
Create a snapshot using this VolumeSnapshotClass with the tag ``"Name=value1"``
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: csi-aws-vsc
driver: ebs.csi.aws.com
deletionPolicy: Delete
parameters:
  tagSpecification_1: "Name=value1"
````
In the AWS console, for the snapshot, the value for the "Name" tag is``<cluster-name>-dynamic-snapshot-<some-id>``which is automatically created.
**What you expected to happen?**
In the AWS console, for the snapshot, the value for the "Name" tag should be "value1". 

**How to reproduce it (as minimally and precisely as possible)?**
1. Follow this example in the repo to create a snapshot, but with the above VolumeSnapshotClass.
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/snapshot
2. Got to AWS console EC2, check the tags for the snapshot.

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): 
```
Client Version: v1.26.3
Kustomize Version: v4.5.7
Server Version: v1.24.10-eks-48e63af
```
- Driver version: v1.17.0

PR #1548, opened 2023-03-28
Fix VolumeSnapshotClass tagging

hanyuel:
Bug fix for #1545 

When users set "Name" tag for a snapshot via `VolumeSnapshotClass.parameters`, the value of the Name tag is overwritten by [the value automatically populated by the Driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/5d75bf4be309d3b554d3959a79cc8f05ed65d4d2/pkg/driver/controller.go#L632), which is unexpected behavior.
This PR fixes the bug by setting the tags from VolumeSnapshotClass.parameters _after_ the Driver automatically populates the Name tag.

**What testing is done?** 
Added unit test & 
1. Built and installed the driver that has the changes. 
2. Followed the repro steps in #1545 to create a snapshot with "Name" tag in ``VolumeSnapshotClass.parameters``.
3.  Confirmed the name of the snapshot and the tags of the snapshot in the EC2 console are the values from ``VolumeSnapshotClass.parameters``.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -622,10 +622,6 @@ func (d *controllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid tag value: %v", err)
 	}
 
-	for k, v := range addTags {
-		snapshotTags[k] = v
-	}
-
 	if d.driverOptions.kubernetesClusterID != "" {
 		resourceLifecycleTag := ResourceLifecycleTagPrefix + d.driverOptions.kubernetesClusterID
 		snapshotTags[resourceLifecycleTag] = ResourceLifecycleOwned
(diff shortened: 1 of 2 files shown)


pkg/driver/controller.go
```diff
@@ -622,15 +622,16 @@ func (d *controllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
 	if d.driverOptions.kubernetesClusterID != "" {
 		resourceLifecycleTag := ResourceLifecycleTagPrefix + d.driverOptions.kubernetesClusterID
 		snapshotTags[resourceLifecycleTag] = ResourceLifecycleOwned
 		snapshotTags[NameTag] = d.driverOptions.kubernetesClusterID + "-dynamic-" + snapshotName
 	}
+
+	for k, v := range addTags {
+		snapshotTags[k] = v
+	}
+
```
ConnorJC3, line 634: Can we move this after the for loop on line 634-637 instead, for consistency with how it works in `CreateVolume`: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/5d75bf4be309d3b554d3959a79cc8f05ed65d4d2/pkg/driver/controller.go#L251-L266

ConnorJC3 (commented):
Might be worth adding a test case for this?

pkg/driver/controller.go
hanyuel: Sure

hanyuel:
> Might be worth adding a test case for the change? (also, see comment below)

Thanks for the suggestions. Added a test case

PR #1568, opened 2023-04-12
Allow customize secret name in Helm Chart

### LINKED ISSUE #1492: Helm: Allow to set custom secret name & keys
We use helm to install the EBS driver on our cluster. However the default secret name `aws-secret` is so generic and can't be changed atm, which makes it hard to use on a larger cluster. Would be nice if one could change the name of the secret and its keys.

/feature

```
            - name: AWS_ACCESS_KEY_ID
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.controller.aws_access_key.secret_name }}
                  key: {{ .Values.controller.aws_access_key.key_for_id }}
                  optional: true
            - name: AWS_SECRET_ACCESS_KEY
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.controller.aws_access_key.secret_name }}
                  key: {{ .Values.controller.aws_access_key.key_for_secret }}
                  optional: true
```

gtxu:
feature
Allow user set customized aws access secret name in Helm values, the default secret name in value set to `aws-secret` so this should not break existing workflows.
close #1492 
**What testing is done?** 
Helm dry-run checked, manifests correctly parsed

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -109,6 +109,11 @@ imagePullSecrets: []
 nameOverride:
 fullnameOverride:
 
+awsAccessSecret: 
+  name: aws-secret
+  keyId: key_id
+  accessKey: access_key
+
 controller:
   additionalArgs: []
   sdkDebugLog: false
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -101,18 +101,20 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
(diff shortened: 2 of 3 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -109,6 +109,9 @@ imagePullSecrets: []
 nameOverride:
 fullnameOverride:
 
+awsAccessSecret: 
+  secretName: aws-secret
```
torredil, line 113: suggest `secretName` -> `name`

ConnorJC3 (changes requested):
Please also allow changing the `key`s as well as the `name` of the secret

ConnorJC3:
I don't understand the security concern - what is the issue with allowing configuration of the field's key, and why does it not apply to the secret name?

PR #1573, opened 2023-04-15
update list of nitro instances

petderek:
bugfix

Updating the list of nonNitro instances to be in sync with the [documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances). There were some types in the list that didn't belong (eg, c7g is a Nitro type), and a few types that needed to be added.

This makes the limit prediction logic a little bit more accurate and helpful for these instance types.

**What testing is done?** 

I ran unit tests --  but it doesn't seem as though I have permission to run any of the other tests. Happy to take suggestions, though.


torredil:
Thanks! 
/lgtm

PR #1580, opened 2023-04-20
provide opportuniti to set empty resources in Daemon Sets

vladimir4862:
provide opportuniti to set empty resources in Daemon Sets

bug fix

to provide opportuniti to set empty resources in Daemon Sets 
in current if you set in your own value file this resources (node.resources) as empty helm will use default (which provide from default values.yaml)
**What testing is done?** 
helm template .

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -184,13 +184,7 @@ controller:
   # rollingUpdate:
   #   maxSurge: 0
   #   maxUnavailable: 1
-  resources:
-    requests:
-      cpu: 10m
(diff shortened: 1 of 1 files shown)


ConnorJC3:
/close

Hi @vladimir4862 ! Helm will allow you to configure an empty object for the resources, which will result in no resources on the final objects. Thus, we will not be accepting this PR, as we do want a default configuration for resources.

If you need assistance in using Helm I recommend reading the [Helm documentation](https://helm.sh/docs/).

vladimir4862:
Thanks a lot. 
you are right it easy to set by helm. 
Sorry for this.

Issue #1232, opened 2022-05-10
Support startup taint


sylr:
**Is your feature request related to a problem?/Why is this needed**

I'd like the [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) to implement a startup taint mechanism.

/feature

**Describe the solution you'd like in detail**

When you are dealing with clusters where nodes come and go often we want to ensure that those new nodes have all the required components before the actual workloads are started.

Cilium for example recommends that all nodes spawned in a cluster be added the taint `node.cilium.io/agent-not-ready=true:NoExecute`. The cilium agent, which is a required component in a cilium based cluster, tolerates that taint and, upon startup completion of the agent, removes the taint on the node so that the workload can be started with a network stack up and running.

When I start new nodes in a EKS / Karpenter / Cilium based cluster, it happens that workloads requiring mounting an EBS volume starts before the aws-ebs-csi-driver and the workloads enter a backoff loop which can be long.

I'd like to prevent that by adding a new startup taint to my new nodes, e.g.: **`ebs.csi.aws.com/agent-not-ready=true:NoExecute`** and I would like that this taint be removed upon aws-ebs-csi-driver startup completion.

torredil:
Thanks for the request, this feature will be available in 1.19.
/close

PR #1581, opened 2023-04-20
Support node start-up taint

gtxu:
 * Feature

This PR added support for start-up taint removal for csi-node daemonset pods. This feature allows cluster admin to set taints on nodes, blocking workload pod to be scheduled before the driver start up and be ready. By automatically remove the taints marks the node ready for any CSI functionalities and workload start to be scheduled to the node. This feature can be configured in driver options and in Helm values.

closes issue #1232 
**What testing is done?** 
Manual Tested taint removal and logs

### DIFF
--- a/cmd/options/node_options.go
+++ b/cmd/options/node_options.go
@@ -31,8 +31,13 @@ type NodeOptions struct {
 	// itself (dynamically discovering the maximum number of attachable volume per EC2 machine type, see also
 	// https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/347).
 	VolumeAttachLimit int64
+
+	// StartupTaintRemoval configures the node service behavior whether to remove customized preset taint `agent-not-ready`
+	// on the node. This taint prevent any workload pods from scheduling prior the csi node service is startup.
+	StartupTaintRemoval bool
 }
 
 func (o *NodeOptions) AddFlags(fs *flag.FlagSet) {
 	fs.Int64Var(&o.VolumeAttachLimit, "volume-attach-limit", -1, "Value for the maximum number of volumes attachable per node. If specified, the limit applies to all nodes. If not specified, the value is approximated from the instance type.")
+	fs.BoolVar(&o.StartupTaintRemoval, "start-up-taint", false, "To enable the node service remove node-ready taint after startup (default to false).")
 }
--- a/docs/install.md
+++ b/docs/install.md
@@ -41,6 +41,19 @@ kubectl create secret generic aws-secret \
 ### Configure driver toleration settings
 By default, the driver controller tolerates taint `CriticalAddonsOnly` and has `tolerationSeconds` configured as `300`; and the driver node tolerates all taints. If you don't want to deploy the driver node on all nodes, please set Helm `Value.node.tolerateAllTaints` to false before deployment. Add policies to `Value.node.tolerations` to configure customized toleration for nodes.
 
+### Configure node taint and driver start-up taint
+In some cases when new node frequesntly join the cluster, workload pods can be scheduled to a new node ahead of the csi-node start-up and ready on that node. This race condition between workload pod and csi-node pod will cause the workload pod fails to mount PVC at first place.
+
+To help overcome this situation, CSI Driver node can manipulate Kubernetes’s taints on a given node to help preventing pods from starting before CSI Driver's node pod runs on this node.
+
+To configure start-up taint, the cluster administrator can places a taint with key `node.ebs.csi.aws.com/agent-not-ready` on a given uninitialized node(or node group). This prevents pods that don’t have a matching toleration from either being scheduled or altogether running on the node until the taint is removed. If use Helm to install CSI Driver, set `.Value.node.startUpTaint` to `true`. Once the CSI Driver pod runs on the node, initializes and ready, it will removes the aforementioned taint. After that, workload pods will start being scheduled and running on the node, with CSI Driver full functional on that node.
+
+The taint effect can be set to `NoSchedule` or `NoExecute`,
+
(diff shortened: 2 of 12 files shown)


pkg/driver/node.go
```diff
@@ -89,6 +89,13 @@ func newNodeService(driverOptions *DriverOptions) nodeService {
 		panic(err)
 	}
 
+	if driverOptions.startupTaintRemoval {
+		err := cloud.RemoveNodeTaint(cloud.DefaultKubernetesAPIClient, AgentNotReadyNodeTaintKey)
+		if err != nil {
+			klog.InfoS("Node agent-not-ready taint error", "error", err)
```
rdpsin, line 95: What if someone enables `startupTaintRemoval` even when running on other COs? This will cause the node service to fail to run.

pkg/cloud/patch_node.go
```diff
@@ -0,0 +1,77 @@
+package cloud
```
rdpsin, line 1: `cloud` is for AWS specific stuff. This should be part of `driver`.

pkg/cloud/patch_node.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+type JSONPatch struct {
+	OP    string      `json:"op,omitempty"`
+	Path  string      `json:"path,omitempty"`
+	Value interface{} `json:"value"`
+}
+
+// RemoveNodeTaint() patched the node, removes the taint that match NodeTaintKey
+func RemoveNodeTaint(k8sAPIClient KubernetesAPIClient, NodeTaintKey string) error {
```
rdpsin, line 22: np: `nodeTaintKey`

pkg/cloud/patch_node.go
```diff
@@ -0,0 +1,77 @@
...
+		if taint.Key != NodeTaintKey {
+			taints = append(taints, taint)
+		} else {
+			hasTaint = true
+			klog.InfoS("Node taint found")
+		}
+	}
+
+	if !hasTaint {
+		return fmt.Errorf("could not find node taint, key: %v, node: %v", NodeTaintKey, nodeName)
```
rdpsin, line 50: Do we want to return an error here? This will cause the driver to fail to run, even if someone explicitly didn't put a taint on the node.

docs/install.md
```diff
@@ -41,6 +41,19 @@ kubectl create secret generic aws-secret \
 ### Configure driver toleration settings
 By default, the driver controller tolerates taint `CriticalAddonsOnly` and has `tolerationSeconds` configured as `300`; and the driver node tolerates all taints. If you don't want to deploy the driver node on all nodes, please set Helm `Value.node.tolerateAllTaints` to false before deployment. Add policies to `Value.node.tolerations` to configure customized toleration for nodes.
 
+### Configure node taint and driver start-up taint
+In some cases when new node frequesntly join the cluster, workload pods can be scheduled to a new node ahead of the csi-node start-up and ready on that node. This race condition between workload pod and csi-node pod will cause the workload pod fails to mount PVC at first place.
```
rdpsin, line 45: np: node->nodes, frequesntly->frequently

docs/install.md
```diff
@@ -41,6 +41,19 @@ kubectl create secret generic aws-secret \
...
+
+To help overcome this situation, CSI Driver node can manipulate Kubernetes’s taints on a given node to help preventing pods from starting before CSI Driver's node pod runs on this node.
+
+To configure start-up taint, the cluster administrator can places a taint with key `node.ebs.csi.aws.com/agent-not-ready` on a given uninitialized node(or node group). This prevents pods that don’t have a matching toleration from either being scheduled or altogether running on the node until the taint is removed. If use Helm to install CSI Driver, set `.Value.node.startUpTaint` to `true`. Once the CSI Driver pod runs on the node, initializes and ready, it will removes the aforementioned taint. After that, workload pods will start being scheduled and running on the node, with CSI Driver full functional on that node.
```
torredil, line 49: Suggested wording:

```
The cluster administrator can place a taint on an uninitialized node (or node group) with the key `node.ebs.csi.aws.com/agent-not-ready`. This taint prevents pods that do not have a corresponding toleration from being scheduled or running on the node until the taint is removed. If the CSI Driver is installed using Helm, .Value.node.removeDriverNodeTaints can be set to true. Once the CSI Driver pod is running on the node, it will remove the aforementioned taint. Afterward, workload pods can be scheduled and run on the node along with the CSI Driver.
```

cmd/options/node_options.go
```diff
@@ -31,8 +31,13 @@ type NodeOptions struct {
...
 	VolumeAttachLimit int64
+
+	// StartupTaintRemoval configures the node service behavior whether to remove customized preset taint `agent-not-ready`
+	// on the node. This taint prevent any workload pods from scheduling prior the csi node service is startup.
+	StartupTaintRemoval bool
 }
 
 func (o *NodeOptions) AddFlags(fs *flag.FlagSet) {
 	fs.Int64Var(&o.VolumeAttachLimit, "volume-attach-limit", -1, "Value for the maximum number of volumes attachable per node. If specified, the limit applies to all nodes. If not specified, the value is approximated from the instance type.")
+	fs.BoolVar(&o.StartupTaintRemoval, "start-up-taint", false, "To enable the node service remove node-ready taint after startup (default to false).")
```
torredil, line 42: Adding this knob provides primarily two benefits:

1. Saving an API call to the K8s API if it is disabled.
2.  Sets intent, which allows us to log a warning if the taint is missing.

Removing the knob allows for the feature to work automatically with no user configuration needed.

Should we move forward with adding this knob? What is the best user experience here?

pkg/cloud/patch_node.go
torredil, line 22: Can we use a helper library here instead so that we don't have to construct the JSON patch manually, etc? 
K8s taints package implements utilities for working with taints: https://pkg.go.dev/k8s.io/kubernetes/pkg/util/taints.

torredil (changes requested):

ConnorJC3:
/close

Superceded by #1588
Thanks for the initial work!

PR #1584, opened 2023-04-21
Allow throughput with defaulted GP3 volume type

### LINKED ISSUE #1583: Throughput parameter doesn't work
/kind bug

**What happened?**
Throughput parameter on StorageClass not being implemented for dynamic provisioning.

I was using a `gp3` EBS type and specified throughput as 250Mb/s, and it came up as 125Mb/s.

Define a StorageClass as follows:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: somename
parameters:
  encrypted: "true"
  iops: "5000"
  kmsKeyId: "kmskeyId"
  throughput: "250"
provisioner: ebs.csi.aws.com
reclaimPolicy: Delete
```
  
**Environment**
```
Client Version: v1.24.10-eks-48e63af
Kustomize Version: v4.5.4
Server Version: v1.25.8-eks-ec5523e
```
- Driver version: v1.18.0

ConnorJC3:
Fixes #1583 and adds a test case for the scenario

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -393,7 +393,7 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	if iops > 0 {
 		requestInput.Iops = aws.Int64(iops)
 	}
-	if throughput > 0 && diskOptions.VolumeType == VolumeTypeGP3 {
+	if throughput > 0 {
 		requestInput.Throughput = aws.Int64(throughput)
 	}
 	snapshotID := diskOptions.SnapshotID
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -175,7 +175,8 @@ func TestCreateDisk(t *testing.T) {
 				AvailabilityZone: defaultZone,
 			},
 			expCreateVolumeInput: &ec2.CreateVolumeInput{
-				Iops: aws.Int64(3000),
+				Iops:       aws.Int64(3000),
+				Throughput: aws.Int64(125),
 			},
 			expErr: nil,
 		},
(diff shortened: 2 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -393,7 +393,7 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	if iops > 0 {
 		requestInput.Iops = aws.Int64(iops)
 	}
-	if throughput > 0 && diskOptions.VolumeType == VolumeTypeGP3 {
+	if throughput > 0 && createType == VolumeTypeGP3 {
```
torredil, line 396: Is this check `createType == VolumeTypeGP3` necessary? throughput will only ever be > 0 if `createType = VolumeTypeGP3` or `createType == ""`: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/dd719c12ef9cbf8e2b86152003b3afb8fa0057e9/pkg/cloud/cloud.go#L332

torredil (commented):
Is this check `createType == VolumeTypeGP3` necessary?

pkg/cloud/cloud.go
ConnorJC3: That's not true, we don't check the volume type when pass through the throughput in controller.go

torredil (approved):
/approve

pkg/cloud/cloud_test.go
```diff
@@ -1961,8 +1980,8 @@ func (m *eqCreateVolumeMatcher) Matches(x interface{}) bool {
 	if input == nil {
 		return false
 	}
-	// Compare only IOPS for now
-	ret := reflect.DeepEqual(m.expected.Iops, input.Iops)
+	// TODO: Check all inputs
+	ret := reflect.DeepEqual(m.expected.Iops, input.Iops) && reflect.DeepEqual(m.expected.Throughput, input.Throughput)
```
rdpsin, line 1984: Can we use `cmp.Diff` here?

hanyuel (approved):

PR #1590, opened 2023-05-04
remove condition on iopspergb key being mandatory for io1 volumes

surian:
Condition on iopsPerGB key being mandatory for io1 volumes introduced by https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/767 is irrelevant and should be removed as it's possible to create a io1 volume by specifying the desired number of iops.

```
$ aws ec2 create-volume --availability-zone eu-west-3a --iops 1337 --size 100 --volume-type io1
{
    "AvailabilityZone": "eu-west-3a",
    "CreateTime": "2023-05-04T12:30:04+00:00",
    "Encrypted": false,
    "Size": 100,
    "SnapshotId": "",
    "State": "creating",
    "VolumeId": "vol-0cba6d099003264f2",
    "Iops": 1337,
    "Tags": [],
    "VolumeType": "io1",
    "MultiAttachEnabled": false
}
```

The documentation updated by https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1409 is also already mentioning that iopPerGB **or** iops can be used for io1 volumes but it's today impossible to use the iops key because of this condition.

### DIFF
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -797,7 +797,7 @@ func TestCreateVolume(t *testing.T) {
 			},
 		},
 		{
-			name: "success with volume type io1",
+			name: "success with volume type io1 using iopsPerGB",
 			testFunc: func(t *testing.T) {
 				req := &csi.CreateVolumeRequest{
 					Name:               "vol-test",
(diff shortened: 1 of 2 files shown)


pkg/driver/controller_test.go
```diff
@@ -1521,45 +1521,6 @@ func TestCreateVolume(t *testing.T) {
 				checkExpectedErrorCode(t, err, codes.Aborted)
 			},
 		},
-		{
-			name: "fail with missing iopsPerGB parameter",
```
rdpsin, line 1525: Can we add a test to check for success when using `io1/io2` volume with no `iopsPerGb` parameter?

surian: I added 2 tests (one for io1, one for io2) to check for success when iops parameter is used instead of iopspergb.

torredil (approved):

ConnorJC3:
/retest

flaky test

Could you please squash your commits? Other than that, this PR looks good to me as written.

ConnorJC3:
/hold

needs squash

surian:
> Could you please squash your commits? Other than that, this PR looks good to me as written.

Done 👍

torredil (approved):
Thanks!
/unhold

PR #1596, opened 2023-05-12
Remove CPU limits; Decrease node pod CPU usage

torredil:
**Is this a bug fix or adding new feature?**
- Bug fix.

**What is this PR about? / Why do we need it?**
- This PR removes CPU limits to prevent unnecessary throttling and decreases the livenessprobe frequency for the node-driver-registrar sidecar container in the node pod.
- closes #1591

PR #1598, opened 2023-05-15
Drop support for Windows 20H2

torredil:

As mentioned in the [release notes](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG.md), v1.18.0 is the last minor release of the AWS EBS CSI Driver to support Windows 20H2.


ConnorJC3:
/lgtm

I will be approving this after #1596  merges if there is no negative feedback and CI passes.

PR #1599, opened 2023-05-15
Add option to set custom user-agent

torredil:

New feature

Add an option to set a custom user agent.

It is recommended as a best practice to include a User-Agent header with every request. Doing this helps to more effectively diagnose and fix problems.

**What testing is done?** 

Manual/CI/Unit Testing


docs/options.md
```diff
@@ -9,3 +9,4 @@ There are a couple of driver options that can be passed as arguments when starti
 | k8s-tag-cluster-id          | aws-cluster-id-1                                  |                                                     | ID of the Kubernetes cluster used for tagging provisioned EBS volumes|
 | aws-sdk-debug-log           | true                                              | false                                               | If set to true, the driver will enable the aws sdk debug log level|
 | logging-format              | json                                              | text                                                | Sets the log format. Permitted formats: text, json|
+| user-agent-extra            | csi-ebs                                           | helm                                                | Sets the user agent string|
```
ConnorJC3, line 12: "Sets the user agent string" -> "Extra string appended to user agent"

pkg/cloud/cloud.go
```diff
@@ -171,6 +171,8 @@ var (
 // Set during build time via -ldflags
 var driverVersion string
 
+var userAgentExtra string
```
ConnorJC3, line 174: Remove this line

pkg/cloud/cloud.go
```diff
@@ -236,7 +238,7 @@ var _ Cloud = &cloud{}
 
 // NewCloud returns a new instance of AWS cloud
 // It panics if session is invalid
-func NewCloud(region string, awsSdkDebugLog bool) (Cloud, error) {
+func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string) (Cloud, error) {
 	RegisterMetrics()
 	return newEC2Cloud(region, awsSdkDebugLog)
```
ConnorJC3, line 243: `userAgentExtra` needs to be passed into this, no?

pkg/driver/driver_test.go
```diff
@@ -94,3 +94,12 @@ func TestWithAwsSdkDebugLog(t *testing.T) {
 		t.Fatalf("expected awsSdkDebugLog option got set to %v but is set to %v", enableSdkDebugLog, options.awsSdkDebugLog)
 	}
 }
+
+func TestWithUserAgentExtra(t *testing.T) {
+	var userAgentExtra string = "test-user-agent"
+	options := &DriverOptions{}
+	WithUserAgentExtra(userAgentExtra)(options)
+	if options.userAgentExtra != userAgentExtra {
+		t.Fatalf("expected userAgent option got set to %s but is set to %s", userAgentExtra, options.userAgentExtra)
```
ConnorJC3, line 103: "expected userAgent option" -> "expected userAgentExtra option"

ConnorJC3 (changes requested):
Mostly good, but needs some fixes

Also, please change commit/PR name to something like "Add option to append extra string to user agent"

ConnorJC3 (approved):

PR #1600, opened 2023-05-15
Enable volume modification through annotations

rdpsin:
The commit allows users to modify volume properties through PVC annotations. For more details, refer to the modify-volume doc.

**What testing is done?**


ConnorJC3 (approved):
Long term, we should have unit/e2e tests for this, but otherwise lgtm

torredil (approved):

Issue #1608, opened 2023-05-19
Using Fast Snapshot Restores could not create the second volume's fast snapshot


Phaow:
/kind bug

**What happened?**

1. Create VolumeSnapshotClass with fastSnapshotRestoreAvailabilityZones parameter

```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: csi-aws-vsc-fast
driver: ebs.csi.aws.com
deletionPolicy: Delete
parameters:
  fastSnapshotRestoreAvailabilityZones: "us-east-2a,us-east-2b,us-east-2c"
```
2. Create 2 namespaces `fsr-test1` and `fsr-test2`
3. Create `pvc1`,`pod1` in namespace `fsr-test1` and wait for pod running then create snapshot for the `pvc1`, the volumesnapshot/volumesnapshotcontent could become `readyToUse:true`
4. Create `pvc2`,`pod2` in namespace `fsr-test2` and wait for pod running then create snapshot for the `pvc2`, the volumesnapshot/volumesnapshotcontent stuck at  `readyToUse:false`

**What you expected to happen?**
In step 4 the volumesnapshot/volumesnapshotcontent could become `readyToUse:true`

**How to reproduce it (as minimally and precisely as possible)?**
Always

**Anything else we need to know?**:
Check the controller csi-snapshotter container logs, lots of `Could not create snapshot  SnapshotCreationPerVolumeRateExceeded: The maximum per volume CreateSnapshot request rate has been exceeded. Use an increasing or variable sleep interval between requests.` error
```console
$ kubectl logs -c csi-snapshotter ebs-csi-controller-5f5cb78cf6-98z8c
...
[...]

PR #1611, opened 2023-05-22
Add 30 second rate limit to CreateSnapshot

ConnorJC3:

Bug fix

See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1608
See https://github.com/kubernetes-csi/external-snapshotter/issues/778

This does not seek to be a comprehensive rate-limiting solution, but rather to add a temporary workaround for the bug in the snapshotter sidecar by refusing to call the CreateSnapshot for a specific volume unless it has been 30 seconds since the last attempt.

**What testing is done?** 

Manual/New unit test/CI

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/awslabs/volume-modifier-for-k8s/pkg/rpc"
@@ -667,6 +668,13 @@ func (d *controllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 
 	snapshot, err = d.cloud.CreateSnapshot(ctx, volumeID, opts)
 	if err != nil {
+		// TODO: Workaround for https://github.com/kubernetes-csi/external-snapshotter/issues/778
+		// and for https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1608#issuecomment-1554748900
+		//
+		// Wait 15 seconds before returning to the caller if the AWS CreateSnapshot call failed
+		// Because of the inflight check, this will prevent retries while waiting
+		klog.V(4).InfoS("Waiting 15 seconds to delay snapshot retries", "volumeID", volumeID)
+		time.Sleep(time.Second * 15)
 		return nil, status.Errorf(codes.Internal, "Could not create snapshot %q: %v", snapshotName, err)
 	}
 


pkg/cloud/cloud.go
```diff
@@ -167,6 +168,10 @@ var (
 
 	// VolumeNotBeingModified is returned if volume being described is not being modified
 	VolumeNotBeingModified = fmt.Errorf("volume is not being modified")
+
+	// ErrSnapshotRateLimit is returned when cloud.go refuses to send a CreateSnapshot AWS API request
+	// Used to work around https://github.com/kubernetes-csi/external-snapshotter/issues/778
+	ErrSnapshotRateLimit = fmt.Errorf("Refusing to send CreateSnapshot to AWS (see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1608#issuecomment-1554748900)")
```
rdpsin, line 174: Error messages should not contain external links; they should be descriptive enough on their own. Suggest changing this to something like

```
fmt.Errorf("snapshot creation rate-limited")
```

pkg/cloud/cloud.go
```diff
@@ -776,7 +781,17 @@ func (c *cloud) IsExistInstance(ctx context.Context, nodeID string) bool {
 	return true
 }
 
+var lastCreateSnapshot sync.Map
```
rdpsin, line 784: Why use `sync.Map` instead of a normal `map` with a `mutex`?

pkg/cloud/cloud.go
```diff
@@ -776,7 +781,17 @@ func (c *cloud) IsExistInstance(ctx context.Context, nodeID string) bool {
...
+
 func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOptions *SnapshotOptions) (snapshot *Snapshot, err error) {
+	if lastTimeUncast, ok := lastCreateSnapshot.Load(volumeID); ok {
+		if lastTime, ok := lastTimeUncast.(time.Time); ok {
+			if time.Since(lastTime) < time.Second*30 {
```
rdpsin, line 789: Should be 15s. https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html

pkg/cloud/cloud.go
```diff
@@ -796,6 +811,7 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
 		Description:       aws.String(descriptions),
 	}
 
+	lastCreateSnapshot.Store(volumeID, time.Now())
```
rdpsin, line 814: May also need to delete entry at some point, otherwise memory consumed by driver would balloon.

pkg/cloud/cloud.go
rdpsin: Why is this not part of `cloud` struct?

PR #1612, opened 2023-05-23
Update logline to remove "formatted"

odinuge:

Just updating a logline.

This code path is always hit when mounting, even when the filesystem was already present, and we just mount it.

Reading these log lines for existing PVCs makes my heart miss a beat, when I realize that the volume was indeed not formatted before mounting. If its being formatted, its already present in logs above when on v=4.

**What testing is done?** 

n/a

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -238,7 +238,7 @@ func (d *nodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	}
 
 	// FormatAndMount will format only if needed
-	klog.V(4).InfoS("NodeStageVolume: formatting and mounting with fstype", "source", source, "volumeID", volumeID, "target", target, "fstype", fsType)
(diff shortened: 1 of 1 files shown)


torredil (commented):
Can we also make this change in [#241](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/771e745446ce12f17a523ca4a95fd673399c308f/pkg/driver/node.go#L241)? If the disk is already formatted or it is being mounted as read-only, it will be mounted without formatting.

suggest: `NodeStageVolume: staging volume`

torredil:
Looks great, one last request before merging this change; lets squash the commits: https://kubernetes.io/docs/contribute/new-content/open-a-pr/#squashing-commits

odinuge:
Thanks for the suggestion @torredil! Rebased and pushed now!

torredil (approved):
@odinuge Thank you for your contribution 👍

PR #1614, opened 2023-05-24
Namespace-scoped leases permissions

### LINKED ISSUE #1564: overly broad permissions: `leases` permissions should be namespace scoped
/kind bug

**What happened?**

The current deployment creates a ClusterRole/ClusterRoleBinding with the `leases` permission.

The deployment uses the minimal set of permissions necessary. This set of permissions should be able to be moved to a Role/RoleBinding in the namespace where the driver is deployed. As best I can tell, the code only needs to create/use a lease in the containing namespace and not in all namespaces cluster wide.

Follow the install instructions for either a kustomize or helm install.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/cbd46975f3f423cfb857250f1c13c2987f9eb56f/deploy/kubernetes/base/clusterrole-provisioner.yaml#L34-L36

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/cbd46975f3f423cfb857250f1c13c2987f9eb56f/charts/aws-ebs-csi-driver/templates/clusterrole-provisioner.yaml#L33-L35

**Environment**
- Kubernetes version (use `kubectl version`): 1.22
- Driver version: v1.17.0

torredil:
This PR addresses the overly broad permissions previously granted to the 
AWS EBS CSI driver for managing 'leases'. It replaces the `ClusterRole` and 
`ClusterRoleBinding`, which provided cluster-wide access, with a `Role` and 
`RoleBinding`, limiting access to the namespace in which the driver is deployed.

The original configuration granted the driver access to leases across all 
namespaces, which was unnecessary. By limiting access to the namespace in which the driver is deployed, we can enhance our security posture.

closes #1564 

**What testing is done?** 
- CI
- Manual testing


deploy/kubernetes/base/role-leases.yaml
```diff
@@ -0,0 +1,11 @@
+---
+# Source: aws-ebs-csi-driver/templates/role-leases.yaml
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  namespace: default
```
ConnorJC3, line 6: Please add ` | sed -e "/namespace: /d" ` to the `generate-kustomize` line that makes this file, so that this namespace gets tossed

deploy/kubernetes/base/rolebinding-leases.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# Source: aws-ebs-csi-driver/templates/rolebinding-leases.yaml
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ebs-csi-leases-rolebinding
+  namespace: default
```
ConnorJC3, line 7: Same as previous comment

ConnorJC3 (commented):
One nit, otherwise lgtm

deploy/kubernetes/base/role-leases.yaml
torredil: Done.

PR #1615, opened 2023-05-25
Revert Increase external test pod start timeout #1586

torredil:

Pod start timeout was increased in #1586 due to constant CI flakes, that issue has now been resolved.

This PR reverts the pod start timeout change to use the default timeout of 5 minutes.


torredil:
/hold
I will unhold when this is ready to go in.

PR #1619, opened 2023-05-29
chore: Allow to set automountServiceAccountToken in ServiceAccount

kahirokunn:

new feature

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes

> The LegacyServiceAccountTokenNoAutoGeneration feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets). (https://github.com/kubernetes/kubernetes/pull/108309, [@zshihang](https://github.com/zshihang))

Since k8s 1.24, TOKEN is not mounted automatically.
If you want to access with IRSA, you need to use a token.

Allow to set automountServiceAccountToken in ServiceAccount
https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

**What testing is done?** 

done.

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -227,6 +227,7 @@ controller:
     create: true
     name: ebs-csi-controller-sa
     annotations: {}
+    automountServiceAccountToken: true
   tolerations:
     - key: CriticalAddonsOnly
       operator: Exists
@@ -300,6 +301,7 @@ node:
     create: true
     name: ebs-csi-node-sa
     annotations: {}
+    automountServiceAccountToken: true
   enableWindows: false
   # The "maximum number of attachable volumes" per node
   volumeAttachLimit:
--- a/docs/install.md
+++ b/docs/install.md
@@ -29,7 +29,8 @@ For more information, review ["Creating the Amazon EBS CSI driver IAM role for s
 
 There are several methods to grant the driver IAM permissions:
 * Using IAM [instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html) - attach the policy to the instance profile IAM role and turn on access to [instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for the instance(s) on which the driver Deployment will run
-* EKS only: Using [IAM roles for ServiceAccounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) - create an IAM role, attach the policy to it, then follow the IRSA documentation to associate the IAM role with the driver Deployment service account, which if you are installing via Helm is determined by value `controller.serviceAccount.name`, `ebs-csi-controller-sa` by default
+* EKS only: Using [IAM roles for ServiceAccounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) - create an IAM role, attach the policy to it, then follow the IRSA documentation to associate the IAM role with the driver Deployment service account, which if you are installing via Helm is determined by value `controller.serviceAccount.name`, `ebs-csi-controller-sa` by default. If you are using k8s 1.24 or higher, the ServiceAccountToken is not mounted because the `LegacyServiceAccountTokenNoAutoGeneration` feature gate is enabled.
+Therefore, if you are using k8s 1.24 or higher, you need to set `true` to `controller.serviceAccount.autoMountServiceAccountToken`.
 * Using secret object - create an IAM user, attach the policy to it, then create a generic secret called `aws-secret` in the `kube-system` namespace with the user's credentials
 ```sh
 kubectl create secret generic aws-secret \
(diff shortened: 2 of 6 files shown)


ConnorJC3 (commented):
Needs testing with IRSA:

If this breaks IRSA, it should default on

If this *doesn't* break IRSA, I think we should consider just setting it to false directly, rather than exposing an option (as the option should never matter)

Otherwise lgtm, I'll test and report back on IRSA interactions

kahirokunn:
@ConnorJC3 
In ServiceAccounts that use IRSA in the world, everyone has the accountServiceAccountToken set to True by default.
I actually think it is better that way.
Would you change it that way?

ConnorJC3 (approved):
Actually, upon further reflection, I think this is fine as is. There are usecases where we need the creds, such as the node untainting feature, so this config should be left up to the user.

lgtm, cc @torredil for review

torredil:
/retest
/lgtm

Hey there, thanks for your contribution. Just one quick thing from me:
- Could we update the docs? the parameter name is pretty intuitive, but ideally new configurable parameters should be documented as it helps other users understand what the parameter does (and why they might want to set it to true or false). 

Some general feedback: Suggest providing more context in the PR description. We mention that we're allowing users to configure the parameter and link to the relevant docs -- which is very helpful -- but it would be excellent to explain why a user might want to configure this setting and the implications for doing so. I've had some success in situations like this in the past by spending a few extra minutes providing clarity which might help reviewers get your changes merged in a timely manner 👍

torredil (approved):

kahirokunn:
@torredil Okay. How about now?

torredil (approved):
Thanks for this @kahirokunn /lgtm

kahirokunn:
@torredil https://github.com/aws/eks-charts/blob/620e0169ec96d79086f004235d7b1764a60dbc40/stable/aws-load-balancer-controller/values.yaml#L25
By the way, aws-load-balancer-controller was true by default.
I think the same default of true would be fine, what do you think?

kahirokunn:
Since there are other environments than the EKS environment, default false may be acceptable.

kahirokunn:
Either is fine.
Can't wait to be merged. 🙌

ConnorJC3:
/hold

@kahirokunn yeah, I think we should set this to default `true`. The large majority of our users use IRSA, so changing this out from under them would be a poor experience.

Once the default is changed (don't forget to regenerate the `kustomize` files too), I can merge this in :+1:

kahirokunn:
@ConnorJC3 thx. done. 👍

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -227,6 +227,7 @@ controller:
     create: true
     name: ebs-csi-controller-sa
     annotations: {}
+    automountServiceAccountToken: false
```
ConnorJC3, line 230: @kahirokunn you forgot to change this one to true too

kahirokunn: sorry, fixed.

ConnorJC3:
I went ahead and manually cleaned the commit message to make the bot happy.

/remove-hold
/lgtm
/approve

liggitt:
> Since k8s 1.24, TOKEN is not mounted automatically.

This is not accurate, kubernetes 1.24 still mounts tokens as it did before. Those tokens are no longer stored in Secret API objects, but are still mounted by default.

docs/install.md
```diff
@@ -29,7 +29,8 @@ For more information, review ["Creating the Amazon EBS CSI driver IAM role for s
...
-* EKS only: Using [IAM roles for ServiceAccounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) - create an IAM role, attach the policy to it, then follow the IRSA documentation to associate the IAM role with the driver Deployment service account, which if you are installing via Helm is determined by value `controller.serviceAccount.name`, `ebs-csi-controller-sa` by default
+* EKS only: Using [IAM roles for ServiceAccounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) - create an IAM role, attach the policy to it, then follow the IRSA documentation to associate the IAM role with the driver Deployment service account, which if you are installing via Helm is determined by value `controller.serviceAccount.name`, `ebs-csi-controller-sa` by default. If you are using k8s 1.24 or higher, the ServiceAccountToken is not mounted because the `LegacyServiceAccountTokenNoAutoGeneration` feature gate is enabled.
+Therefore, if you are using k8s 1.24 or higher, you need to set `true` to `controller.serviceAccount.autoMountServiceAccountToken`.
```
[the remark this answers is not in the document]
kahirokunn: I'd like to know too.

PR #1621, opened 2023-06-02
Fix context handling in WaitForVolumeAttachment & add in-flight checks to attachment/detachment operations

torredil:
----
**`cloud.go`** changes:

This PR addresses incorrect handling of the Context in `WaitForAttachmentState` which implements a retry mechanism using exponential backoff to wait for an attachment to reach the expected state via the `DescribeVolumesWithContext` API.

The context is used to carry deadlines and cancellation signals across API boundaries. When polling for volume state, its possible for the context to be cancelled due to timeout etc. If this happens while we're waiting for the volume to reach the expected state, it is desirable to stop waiting and immediately return an error to the caller. This change allows us to better handle cases where the context is cancelled or reaches its deadline by stopping execution when no longer necessary.

----

**`controller.go`** changes:

Added in-flight checks to both `ControllerPublishVolume` and `ControllerUnpublishVolume`. The in-flight check acts as a synchronization mechanism by providing mutual exclusion to ensure idempotency.

----
Testing changes:

Added tests to validate code paths and logic when attaching / detaching volumes. Also did some refactoring to move towards table-driven tests which vastly improves readability and maintainability of these tests.

**What testing is done?** 
----
- Manual
- CI

**With this patch:**
```
I0602 17:16:29.032300       1 controller.go:370] "ControllerPublishVolume: attached" volumeID="vol-0163b0b755696dcdc" nodeID="i-07532540ce852721b" devicePath="/dev/xvdal"
I0602 17:16:47.545514       1 controller.go:362] "ControllerPublishVolume: attaching" volumeID="vol-0bf9e9d3205eb9fe7" nodeID="i-07532540ce852721b"
I0602 17:16:49.352063       1 controller.go:370] "ControllerPublishVolume: attached" volumeID="vol-0bf9e9d3205eb9fe7" nodeID="i-07532540ce852721b" devicePath="/dev/xvdap"
I0602 17:16:49.358931       1 controller.go:362] "ControllerPublishVolume: attaching" volumeID="vol-0bf9e9d3205eb9fe7" nodeID="i-07532540ce852721b"
I0602 17:16:49.519855       1 controller.go:370] "ControllerPublishVolume: attached" volumeID="vol-0bf9e9d3205eb9fe7" nodeID="i-07532540ce852721b" devicePath="/dev/xvdap"
 ```

**Without this patch:**

```
I0524 20:56:35.690733       1 inflight.go:74] "Node Service: volume operation finished" key="snapshot-299d9d27-a925-4954-93df-80ccd2105e85"
I0524 20:56:35.711511       1 cloud.go:713] "Waiting for volume state" volumeID="vol-05ee7370b8091e868" actual="detaching" desired="detached"
I0524 20:56:36.174965       1 inflight.go:74] "Node Service: volume operation finished" key="pvc-2a9b6b41-dc69-42ff-aa57-ec6bfcbd73e0"
I0524 20:56:36.176699       1 cloud.go:713] "Waiting for volume state" volumeID="vol-08a1528c35dfeb1ff" actual="detaching" desired="detached"
I0524 20:56:36.205657       1 cloud.go:654] "Ignoring error from describe volume, will retry" volumeID="vol-0aecdf207857b8263" err=<
	RequestCanceled: request context canceled
	caused by: context deadline exceeded
 >
I0524 20:57:31.4134426       1 cloud.go:654] "Ignoring error from describe volume, will retry" volumeID="vol-0aecdf207857b8263" err=<
	RequestCanceled: request context canceled
	caused by: context deadline exceeded
 >
...
 ```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -628,7 +628,7 @@ func (c *cloud) WaitForAttachmentState(ctx context.Context, volumeID, expectedSt
 
 	var attachment *ec2.VolumeAttachment
 
-	verifyVolumeFunc := func() (bool, error) {
+	verifyVolumeFunc := func(ctx context.Context) (bool, error) {
 		request := &ec2.DescribeVolumesInput{
 			VolumeIds: []*string{
 				aws.String(volumeID),
@@ -710,11 +710,11 @@ func (c *cloud) WaitForAttachmentState(ctx context.Context, volumeID, expectedSt
 			return true, nil
 		}
 		// continue waiting
-		klog.V(2).InfoS("Waiting for volume state", "volumeID", volumeID, "actual", attachmentState, "desired", expectedState)
+		klog.V(4).InfoS("Waiting for volume state", "volumeID", volumeID, "actual", attachmentState, "desired", expectedState)
 		return false, nil
 	}
 
-	return attachment, wait.ExponentialBackoff(backoff, verifyVolumeFunc)
+	return attachment, wait.ExponentialBackoffWithContext(ctx, backoff, verifyVolumeFunc)
 }
 
 func (c *cloud) GetDiskByName(ctx context.Context, name string, capacityBytes int64) (*Disk, error) {
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -354,27 +354,20 @@ func (d *controllerService) ControllerPublishVolume(ctx context.Context, req *cs
 	volumeID := req.GetVolumeId()
 	nodeID := req.GetNodeId()
 
-	if !d.cloud.IsExistInstance(ctx, nodeID) {
-		return nil, status.Errorf(codes.NotFound, "Instance %q not found", nodeID)
+	if !d.inFlight.Insert(volumeID) {
+		return nil, status.Error(codes.Aborted, fmt.Sprintf(internal.VolumeOperationAlreadyExistsErrorMsg, volumeID))
 	}
-	disk, err := d.cloud.GetDiskByID(ctx, volumeID)
-	if err != nil {
-		if errors.Is(err, cloud.ErrNotFound) {
-			return nil, status.Error(codes.NotFound, "Volume not found")
-		}
-		return nil, status.Errorf(codes.Internal, "Could not get volume with ID %q: %v", volumeID, err)
-	}
+	defer d.inFlight.Delete(volumeID)
 
-	// If given volumeId already assigned to given node, will directly return current device path
+	klog.V(2).InfoS("ControllerPublishVolume: attaching", "volumeID", volumeID, "nodeID", nodeID)
 	devicePath, err := d.cloud.AttachDisk(ctx, volumeID, nodeID)
 	if err != nil {
 		if errors.Is(err, cloud.ErrVolumeInUse) {
-			return nil, status.Error(codes.FailedPrecondition, strings.Join(disk.Attachments, ","))
+			return nil, status.Errorf(codes.FailedPrecondition, "Volume %q is already attached to a different node, expected node: %q", volumeID, nodeID)
 		}
-		// TODO: Check volume capability matches for ALREADY_EXISTS
 		return nil, status.Errorf(codes.Internal, "Could not attach volume %q to node %q: %v", volumeID, nodeID, err)
 	}
-	klog.V(5).InfoS("[Debug] ControllerPublishVolume: attached to node", "volumeID", volumeID, "nodeID", nodeID, "devicePath", devicePath)
+	klog.V(2).InfoS("ControllerPublishVolume: attached", "volumeID", volumeID, "nodeID", nodeID, "devicePath", devicePath)
 
 	pvInfo := map[string]string{DevicePathKey: devicePath}
 	return &csi.ControllerPublishVolumeResponse{PublishContext: pvInfo}, nil
@@ -413,13 +406,20 @@ func (d *controllerService) ControllerUnpublishVolume(ctx context.Context, req *
 	volumeID := req.GetVolumeId()
 	nodeID := req.GetNodeId()
 
+	if !d.inFlight.Insert(volumeID) {
+		return nil, status.Error(codes.Aborted, fmt.Sprintf(internal.VolumeOperationAlreadyExistsErrorMsg, volumeID))
+	}
+	defer d.inFlight.Delete(volumeID)
+
+	klog.V(2).InfoS("ControllerUnpublishVolume: detaching", "volumeID", volumeID, "nodeID", nodeID)
 	if err := d.cloud.DetachDisk(ctx, volumeID, nodeID); err != nil {
 		if errors.Is(err, cloud.ErrNotFound) {
+			klog.V(2).InfoS("ControllerUnpublishVolume: attachment not found", "volumeID", volumeID, "nodeID", nodeID)
 			return &csi.ControllerUnpublishVolumeResponse{}, nil
 		}
 		return nil, status.Errorf(codes.Internal, "Could not detach volume %q from node %q: %v", volumeID, nodeID, err)
 	}
-	klog.V(5).InfoS("[Debug] ControllerUnpublishVolume: detached from node", "volumeID", volumeID, "nodeID", nodeID)
+	klog.V(2).InfoS("ControllerUnpublishVolume: detached", "volumeID", volumeID, "nodeID", nodeID)
 
 	return &csi.ControllerUnpublishVolumeResponse{}, nil
 }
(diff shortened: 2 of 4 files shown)


pkg/cloud/cloud.go
```diff
@@ -576,15 +590,22 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
...
 	device, err := c.dm.GetDevice(instance, volumeID)
 	if err != nil {
 		return err
 	}
 	defer device.Release(true)
 
-	if !device.IsAlreadyAssigned {
-		klog.InfoS("DetachDisk: called on non-attached volume", "volumeID", volumeID)
+	if !isDiskAttached(disk, nodeID) && !device.IsAlreadyAssigned {
+		return nil
```
ConnorJC3, line 608: I am VERY worried that this is fundamentally unsafe, especially in a scenarios involving parallelism or multiple instances of the controller (which is, by my understanding, *legal* according to the CSI spec)

Can we please add a justification explaining why this is safe as a comment in the code, both to
1. Prove to us, on this PR, that this change is actually safe, and
2. To convey the reasoning for that to future maintainers

pkg/cloud/cloud.go
```diff
@@ -710,11 +731,20 @@ func (c *cloud) WaitForAttachmentState(ctx context.Context, volumeID, expectedSt
 			return true, nil
 		}
 		// continue waiting
-		klog.V(2).InfoS("Waiting for volume state", "volumeID", volumeID, "actual", attachmentState, "desired", expectedState)
+		klog.V(4).InfoS("Waiting for volume state", "volumeID", volumeID, "actual", attachmentState, "desired", expectedState)
```
ConnorJC3, line 734: Why? This log message has been historically very useful for debugging attach-related issues

pkg/driver/controller_test.go
```diff
@@ -3112,450 +3114,251 @@ func TestControllerPublishVolume(t *testing.T) {
...
-				mockCloud := cloud.NewMockCloud(mockCtl)
+			awsDriver, mockCtl, mockCloud := createControllerService(t)
+			defer mockCtl.Finish()
 
-				awsDriver := controllerService{
-					cloud:         mockCloud,
-					inFlight:      internal.NewInFlight(),
-					driverOptions: &DriverOptions{},
-				}
+			if tc.name == "Aborted error when operation already in-flight" {
```
ConnorJC3, line 3344: Never do this, needs to be changed. Add a setup function or similar to the test parameters, do not hardcode behavior based on test name.

Changing the test name should NEVER affect the test's behavior.

pkg/cloud/cloud.go
```diff
@@ -515,12 +517,24 @@ func (c *cloud) AttachDisk(ctx context.Context, volumeID, nodeID string) (string
...
+	}
+
 	device, err := c.dm.NewDevice(instance, volumeID)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("could not create device for volume %q: %w", volumeID, err)
 	}
 	defer device.Release(false)
 
+	if isDiskAttached(disk, nodeID) && device.IsAlreadyAssigned {
```
ConnorJC3, line 534: Is this not vulnerable to the "time travel" problem documented here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L680-L683 ?

torredil: The `WaitForAttachmentState` function is polling AWS for the current state of the volume attachment and waits until the volume reaches the expected state (attached or detached). It checks the state reported by AWS directly. If AWS's eventual consistency leads to temporarily incorrect state reports, this function is designed to handle that. It retries until the state matches the expected state, or until the maximum number of retries is exhausted.

When the `AttachDisk` function checks `if isDiskAttached(disk, nodeID) && device.IsAlreadyAssigned`, it's checking the local state before calling `WaitForAttachmentState`. This local state check is "safe" in the sense that it's based on the state as understood by the cloud controller, which should be internally consistent.

However, the real concern with AWS's eventual consistency is how it impacts interactions with AWS's APIs, such as when issuing attach requests or when polling for the attachment state. These are the points where the cloud controller's understanding of the state needs to align with AWS's understanding of the state. That's why error checking and retries are used in these parts of the code.

For the specific case of the device already being assigned and the disk already being attached (as per `isDiskAttached(disk, nodeID) && device.IsAlreadyAssigned`), this should be a stable state. If AWS reports a different state due to eventual consistency issues, the `WaitForAttachmentState` function handles this by continuously polling until the reported state matches the expected state.

So, the line is safe because it relies on internal consistency within the cloud controller. Any potential inconsistencies arising from AWS's eventual consistency are handled in the parts of the code that interact with AWS's APIs directly, including the `WaitForAttachmentState` function.

pkg/cloud/cloud.go
torredil: Higher verbosity levels are typically used for this category of logs which are very frequent, logging volume state results in a high volume of log entries. 

Instead, we now log the actual attachment / detachment times which also provides useful information for debugging, a more comprehensive view of a volume's attachment/detachment lifecycle, and substantially reduces the overall volume of log entries at the default verbosity level.

pkg/cloud/cloud.go
torredil: Added a comment. 

This early exit condition is safe because if the disk is not attached to the node and the device is not assigned there is nothing more to do in `DetachDisk`. This operation is idempotent, which means it can be repeated without changing the final outcome, thus even in a scenario with parallelism or multiple controller instances this function is safe due to the idempotency.

There is one potential race condition: its possible that between the time we check if the disk is attached and when `DetachVolumeWithContext` is called another process has already detached the volume. In the worst case this would simply result in an extra unnecessary API call. This scenario is handled gracefully by the error checks following the `DetachVolumeWithContext` call.

> multiple instances of the controller (which is, by my understanding, legal according to the CSI spec)

By my understanding more than one controller running concurrently is not supported, see https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/516#issuecomment-635453521 has there been a change or update to the CSI spec that allows for this?

pkg/driver/controller_test.go
torredil: Done.

pkg/cloud/cloud.go
ConnorJC3: > This early exit condition is safe because if the disk is not attached to the node and the device is not assigned there is nothing more to do in DetachDisk

My point is that I don't think the code as written actually verifies that "the disk is not attached to the node"

I'm not talking about k8s, I'm talking about CSI

pkg/cloud/cloud.go
ConnorJC3: I'm worried about the case where the API reports the same (**but incorrect**) state due to eventual consistency issues, as described here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L680-L683

pkg/cloud/cloud.go
torredil: Let's take the extra API call and attempt to detach regardless of the disk's perceived attachment status to address eventual consistency.

ConnorJC3 (approved):
lgtm

pkg/driver/controller.go
```diff
@@ -354,27 +354,20 @@ func (d *controllerService) ControllerPublishVolume(ctx context.Context, req *cs
...
 	}
+	defer d.inFlight.Delete(volumeID)
 
-	// If given volumeId already assigned to given node, will directly return current device path
+	klog.V(2).InfoS("ControllerPublishVolume: attaching", "volumeID", volumeID, "nodeID", nodeID)
 	devicePath, err := d.cloud.AttachDisk(ctx, volumeID, nodeID)
 	if err != nil {
 		if errors.Is(err, cloud.ErrVolumeInUse) {
-			return nil, status.Error(codes.FailedPrecondition, strings.Join(disk.Attachments, ","))
+			return nil, status.Errorf(codes.FailedPrecondition, "Volume %q is already attached to a different node", volumeID)
```
hanyuel, line 366: Do we need to include node id in the error message so we know which node was requested to have the volume attach to?

pkg/driver/controller.go
```diff
@@ -354,27 +354,20 @@ func (d *controllerService) ControllerPublishVolume(ctx context.Context, req *cs
 	volumeID := req.GetVolumeId()
 	nodeID := req.GetNodeId()
 
-	if !d.cloud.IsExistInstance(ctx, nodeID) {
-		return nil, status.Errorf(codes.NotFound, "Instance %q not found", nodeID)
-	}
-	disk, err := d.cloud.GetDiskByID(ctx, volumeID)
```
hanyuel, line 360: Why is it safe to remove ``ec2.DescribeVolume`` call?

ConnorJC3: It should probably be included in the PR description, but by my understanding this check is superfluous and removed for efficiency.

hanyuel: Why was  this check added before?

pkg/driver/controller.go
torredil: We log that information right before calling `AttachDisk`: https://github.com/torredil/aws-ebs-csi-driver/blob/62c55109900ee31e9d1b3c9d3e96feeabdda7714/pkg/driver/controller.go#L362 but more clarity is always helpful. Changed to `"Volume %q is already attached to a different node, expected node: %q", volumeID, nodeID)`.

PR #1626, opened 2023-06-14
Update DM allocator to use all available names

ConnorJC3:

Bug fix

Changes the device name allocator to use a list of all legal EBS device names.

Supersedes #1622 

**What testing is done?** 

Updated unit tests, manually tested, CI


hanyuel:
Can you please add in your PR description explaining what is the bug/problem?

pkg/cloud/devicemanager/device_names.go
```diff
@@ -0,0 +1,152 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package devicemanager
+
+// EC2 allowed EBS device names, discovered by trial and error
+// Notable (undoumented) restrictions include:
```
hanyuel, line 20: typo: ``undocumented``

pkg/cloud/devicemanager/allocator.go
```diff
@@ -34,34 +34,19 @@ type ExistingNames map[string]string
...
-		for c2 := 'a'; c2 <= c2end; c2++ {
-			name := fmt.Sprintf("%s%s%s", prefix, string(c1), string(c2))
-			if _, found := existingNames[name]; !found {
-				return name, nil
-			}
+// GetNext returns a free device name or error when there is no free device name
+// It does this by using a list of legal EBS device names from device_names.go
+func (d *nameAllocator) GetNext(existingNames ExistingNames) (string, error) {
+	for _, name := range deviceNames {
+		if _, found := existingNames[name]; !found {
```
hanyuel, line 48: I know this PR doesn't change it, but why does ``ExistingNames`` need to be a map? We care about only the keys, not the values.

ConnorJC3: `ExistingNames` is a map because that is how it is stored and passed in `manager.go`.

torredil (approved):

PR #1630, opened 2023-06-14
Avoid generating manifests with empty envFrom fields

### LINKED ISSUE #1629: Kustomize fails to patch controller Deployment
/kind bug

## What happened?
When applying multiple patches to the [deploy manifests](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/deploy/kubernetes) via Kustomize (v5.0.3+), it fails to build with the following error:

```bash
$ kustomize build ./deploy/kubernetes/overlays/stable 
Error: updating name reference in 'spec/template/spec/containers/envFrom/configMapRef/name' field of 'Deployment.v1.apps/ebs-csi-controller.kube-system': considering field 'spec/template/spec/containers/envFrom/configMapRef/name' of object Deployment.v1.apps/ebs-csi-controller.kube-system: visit traversal on path: [envFrom configMapRef name]: expected sequence or mapping node
```

It expects the `spec/template/spec/containers/envFrom` field to be a sequence or mapping node, however, it's null.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ed3c33d6733db26c04bc975e708122f6b144fa99/deploy/kubernetes/base/controller.yaml#L97

## What you expected to happen?

I expected both patches to be applied.

## How to reproduce it (as minimally and precisely as possible)?

Add the following patches to the [deploy/kubernetes/overlays/stable/kustomization.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/overlays/stable/kustomization.yaml):

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - ./ecr-public
patches:
  - patch: '[{"op": "add","path": "/spec/template/spec/containers/0/args/-","value": "--http-endpoint=:8080"}]'
    target:
      group: apps
      version: v1
      kind: Deployment
      name: ebs-csi-controller
  - patch: |-
      apiVersion: apps/v1
      kind: Deployment
      metadata:
        namespace: kube-system
        name: ebs-csi-controller
      spec:
        template:
          metadata:
            annotations:
              prometheus.io/port: '8080'
              prometheus.io/scrape: 'true'
```

Then run:
```bash
kustomize build ./deploy/kubernetes/overlays/stable 
```

## Anything else we need to know?

This problem can be solved by avoiding generating empty `envFrom` fields from the Helm charts.
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ed3c33d6733db26c04bc975e708122f6b144fa99/deploy/kubernetes/base/controller.yaml#L97
**Environment**
- Kubernetes version (use `kubectl version`): N/A
- kustomize version: v5.0.3
- Driver version: v1.19.0

mvgmb:
Closes #1629

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -132,10 +132,10 @@ spec:
             {{- with .Values.controller.env }}
             {{- . | toYaml | nindent 12 }}
             {{- end }}
+          {{- with .Values.controller.envFrom }}
           envFrom:
-            {{- with .Values.controller.envFrom }}
             {{- . | toYaml | nindent 12 }}
-            {{- end }}
+          {{- end }}
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
@@ -210,10 +210,10 @@ spec:
             {{- with .Values.sidecars.provisioner.env }}
             {{- . | toYaml | nindent 12 }}
             {{- end }}
+          {{- with .Values.controller.envFrom }}
           envFrom:
-            {{- with .Values.controller.envFrom }}
             {{- . | toYaml | nindent 12 }}
-            {{- end }}
+          {{- end }}
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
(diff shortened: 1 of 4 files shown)


torredil (approved):
Thanks @mvgmb /lgtm

PR #1659, opened 2023-06-26
Add 15 second rate limit to CreateSnapshot

ConnorJC3:

Bug fix

See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1608
See https://github.com/kubernetes-csi/external-snapshotter/issues/778

Adds a 15 second rate limit to CreateSnapshot when the failure originates from CreateSnapshot in cloud (i.e. the error likely originates from the AWS API).

This prevents the driver from getting stuck in an infinite loop if snapshot creation fails, where it will indefinately retry creating a snapshot and continue to receive an error because it is going too fast.

**What testing is done?** 

Manual/CI/Inflight already has unit tests

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/awslabs/volume-modifier-for-k8s/pkg/rpc"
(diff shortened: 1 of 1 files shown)


hanyuel:
The change in this PR is identical to #1611 . What was the reason we need a new PR?

hanyuel:
> The change in this PR is identical to #1611 . What was the reason we need a new PR?

Never mind. I figured out it's because both are merging from the same branch.

ConnorJC3:
To give more context, #1611 was created on a previous version of this change, but this methodology differs so drastically from the version that @rdpsin reviewed in #1611 that I figured starting fresh with a new PR was the appropriate choice.

This PR supersedes and replaces #1611.

torredil:
This PR can be closed now that https://github.com/kubernetes-csi/external-snapshotter/pull/871 has (hopefully) been fixed.

PR #1661, opened 2023-06-28
Support specifying inode size for filesystem format

### LINKED ISSUE #1660: Support specifying inode size for filesystem format
In some cases, we may run out of inodes, so I would like aws-ebs-csi-driver to support setting the inode size during formatting the filesystem, so that the number of inodes can be adjusted.

Just like #1452 did, add a `inodeSize` parameter to the storageclass.

N/A

**Additional context**
I'd like to submit PR to implement it.
/kind feature
/assign

fgksgf:
New feature

Close #1660

**What testing is done?** 
UT and manual test

## Manual Test
- Create a EKS cluster, deploy ebs-csi-driver, replace the image with the one built from this PR
- Create a pod with the manifests (no parameters):
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 2Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: app
spec:
  containers:
  - name: app
    image: ubuntu
    securityContext:
      privileged: true
    tty: true
    command: ["/bin/sh"]
    args: ["-c", "while true; do echo $(date -u) >> /data/out.txt; sleep 30; done"]
    volumeMounts:
    - name: persistent-storage
      mountPath: /data
    - mountPath: /host/var/lib/kubelet
      mountPropagation: Bidirectional
      name: kubelet-dir
    - mountPath: /host/dev
      name: device-dir
  volumes:
  - name: persistent-storage
    persistentVolumeClaim:
      claimName: ebs-claim
  - hostPath:
      path: /var/lib/kubelet
      type: Directory
    name: kubelet-dir
  - hostPath:
      path: /dev
      type: Directory
    name: device-dir
``` 
- Run `kubectl exec -it app -- tune2fs -l /dev/nvme1n1`, get the output, save it as `base.txt`, then delete them.

### `inodeSize`
- Create a pod with the manifests but do some modification (set the parameter `inodeSize: "512"`), get the output of `tune2fs`, save it as `inode-size.txt`, then delete the resources
- Diff `base.txt` with `inode-size.txt`:
```diff
- Inode blocks per group:   512
+ Inode blocks per group:   1024
 
- Inode size:	        256
+ Inode size:	        512
```

### `bytesPerINode`
- Create a pod with the manifests but do some modification (set the parameter `bytesPerINode: "8192"`), get the output of `tune2fs`:
```
tune2fs 1.46.5 (30-Dec-2021)
...
Inode count:              262144
Block count:              524288
Block size:               4096
Fragment size:            4096
Group descriptor size:    64
Reserved GDT blocks:      255
Blocks per group:         32768
Fragments per group:      32768
Inodes per group:         16384
Inode blocks per group:   1024
Flex block group size:    16
...
Inode size:	        256
...
```

- Since `bytes-per-inode = (total blocks * block size) / inode count`, i.e. `524288 * 4096 / 262144 = 8192`, we can confirm the parameter works.

### `numberOfINodes`
- Create a pod with the manifests but do some modification (set the parameter `numberOfINodes: "200000"`), get the output of `tune2fs`, save it as `inode-num.txt`, then delete the resources
- Diff `base.txt` with `inode-num.txt`:
```diff
- Inode count:              131072
+ Inode count:              200192

- Inodes per group:         8192
+ Inodes per group:         12512

- Inode blocks per group:   512
+ Inode blocks per group:   782
```


hanyuel:
Hi @fgksgf , thanks for submitting the PR. Can you add (in the PR description) what testings have been done?

ConnorJC3:
Feel free to ignore the Windows test failure, it's due to a separate ongoing issue we're working on.

Based on a rough skim of the PR it looks good to me, but I'd like to do an in-depth review and manually test this locally before I give the official lgtm - I'll report back when I get to do that (probably later this week)

fgksgf:
> Hi @fgksgf , thanks for submitting the PR. Can you add (in the PR description) what testings have been done?

@hanyuel Ok, before that I want to add other inode related parameters those are only supported by ext2, ext3, and ext4, including `-i bytes-per-inode` and `-N number-of-inodes`, what do you think? Do i need to file an another PR?

/cc @ConnorJC3 

> Feel free to ignore the Windows test failure, it's due to a separate ongoing issue we're working on.

Thank you for the kind reminder.

fgksgf:
@hanyuel @ConnorJC3  Hello, I just added what testings I have been done, PTAL :)

torredil (changes requested):
I think the current approach is fine for a few exclusion types, but it's not very scalable. I'm also not a big fan of exposing these maps globally.  Is it unreasonable to expect that we will add more exclusion types in the future?

Rather than having multiple maps each tracking the support for a single parameter, we should use a single map that tracks the support for all parameters:

```
type FileSystemConfig struct {
	SupportedParams map[string]bool
}

var FileSystemConfigs = make(map[string]FileSystemConfig)

// FSTypeExt2 config
// FSTypeExt3 config
// FSTypeExt4 config
// FSTypeXfs config

// FSTypeNtfs config
FileSystemConfigs[FSTypeNtfs] = FileSystemConfig{
	SupportedParams: map[string]bool{
		"blockSize":       false,
		"inodeSize":       false,
		"bytesPerInode":   false,
		"numberOfInodes":  false,
	},
}
```

and then we can just do 

```
func (fsConfig FileSystemConfig) isParameterSupported(paramName string) bool {
	supported, _ := fsConfig.SupportedParams[paramName]
	return supported
}
```

This is easier to reason about and extend. Whenever a new param or fsType is introduced we only have to update `FileSystemConfigs`,  everything else should work with no further modifications.

ConnorJC3 (approved):
This is awesome, lgtm as far as I'm concerned

cc @torredil for review

torredil (approved):
This is looking very clean now @fgksgf, perfect! 

Thank you for your contribution and the very detailed testing section in the PR description, appreciate that.

torredil:
One last request: Will merge this PR once the commits have been [squashed](https://www.kubernetes.dev/docs/guide/github-workflow/#squash-commits).

PR #1670, opened 2023-07-05
Allow additional volume mounts on node pods

### LINKED ISSUE #1544: Add SELinux mounts
Bottlerocket worked through an [issue](https://github.com/bottlerocket-os/bottlerocket/issues/2656) where there are advantages to avoiding an SELinux relabel of all files when mounting a volume. The approach is to make the driver "SELinux aware" and then set the mount context in `mountOptions` to specify the label rather than enumerate and relabel all the files on attachment. The way it stands, it appears one has to patch the CSI driver to make the SELinux sysfs and config available to make this work.

I am proposing that we add these mounts to the driver so systems with SELinux will automatically gain access to the SELinux config to be able to use the `context` in `mountOptions`

Essentially we would add this to the node definition:
for `volumes`:
```
# SELinux specific volumes
- hostPath:
    path: /sys/fs/selinux
    type: Directory
  name: selinuxfs
- hostPath:
    path: /etc/selinux/config
    type: FileOrCreate
  name: selinux-config
```
and then the `volumeMounts`:
```
# SELinux specific volume mounts
- mountPath: /sys/fs/selinux
  name: selinuxfs
- mountPath: /etc/selinux/config
  name: selinux-config
```

Do nothing, the user can patch the driver with the mounts and gain this functionality. This is a bit more work but does allow users to make the decision on if they want SELinux contexts in the driver.

**Additional context**
A lot of this context was gained from this [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1710-selinux-relabeling) and this essentially starts down the path of "opting-in" support for this support.

AndrewSirenko:

Feature

Closes #1544 

This PR adds a helm parameter that allows additional volume mounts on the node pods (and adds documentation on how to add additional volume mounts on the controller).  

This would allow users to add their own mounts to the driver via the Helm parameters `node.volumes` and `node.volumeMounts` in [this PR's values.yaml](https://github.com/AndrewSirenko/aws-ebs-csi-driver/blob/3de7fded5908a4abc731227594e76255ce4b9cdc/charts/aws-ebs-csi-driver/values.yaml#L342-L351) without needing to patch the driver. 

**What testing is done?** 

Tested on live k8s cluster. 

1. Installed the ebs-csi-driver to a test cluster with two additional volume mounts for node pods. 
values.yaml

2. Describing one of the node pods "kubectl describe  pod ebs-csi-node-hstbp -n kube-system" the additional two volume mounts that were requested. 
Additional Volumes Mounted

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -264,8 +264,18 @@ controller:
     runAsUser: 1000
     runAsGroup: 1000
     fsGroup: 1000
+  # Add additional volume mounts on the controller with controller.volumes and controller.volumeMounts
   volumes: []
+  # Add additional volumes to be mounted onto the controller:
+  # - name: custom-dir
+  #   hostPath:
+  #     path: /path/to/dir
+  #     type: Directory
   volumeMounts: []
+  # And add mount paths for those additional volumes:
+  # - name: custom-dir
+  #   mountPath: /mount/path
+  # ---
   # securityContext on the controller container (see sidecars for securityContext on sidecar containers)
   containerSecurityContext:
     readOnlyRootFilesystem: true
@@ -328,6 +338,18 @@ node:
     runAsUser: 0
     runAsGroup: 0
     fsGroup: 0
+  # Add additional volume mounts on the node pods with node.volumes and node.volumeMounts
+  volumes: []
+  # Add additional volumes to be mounted onto the node pods:
+  # - name: custom-dir
+  #   hostPath:
+  #     path: /path/to/dir
+  #     type: Directory
+  volumeMounts: []
+  # And add mount paths for those additional volumes:
+  # - name: custom-dir
+  #   mountPath: /mount/path
+  # ---
   # securityContext on the node container (see sidecars for securityContext on sidecar containers)
   containerSecurityContext:
     readOnlyRootFilesystem: true
--- a/charts/aws-ebs-csi-driver/templates/node.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node.yaml
@@ -89,6 +89,9 @@ spec:
               mountPath: /csi
             - name: device-dir
               mountPath: /dev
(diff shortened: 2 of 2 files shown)


ConnorJC3 (approved):
/lgtm

Given that this is the config for the EBS CSI Driver chart, "driver's node pods" and "node pods" mean the same thing - there are no other node pods that users could be confused about.

Either wording is good in my opinion, as a matter of fact "node pods" is probably slightly better as it doesn't include obvious/redundant information.

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -328,6 +338,18 @@ node:
     runAsUser: 0
     runAsGroup: 0
     fsGroup: 0
+  # Add additional volume mounts on the node pods with node.volumes and node.volumeMounts
+  volumes: [ ]
```
torredil, line 342: Let's do `[  ]` -> `[]` for consistency, similar to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/280ac42f577409c1976b88cb2ebd0d9c3d161725/charts/aws-ebs-csi-driver/values.yaml#L267

Same request at 348

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -264,8 +264,18 @@ controller:
     runAsUser: 1000
     runAsGroup: 1000
     fsGroup: 1000
+  # Add additional volume mounts on the controller with controller.volumes and controller.volumeMounts
   volumes: []
+  # Add additional volumes to be mounted onto the controller like:
```
torredil, line 269: nit: ~like~

torredil (changes requested):
Requested small changes, otherwise largely lgtm

AndrewSirenko:
> Requested small changes, otherwise largely lgtm

Fixed in latest change @torredil

torredil (approved):
Thanks for the quick turnaround, will let hanyuel take a look at this before approving /lgtm

PR #1681, opened 2023-07-13
Update csidriver.yaml

ashish-dua:
Want to keep fsGroupPolicy fixed as per project requirement


ConnorJC3:
/close

Hello @ashish-dua, if you do not want `fsGroupPolicy to be set, please use the provided `useOldCSIDriver` parameter. By setting that parameter to true (for example, via `--set useOldCSIDriver=true`, `fsGroupPolicy` will not be set).

PR #1689, opened 2023-07-13
Use SSM to get latest stable AMI for EC2 nodes

torredil:

Use AWS SSM to get the latest AL2023 AMI ID.

**What testing is done?** 
- CI

### DIFF
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -55,6 +55,7 @@ KOPS_VERSION=${KOPS_VERSION:-1.27.0-beta.3}
 KOPS_STATE_FILE=${KOPS_STATE_FILE:-s3://k8s-kops-csi-e2e}
 KOPS_PATCH_FILE=${KOPS_PATCH_FILE:-./hack/kops-patch.yaml}
 KOPS_PATCH_NODE_FILE=${KOPS_PATCH_NODE_FILE:-./hack/kops-patch-node.yaml}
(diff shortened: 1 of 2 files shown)


hanyuel (approved):

hack/e2e/run.sh
```diff
@@ -55,6 +55,7 @@ KOPS_VERSION=${KOPS_VERSION:-1.27.0-beta.3}
 KOPS_STATE_FILE=${KOPS_STATE_FILE:-s3://k8s-kops-csi-e2e}
 KOPS_PATCH_FILE=${KOPS_PATCH_FILE:-./hack/kops-patch.yaml}
 KOPS_PATCH_NODE_FILE=${KOPS_PATCH_NODE_FILE:-./hack/kops-patch-node.yaml}
+AMI_ID=$(aws ssm get-parameters --names /aws/service/ami-amazon-linux-latest/al2023-ami-kernel-default-x86_64 --region ${REGION} | jq -r '.Parameters[0].Value')
```
AndrewSirenko, line 58: Add `--output json` to the `get-parameters` call in case AWS CLI configured to not JSON.

ConnorJC3, line 58:
```suggestion
AMI_ID=$(aws ssm get-parameters --names /aws/service/ami-amazon-linux-latest/al2023-ami-kernel-default-x86_64 --region ${REGION} --query 'Parameters[0].Value' --output text)
```

Even better: no `jq` needed

torredil: Even if the default output is JSON, being explicit makes this more resilient -- good mindset. 

Streamlining this further by filtering the output to just the value of the parameter and outputting the result as plain text  as per https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1689#discussion_r1264075194.

PR #1705, opened 2023-08-01
Remove error_handler breaking CI

ConnorJC3:
`error_handler` breaks CI - removing it until we can find a better solution
See, for example, the end of the logs for https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/1703/pull-aws-ebs-csi-driver-test-e2e-external-eks-windows/1686370764527243264


torredil (approved):
/approve

PR #1711, opened 2023-08-08
Bump golang.org/x/net/html to fix CVE-2023-3978

jsafrane:
/kind bug

The CVE is about HTML rendering and I don't think this CSI driver renders any HTML. Trying to silence security scanners.

**What testing is done?** 
`make test`


torredil (approved):
Thanks @jsafrane /lgtm
/approve

PR #1734, opened 2023-09-06
Temporarily workaround Windows e2e failures with larger instance type and slimmer AMI

ConnorJC3:

"Bug fix"

Windows e2e is broken since the June AMI, this temporarily works around that breakage so the Windows e2e tests will be a reliable signal again.

**What testing is done?** 

CI


torredil (approved):
/lgtm

PR #1736, opened 2023-09-11
PreStop lifecycle hook feature

torredil:

New feature.

- When a node is terminated, persistent workflows using EBS volumes can take 6+ minutes to start up again. This happens when a volume is not cleanly unmounted, which causes the [Attach/Detach controller](https://github.com/kubernetes/kubernetes/issues/20262) (in kube-controller-manager) to wait for [6 minutes](https://github.com/kubernetes/kubernetes/blob/3f579d8971fcce96d6b01b968a46c720f10940b8/pkg/controller/volume/attachdetach/attach_detach_controller.go#L94) before issuing a force detach and allowing the volume to be attached to another node.  For context, the 6 minute interval is a non-configurable value used as a safety mechanism to handle uncertainty during workflow interruptions with the goal of preventing data corruption caused by forced detachments.
- The lifecycle hook aims to ensure that before the node (and the CSI driver node pod running on it) is shut down, all VolumeAttachment objects associated with that node are removed, indicating that all volumes have been successfully unmounted and detached. This way, we can be sure that all race conditions during the process are taken care of (node side and controller side).
- No unnecessary delay is added to the termination workflow, as the PreStop hook logic is only executed when the node is being drained (thus preventing delays in termination where the node pod is killed due to a rolling restart, or during driver upgrades, but the workload pods are expected to be running).
- By leveraging the PreStop hook, there's no need to rely on external tools or handlers like the AWS Node Termination Handler to prevent 6 minute delays.

**What testing is done?** 
- Unit testing.
- Manual testing:

1. Setup dev cluster.

2. Build and deploy new driver image with changes introduced in this PR.

3. Apply dynamic provisioning manifests `kubectl apply -f manifests`:
```
persistentvolumeclaim/ebs-claim created
pod/app created
storageclass.storage.k8s.io/ebs-sc created
```

4. Verify VolumeAttachment object is present `kubectl get volumeattachment`:
```
kubectl get volumeattachment
NAME                                                                   ATTACHER          PV                                         NODE                  ATTACHED   AGE
csi-e34944e5bc014158229c0d1cb82f38d803f9f1eab0369280fddfb206c3d57b5d   ebs.csi.aws.com   pvc-90f38262-3d71-40be-beb4-553a773c74d9   i-01cf45161c87325e8   true       20s
```

5. Cordon node to simulate drain event, `kubectl cordon <node-name>`.

6. Describe node to confirm node is unschedulable. `kubectl describe <node-name>`.
```
Taints:             node.kubernetes.io/unschedulable:NoSchedule
Unschedulable:      true

Normal   NodeNotSchedulable       1s                 kubelet                Node i-01cf45161c87325e8 status is now: NodeNotSchedulable
```

7. Delete driver node pod on unschedulable node with associated VolumeAttachment object and confirm driver node pod is not taken down until either `terminationGracePeriodSeconds` elapses or VolumeAttachment object is removed.


cmd/options.go
```diff
@@ -95,6 +98,18 @@ func GetOptions(fs *flag.FlagSet) *Options {
 			nodeOptions.AddFlags(fs)
 			args = os.Args[1:]
 
+		case cmd == "pre-stop-hook":
+			clientset, clientErr := cloud.DefaultKubernetesAPIClient()
+			if clientErr != nil {
+				klog.ErrorS(err, "unable to communicate with k8s API")
+			} else {
+				err = hooks.PreStop(clientset, 1*time.Minute)
```
ConnorJC3, line 106: Firstly, can we move the special time value to a constant at the top of this file called something like `preStopTimeout` so it isn't just a magic value?

Secondly, this value should be 30 seconds, as the default `terminationGracePeriod` (which we are using) is 30 seconds: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+	if isNodeBeingDrained(node) {
+		klog.InfoS("PreStop: node is being drained, checking for remaining VolumeAttachments", "node", nodeName)
+		return waitForVolumeAttachments(ctx, clientset, nodeName)
+	}
+
+	klog.InfoS("PreStop: node is not being drained, skipping VolumeAttachments check", "node", nodeName)
+	return nil
+}
+
+func fetchNodeInfo(ctx context.Context, clientset kubernetes.Interface, nodeName string) (*v1.Node, error) {
```
ConnorJC3, line 42: Suggest renaming this to `fetchNode`, as it is fetching the entire node (object), not "node info"

ConnorJC3: Also, not in scope for this PR, but I wonder if it's worth creating a "k8s utils" style package that does this kind of common operation? We probably have nearly identical code to this in the taint removal feature.

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+	defer close(stopCh)
+
+	allAttachmentsDeleted := make(chan struct{})
+
+	factory := informers.NewSharedInformerFactory(clientset, 0)
+	informer := factory.Storage().V1().VolumeAttachments().Informer()
+
+	_, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			klog.V(4).InfoS("DeleteFunc: VolumeAttachment deleted", "node", nodeName)
```
ConnorJC3, line 70: This can be very noisy, especially on large clusters - and is unlikely to produce log messages of value to debugging. Probably best to use `V(5)` or `V(6)`

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+			if err := checkVolumeAttachments(ctx, clientset, nodeName, allAttachmentsDeleted); err != nil {
```
ConnorJC3, line 71: This costs us a k8s API call every time it runs. It feels like we can be smarter about this - can we at least confirm the deleted volume attachment object references this node before spending an API call?

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to add event handler to VolumeAttachment informer: %w", err)
+	}
+
+	go informer.Run(stopCh)
+
+	if err := checkVolumeAttachments(ctx, clientset, nodeName, allAttachmentsDeleted); err != nil {
+		return err
```
ConnorJC3, line 85: This could be a transient failure, and doesn't match the behavior of any other failure of this function. We should log the error and continue on.

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+
+func checkVolumeAttachments(ctx context.Context, clientset kubernetes.Interface, nodeName string, allAttachmentsDeleted chan struct{}) error {
+	allAttachments, err := clientset.StorageV1().VolumeAttachments().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("checkVolumeAttachments: failed to list VolumeAttachments: %w", err)
+	}
+
+	for _, attachment := range allAttachments.Items {
+		if attachment.Spec.NodeName == nodeName {
+			klog.InfoS("isVolumeAttachmentEmpty: found VolumeAttachment", "attachment", attachment, "node", nodeName)
```
ConnorJC3, line 106: Suggest changing to something like `found VolumeAttachment` to something like `not ready to exit, found VolumeAttachment`. A user viewing the logs should clearly understand that this message indicates the pre-stop hook has decided it needs to wait more.

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+				return
+			}
+		},
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to add event handler to VolumeAttachment informer: %w", err)
+	}
+
+	go informer.Run(stopCh)
```
ConnorJC3, line 82: I think you should be able to use the same `allAttachmentsDeleted` channel here (as we close it on finish), thus removing the need for the `stopCh`.

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+func PreStop(clientset kubernetes.Interface, timeout time.Duration) error {
```
AndrewSirenko, line 17: Could we add a doc or file comment about what the point of this hook is, and what the happy & bad cases look like? You described it well in the PR description.

What I would like to see:
- 1-2 sentence description of the 6 min issue (like the one you have in PR description)
- Happy case: This hook attempts to force the node to wait until all volumeAttachments are detached
- Unhappy case: Context times out, customer forced to wait for 6 min force detach.

AndrewSirenko (changes requested):

cmd/options.go
```diff
@@ -95,6 +98,18 @@ func GetOptions(fs *flag.FlagSet) *Options {
 			nodeOptions.AddFlags(fs)
 			args = os.Args[1:]
 
+		case cmd == "pre-stop-hook":
+			clientset, clientErr := cloud.DefaultKubernetesAPIClient()
```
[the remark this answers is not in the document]
torredil: There is an error variable in this context to confuse it with: 
```
cmd/options.go:102:15: shadow: declaration of "err" shadows declaration at line 70 (govet)
clientset, err := cloud.DefaultKubernetesAPIClient()
```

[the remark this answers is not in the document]
ConnorJC3: Ah this is some go nonsense, just leave it as is not worth fixing.

cmd/hooks/prestop.go
torredil: Added relevant context as a comment.

cmd/hooks/prestop.go
torredil: Done.

cmd/hooks/prestop.go
torredil: Done.

cmd/hooks/prestop.go
torredil: Done.

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,113 @@
...
+
+	factory := informers.NewSharedInformerFactory(clientset, 0)
+	informer := factory.Storage().V1().VolumeAttachments().Informer()
+
+	_, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			klog.V(4).InfoS("DeleteFunc: VolumeAttachment deleted", "node", nodeName)
+			if err := checkVolumeAttachments(ctx, clientset, nodeName, allAttachmentsDeleted); err != nil {
+				klog.ErrorS(err, "DeleteFunc: error checking VolumeAttachments")
+				return
```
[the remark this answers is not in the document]
torredil: Done.

cmd/hooks/prestop.go
torredil: Good call, added the extra check.

cmd/hooks/prestop.go
torredil: Good call, changed to `V(5)` in newest rev: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use

cmd/hooks/prestop.go
torredil: > I wonder if it's worth creating a "k8s utils" style package that does this kind of common operation

This is an excellent idea 👍 

> Suggest renaming this to fetchNode

Done.

cmd/options.go
```diff
@@ -95,6 +98,18 @@ func GetOptions(fs *flag.FlagSet) *Options {
...
+			if clientErr != nil {
+				klog.ErrorS(err, "unable to communicate with k8s API")
+			} else {
+				err = hooks.PreStop(clientset, 1*time.Minute)
+				if err != nil {
+					klog.ErrorS(err, "failed to execute PreStop lifecycle hook")
```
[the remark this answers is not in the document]
torredil: Done.

cmd/options.go
torredil: Done.

torredil:
Pending rebase

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,126 @@
...
+	allAttachmentsDeleted := make(chan struct{})
+
+	factory := informers.NewSharedInformerFactory(clientset, 0)
+	informer := factory.Storage().V1().VolumeAttachments().Informer()
+
+	_, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			klog.V(5).InfoS("DeleteFunc: VolumeAttachment deleted", "node", nodeName)
+
+			if obj, ok := obj.(*storagev1.VolumeAttachment); ok && obj.Spec.NodeName == nodeName {
```
ConnorJC3, line 83: Potentially dumb question: does the parameter to this function have to be `interface{}`? Can we just change it there rather than casting here?

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,126 @@
...
+When a node is terminated, persistent workflows using EBS volumes can take 6+ minutes to start up again.
+This happens when a volume is not cleanly unmounted, which causes the Attach/Detach controller (in kube-controller-manager)
+to wait for 6 minutes before issuing a force detach and allowing the volume to be attached to another node.
+
+This PreStop lifecycle hook aims to ensure that before the node (and the CSI driver node pod running on it) is shut down,
+all VolumeAttachment objects associated with that node are removed, thereby indicating that all volumes have been successfully unmounted and detached.
+
+No unnecessary delay is added to the termination workflow, as the PreStop hook logic is only executed when the node is being drained
+(thus preventing delays in termination where the node pod is killed due to a rolling restart, or during driver upgrades, but the workload pods are expected to be running).
+If the PreStop hook hangs during its execution, the driver node pod will be forcefully terminated after 30 seconds (the default terminationGracePeriodSeconds for a pod).
```
ConnorJC3, line 28: nitpick: 30 seconds should actually be 32 seconds in this comment

see https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination:
> If the preStop hook is still running after the grace period expires, the kubelet requests a small, one-off grace period extension of 2 seconds.

cmd/hooks/prestop.go
```diff
@@ -0,0 +1,126 @@
...
+*/
+
+func PreStop(clientset kubernetes.Interface, timeout time.Duration) error {
+	klog.InfoS("PreStop: executing PreStop lifecycle hook")
```
ConnorJC3, line 32: Suggestion: add `timeout` to this log message

ConnorJC3 (approved):
some minor comments but largely lgtm

cmd/hooks/prestop.go
AndrewSirenko: For posterity in case my original comment was misunderstood:

If I were not in the meetings that discussed the 6+ minute delays, I would not have the context to understand why this prestop hook should be here. Therefore we should provide a bit of context as a comment, especially if it is added in a clean way as was done in the edit to this PR. 

@torredil the comment you added meets my standards thank you!

cmd/hooks/prestop.go
torredil: Implemented this way as it seems to be a pretty well established pattern, but I refactored this to improve readability 👍

cmd/hooks/prestop.go
torredil: Done.

PR #1741, opened 2023-09-13
Do not call ModifyVolume if the volume is already in the desired state

ConnorJC3:

Bug fix

Today, the EBS CSI Driver will call `ModifyVolume` in cases when the volume is already in the desired state (which will trigger a modification in EBS). This PR changes the behavior so that the driver instantly reports a success in the case that the volume already matches the desired state.

**What testing is done?** 

Added unit tests, Manual

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -479,18 +479,14 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
 	}
 
 	newSizeGiB := util.RoundUpGiB(newSizeBytes)
-	var oldSizeGiB int64
-	if newSizeBytes != 0 {
-		volumeSize, err := c.validateVolumeSize(ctx, volumeID, newSizeGiB)
-		if err != nil || volumeSize != 0 {
-			return volumeSize, err
-		}
+	needsModification, volumeSize, err := c.validateModifyVolume(ctx, volumeID, newSizeGiB, options)
+	if err != nil || !needsModification {
+		return volumeSize, err
 	}
 
 	req := &ec2.ModifyVolumeInput{
 		VolumeId: aws.String(volumeID),
 	}
-	// Only set req.size for resizing volume
 	if newSizeBytes != 0 {
 		req.Size = aws.Int64(newSizeGiB)
 	}
@@ -500,7 +496,7 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
 	if options.VolumeType != "" {
 		req.VolumeType = aws.String(options.VolumeType)
 	}
-	if options.VolumeType == VolumeTypeGP3 {
+	if options.Throughput != 0 {
 		req.Throughput = aws.Int64(int64(options.Throughput))
 	}
 
@@ -509,26 +505,17 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
 		return 0, fmt.Errorf("unable to modify AWS volume %q: %w", volumeID, err)
 	}
 
-	mod := response.VolumeModification
-	state := aws.StringValue(mod.ModificationState)
-
-	if volumeModificationDone(state) {
-		if newSizeBytes != 0 {
-			return c.checkDesiredSize(ctx, volumeID, newSizeGiB)
-		} else {
-			return 0, nil
-		}
-	}
-
-	err = c.waitForVolumeSize(ctx, volumeID)
-	if newSizeBytes != 0 {
+	// If the volume modification isn't immediately completed, wait for it to finish
+	state := aws.StringValue(response.VolumeModification.ModificationState)
+	if !volumeModificationDone(state) {
+		err = c.waitForVolumeModification(ctx, volumeID)
 		if err != nil {
-			return oldSizeGiB, err
+			return 0, err
 		}
-		return c.checkDesiredSize(ctx, volumeID, newSizeGiB)
-	} else {
-		return 0, c.waitForVolumeSize(ctx, volumeID)
 	}
+
+	// Perform one final check on the volume
+	return c.checkDesiredState(ctx, volumeID, newSizeGiB, options)
 }
 
 func (c *cloud) DeleteDisk(ctx context.Context, volumeID string) (bool, error) {
@@ -1170,7 +1157,7 @@ func isAWSErrorIdempotentParameterMismatch(err error) bool {
 // Checks for desired size on volume by also verifying volume size by describing volume.
 // This is to get around potential eventual consistency problems with describing volume modifications
 // objects and ensuring that we read two different objects to verify volume state.
-func (c *cloud) checkDesiredSize(ctx context.Context, volumeID string, newSizeGiB int64) (int64, error) {
+func (c *cloud) checkDesiredState(ctx context.Context, volumeID string, desiredSizeGiB int64, options *ModifyDiskOptions) (int64, error) {
 	request := &ec2.DescribeVolumesInput{
 		VolumeIds: []*string{
 			aws.String(volumeID),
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -509,26 +505,17 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
-			return oldSizeGiB, err
+			return 0, err
 		}
-		return c.checkDesiredSize(ctx, volumeID, newSizeGiB)
-	} else {
-		return 0, c.waitForVolumeSize(ctx, volumeID)
 	}
+
+	// Perform one final check on the volume
+	return c.checkDesiredState(ctx, volumeID, newSizeGiB, options)
```
AndrewSirenko, line 518: Dumb Q: Is the extra trip to EC2 API worth it? What is being checked here that wasn't checked earlier?

ConnorJC3: If you look at the `checkDesiredState` function, there is a comment above it explaining the potential eventual consistency concerns the previous maintainers implemented it for.

Even if it really isn't necessary, I don't think this PR is the appropriate place to be tearing out such a check.

pkg/cloud/cloud.go
```diff
@@ -1271,49 +1261,71 @@ func (c *cloud) AvailabilityZones(ctx context.Context) (map[string]struct{}, err
...
 
-	// if there was an error fetching volume modifications and it was anything other than VolumeNotBeingModified error
-	// that means we have an API problem.
-	if modFetchError != nil && !errors.Is(modFetchError, VolumeNotBeingModified) {
-		return oldSizeGiB, fmt.Errorf("error fetching volume modifications for %q: %w", volumeID, modFetchError)
+	// At this point, we know we are starting a new volume modification
+	// If we're asked to modify a volume to its current state, ignore the request and immediately return a success
+	needsModification := false
+
+	if newSizeGiB != 0 && oldSizeGiB < newSizeGiB {
```
torredil, line 1313: Can this be simplified to `if newSizeGiB > oldSizeGiB` ? if that statement is true, then `newSizeGiB` should always be greater than 0.

pkg/cloud/cloud.go
```diff
@@ -1271,49 +1269,71 @@ func (c *cloud) AvailabilityZones(ctx context.Context) (map[string]struct{}, err
...
-	if oldSizeGiB >= newSizeGiB {
-		klog.V(5).InfoS("[Debug] Volume", "volumeID", volumeID, "oldSizeGiB", oldSizeGiB, "newSizeGiB", newSizeGiB)
-		err = c.waitForVolumeSize(ctx, volumeID)
-		if err != nil && !errors.Is(err, VolumeNotBeingModified) {
-			return oldSizeGiB, err
+	if !needsModification {
+		klog.V(5).InfoS("[Debug] Skipping modification for volume due to matching stats", "volumeID", volumeID)
+		// Wait for any existing modifications to prevent race conditions where DescribeVolume(s) returns the new
+		// state before the volume is actually finished modifying
+		err = c.waitForVolumeModification(ctx, volumeID)
```
torredil, line 1330: 👍

pkg/cloud/cloud.go
```diff
@@ -1271,49 +1269,71 @@ func (c *cloud) AvailabilityZones(ctx context.Context) (map[string]struct{}, err
...
+			return true, 0, fmt.Errorf("volume %q in OPTIMIZING state, cannot currently modify", volumeID)
 		}
 	}
 
-	// if there was an error fetching volume modifications and it was anything other than VolumeNotBeingModified error
-	// that means we have an API problem.
-	if modFetchError != nil && !errors.Is(modFetchError, VolumeNotBeingModified) {
-		return oldSizeGiB, fmt.Errorf("error fetching volume modifications for %q: %w", volumeID, modFetchError)
+	// At this point, we know we are starting a new volume modification
+	// If we're asked to modify a volume to its current state, ignore the request and immediately return a success
```
AndrewSirenko, line 1308: Nit / Possibly wrong question: Can we refactor this into a helper function? Maybe named something like  "modifyVolumeNoOpCheck"

AndrewSirenko (changes requested):

pkg/cloud/cloud.go
```diff
@@ -1182,15 +1169,23 @@ func (c *cloud) checkDesiredSize(ctx context.Context, volumeID string, newSizeGi
 	}
 
 	// AWS resizes in chunks of GiB (not GB)
-	oldSizeGiB := aws.Int64Value(volume.Size)
-	if oldSizeGiB >= newSizeGiB {
-		return oldSizeGiB, nil
+	realSizeGiB := aws.Int64Value(volume.Size)
+
+	if realSizeGiB < desiredSizeGiB {
```
AndrewSirenko, line 1174: Nit: This chunk was a bit hard to read on first glance, could you add a comment or refactor this into a helper function? 

Something Like "modifyVolumeIdempotencyCheck"

AndrewSirenko:
Great extra unit tests cases, thank you @ConnorJC3!

pkg/cloud/cloud.go
ConnorJC3: Isn't that already what this function is? I don't think making a function that will never be reused is a good idea, I'll go with your second suggestion and add a comment unless there are other objections.

pkg/cloud/cloud.go
ConnorJC3: We _could_ but it would be introducing a function that will almost certainly never be called more than in a single place, which feels weird to me.

pkg/cloud/cloud.go
```diff
@@ -1271,49 +1269,71 @@ func (c *cloud) AvailabilityZones(ctx context.Context) (map[string]struct{}, err
...
-	// if there was an error fetching volume modifications and it was anything other than VolumeNotBeingModified error
-	// that means we have an API problem.
-	if modFetchError != nil && !errors.Is(modFetchError, VolumeNotBeingModified) {
-		return oldSizeGiB, fmt.Errorf("error fetching volume modifications for %q: %w", volumeID, modFetchError)
+	// At this point, we know we are starting a new volume modification
+	// If we're asked to modify a volume to its current state, ignore the request and immediately return a success
+	needsModification := false
+
+	if newSizeGiB != 0 && oldSizeGiB < newSizeGiB {
+		needsModification = true
```
torredil, line 1314: Suggest  moving this logic to a separate function instead: 
`func needsVolumeModification(oldSizeGiB int64, volume *VolumeType, options ModificationOptions) bool {}`

pkg/cloud/cloud.go
AndrewSirenko: Gotcha, my personal preference has been to abstract away multi-step checks like this or other hard to read blocks of code into its own function to increase readability to those unfamiliar with code base. But it's just a preference and I see your point of it being only called once. Thank you!

pkg/cloud/cloud.go
```diff
@@ -500,7 +496,7 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
 	if options.VolumeType != "" {
 		req.VolumeType = aws.String(options.VolumeType)
 	}
-	if options.VolumeType == VolumeTypeGP3 {
+	if options.Throughput != 0 {
```
torredil, line 499: What happens if you attempt to modify throughput on volume types that don't support it?

torredil (commented):
Some questions and a few minor things but otherwise lgtm, thanks.

pkg/cloud/cloud.go
ConnorJC3: Yes, fixed

pkg/cloud/cloud.go
ConnorJC3: Fixed

For the record, I think this makes the code more clunky and it's weird to add a function that will only ever be called once

pkg/cloud/cloud.go
ConnorJC3: AWS API returns an error -> we return that to the caller

pkg/cloud/cloud.go
ConnorJC3: Eddie made the same request below, so this is fixed, resolving

PR #1742, opened 2023-09-15
Add volume limits for m7a, c7i, r7a, r7iz instance families

torredil:
> **Is this a bug fix or adding new feature?**

Updating hardcoded values.

> **What is this PR about? / Why do we need it?**

This PR adds volume limits for `m7a`, `c7i`, `r7a`, `and r7iz` instance families. See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#nitro-system-limits.

> **What testing is done?** 

`make test`

### DIFF
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -14,6 +14,26 @@ const (
 	nitroMaxAttachments                  = 28
 )
 
+func init() {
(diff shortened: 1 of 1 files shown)


AndrewSirenko (commented):
/lgtm

PR #1743, opened 2023-09-15
Add dependancy upgrades for driver v1.23

AndrewSirenko:
bug fix

**What testing is done?** 

make test, make verify. Dependency upgrade runbook. 

**NOTE:** Does not upgrade opentelemetry related dependencies due to breaking change.


torredil (approved):
/approve

PR #1751, opened 2023-09-21
Update default API version for PodDisruptionBudget

### LINKED ISSUE #1750: Update default API version for PDB
The policy/v1beta1 PodDisruptionBudget was Deprecated in v1.21 (released: 2021-04-08 End of Life: 2022-04-28), and became unavailable in v1.25 (released: 2023-08-23 End of Life: 2023-10-28). The default for the PDB is policy/v1beta1, which can be an issue while using helm template or other non-api related templating. 

This change will maintain backward compatibility while applying to the kubernetes api.

/feature

Solution:
swap the If and else in these lines https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/6d4ff98b4d9a14594ed5fde9ab9fdb39f25b99f4/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml#L1-L5 to look like 

```
{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1") }}
apiVersion: policy/v1
{{- else }}
apiVersion: policy/v1beta1
{{- end }}
...
```

Alternatives: remove the if/else and go with the policy/v1 

I came across this issue by templating out this chart before putting them into my CD system for the yaml to be applied.

AndrewSirenko:
bug fix

Close #1750 

Removes deprecated PodDisruptionBudget API version policy/v1beta1. 

**What testing is done?** 
Install default helm chart to test cluster, verifying the chart renders and uses policy/v1. 

```
❯ kubectl api-resources | grep -i "disruption"
poddisruptionbudgets              pdb          policy/v1                              true         PodDisruptionBudget
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
@@ -1,8 +1,4 @@
-{{- if not (.Capabilities.APIVersions.Has "policy/v1") }}
-apiVersion: policy/v1beta1
-{{- else }}
 apiVersion: policy/v1
-{{- end }}
 kind: PodDisruptionBudget
 metadata:
   name: ebs-csi-controller


torredil (approved):
/approve

PR #1755, opened 2023-09-22
Update static-provisioning example README.md

### LINKED ISSUE #1733: Add line in README.md for static provisioning ebs
Hello!
In your documentation https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/examples/kubernetes/static-provisioning/README.md need to add line with storageClassName for pv same as in pvc
because if you dont describe it you will have an error that claim can't attach to pv
And the error will be received: "Cannot bind to requested volume "name-pv": storageClassName does not match"
Thank you

AndrewSirenko:
bug fix

Small documentation update for the static-provisioning example. Closes #1733.

**What testing is done?** 
N/A


torredil (approved):
/approve

PR #1758, opened 2023-09-26
Fix go.mod indirect versions

AndrewSirenko:
bug fix

Certain indirect dependencies were v0.0.0 in go.mod which caused some issues when setting up the repository locally. 

**What testing is done?** 
`make verify`


torredil (approved):
Thanks @AndrewSirenko /lgtm.

For context:`k8s.io/endpointslice` was added to [staging](https://github.com/kubernetes/kubernetes/tree/master/staging) in 1.28: https://github.com/kubernetes/enhancements/issues/3685.

PR #1762, opened 2023-09-28
Add Github actions code coverage CI job

ConnorJC3:

"New feature" for CI

Adds a GitHub Actions CI job to diff code coverage and comment the difference on the PR.

**What testing is done?** 

Note: although the PR jobs will run on this PR, the comment job must be committed to `master` before it can run. See example runs below:

Example with decrease: https://github.com/ConnorJC3/aws-ebs-csi-driver/pull/2
Example with no change: https://github.com/ConnorJC3/aws-ebs-csi-driver/pull/3
Example with increase: https://github.com/ConnorJC3/aws-ebs-csi-driver/pull/5
Example with external contributor: https://github.com/ConnorJC3/aws-ebs-csi-driver/pull/6

### DIFF
--- a/.github/workflows/generate-code-coverage.yaml
+++ b/.github/workflows/generate-code-coverage.yaml
@@ -1,0 +1,51 @@
+name: Generate Code Coverage
+on: [pull_request]
+  
+jobs:
+  cover-base:
+    name: Generate Base Coverage
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout base
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+
+      - name: Set up go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '^1.20.2'
+
+      - name: Generate report
+        run: |
+          go test -coverprofile base-coverage.tmp ./cmd/... ./pkg/...
+          cat base-coverage.tmp | grep -v "mock_" > base-coverage.out
+
+      - name: Upload report
+        uses: actions/upload-artifact@v3
+        with:
+          name: base-coverage
+          path: base-coverage.out
+
+  cover-pr:
+    name: Generate PR Coverage
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v3
+
+      - name: Set up go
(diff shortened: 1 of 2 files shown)


.github/workflows/generate-code-coverage.yaml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - name: Checkout base
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+
+      - name: Set up go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '^1.20.2'
```
AndrewSirenko, line 17: Maybe overkill, but can we avoid hardcoding with an env var? https://stackoverflow.com/questions/74021760/how-to-use-environment-variables-on-github-actions-without-hard-coding-them-is

.github/workflows/generate-code-coverage.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Generate report
+        run: |
+          go test -coverprofile base-coverage.tmp ./cmd/... ./pkg/...
```
AndrewSirenko, line 21: Any chance we can get the `-html=coverage.out` as an artifact and link the file in the comment?

.github/workflows/generate-code-coverage.yaml
ConnorJC3: I'm not sure I understand what you're suggesting? Can you clarify?

.github/workflows/generate-code-coverage.yaml
ConnorJC3: I don't believe so, as we would have to host it somewhere. As far as I can tell GitHub does not host artifacts for you.

I do have an idea to show the code coverage output on the prow test runs, but didn't think it was appropriate for this PR (and might not be possible? needs more investigation)

.github/workflows/generate-code-coverage.yaml
torredil: We currently hardcode the go version for other workflows as well but its a good idea to grab it from go mod using `go-version-file` : https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file.

PR #1767, opened 2023-09-29
Metric Instrumentation Framework

torredil:

This PR implements a framework that can be utilized to instrument driver internals and AWS API calls. Collected metrics are exposed via a Prometheus endpoint.

**Changes:**

1. **Metrics package**: For clearer separation of concerns and encapsulating metric-related functionality, metrics collection has been moved from `cloud` to a dedicated `metrics` package.
   
2. **Singleton Metric Recorder**: A singleton `metricRecorder` has been implemented to provide an organized way of recording and managing metrics across the driver. The recorder takes care of registering the metrics only once throughout the driver's lifetime and provides exported methods to record operation durations, errors, throttle events, and values. 

    The singleton is initialized lazily when `Recorder()` is called for the first time, (which would be during driver startup if metrics are enabled via the `enableMetrics` parameter in our Helm chart):
    ```
    if options.ServerOptions.HttpEndpoint != "" {
		r := metrics.Recorder()
		r.InitializeHttpHandler(options.ServerOptions.HttpEndpoint, "/metrics")
	}
    ```
     Any subsequent calls to `Recorder()` simply return the already initialized instance.
     
     As an alternative solution, DI was also considered but ultimately having to explictly pass the metric recorder instsance everywhere needed limits the scope of what can be instrumented. We could technically inject the recorder into the controller and node service but -- as a theoretical example -- if future maintainers / users were to introduce a batching framework or interested in instrumenting deeply integrated code in the driver, they'd run into a wall. The singleton approach allows for simply importing the metrics package wherever needed and gives a consistent interface for metric recording throughout the driver. Also removes the possibility of registering metrics multiple times and other discrepancies.
     
3. **HTTP Handler Initialization**: metrics package now takes care of setting up the HTTP server for metrics via `InitializeHttpHandler`. This abstracts the process of starting the metric server away from main and lets us remove the cloud import there as well.

---

**What testing is done?** 
- `make test`
- Manual testing:

```
$ kubectl logs ebs-csi-controller-79f798ffbc-kxx6d -n kube-system

I0929 14:58:45.071800       1 driver.go:77] "Driver Information" Driver="ebs.csi.aws.com" Version="v1.23.0"
I0929 14:58:45.071968       1 metadata.go:85] "retrieving instance data from ec2 metadata"
I0929 14:58:45.072450       1 metrics.go:134] "Metric server listening" address="0.0.0.0:3301" path="/metrics"
```

```
$ helm upgrade --install aws-ebs-csi-driver --namespace kube-system ./charts/aws-ebs-csi-driver --values ./charts/aws-ebs-csi-driver/values.yaml --set controller.enableMetrics=true

$ kubectl port-forward ebs-csi-controller-79f798ffbc-kxx6d 3301:3301 -n kube-system &

$ kubectl apply -f examples/kubernetes/dynamic-provisioning/manifests

$ curl 127.0.0.1:3301/metrics

$ # HELP cloudprovider_aws_api_request_duration_seconds [ALPHA] Latency of AWS API calls
# TYPE cloudprovider_aws_api_request_duration_seconds histogram
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.005"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.01"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.025"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.05"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.1"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.25"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.5"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="1"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="2.5"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="5"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="10"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="+Inf"} 1
cloudprovider_aws_api_request_duration_seconds_sum{request="AttachVolume"} 0.38017901
cloudprovider_aws_api_request_duration_seconds_count{request="AttachVolume"} 1
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.005"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.01"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.025"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.05"} 0
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.1"} 2
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.25"} 3
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="0.5"} 3
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="1"} 3
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="2.5"} 3
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="5"} 3
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="10"} 3
cloudprovider_aws_api_request_duration_seconds_bucket{request="DescribeInstances",le="+Inf"} 3
cloudprovider_aws_api_request_duration_seconds_sum{request="DescribeInstances"} 0.270916661
cloudprovider_aws_api_request_duration_seconds_count{request="DescribeInstances"} 3
...
```


AndrewSirenko:
Can we add a PR Description? 

Maybe something like: 

Feature

Today, the EBS CSI Driver exposes AWS API metrics via a [Prometheus](https://prometheus.io/) endpoint. 

However, currently we only expose AWS API Call latency, not driver operation latency numbers. 

This PR will let the driver expose metrics for CSI RPC Call Latency (i.e. time between initiation of the ControllerPublishVolume RPC call (which triggers the attachment) to the moment the volume's "attached" state is confirmed). 

Furthermore, this PR decouples the metrics implementation from the cloud package, instead instantiating a metricsRecorder singleton for the life of the driver. 

What testing is done?
??

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+		),
+		apiThrottle: metrics.NewCounterVec(
+			&metrics.CounterOpts{
+				Name:           "cloudprovider_aws_api_throttled_requests_total",
+				Help:           "AWS API throttled requests",
+				StabilityLevel: metrics.ALPHA,
+			},
+			[]string{"operation_name"},
+		),
+		internalValue: metrics.NewHistogramVec(
```
AndrewSirenko, line 60: Nit: Rename internalValue to something like csiOperationDuration?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+			&metrics.HistogramOpts{
+				Name:           "ebs_csi_operation",
```
AndrewSirenko, line 62: Nit: Rename to "ebs_csi_operation_duration_seconds" to match above AWS API one.

pkg/metrics/metrics.go
AndrewSirenko: Better question, what is internalValue? And if it's not measuring csiOperationDuration, why aren't we measuring it?

pkg/metrics/metrics.go
torredil: Went with `ebs_csi_aws_com` in latest rev for the prefix since that's our official driver name: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/8f246ed15e1aec91c4e5c72cda46f37c7533b5bf/pkg/driver/driver.go#L45

(`.` characters are illegal hence the `_`).

How does this look?

```
internalDurationMetric = "ebs_csi_aws_com_duration_seconds"
internalValueMetric    = "ebs_csi_aws_com_operation"
internalErrorMetric    = "ebs_csi_aws_com_errors_total"
```
Been considering `ebs_csi_aws_com_operation` -> `ebs_csi_aws_com_operation_values`.

Users will be relying on these names when scraping metrics so lets make sure we get this right before PR is merged.

pkg/metrics/metrics.go
torredil: `internalValue` here was meant to be a generic metric that could be used for capturing both duration and values in a histogram -- latest rev reworked things quite a bit and incorporates this feedback.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,208 @@
...
+	"sync"
+
+	"k8s.io/component-base/metrics"
+	"k8s.io/klog/v2"
+)
+
+const (
+	// Metric types
+	AWS metricType = iota
+	Internal
```
rdpsin, line 16: np: This enum should be in its own `const` block.

torredil: Fixed ✅

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,209 @@
...
+		klog.InfoS("Metric server listening", "address", address, "path", path)
+
+		if err := http.ListenAndServe(address, mux); err != nil {
+			klog.ErrorS(err, "Failed to start metric server", "address", address, "path", path)
+			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+		}
+	}()
+}
+
+// newRecorder returns a new instance of metricRecorder.
```
rdpsin, line 144: major np (doesn't require new PR): don't really need these comments since it's an unexported method and the method name is self explanatory. Same with other docstrings in this file.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,209 @@
...
+	if metric, ok := m.metrics[t]; ok {
+		metric.throttle.With(metrics.Labels{"operation_name": o}).Inc()
+	} else {
+		klog.Warningf("Metric type %d not found", t)
+	}
+}
+
+// Value logs a value observation for the specified metric type.
+// Parameters:
+// - t: The type of the metric.
```
rdpsin, line 110: major np (doesn't require a new PR): The Go-way of writing docstrings is to put named arguments directly in the comment. See: https://tip.golang.org/doc/comment

pkg/metrics/metrics.go
torredil: Good feedback, +1 and prefer no comments unless necessary, typically don't write so calibrating that scale.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,181 @@
...
+)
+const (
+	// AWS API metrics
+	awsDurationMetric = "cloudprovider_aws_api_request_duration_seconds"
+	awsThrottleMetric = "cloudprovider_aws_api_throttled_requests_total"
+	awsErrorMetric    = "cloudprovider_aws_api_request_errors"
+
+	// Internal driver metrics
+	internalDurationMetric = "ebs_csi_aws_com_duration_seconds"
+	internalValueMetric    = "ebs_csi_aws_com_operation"
```
ConnorJC3, line 26: So I know there was discussion about this above, but I want to bring it up again because I think it's still relevant in the latest version of the PR:

What is an "internal value metric"? I don't understand what you mean and I'm an EBS CSI Driver maintainer, imagine the user confusion.

This needs to be renamed to something that clearly indicates what it will be used for in a way that doesn't require reading the source code to find out.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,181 @@
...
+	internalErrorMetric    = "ebs_csi_aws_com_errors_total"
```
ConnorJC3, line 27: Drop the `_total` suffix, as:

1. That implies it records every error the driver encounters, which is false
2. It's inconsistent with the AWS API metric names

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,181 @@
...
+type metricType int
+
+type metricRecorder struct {
+	registry metrics.KubeRegistry
+	metrics  map[metricType]metricFields
+}
+
+type metricFields struct {
+	duration *metrics.HistogramVec
+	value    *metrics.HistogramVec
```
ConnorJC3, line 46: Probably related to the `internalValue` above, but is this meant to be a count of all occurrences? Should this value be called `count`?

ConnorJC3 (commented):
Top level comment: this current iteration feels weird to me.

It essentially enforces an "operation" style recording of metrics, and tries to stuff everything into that.

Does it really make sense to have one metric for every possible "duration" that could occur within the EBS CSI Driver?

This is just me spitballing, but I wonder if it would make sense for the caller to supply their own name of what the metric should be.

We could have a metric for `csi_rpc_duration`, for `k8s_api_duration`, for `mkfs_duration`, etc - stuffing all those into a generic `duration_seconds` bucket feels very wrong and is antiethical to standard Prometheus design.

pkg/metrics/metrics.go
torredil: > This is just me spitballing, but I wonder if it would make sense for the caller to supply their own name of what the metric should be.

Thought about this initially but after deliberation figured one metric and distinguishing via the use of labels would be a cleaner solution. Can certainly rework this and push in that front if you recommend. FWIW we won't be stuffing CSI rpc or K8s API measurements under this metric -- will be exposed via `csi_operations_seconds` instead:

```
Container Storage Interface operation duration with gRPC error code status total
Type: Histogram
Labels: driver_name, grpc_status_code, method_name, migrated
```

https://kubernetes.io/docs/reference/instrumentation/metrics/

> This needs to be renamed to something that clearly indicates what it will be used for

Suggestions?

ConnorJC3: Can we clarify what we intend to put under this metric as is? It's hard for me to make recommendations here because it's not clear to me what the `ebs_csi_aws_com_operation` metric will contain.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Recorder returns the singleton instance of metricRecorder.
+// nil is returned if the recorder is not initialized.
+func Recorder() *metricRecorder {
+	return r
+}
+
+// NewRecorder initializes a new metricRecorder instance if it hasn't been initialized.
+func NewRecorder() *metricRecorder {
```
ConnorJC3, line 28: Thoughts on renaming this to something like `InitializeRecorder`? As-is, I'm not sure the name conveys that not only are you creating a new instance of the recorder object, but that it's going to be set as the singleton.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	mux.Handle(path, metrics.HandlerFor(
+		m.registry,
+		metrics.HandlerOpts{
+			ErrorHandling: metrics.ContinueOnError,
+		}))
+
+	go func() {
+		klog.InfoS("Metric server listening", "address", address, "path", path)
+
+		if err := http.ListenAndServe(address, mux); err != nil {
```
ConnorJC3, line 89: Suggestion: harden security by adding a timeout. The default `http` package server comes with no timeout and is susceptible to DoS attacks.

```suggestion
		server := &http.Server{
				ReadTimeout: 3 * time.Second,
		}
		if err := server.ListenAndServe(address, mux); err != nil {
```

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,82 @@
+package metrics
```
ConnorJC3, line 1: Top level comment: maybe worth adding very basic tests for the `== nil` cases?

ConnorJC3 (commented):
/lgtm

but consider the suggestions below:

pkg/metrics/metrics_test.go
torredil: ✅

pkg/metrics/metrics.go
torredil: ✅

PR #1763, opened 2023-09-28
[Hold]  Add format option e2e tests

AndrewSirenko:
**Is this a bug fix or adding new feature?**
Feature

**What is this PR about? / Why do we need it?**
Incomplete PR for adding format option e2e tests. They test that a volume created with a supported custom format option successfully mount and are later resizable. 

Considering refactoring part of format_options.go into a tester file like other e2e tests. 

See individual commits for conceptual breakdown. 

For this PR I'm looking for feedback on:
- Any TODO question
- Test Readability

**What testing is done?** 
ginkgo run -v --focus='\[format-options'

PR #1769, opened 2023-10-04
Add e2e tests for format options

AndrewSirenko:
feature

Adds end to end tests for format option parameters like `blocksize` and `bytesperinode`

**What testing is done?** 
ginkgo run --focus="\[format]"


tests/e2e/format_options.go
```diff
@@ -0,0 +1,89 @@
...
+	BeforeEach(func() {
+		cs = f.ClientSet
+		ns = f.Namespace
+		ebsDriver = driver.InitEbsCSIDriver()
+	})
+
+	for _, fsType := range testedFsTypes {
+		Context(fmt.Sprintf("with an %s filesystem", fsType), func() {
+			for _, formatOptionTestCase := range formatOptionTests {
+				formatOptionTestCase := formatOptionTestCase // Go trap
```
torredil, line 71: np: Remove comment or add more context.

tests/e2e/testsuites/format_options_tester.go
```diff
@@ -0,0 +1,93 @@
...
+// FormatOptionTest will provision required StorageClass(es), PVC(s) and Pod(s) in order to test that volumes with
+// a specified custom format options will mount and are able to be later resized.
+type FormatOptionTest struct {
+	CreateVolumeParameterKey   string
+	CreateVolumeParameterValue string
+}
+
+const (
+	volumeSizeIncreaseAmtGi = 1
+	volumeMountPath         = "/mnt/test-format-option" // TODO should I keep this as mnt/test-1, and refactor to be `DefaultMountPath` globally in testsuites?
```
torredil, line 35: No need to have a custom mount path here, for consistency `mnt/test-1` and thinking about a default mount path variable which is very commonly used throughout the test suite seem like a good idea.

tests/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go
```diff
@@ -92,3 +74,28 @@ func WaitForPvToResize(c clientset.Interface, ns *v1.Namespace, pvName string, d
 	}
 	return fmt.Errorf("Gave up after waiting %v for pv %q to complete resizing", timeout, pvName)
 }
+
+// ResizeTestPvc increases size of given persistent volume claim by `sizeIncreaseGi` Gigabytes
+func ResizeTestPvc(client clientset.Interface, namespace *v1.Namespace, testPvc *TestPersistentVolumeClaim, sizeIncreaseGi int64) (updatedPvc *v1.PersistentVolumeClaim, updatedSize resource.Quantity) {
```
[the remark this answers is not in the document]
AndrewSirenko: Isn't that what this helper function does?

tests/e2e/format_options.go
```diff
@@ -0,0 +1,89 @@
...
+	)
+
+	BeforeEach(func() {
+		cs = f.ClientSet
+		ns = f.Namespace
+		ebsDriver = driver.InitEbsCSIDriver()
+	})
+
+	for _, fsType := range testedFsTypes {
+		Context(fmt.Sprintf("with an %s filesystem", fsType), func() {
```
ConnorJC3, line 69: np "with" -> "using" (on this line only, not the below context

tests/e2e/testsuites/format_options_tester.go
```diff
@@ -0,0 +1,87 @@
...
+func (t *FormatOptionTest) Run(client clientset.Interface, namespace *v1.Namespace, ebsDriver driver.PVTestDriver, fsType string) {
+	By("setting up pvc with custom format option")
+	volumeDetails := createFormatOptionVolumeDetails(fsType, t)
+	testPvc, _ := volumeDetails.SetupDynamicPersistentVolumeClaim(client, namespace, ebsDriver)
+	defer testPvc.Cleanup()
+
+	By("deploying pod with custom format option")
+	getFsInfoTestPod := createPodWithVolume(client, namespace, PodCmdWriteToVolume(DefaultMountPath), testPvc, volumeDetails)
+	defer getFsInfoTestPod.Cleanup()
+	getFsInfoTestPod.WaitForSuccess()
```
ConnorJC3, line 45: How does the `getFsInfoTestPod` actually get fs info? It doesn't even seem to be created with the necessary privileges to do that?

ConnorJC3 (changes requested):
Half the PR seems to be missing, namely the part where we check if the filesystem was actually formatted correctly

AndrewSirenko:
As discussed with @ConnorJC3 @torredil we will validate that custom formatting options were actually applied by mkfs via privileged ubuntu containers running the tune2fs command and regex-Finding their logs in a future PR once testsuite utils are refactored. 

See #1763 for that approach.

PR #1706, opened 2023-08-01
Support clustered allocation when formatting ext4 filesystem

AndrewSirenko:
**Is this a bug fix or adding new feature?**
New feature

**What is this PR about? / Why do we need it?**
Close #1687 

This PR:
- Adds `ext4BigAlloc` and `ext4ClusterSize` storage class parameters that expose enabling clustered allocation on `ext4` filesystems. 
- Allows developers to enable [torn write prevention](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/storage-twp.html) on their dynamically provisioned volumes by setting by setting `ext4BigAlloc` to `true` and 'ext4ClusterSize` to the desired cluster size. 

The `ext4BigAlloc` parameter enables clustered allocation on ext4 filesystems by enabling the bigalloc feature (ie performing `mkfs.ext4 -O bigalloc`) when formatting a dynamically-provisioned Persistent Volume (PV). See [this ext4 man page](https://man7.org/linux/man-pages/man5/ext4.5.html) for more details about bigalloc.

The `ext4ClusterSize` parameter allows users to set a custom cluster size when formatting a PV with the `ext4` filesystem IF they have set `ext4BigAlloc` to true. (appends `-C ext4ClusterSize` to other mkfs.ext4 formatting options).

One benefit to exposing this formatting option is that developers can enable [torn write prevention](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/storage-twp.html) on their volumes to improve the performance of their I/O-intensive relational database workloads and reduce latency without negatively impacting data resiliency. Relational databases that use InnoDB or XtraDB as the database engine, such as MySQL and MariaDB, will benefit from torn write prevention.

**What testing is done?** 
Unit Tests + Manual Testing on EKS Cluster. 

---

### Manual Testing
1. Create an EKS cluster, deploy the ebs-csi-driver a version of the driver with an image built from this PR
1. Test Happy Path (valid `ext4ClusterSize` parameter value) and Unhappy Path (Invalid `ext4ClusterSize` parameter value)
   
#### Happy Path: Storage class with valid `ext4ClusterSize` parameter value yields properly formatted filesystem on PV
1. Create a pod with the following manifest (Note the `ext4ClusterSize: "16384"` in the storage class definition):

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
[...]

PR #1770, opened 2023-10-04
Refactor format option validation and tests

AndrewSirenko:
bug fix

Split off parts of #1706 that refactor the format option parameter validation and controller unit tests. 

**What testing is done?** 
make test

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -21,6 +21,7 @@ package util
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


pkg/util/util_test.go
```diff
@@ -154,3 +155,28 @@ func TestGetAccessModes(t *testing.T) {
 		t.Fatalf("Wrong values returned for volume capabilities. Expected %v, got %v", expectedModes, actualModes)
 	}
 }
+
+func TestIsAlphanumeric(t *testing.T) {
+	testCases := []struct {
+		name       string
+		testString string
+		expected   bool
```
ConnorJC3, line 163: Probably should rename this to `expResult` for conformity with other tests

PR #1778, opened 2023-10-09
Update GitHub workflows

torredil:

This PR upgrades GitHub actions to latest versions, sets Go version dynamically via `go-version-file`, and removes the verify action as this test is already exercised by Prow.

**What testing is done?** 

CI


.github/workflows/trivy-containers.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
     
     steps:
       - name: Scan container image
-        uses: aquasecurity/trivy-action@master
```
AndrewSirenko, line 34: To clarify: We're hardcoding the version here to not be caught off guard by a breaking change?

AndrewSirenko (commented):
/lgtm 

Thanks for yet another automation

PR #1779, opened 2023-10-10
CI: Fix 'error: externally-managed-environment'

torredil:

Bug fix

See [PEP 668 – Marking Python base environments as “externally managed”](https://peps.python.org/pep-0668/) and https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/1706/pull-aws-ebs-csi-driver-test-helm-chart/1711789551590051840/build-log.txt for context.

**What testing is done?** 

CI


AndrewSirenko (commented):
/lgtm

PR #1789, opened 2023-10-18
Add Helm compatibility for --reuse-values upgrades

ConnorJC3:

Adds support for upgrading with `--reuse-values` with Helm from any prior version

The team decided we want to support upgrades when using `--reuse-values`. This PR adds support for that from any prior version. Unfortunately, already released versions cannot be fixed, but this will fix any upgrades to the next or any later version.

**What testing is done?** 

Testing on master (doesn't work):
```bash
# Oldest release of the EBS CSI Driver chart, thus the worst case scenario for --reuse-values
$ git checkout aws-ebs-csi-driver-0.6.2

# Note: depending on your Kubernetes version, you may need to make some changes to the Chart to bump removed API versions
$ helm install aws-ebs-csi-driver .
NAME: aws-ebs-csi-driver
LAST DEPLOYED: Wed Oct 18 18:29:05 2023
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:

    kubectl get pod -n kube-system -l "app.kubernetes.io/name=aws-ebs-csi-driver,app.kubernetes.io/instance=aws-ebs-csi-driver"
    
# Switch to latest release
$ git checkout release-1.23

# Note: useOldCSIDriver=true is required to prevent immutable change (although it won't matter here because the upgrade is broken anyways)
# It breaks!
$ helm upgrade --reuse-values --set useOldCSIDriver=true aws-ebs-csi-driver .
Error: UPGRADE FAILED: template: aws-ebs-csi-driver/templates/volumesnapshotclass.yaml:1:17: executing "aws-ebs-csi-driver/templates/volumesnapshotclass.yaml" at <.Values.sidecars.snapshotter.forceEnable>: nil pointer evaluating interface {}.forceEnable
```

Testing with PR (works):
```bash
# Oldest release of the EBS CSI Driver chart, thus the worst case scenario for --reuse-values
$ git checkout aws-ebs-csi-driver-0.6.2

# Note: depending on your Kubernetes version, you may need to make some changes to the Chart to bump removed API versions
$ helm install aws-ebs-csi-driver .
NAME: aws-ebs-csi-driver
LAST DEPLOYED: Wed Oct 18 18:31:53 2023
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:

    kubectl get pod -n kube-system -l "app.kubernetes.io/name=aws-ebs-csi-driver,app.kubernetes.io/instance=aws-ebs-csi-driver"

# Switch to PR
$ git checkout helm-reuse-values-support

# Note: useOldCSIDriver=true is required to prevent immutable change
# It works!
$ helm upgrade --reuse-values --set useOldCSIDriver=true aws-ebs-csi-driver .          
Release "aws-ebs-csi-driver" has been upgraded. Happy Helming!
NAME: aws-ebs-csi-driver
LAST DEPLOYED: Wed Oct 18 18:33:57 2023
NAMESPACE: default
STATUS: deployed
REVISION: 3
NOTES:

    kubectl get pod -n default -l "app.kubernetes.io/name=aws-ebs-csi-driver,app.kubernetes.io/instance=aws-ebs-csi-driver"

NOTE: The [CSI Snapshotter](https://github.com/kubernetes-csi/external-snapshotter) controller and CRDs will no longer be installed as part of this chart and moving forward will be a prerequisite of using the snap shotting functionality.

# All pods are running etc
$ kubectl -n kube-system get pods                              
NAME                                            READY   STATUS    RESTARTS      AGE
aws-cloud-controller-manager-dt766              1/1     Running   0             41m
aws-node-termination-handler-68d9cbd5dc-xfzcx   1/1     Running   0             41m
coredns-78c4944598-8zs9h                        1/1     Running   0             40m
coredns-78c4944598-bbhqv                        1/1     Running   0             41m
coredns-autoscaler-fcf87bf56-h5j9h              1/1     Running   0             41m
dns-controller-c848f4d4d-kvxx4                  1/1     Running   0             41m
etcd-manager-events-i-00dc16e2d485fab66         1/1     Running   0             42m
etcd-manager-main-i-00dc16e2d485fab66           1/1     Running   0             42m
kops-controller-vmfll                           1/1     Running   0             41m
kube-apiserver-i-00dc16e2d485fab66              2/2     Running   2 (43m ago)   41m
kube-controller-manager-i-00dc16e2d485fab66     1/1     Running   4 (42m ago)   42m
kube-proxy-i-008703fa7cb2d77b7                  1/1     Running   0             39m
kube-proxy-i-00dc16e2d485fab66                  1/1     Running   0             41m
kube-scheduler-i-00dc16e2d485fab66              1/1     Running   0             41m
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/_node-windows.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_node-windows.tpl
@@ -1,5 +1,6 @@
 {{- define "node-windows" }}
-{{- if .Values.node.enableWindows }}
+{{- if (.Values.node).enableWindows }}
+---
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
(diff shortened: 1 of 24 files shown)


charts/aws-ebs-csi-driver/templates/_node-windows.tpl
```diff
@@ -1,5 +1,6 @@
 {{- define "node-windows" }}
-{{- if .Values.node.enableWindows }}
+{{- if (.Values.node).enableWindows }}
+---
```
AndrewSirenko, line 3: This `---` is just for styling yes?

charts/aws-ebs-csi-driver/templates/_node-windows.tpl
```diff
@@ -117,23 +148,27 @@ spec:
               exec:
                 command: ["/bin/aws-ebs-csi-driver", "pre-stop-hook"]
         - name: node-driver-registrar
-          image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.sidecars.nodeDriverRegistrar.image.repository .Values.sidecars.nodeDriverRegistrar.image.tag }}
-          imagePullPolicy: {{ default .Values.image.pullPolicy .Values.sidecars.nodeDriverRegistrar.image.pullPolicy }}
+          image: {{ printf "%s%s:%s" (default "" (.Values.image).containerRegistry) (default "public.ecr.aws/eks-distro/kubernetes-csi/node-driver-registrar" (((.Values.sidecars).nodeDriverRegistrar).image).repository) (default "v2.9.0-eks-1-28-6" (((.Values.sidecars).nodeDriverRegistrar).image).tag) }}
```
AndrewSirenko, line 151: I'm against putting default sidecar tags in the chart that we would have to risk changing every release.

deploy/kubernetes/base/serviceaccount-csi-controller.yaml
```diff
@@ -6,7 +6,4 @@ metadata:
   name: ebs-csi-controller-sa
   labels:
     app.kubernetes.io/name: aws-ebs-csi-driver
-  #Enable if EKS IAM roles for service accounts (IRSA) is used. See https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html for details.
```
AndrewSirenko, line 9: Why are we removing this comment? Is it no longer relevant?

AndrewSirenko (changes requested):

charts/aws-ebs-csi-driver/templates/_node-windows.tpl
AndrewSirenko: It seems like this default value is not relevant to the chart tester CI job failing: [Additional DaemonSets feature #1722](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1722)

ConnorJC3: I dislike it, but all default values (including tags) must be duplicated into the chart for this support to work. Otherwise, users upgrading with `--reuse-values` will end up with an inconsistent (or in the case of tags, broken) experience.

charts/aws-ebs-csi-driver/templates/_node-windows.tpl
ConnorJC3: No, it's for separating multiple YAML documents in the same file. It used to be inside `node.yaml` and `node-windows.yaml` but I moved it here so it's only rendered if the relevant `DaemonSet` is enabled, otherwise extra `---`s will be rendered when the `DaemonSet` is disabled.

deploy/kubernetes/base/serviceaccount-csi-controller.yaml
ConnorJC3: They aren't critical to this change, so if you feel strongly I can put them back in, but I took the opportunity to remove all remaining `if eq .Release.Name "kustomize"` invocations in the chart as they were directly next to what I was already working on.

This particular comment is bad because:
1. It encourages manually editing the manifests, which is bad practice and we should not be encouraging (instead, users should use the Helm chart or Kustomize to make changes)
2. It creates a small additional burden on maintainers that they must understand the special properties of the `kustomize` release name
3. Explaining how IAM permissions interact with the driver is better left to `install.md`, where we document the various methods of granting IAM permissions, what policies the role needs, the existence of the managed policy, and other crucial information this short comment does not contain

ConnorJC3:
/close

After further evaluation, the team decided to deprecate `--reuse-values` in the helm chart in the future instead. Closing this PR out.

PR #1805, opened 2023-10-25
Add compatibility workaround for A1 instance family

ConnorJC3:

Bug fix

Adds compatibility for running on instances with the a1 instance family

**What testing is done?** 

```
$ helm install aws-ebs-csi-driver . --set a1CompatibilityDaemonSet=true --set image.repository=368597081700.dkr.ecr.us-west-2.amazonaws.com/aws-ebs-csi-driver --set image.tag=fdb98666fa056862748ca337049f907d7b080e25
NAME: aws-ebs-csi-driver
LAST DEPLOYED: Wed Oct 25 15:15:29 2023
NAMESPACE: default
STATUS: deployed
REVISION: 1
NOTES:

    kubectl get pod -n default -l "app.kubernetes.io/name=aws-ebs-csi-driver,app.kubernetes.io/instance=aws-ebs-csi-driver"

NOTE: The [CSI Snapshotter](https://github.com/kubernetes-csi/external-snapshotter) controller and CRDs will no longer be installed as part of this chart and moving forward will be a prerequisite of using the snap shotting functionality.

$ kubectl get daemonsets.apps                   
NAME                    DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
ebs-csi-node            0         0         0       0            0           kubernetes.io/os=linux   3m30s
ebs-csi-node-a1compat   1         1         1       1            1           kubernetes.io/os=linux   3m30s
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/node.yaml
+++ b/charts/aws-ebs-csi-driver/templates/node.yaml
@@ -9,6 +9,38 @@
     "node" (deepCopy $.Values.node | mustMerge $values)
   )
 }}
----
+{{- include "node" (deepCopy $ | mustMerge $args) -}}
+{{- end }}
+{{- if .Values.a1CompatibilityDaemonSet }}
+{{$args := dict
+  "NodeName" "ebs-csi-node-a1compat"
+  "Values" (dict
+    "image" (dict
+      "tag" (printf "%s-a1compat" (default (printf "v%s" .Chart.AppVersion) (.Values.image.tag | toString)))
+    )
+    "node" (dict
+      "affinity" (dict
+        "nodeAffinity" (dict
+          "requiredDuringSchedulingIgnoredDuringExecution" (dict
+            "nodeSelectorTerms" (list
+              (dict "matchExpressions" (list
+                (dict
+                  "key" "eks.amazonaws.com/compute-type"
+                  "operator" "NotIn"
+                  "values" (list "fargate")
+                )
+                (dict
+                  "key" "node.kubernetes.io/instance-type"
+                  "operator" "In"
+                  "values" (list "a1.medium" "a1.large" "a1.xlarge" "a1.2xlarge" "a1.4xlarge")
+                )
+              ))
+            )
+          )
+        )
+      )
+    )
+  )
+}}
 {{- include "node" (deepCopy $ | mustMerge $args) -}}
 {{- end }}
--- a/hack/e2e/ecr.sh
+++ b/hack/e2e/ecr.sh
@@ -27,8 +27,8 @@ function ecr_build_and_push() {
       docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
       make all-push
     else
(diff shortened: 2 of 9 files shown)


AndrewSirenko:
/lgtm

Excited for optimize makefile & run.sh task to clean up :spaghetti: though...

charts/aws-ebs-csi-driver/templates/node.yaml
```diff
@@ -9,6 +9,38 @@
...
+    "node" (dict
+      "affinity" (dict
+        "nodeAffinity" (dict
+          "requiredDuringSchedulingIgnoredDuringExecution" (dict
+            "nodeSelectorTerms" (list
+              (dict "matchExpressions" (list
+                (dict
+                  "key" "eks.amazonaws.com/compute-type"
+                  "operator" "NotIn"
+                  "values" (list "fargate")
```
wmesard, line 30: Why is this Fargate-specific check necessary?

hack/e2e/ecr.sh
```diff
@@ -27,8 +27,8 @@ function ecr_build_and_push() {
       docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
       make all-push
     else
-      IMAGE=${IMAGE_NAME} TAG=${IMAGE_TAG} OS=linux ARCH=amd64 OSVERSION=amazon make image
-      docker tag "${IMAGE_NAME}":"${IMAGE_TAG}"-linux-amd64-amazon "${IMAGE_NAME}":"${IMAGE_TAG}"
+      IMAGE=${IMAGE_NAME} TAG=${IMAGE_TAG} OS=linux ARCH=amd64 OSVERSION=al2023 make image
+      docker tag "${IMAGE_NAME}":"${IMAGE_TAG}"-linux-amd64-al2023 "${IMAGE_NAME}":"${IMAGE_TAG}"
```
wmesard, line 31: (Out of scope for this PR, but it would be better if this extracted the information from the resulting image, instead of passing it in. Having it hardcoded here and in the Makefile is tech debt that will bite us eventually.)

wmesard (commented):
One question

charts/aws-ebs-csi-driver/templates/node.yaml
ConnorJC3: I'm not sure if it's absolutely necessary here, in order to make the minimal change necessary I pulled it from our defaults: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L319-L327
(so the resulting change is the default + the a1 restriction, rather than the default replaced by the a1 restriction)

Even if it's not actually necessary, it definitely won't hurt, and considering we don't control what EKS fargate does it would be bad if in some future they started vending nodes with an a1 "instance type" and we tried to run on them.

wmesard (approved):

hack/e2e/ecr.sh
AndrewSirenko: Adding to our `optimize makefile & run.sh task` backlog task.

PR #1817, opened 2023-10-30
Release aws-ebs-csi-driver v1.23.2

AndrewSirenko:
Release

Release aws-ebs-csi-driver v1.23.2

**What testing is done?** 
`make verify`


charts/aws-ebs-csi-driver/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Helm chart
+## v2.23.2
+* Bump driver version to `v1.23.2`
```
ConnorJC3, line 3: Should probably also include sidecar bumps?

PR #1819, opened 2023-10-31
Implement Generic Batching Framework + Batch DescribeVolume API requests

torredil:

New feature.

This PR implements a generic batching framework, designed to facilitate the aggregation and execution of tasks -- most notably batched API requests for `DescribeVolume` -- used by the driver to poll for volume state and determine attachment status. 

Today, the driver makes one DV API call per volume, however, DV can return up to 500 results. These changes allow for utilizing the API's capacity to handle up to 500 volumes at once, which significantly helps mitigate rate limit issues with DV  and enhance performance when using the driver under heavy load scenarios as a result.

There are two parts to this PR, separated by commits:

1. **Generic Batching Framework**.

2. **Batch DescribeVolume API Requests**.
  
**What testing is done?** 

- `make test`
- CI
- Manual testing


torredil:
Kustomize CI job needs to be updated. Driver fails to enter a running state because `--batching` is passed into the controller container for driver version 1.24 (need to deploy driver using image built for CI instead).

torredil:
Re Kustomize CI job needs to be updated -- we decided to run `pull-aws-ebs-csi-driver-external-test-kustomize` only on release branches. This is better than doing in-place modifications of the image repository / tag values. 

An additional advantage in doing the above is that we implicitly validate image promotion for new releases moving forward since we pull from the official public ECR gallery.

pkg/cloud/cloud.go
```diff
@@ -463,6 +618,7 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 			return nil, fmt.Errorf("could not attach tags to volume: %v. %w", volumeID, err)
 		}
 	}
+	klog.InfoS("volume created", "volumeID", volumeID)
```
ConnorJC3, line 621: Debug log left in?

ConnorJC3 (commented):
Needs rebase, one comment, otherwise lgtm

docs/options.md
```diff
@@ -12,3 +12,4 @@ There are a couple of driver options that can be passed as arguments when starti
 | logging-format              | json                                              | text                                                | Sets the log format. Permitted formats: text, json|
 | user-agent-extra            | csi-ebs                                           | helm                                                | Extra string appended to user agent|
 | enable-otel-tracing         | true                                              | false                                               | If set to true, the driver will enable opentelemetry tracing. Might need [additional env variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration) to export the traces to the right collector|
+| batching                    | true                                              | true                                                | If set to true, the driver will enable batching of API calls. This is especially helpful for improving performance in workloads that are sensitive to EC2 rate limits|
```
ConnorJC3, line 15: Additional comment: Can we add something like "at the cost of a small increase to worst-case latency"? Just so customers that want to hyper-optimize latency understand the impact of the option.

AndrewSirenko (approved):
/lgtm
(Minus the cloud_test merge conflict and 1 question)

docs/options.md
torredil: ✅

pkg/cloud/cloud.go
torredil: Good catch, fixed.

ConnorJC3:
/lgtm
(reapplying @AndrewSirenko's lgtm)

/approve

torredil:
For peace of mind, investigated the resizing failure in `pull-aws-ebs-csi-driver-e2e-single-az`:

```
1 driver.go:125] "GRPC error" err="rpc error: code = Internal desc = Could not resize volume \"vol-003bbe6304464240d\": rpc error: code = Internal desc = Could not modify volume \"vol-003bbe6304464240d\": volume \"vol-003bbe6304464240d\" in OPTIMIZING state, cannot currently modify"
```

In conclusion, the error observed above is unrelated to this PR and occurs when a volume takes longer than 1 minute to transition from an `optimizing` state to `completed`. See a prior example [here](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/1810/pull-aws-ebs-csi-driver-e2e-single-az/1717907738748522496). This behavior can be easily reproduced outside of the driver by simply modifying a volume via the CLI or console. 

Based on my observation, it is fairly common to see a volume remain in `optimizing 99%` for up to 5 minutes before transitioning to complete. However, the single AZ tests expect resizing to succeed within one minute:

```
Nov  7 17:54:03.198: INFO: Unexpected error: 
  <*errors.errorString | 0xc000a9b9a0>: 
    Gave up after waiting 1m0s for pv "pvc-af4167dc-d33b-4b78-a3f1-35d6e2856d2d" to complete resizing
  {
       s: "Gave up after waiting 1m0s for pv \"pvc-af4167dc-d33b-4b78-a3f1-35d6e2856d2d\" to complete resizing",
  }
```

Given that we recently introduced more tests that exercise this code path in single-az, we are more likely to observe the error moving forward if that timeout value is not relaxed.

/retest

torredil:
`ResizeOrModifyDisk` needs to be revisited. There is a discrepancy in the logic. 

According to cloud, a modification is considered done if the volume is an `optimizing` state:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L1320-L1323

In the case where `ResizeOrModifyDisk` hasn't been previously called for a volume, we simply move on to check if the volume is in the [desired state](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L523). If it is, `ResizeOrModifyDisk` succeeds even if the volume is still `optimizing`. However, if `checkDesiredState` returns an error or the sidecar times out, [latestMod](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L1287) won't be nil the second time around and we error out if the volume is `optimizing`:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L1298-L1300

PR #1820, opened 2023-10-31
bug hunting part 1

torredil:

**What testing is done?**


AndrewSirenko (approved):

AndrewSirenko:
Would appreciate a small PR description on these, such as labeling bug fix and a few words as to why. Or a more meaningful PR name.

PR #1824, opened 2023-11-01
Update sidecar timeout values

### LINKED ISSUE #1671: CSI RPC call timeouts are non-optimal
Currently we are using default timeout for most operations in helm charts and I do not think they are optimal. For example, default provisioner timeout is 10s and default attacher timeout is 15s. When last we measured these numbers, we found them to be too low for EBS driver.

torredil:
- This PR changes the default timeout value (15s) for the attacher sidecar, responsible for exercising `ControllerPublishVolume` / `ControllerUnpublishVolume`. The default value of 15s used today is not a sensible default, as a result the following error is observed frequently:

```
E1101 14:21:54.345145       1 driver.go:124] "GRPC error" err="rpc error: code = Internal desc = Could not detach volume \"vol-07ec886465a316b9b\" from node \"i-0e137107f5012fb1f\": context deadline exceeded"
```

In the vast majority of cases the volume is successfully detached just mere seconds after the attacher times out.

closes #1671

For context:
```
--timeout <duration>: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of ControllerPublish and ControllerUnpublish calls. See [CSI error and timeout handling](https://github.com/kubernetes-csi/external-attacher#csi-error-and-timeout-handling) for details. 15 seconds is used by default.
```

**What testing is done?** 
- Manual testing
- CI

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -190,6 +190,9 @@ spec:
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.sidecars.provisioner.image.repository .Values.sidecars.provisioner.image.tag }}
           imagePullPolicy: {{ default .Values.image.pullPolicy .Values.sidecars.provisioner.image.pullPolicy }}
           args:
+            {{- if not (regexMatch "(-timeout)" (join " " .Values.sidecars.provisioner.additionalArgs)) }}
+            - --timeout=60s
+            {{- end }}
             - --csi-address=$(ADDRESS)
             - --v={{ .Values.sidecars.provisioner.logLevel }}
             - --feature-gates=Topology=true
@@ -245,6 +248,9 @@ spec:
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.sidecars.attacher.image.repository .Values.sidecars.attacher.image.tag }}
           imagePullPolicy: {{ default .Values.image.pullPolicy .Values.sidecars.attacher.image.pullPolicy }}
           args:
+            {{- if not (regexMatch "(-timeout)" (join " " .Values.sidecars.attacher.additionalArgs)) }}
+            - --timeout=60s
+            {{- end }}
             - --csi-address=$(ADDRESS)
             - --v={{ .Values.sidecars.attacher.logLevel }}
             - --leader-election={{ .Values.sidecars.attacher.leaderElection.enabled | required "leader election state for csi-attacher is required, must be set to true || false." }}
@@ -339,6 +345,9 @@ spec:
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.sidecars.volumemodifier.image.repository .Values.sidecars.volumemodifier.image.tag }}
           imagePullPolicy: {{ default .Values.image.pullPolicy .Values.sidecars.volumemodifier.image.pullPolicy }}
           args:
+            {{- if not (regexMatch "(-timeout)" (join " " .Values.sidecars.volumemodifier.additionalArgs)) }}
+            - --timeout=60s
+            {{- end }}
             - --csi-address=$(ADDRESS)
             - --v={{ .Values.sidecars.volumemodifier.logLevel }}
             - --leader-election={{ .Values.sidecars.volumemodifier.leaderElection.enabled | required "leader election state for csi-volumemodifier is required, must be set to true || false." }}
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/base/controller.yaml
```diff
@@ -156,6 +156,7 @@ spec:
           imagePullPolicy: IfNotPresent
           args:
             - --csi-address=$(ADDRESS)
+            - --timeout=60s
```
AndrewSirenko, line 159: Did you measure any other values (i.e. why not 30s?), or is this PR just a quick fix for a common customer pain point.

AndrewSirenko (approved):
/lgtm 
Left non-blocking comment.

pkg/cloud/cloud.go
```diff
@@ -696,7 +696,7 @@ func (c *cloud) WaitForAttachmentState(ctx context.Context, volumeID, expectedSt
 			attachmentState = volumeDetachedState
 		}
 		if attachment != nil {
-			// AWS eventual consistency can go back in time.
+			// AWS eventual consistency can go back in TIME.
```
ConnorJC3, line 699: Drop this diff

wmesard (commented):
When an operation does take a long time, how will the cx change as a result of this PR? Fewer annoying error messages in the log, sure. But what else? Will it reduce the number of AWS API calls? Increase it? Will it reduce the length of time that it takes for K8s to notice that the volume attachment state has changed? Increase it?

charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -241,6 +241,7 @@ spec:
           imagePullPolicy: {{ default .Values.image.pullPolicy .Values.sidecars.attacher.image.pullPolicy }}
           args:
             - --csi-address=$(ADDRESS)
+            - --timeout=60s
```
rdpsin, line 244: Why not make it configurable?

ConnorJC3:
How will this behave if the user is already passing `--timeout` via `additionalArgs`? Do we need a guard against that?

charts/aws-ebs-csi-driver/templates/controller.yaml
torredil: This parameter is configurable via `additionalArgs`.

deploy/kubernetes/base/controller.yaml
torredil: Great question! - measured against 30s and saw a significant improvement compared to the default, but still observed context cancellations for attach and resize operations in several external test runs. For 60s, out of 25 external runs, **0** context cancellations were extracted from the logs.

Not directly relevant to this discussion but for perspective, azuredisk CSI driver sets a timeout value of `1200s` for the external attacher. Having said that, based on that data and given this is not a one way door, I suggest updating the timeout values to 60s as a more sensible default without potentially introducing excessively long waits in cases where an operation is genuinely stalled.

Will post one more comment with a few data points answering outstanding questions before this is merged.

AndrewSirenko:
/lgtm 
as two way door

charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -339,6 +345,9 @@ spec:
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.sidecars.volumemodifier.image.repository .Values.sidecars.volumemodifier.image.tag }}
           imagePullPolicy: {{ default .Values.image.pullPolicy .Values.sidecars.volumemodifier.image.pullPolicy }}
           args:
+            {{- if not (regexMatch "(-volumemodifier)" (join " " .Values.sidecars.resizer.additionalArgs)) }}
```
ConnorJC3, line 348: Wrong additional args, parameter, using resizer

ConnorJC3 (commented):
Missing snapshotter?

charts/aws-ebs-csi-driver/templates/controller.yaml
torredil: Fixed.

torredil:
> Missing snapshotter?

The snapshotter sidecar already uses a timeout value of 60s by default.

torredil:
The primary goal of adjusting the timeout values in this PR is to prevent premature context cancellations for CSI operations. With longer timeout values, Kubernetes (more specifically the sidecars) will wait a longer duration (60s) before retrying cancelled operations - this does not inherently change the speed at which (as an example) the volume attachment state changes - it simply provides the respective EC2 API call more time to complete before the sidecar retries. 

In terms of operational performance, the increased timeout values would result in delays for operations that are genuinely stuck (which would likely be an indicator of a bug or inefficiency in the driver code. In this regard, increasing the timeout values improves the resiliency of our code because bugs can hide underneath the current set of values). However, in all other cases the operational performance is **improved** as the state is updated sooner because there is no need to wait for a retry.

/approve

PR #1828, opened 2023-11-02
Chart: add egress network policy for controller

calvix:
**Is this a bug fix or adding a new feature?**
Bug fix
In a secure environment with strict network policy mode, all traffic is denied except traffic allowed by the network policies.

In order to be able to run the `aws-ebs-csi-driver ` in such an environment  it is necessary to specify the network policy for the controller otherwise it will end with the error:
```
I1102 15:12:02.790807       1 event.go:298] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"example-pvc", UID:"81413a15-1473-4778-94d4-6a2908249527", APIVersion:"v1", ResourceVersion:"37899", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "gp2": rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

**What testing is done?**

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller-network-policy.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller-network-policy.yaml
@@ -1,0 +1,13 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "aws-ebs-csi-driver.name" . }}-controller
+  namespace: kube-system
+spec:
+  egress:
+  - {}
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "aws-ebs-csi-driver.name" . }}
+  policyTypes:
+  - Egress


rdpsin:
What's the effect of this on non-airgapped environments? Should this be put behind a conditional?

calvix:
> What's the effect of this on non-airgapped environments? Should this be put behind a conditional?

if the environment is not running any network policy controller it should have no effect

If the environment is running a network policy controller in non-strict mode (all traffic is allowed unless network policy specifies otherwise) it should work the same


But if you feel this should be conditional and enabled only when a certain value is set than I am fine to adjust the changes.

rdpsin:
My personal preference would be to put it behind a flag so that it doesn't surprise users updating to a new version of the chart (even though there's no functional change). @torredil @ConnorJC3 @AndrewSirenko are the arbiters, though. :)

torredil (changes requested):
Hi @calvix, thank you for your contribution.

Let's consider the case where a cluster administrator has already deployed a network policy for the controller. 

By deploying this network policy we could be unintentionally overriding restrictions set by that existing policy because network policies are additive. This would allow for broad egress traffic, which in environments where traffic needs to be controlled for compliance or security reasons would be very undesirable.

+1 to @rdpsin's suggestion. See an example in minio's chart here: https://github.com/minio/minio/blob/754f7a8a395e87a744050476d8b16c16af50a800/helm/minio/values.yaml#L500 where they provide this configuration, disabled by default.

charts/aws-ebs-csi-driver/templates/controller-network-policy.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "aws-ebs-csi-driver.name" . }}-controller
+  namespace: kube-system
```
ConnorJC3, line 5: This shouldn't be hardcoded, it should use `Release.Namespace`

charts/aws-ebs-csi-driver/templates/controller-network-policy.yaml
```diff
@@ -0,0 +1,13 @@
...
+spec:
+  egress:
+  - {}
+  podSelector:
+    matchLabels:
```
ConnorJC3, line 10: This should use `aws-ebs-csi-driver.selectorLabels`

ConnorJC3 (changes requested):
Sorry about the late review on this, but agree with above. We definitely cannot have this enabled by default, our security-conscious users would be very mad if we suddenly deploy a default-enabled `NetworkPolicy` that bypasses their configurations.

To be honest, I'm not sure if there's even a strong need for this at all - in environments so security conscious the user has defaulted the `NetworkPolicy` to egress disabled, would they really want a policy with an unlimited egress policy?

ConnorJC3:
/close

Hi, the team talked about this, and we decided that we don't want to move forwards with this change at the moment.

Deploying a default-enabled `NetworkPolicy` is extremely likely to surprise security-conscious users. Users running clusters with special requirements like strict network policy mode are likely to have a better idea of what policy is appropriate than a blanket policy we apply to all clusters.

PR #1833, opened 2023-11-08
Do not return an error if volume is optimizing in validateModifyVolume

torredil:

Bug

This PR fixes a bug where an error is incorrectly returned if a volume is in an optimizing state.

According to cloud, a modification is considered done if the volume is an `optimizing` state:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L1320-L1323

In the case where `ResizeOrModifyDisk` hasn't been previously called for a volume, we simply move on to check if the volume is in the [desired state](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L523). If it is, `ResizeOrModifyDisk` succeeds even if the volume is still `optimizing`. However, if `checkDesiredState` returns an error or the sidecar times out, [latestMod](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L1287) won't be nil the second time around and we error out if the volume is `optimizing`:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d071db859707479a4c5ee5ba93cca874507519f3/pkg/cloud/cloud.go#L1298-L1300

**What testing is done?**

- `make test`
- CI


ConnorJC3 (changes requested):
Not attempting to _start_ a modification when a volume is in the OPTIMIZING state is intentional

AndrewSirenko:
> Not attempting to _start_ a modification when a volume is in the OPTIMIZING state is intentional

Why are we validating this instead of relying on the EC2 API? Won't EC2 ModifyVolume error out on this case?

torredil:
> Not attempting to start a modification when a volume is in the OPTIMIZING state is intentional

By my understanding not starting a modification if the volume is already in the desired state would be a more appropriate check, could you elaborate?

ConnorJC3:
> Why are we validating this in our driver instead of relying on the EC2 API? Won't EC2 ModifyVolume error out on this case?

Because it produces spurious failed API calls. A volume in the OPTIMIZING state cannot be modified, attempting to make an API call is 100% guaranteed to fail.

> By my understanding not starting a modification if the volume is already in the desired state would be a more appropriate check, could you elaborate?

The appropriate fix here is that this check should not be running until we are sure we need to make a modification. Removing an entire block of checks is not the appropriate fix for this bug.

AndrewSirenko:
> The appropriate fix here is that this check should not be running until we are sure we need to make a modification. Removing an entire block of checks is not the appropriate fix for this bug.

If we really want to prevent an extra API call, I think the most appropriate check would be something along the lines of:

```
if lastModification.State == OPTIMIZING
    if lastModification.NewValues == ModifyVolumeRequest.DesiredValues
           return modification success
    else
        return error("Cannot modify volume in optimizing state")
```

ConnorJC3:
> If we really want to prevent an extra ModifyVolume API call, I think the most appropriate check would be something along the lines of:

Agree that would work, but I think the simplest fix given the existing code is just to move this `if` block to AFTER the `needsVolumeModification` check and leave everything else the same

```
else if state == ec2.VolumeModificationStateOptimizing {
  return true, 0, fmt.Errorf("volume %q in OPTIMIZING state, cannot currently modify", volumeID)
}
```

AndrewSirenko (approved):

PR #1834, opened 2023-11-09
Increase default QPS and worker threads of sidecars

ConnorJC3:

Mostly a bug fix

Now that #1819 is going to land, the default QPS values are way too low. Bump them up so the driver does not come by default in slow mode.

Includes a check to prevent setting these values if the user has already done so via `additionalArgs`.

**What testing is done?** 

CI, local rendering

Use `helm template . --set 'sidecars.provisioner.additionalArgs[0]=--kube-api-qps=999'` to confirm that the `additionalArgs` check is working

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -209,6 +209,11 @@ spec:
             {{- end }}
             {{- end }}
             - --default-fstype={{ .Values.controller.defaultFsType }}
+            {{- if not (regexMatch "(-kube-api-qps)|(-kube-api-burst)|(-worker-threads)" (join " " .Values.sidecars.provisioner.additionalArgs)) }}
+            - --kube-api-qps=20
+            - --kube-api-burst=100
+            - --worker-threads=100
+            {{- end }}
             {{- range .Values.sidecars.provisioner.additionalArgs }}
             - {{ . }}
             {{- end }}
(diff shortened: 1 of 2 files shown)


AndrewSirenko (approved):

torredil (changes requested):
```
$ kubectl get pods -A | grep "ebs"

kube-system   ebs-csi-controller-68bbbf668b-vbzd2    5/6     CrashLoopBackOff   4 (59s ago)   2m27s
kube-system   ebs-csi-controller-68bbbf668b-wr88p    5/6     CrashLoopBackOff   4 (63s ago)   2m27s
kube-system   ebs-csi-node-2jr6c                     3/3     Running            0             2m27s
kube-system   ebs-csi-node-kbzcm                     3/3     Running            0             2m27s
```

```
$ kubectl logs ebs-csi-controller-68bbbf668b-vbzd2 -n kube-system -c csi-resizer

flag provided but not defined: -worker-threads
```

resizer uses `workers`: https://github.com/kubernetes-csi/external-resizer

torredil:
Small change otherwise lgtm

torredil (approved):
Thanks!

/lgtm
/approve

AbeOwlu:
@ConnorJC3 , quick inquiry on the trajectory of the ebs controller regarding client side qps and burst config;
- the upstream idea is to eventually let the server P&F decide on load handling and rate limiting response. Is this on the path for ebs csi?

ConnorJC3:
@AbeOwlu we take the upstream sidecars essentially as-is, so if this change is made there it will automatically be integrated in the EBS CSI Driver.

Issue #1808, opened 2023-10-25
The driver reports incorrect number of allocatables count


zerkms:
/kind bug

**What happened?**

I have a 1.26.4 cluster that runs 4 worker nodes each of which is `t3a.xlarge`. All 4 joined the cluster the same day (within couple hours). All have 2 network interfaces attached.

Yet all 4 report different number of `Allocatables.Count`: 9, 10, 16, 26.

**What you expected to happen?**

I expect all of them to have it 26.

**How to reproduce it (as minimally and precisely as possible)?**

I don't know :shrug: 

**Anything else we need to know?**:

**Environment**: aws
- Kubernetes version (use `kubectl version`): 1.26.4
- Driver version: 1.24.0

zerkms:
Further analysis: the driver uses `GetNumBlockDeviceMappings` via metadata service.

And metadata service reports the number of mapped devices wrong!!

On the machine: 

```
$ curl -s http://169.254.169.254/latest/meta-data/block-device-mapping/|grep ebs| wc -l
16
```

Yet, aws cli:

```
$ aws ec2 describe-instance-attribute --instance-id xxx --attribute=blockDeviceMapping | jq '.BlockDeviceMappings | length'
5
```
(of which 4 are ebs + 1 for the root block devices)

And the aws UI console shows the same 4+1 attached block devices.

zerkms:
After vm full shutdown (via console), it finally reports the matching numbers. Yet, `csinode` still shows 10 as capacity.

Is that value only determined on node register and then not refreshed? If so - how would one update it?

ConnorJC3:
> And metadata service reports the number of mapped devices wrong!!

I'd recommend reporting this to AWS support (especially if it's reproducible), they should be able to get this to the appropriate team that works on IMDS. Unfortunately we don't have a good way to know if/when the metadata service is wrong so we trust its output.

> Is that value only determined on node register and then not refreshed? If so - how would one update it?

Generally, restarting the EBS CSI Driver pod should cause Kubernetes to re-query the limit and update it. See if that fixes it - usually rebooting the node would be equivalent, my best guess is that there might have been a short period of time during the node's startup where the wrong value was still being displayed.

zerkms:
[...]

PR #1843, opened 2023-11-21
Remove block-device-mapping from attach limit

jsafrane:
/kind bug
Fixes: #1808

`meta-data/block-device-mapping` reports devices that were attached to the VM at the time the VM was started. It does not reflect devices attached/detached later.

In addition, even when it showed the attachments accurately, the CSI driver uses this information only at the driver startup. At this time, there may be some CSI volumes attached from the previous node boot. Those volumes are then counted as permanently attached to the node, while Kubernetes may detach them later.

Therefore do not count block-device-mapping at all and always reserve just 1 attachment for the root disk. Assume all other attachments are CSI volumes, and let Kubernetes do the calculations of attached volumes instead.

**What testing is done?** 
Tested with Kubernetes 1.28 (OpenShift) + a node shut down without draining it first and starting it with some CSI driver volumes attached.

Before this PR:

```
$ kubectl get csinode -o yaml
...
kind: CSINode
spec:
  drivers:
  - allocatable:
      count: 22   # 4 CSI volumes are counted as permanently attached by the driver + 1 root disk
    name: ebs.csi.aws.com
    nodeID: i-0d86f24259c8ce1bd
    topologyKeys:
```

After this PR:

```
kind: CSINode
spec:
  drivers:
  - allocatable:
      count: 26 # just the root disk is counted as permanently attached.
    name: ebs.csi.aws.com
    nodeID: i-0d86f24259c8ce1bd
    topologyKeys:
```

### DIFF
--- a/pkg/cloud/metadata.go
+++ b/pkg/cloud/metadata.go
@@ -26,13 +26,12 @@ import (
 
 // Metadata is info about the ec2 instance on which the driver is running
 type Metadata struct {
-	InstanceID             string
-	InstanceType           string
-	Region                 string
-	AvailabilityZone       string
-	NumAttachedENIs        int
-	NumBlockDeviceMappings int
-	OutpostArn             arn.ARN
+	InstanceID       string
+	InstanceType     string
+	Region           string
+	AvailabilityZone string
+	NumAttachedENIs  int
+	OutpostArn       arn.ARN
 }
 
 const (
@@ -70,10 +69,6 @@ func (m *Metadata) GetAvailabilityZone() string {
 
 func (m *Metadata) GetNumAttachedENIs() int {
 	return m.NumAttachedENIs
-}
-
-func (m *Metadata) GetNumBlockDeviceMappings() int {
-	return m.NumBlockDeviceMappings
 }
 
 // GetOutpostArn returns outpost arn if instance is running on an outpost. empty otherwise.
--- a/pkg/cloud/metadata_ec2.go
+++ b/pkg/cloud/metadata_ec2.go
@@ -61,23 +61,13 @@ func EC2MetadataInstanceInfo(svc EC2Metadata, regionFromSession string) (*Metada
 	}
 
 	attachedENIs := strings.Count(enis, "\n") + 1
-	blockDevMappings := 0
-
-	if !util.IsSBE(doc.Region) {
-		mappings, mapErr := svc.GetMetadata(blockDevicesEndpoint)
-		if mapErr != nil {
-			return nil, fmt.Errorf("could not get number of block device mappings: %w", err)
-		}
-		blockDevMappings = strings.Count(mappings, "ebs")
-	}
 
 	instanceInfo := Metadata{
-		InstanceID:             doc.InstanceID,
-		InstanceType:           doc.InstanceType,
-		Region:                 doc.Region,
-		AvailabilityZone:       doc.AvailabilityZone,
-		NumAttachedENIs:        attachedENIs,
-		NumBlockDeviceMappings: blockDevMappings,
+		InstanceID:       doc.InstanceID,
+		InstanceType:     doc.InstanceType,
+		Region:           doc.Region,
+		AvailabilityZone: doc.AvailabilityZone,
+		NumAttachedENIs:  attachedENIs,
 	}
 
 	outpostArn, err := svc.GetMetadata(outpostArnEndpoint)
--- a/pkg/cloud/metadata_interface.go
+++ b/pkg/cloud/metadata_interface.go
@@ -12,7 +12,6 @@ type MetadataService interface {
 	GetRegion() string
 	GetAvailabilityZone() string
 	GetNumAttachedENIs() int
-	GetNumBlockDeviceMappings() int
 	GetOutpostArn() arn.ARN
 }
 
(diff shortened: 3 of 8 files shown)


torredil (approved):
Thank you @jsafrane /lgtm

At the very least, `blockDevMappings` should definitely not take into account CSI-managed volumes. Users relying on the current behavior can make use of the `node.volumeAttachLimit` parameter to "reserve" attachment slots.

torredil:
cc: @ConnorJC3 @AndrewSirenko to review before approval.

AndrewSirenko (approved):

AndrewSirenko:
Big thank you @jsafrane for diving deep on #1808.  

> It's quite hard and error prone to recover from this situation. Cluster admin must drain + shut down a node that has misleading capacity count and start it again. It's not enough to restart the CSI driver.

^^ is quite the workload interruption to pay without this PR.

zerkms:
@jsafrane I don't fully understand this particular fix, but does it take into account network interfaces? They also contribute to the same limit. Eg: if you attach 10 more network interfaces, you will be short by number of EBS to be attached by the same number.

kon-angelo:
I also need to express concerns. Wouldn't this PR break the use of any non-boot volume if it is not done kubernetes ? 

For example some of our users in [gardener](https://gardener.cloud) do opt for additional data volumes which will completely break. It is quite the assumption to make that no other volume can be used at creation and it is a highly breaking change that should be marked as such.

torredil:
Hi @kon-angelo thanks for chiming in and starting this discussion, really appreciate the feedback. 

How are you folks solving this problem with other providers? It would be great to work on a solution that works well with the broader community - currently, K8s does not support managing both non-CSI and CSI volumes simultaneously. Other CSI drivers do not take into account non-CSI managed volumes.

Not at all familiar with gardener so pardon if this is completely unfeasible or does not meet your requirements - given that the driver is deployed by [default](https://github.com/gardener/gardener-extension-provider-aws/blob/da029590d6bb550a6c74634903c47caaa30a05b6/docs/usage/usage.md?plain=1#L571) on every cluster already: would it be possible to follow the standard and use the driver to manage these additional data volumes?

---
I'm not inclined to revert the changes made by this PR - it fixes a critical bug in the driver that affects all users relying on it to manage the lifecycle of their volumes (the supported, common case) - and as previously mentioned, the resulting failure mode requires manual intervention to fix (once the scheduler incorrectly schedules, the stateful pod needs to be deleted + node cordoned so that the same mistake does not happen again).

> _Wouldn't this PR break the use of any non-boot volume if it is not done via kubernetes ?_

> _it breaks the ability to use non-CSI volumes (other than the root volume) on nodes where the EBS CSI driver is running._

For clarity: the changes in this PR do not inherently break the ability to use non-CSI volumes on nodes where the EBS CSI driver is running. If the node has capacity to support the attachment, it will succeed (these data volumes are practically guaranteed to successfully grab a slot - I'd be more worried about the consequences of that). That said, this has been actively discouraged in the community: https://github.com/kubernetes/kubernetes/issues/81533#issuecomment-522601378 cc: @gnufied 

As Jan already mentioned, the driver reports the allocatable count on the node _only once_ during driver initialization. There is no reconciliation. Even when reverting, if users attach (or detach) volumes or ENIs outside of K8s such as via the CLI, the allocatable count will not be updated after plugin registration.

Finally, also wanted to highlight that since the driver is unaware of block devices attached outside of K8s, attempting to force the **CSI** driver to be aware of **non-CSI** attachments means the hard-coded table of values specifying [device names](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/61c721b88933faec8fbb9c6610d5fc9d265ef5d2/pkg/cloud/devicemanager/device_names.go#L32) is unreliable - if the driver attempts to re-use a device name, the attachment will fail [indefinitely](https://repost.aws/knowledge-center/ebs-stuck-attaching). This async failure mode is extra troublesome because it's not very obvious and quite hard to track down.

jsafrane:
There is no Kubernetes way to count volume attachments done outside of Kubernetes. Different CSI drivers solve it in different way, a very quick check of few random cloud based CSI drivers shows:

* OpenStack Cinder reads attach limit from a config file.
* vSphere and Alibaba Disk can optionally read attach limit from env. var.
* Azure Disk can optionally read attach limit from node labels (that defeats CSI purpose to be container orchestrator agnostic!).
* IBM VPC and Google Persistent Disk reserves 1 slot for the root disk and does not have any way to override the limit.

None of these CSI drivers counts existing attachments at the driver startup.

Does AWS provide a reliable way how to distinguish CSI and non-CSI attachment in instance metadata or with very limited AWS permissions? And is driver pod start acceptable time to count the non-CSI attachments? It can happen at arbitrary time, so those non-CSI attachments basically need to be attached forever, which does not sound very reliable to me. But it would work for Gardener's data disks, if I understood it correctly.

I agree this PR could break some clusters and it should be treated as a breaking change. In the worst case, gardener and other clusters that need reserved attachment slots could set `--volume-attach-limit` to override driver's calculation.

jsafrane:
> [...] as this allows you to separate CSI and non-CSI volumes by virtue of the first being there on VM creation and the metadata service accounting it.

That is actually the problem here. AWS metadata report attachments not at VM creation time, but at node startup time. If there were CSI volume attached to it at the startup time, then the instance metadata contain both CSI and non-CSI volume and the driver has no way to tell which is which. We can teach cluster admins to always drain nodes before shutdown / reboot, but in reality there are always some nodes that get shut down ungracefully.

PR #1845, opened 2023-11-24
Refactor E2E: Streamline setting of CreateVolume Parameters in test-cases. 

AndrewSirenko:
Refactor

The aws-ebs-csi-driver's E2E tests currently rely on test-specific helper-function behavior and kitchen-sink wrapper-structs (See the git-blame for the [VolumeDetails](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blame/97688f8549feb99228f442af725c0bd492ac9a60/tests/e2e/testsuites/specs.go#L35-L52)). While test-specific extra helper-function-logic and wrapper-struct-fields helped increase the driver's test-coverage sooner, they negatively impact the discoverability and readability of these utility functions when writing more e2e tests.   

This PR begins to standardize these disparate ways of writing e2e tests by having any test that dynamically provisions volumes rely on explicitly setting [CreateVolume (StorageClass) Parameters](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/parameters.md) within the test-cases.  

These commits are bundled into one PR to simplify code review + CI (Prow tests + Merging).

The purposes of each commit are:
- [Refactor E2E: Remove hidden volume performance parameter setting](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/70db58b6cdbe2fc0d979eb0fb1f87e7e97ee6422)
  - Currently, dynamically provisioning volumes via E2E helper functions secretly sets `iops`, `iopsPerGb`, or `throughput` based on the `volumeType` passed into a `VolumeDetails` struct. Worse, E2E packages should not be responsible for maintaining a source of truth for EBS limits like [maximum throughput for each volume type](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/97688f8549feb99228f442af725c0bd492ac9a60/tests/e2e/driver/ebs_csi_driver.go#L154C9-L154C9). Instead, unless E2E tests are explicitly testing these performance parameters, their default should be left to the EC2 API (via the driver) or be explicitly set in test-case parameters via constants like `DefaultE2EIopsIoVolumes`.
- [Fixup E2E: Pre-Provisioned volumes don't use VolumeType](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/cd1074f706f62327bac7dea0c3e963369cb5eac4) and [Fixup E2E: Pre-Provisioned defaultVolumeType typo](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/9a4b21773c3819a88c30ebc24ab576bdc6eda5a2)
  - Small clean-ups of existing code. 
- [Refactor E2E: Standardize adding CreateVolume parameters to testcases](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/b68912b21c0ad89a0401ce6cd334712179464e1e)
  - When dynamically provisioning volumes, test-case fields like `VolumeType`, `FSType`, and `Encrypted` are [CreateVolume parameters](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/parameters.md) and should not be treated differently from parameters like `iops` and `blockSize` when creating tests. 
- [Refactor E2E: Generalize format_options CreateVolumeDetails helper fcn.](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/246706f2f9e88cde037f55077a7a463165381d79)
  - Cleans up format_options E2E tests based on previous refactors, as well as generalizes createFormatOptionVolumeDetails so that future E2E tests like those for the ModifyVolume feature can use it.  

**What testing is done?** 
All E2E tests passed locally *after each commit*.


tests/e2e/testsuites/e2e_utils.go
```diff
@@ -19,8 +19,27 @@ import "fmt"
 const (
 	DefaultVolumeName = "test-volume-1"
 	DefaultMountPath  = "/mnt/default-mount"
+
+	DefaultE2EIopsIoVolumes = "100"
```
ConnorJC3, line 23: Drop `E2E` from the name (change it to `DefaultIopsIoVolumes`), you're in the e2e package so everything here is for the e2e tests, no need to specify.

pkg/driver/constants.go
```diff
@@ -82,6 +82,9 @@ const (
 	// BlockExpressKey increases the iops limit for io2 volumes to the block express limit
 	BlockExpressKey = "blockexpress"
 
+	// FSTypeKey configures the file system type that will be formatted during volume creation.
+	FSTypeKey = "csi.storage.k8s.io/fstype"
```
ConnorJC3, line 86: If this isn't used in the driver itself, it should be in the e2e package.

AndrewSirenko: We should use it in the driver here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/61c721b88933faec8fbb9c6610d5fc9d265ef5d2/pkg/driver/controller.go#L160

If all other parameter keys mentioned in our docs are in `constants.go` I think that FSTypeKey belongs there as well. Especially because if one day we do use this key in the driver for something more substantial, we would have a duplicated constant.

tests/e2e/testsuites/e2e_utils.go
```diff
@@ -19,8 +19,27 @@ import "fmt"
...
 )
 
 func PodCmdWriteToVolume(volumeMountPath string) string {
 	return fmt.Sprintf("echo 'hello world' >> %s/data && grep 'hello world' %s/data && sync", volumeMountPath, volumeMountPath)
 }
+
+func CreateVolumeDetails(createVolumeParameters map[string]string, volumeSize string) *VolumeDetails {
+	allowVolumeExpansion := true
```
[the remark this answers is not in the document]
ConnorJC3: Fair enough, I'd probably keep it as is then

torredil (approved):
/lgtm

PR #1861, opened 2023-12-12
Add retry and background run to node taint removal

ConnorJC3:

Bug fix

Sometimes, the node taint removal can fail (for example, due to a temporary networking outage, or if another client tries to remove a taint at the same time as us and gets to the API server first). In that case we should retry the taint removal.

This PR does two things:
- Moves the taint removal to run in a goroutine in the background so it doesn't block node startup
- Adds retries with an exponential backoff to the taint removal

**What testing is done?** 

Manually tested on a local cluster (but please at least one reviewer double check my work), as well as CI

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
@@ -33,6 +34,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8stypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/volume"
 )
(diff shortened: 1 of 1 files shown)


torredil (commented):
`make verify` needs fixing:
```
pkg/driver/node.go:859:30: `removeTaintInBackground` - `k8sClient` is unused (unparam)
func removeTaintInBackground(k8sClient cloud.KubernetesAPIClient) {
```

code changes lgtm, manually testing. Will apply label after manual validation + above is resolved.

pkg/driver/node.go
```diff
@@ -861,8 +883,8 @@ func removeNotReadyTaint(k8sClient cloud.KubernetesAPIClient) error {
 
 	clientset, err := k8sClient()
 	if err != nil {
-		klog.V(4).InfoS("Failed to communicate with k8s API, skipping taint removal")
-		return nil //lint:ignore nilerr Failing to communicate with k8s API is a soft failure
+		klog.V(4).InfoS("Failed to setup k8s client")
+		return nil //lint:ignore nilerr If there are no k8s credentials, treat that as a soft failure
```
torredil, line 887: Thanks for the quick turnaround on this - /lgtm

Q: Are there failure modes here where it would be desirable to retry here as opposed to returning nil?

torredil (approved):

AndrewSirenko (approved):
Solves active customer problem. 

Offline discussion was had about configuration values for taintRemovalBackoff. Decided that currently this is a two-way door solution, and we may alleviate cx painpoints without introducing additional configuration option. Currently any retry/timeout is better than no retry and an infinite timeout.

pkg/driver/node.go
ConnorJC3: All that constructing the client does is try to load k8s credentials (it doesn't even reach out to the API server). Thus, all failures here are permanent failures, the most typical failure would be that no `kubeconfig` could be found.

Issue #1224, opened 2022-05-02
Safeguard against outdated `/dev/disk/by-id/` symlinks that can lead Pod to mount the wrong volume


ialidzhikov:
/sig storage
/kind bug

**What happened?**

For nvme volumes the aws-ebs-csi-driver relies on the `/dev/disk/by-id/` symlink to be able to determine the nvme device attached for the given volume id. 

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/0de2586eab537209dcddb8150db52b2409c996cd/pkg/driver/node_linux.go#L83-L88

This symlink is updated by udev rules that react on kernel attach/detach events.

We also know that during Pod restarts volumes can be detached and quickly reattached at another location (i.e. `/dev/nvme3n1` now, but after detach/attach cycle it would be `/dev/nvme4n1`).

A known [cloud-init bug](https://bugs.launchpad.net/cloud-init/+bug/1946003) causes udev rules to be processed with a huge delay. To demonstrate the impact of the corresponding bug let's compare the `udevadm monitor -s block` output on a "healthy" and "affected" Node:
- "healthy" Node
   ```
   udevadm monitor -s block
   monitor will print the received events for:
   UDEV - the event which udev sends out after rule processing
   KERNEL - the kernel uevent
   
   KERNEL[735.827384] change   /devices/pci0000:00/0000:00:1d.0/nvme/nvme3/nvme3n1 (block)
   UDEV  [735.850150] change   /devices/pci0000:00/0000:00:1d.0/nvme/nvme3/nvme3n1 (block)
   KERNEL[737.628955] change   /devices/pci0000:00/0000:00:1c.0/nvme/nvme4/nvme4n1 (block)
   KERNEL[737.657828] change   /devices/pci0000:00/0000:00:1e.0/nvme/nvme2/nvme2n1 (block)
   UDEV  [737.681569] change   /devices/pci0000:00/0000:00:1c.0/nvme/nvme4/nvme4n1 (block)
   UDEV  [737.695555] change   /devices/pci0000:00/0000:00:1e.0/nvme/nvme2/nvme2n1 (block)
   KERNEL[738.219222] change   /devices/pci0000:00/0000:00:1f.0/nvme/nvme1/nvme1n1 (block)
   UDEV  [738.246151] change   /devices/pci0000:00/0000:00:1f.0/nvme/nvme1/nvme1n1 (block)
   ```

- "affected" Node

   ```
[...]

PR #1878, opened 2023-12-21
Use lsblk to safeguard against outdated symlinks

ConnorJC3:

Bug fix

Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1224

**What testing is done?** 

```text
I1221 18:37:12.769528       1 node_linux.go:140] "Comparing volume serial" expected="vol0b99a3d34724acd2b" actual="vol0b99a3d34724acd2b"
```

CI/Added new unit tests

### DIFF
--- a/pkg/driver/node_linux.go
+++ b/pkg/driver/node_linux.go
@@ -22,7 +22,9 @@ package driver
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -47,6 +49,7 @@ func (d *nodeService) appendPartition(devicePath, partition string) string {
 // if the device is not nvme, return the path directly
 // if the device is nvme, finds and returns the nvme device path eg. /dev/nvme1n1
 func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (string, error) {
+	strippedVolumeName := strings.Replace(volumeID, "-", "", -1)
 	canonicalDevicePath := ""
 
 	// If the given path exists, the device MAY be nvme. Further, it MAY be a
@@ -76,6 +79,9 @@ func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (st
 		}
 
 		klog.V(5).InfoS("[Debug] The canonical device path was resolved", "devicePath", devicePath, "cacanonicalDevicePath", canonicalDevicePath)
+		if err = verifyVolumeSerialMatch(canonicalDevicePath, strippedVolumeName, execRunner); err != nil {
+			return "", err
+		}
 		return d.appendPartition(canonicalDevicePath, partition), nil
 	}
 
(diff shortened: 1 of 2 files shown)


torredil (commented):
Is this fix still necessary? it appears the issue leading to #1224 has been addressed via:
- https://bugs.launchpad.net/cloud-init/+bug/1946003
- https://github.com/canonical/cloud-init/issues/3912
- https://github.com/kubernetes/kubernetes/pull/100183

pkg/driver/node_linux.go
```diff
@@ -116,6 +125,31 @@ func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (st
 	return canonicalDevicePath, nil
 }
 
+func checkVolume(canonicalDevicePath, strippedVolumeName string) error {
```
torredil, line 128: np: Readability can be improved here by using a more descriptive func name, perhaps`verifyVolumeSerialMatch`?

torredil (commented):
> Even if the race condition that causes it is very rarely lost, the bug avoided is catastrophic

That makes sense. In that case, agree that paying the `lsblk` fee is worthwhile.

/lgtm

pkg/driver/node_linux.go
```diff
@@ -116,6 +125,31 @@ func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (st
...
+		volumeRegex := regexp.MustCompile(`vol[a-z0-9]+`)
+		for _, volume := range volumeRegex.FindAllString(string(output), -1) {
+			klog.V(6).InfoS("Comparing volume serial", "canonicalDevicePath", canonicalDevicePath, "expected", strippedVolumeName, "actual", volume)
+			if volume != strippedVolumeName {
+				return fmt.Errorf("Refusing to mount %s because it claims to be %s but should be %s", canonicalDevicePath, volume, strippedVolumeName)
+			}
+		}
+	} else {
+		// If the command fails (for example, because lsblk is not available), silently ignore the error and proceed
+		klog.V(5).ErrorS(err, "Ignoring lsblk failure", "canonicalDevicePath", canonicalDevicePath, "strippedVolumeName", strippedVolumeName)
```
rdpsin, line 147: np: How about increasing the severity? This can help customers (and us) debug this issue faster.

pkg/driver/node_linux.go
```diff
@@ -116,6 +125,31 @@ func (d *nodeService) findDevicePath(devicePath, volumeID, partition string) (st
 	return canonicalDevicePath, nil
 }
 
+func verifyVolumeSerialMatch(canonicalDevicePath, strippedVolumeName string) error {
```
AndrewSirenko, line 128: I know that we said adding unit tests for the file is too difficult, but you cleanly extracted this check's logic into what looks to be a testable function (I see [this article explains how we would mock that lsblk exec](https://stackoverflow.com/a/71166114)). 

Could we consider adding unit tests for `verifyVolumeSerialMatch`? 

Might be overkill, but would help us get closer to node service unit tests if/when we do refactor `node_linux.go`.

AndrewSirenko (approved):
Cleanly done, thank you!

pkg/driver/node_linux.go
ConnorJC3: I debated with myself about this, so I'd like to hear other's opinions.

On one hand, a broken `lsblk` is an arguably incorrect installation environment.

On the other hand, it feels unhelpful to flood the logs with the same error for every staged volume if `lsblk` is missing from the container or similar.

AndrewSirenko (approved):

pkg/driver/node_linux.go
torredil: Thought about this quite a bit an ultimately landed on the following as well:

> it feels unhelpful to flood the logs with the same error for every staged volume if lsblk is missing from the container or similar.

PR #1883, opened 2024-01-02
Remove premature CreateVolume error if requested IOPS is below minimum

AndrewSirenko:
Bug fix

EBS CSI Driver will let [EC2 CreateVolume](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html) enforce minimum IOPS values instead of enforcing the [hardcoded minimum limits in cloud.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/0c459169dd183c1b1fb27443d562cc0f6579a62b/pkg/cloud/cloud.go#L65-L78) (Unless `allowAutoIOPSPerGBIncrease` is set).

Note: This is a volume-type agnostic version of PR #1879, as requested by EBS CSI Driver team. 

Pro: 
- EBS CSI Driver relies more on EC2 API as source of truth for input validation (This fixes the GP3 storage class regression mentioned in #1879, as well as prevents future de-sync between the EC2 and CSI Driver Minimum IOPS experience) 

Con:
- Extra CreateVolume calls are made to EC2 API for misconfigured io1/io2 PVCs (potentially endless unless an `infeasable` PVC status is added to CreateVolume similar to how it was added for [Update ControllerModifyVolume Status API #4298](https://github.com/kubernetes/enhancements/pull/4298/files)). This could increase chance of EC2 throttling, but can be solved at the Kubernetes + CSI level via `infeasable` PVC parameter. 

See #1879

**What testing is done?** 
`make test`, Manual testing

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -511,10 +511,7 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		} else if diskOptions.IOPSPerGB > 0 {
 			requestedIops = int64(diskOptions.IOPSPerGB) * capacityGiB
 		}
-		iops, err = capIOPS(createType, capacityGiB, requestedIops, minIops, maxIops, maxIopsPerGb, diskOptions.AllowIOPSPerGBIncrease)
-		if err != nil {
-			return nil, err
-		}
+		iops = capIOPS(createType, capacityGiB, requestedIops, minIops, maxIops, maxIopsPerGb, diskOptions.AllowIOPSPerGBIncrease)
 	}
 
 	var tags []*ec2.Tag
@@ -1505,10 +1502,10 @@ func getVolumeAttachmentsList(volume *ec2.Volume) []string {
 }
 
 // Calculate actual IOPS for a volume and cap it at supported AWS limits.
-func capIOPS(volumeType string, requestedCapacityGiB int64, requestedIops int64, minTotalIOPS, maxTotalIOPS, maxIOPSPerGB int64, allowIncrease bool) (int64, error) {
+func capIOPS(volumeType string, requestedCapacityGiB int64, requestedIops int64, minTotalIOPS, maxTotalIOPS, maxIOPSPerGB int64, allowIncrease bool) int64 {
 	// If requestedIops is zero the user did not request a specific amount, and the default will be used instead
 	if requestedIops == 0 {
-		return 0, nil
+		return 0
 	}
 
 	iops := requestedIops
@@ -1517,10 +1514,6 @@ func capIOPS(volumeType string, requestedCapacityGiB int64, requestedIops int64,
 		if allowIncrease {
 			iops = minTotalIOPS
 			klog.V(5).InfoS("[Debug] Increased IOPS to the min supported limit", "volumeType", volumeType, "requestedCapacityGiB", requestedCapacityGiB, "limit", iops)
-		} else if volumeType == VolumeTypeGP3 {
-			klog.V(5).InfoS("[Debug] Did not increase IOPS", "volumeType", volumeType, "requestedCapacityGiB", requestedCapacityGiB)
-		} else {
-			return 0, fmt.Errorf("invalid IOPS: %d is too low, it must be at least %d", iops, minTotalIOPS)
 		}
 	}
 	if iops > maxTotalIOPS {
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -1505,10 +1502,10 @@ func getVolumeAttachmentsList(volume *ec2.Volume) []string {
 }
 
 // Calculate actual IOPS for a volume and cap it at supported AWS limits.
-func capIOPS(volumeType string, requestedCapacityGiB int64, requestedIops int64, minTotalIOPS, maxTotalIOPS, maxIOPSPerGB int64, allowIncrease bool) (int64, error) {
+func capIOPS(volumeType string, requestedCapacityGiB int64, requestedIops int64, minTotalIOPS, maxTotalIOPS, maxIOPSPerGB int64, allowIncrease bool) int64 {
```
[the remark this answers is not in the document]
AndrewSirenko: Unfortunately, this would be a breaking change for customers relying on this clamping for misconfigured storage classes.

Consider the misconfigured storage class: 

```
type: "gp3"
iops: "60000"
```

Currently, this storage class would be able to be created by the driver because [IOPS would be clamped down to 16000](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/0c459169dd183c1b1fb27443d562cc0f6579a62b/pkg/cloud/cloud.go#L75C11-L75C11). If we remove this conditional, this same storage class could no longer be used.

pkg/cloud/cloud.go
```diff
@@ -1505,10 +1502,10 @@ func getVolumeAttachmentsList(volume *ec2.Volume) []string {
...
 	// If requestedIops is zero the user did not request a specific amount, and the default will be used instead
 	if requestedIops == 0 {
-		return 0, nil
+		return 0
 	}
```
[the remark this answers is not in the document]
AndrewSirenko: This conditional is required so that the line beneath it, `iops := requestedIops`, does not set iops to 0.

pkg/cloud/cloud.go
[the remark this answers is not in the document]
AndrewSirenko: Not to say we should never do this (I personally think longterm we should), but that this PR is not the venue for it.

pkg/cloud/cloud_test.go
```diff
@@ -650,23 +633,6 @@ func TestCreateDisk(t *testing.T) {
 			},
 			expErr: nil,
 		},
-		{
-			name:       "fail: io2 with too low iopsPerGB",
```
torredil, line 654: Similar request here: update the test case to reflect new expected behavior.

pkg/cloud/cloud_test.go
```diff
@@ -575,23 +575,6 @@ func TestCreateDisk(t *testing.T) {
...
-				IOPSPerGB:     1,
-			},
-			expDisk: &Disk{
-				VolumeID:         "vol-test",
-				CapacityGiB:      4,
-				AvailabilityZone: defaultZone,
-			},
-			expCreateVolumeInput: nil,
-			expErr:               fmt.Errorf("invalid IOPS: 4 is too low, it must be at least 100"),
-		},
```
torredil, line 594: Let's update the test case to reflect the new expected behavior - io1 with too low iopsPerGB should still fail server-side:

```
{
			name:       "fail: io1 with too low iopsPerGB",
			volumeName: "vol-test-name",
			diskOptions: &DiskOptions{
				CapacityBytes: util.GiBToBytes(4),
				Tags:          map[string]string{VolumeNameTagKey: "vol-test", AwsEbsDriverTagKey: "true"},
				VolumeType:    VolumeTypeIO1,
				IOPSPerGB:     1,
			},
			expDisk: &Disk{
				VolumeID:         "vol-test",
				CapacityGiB:      4,
				AvailabilityZone: defaultZone,
			},
			expCreateVolumeInput: &ec2.CreateVolumeInput{
				Iops: aws.Int64(4),
			},
			expErr:             fmt.Errorf("could not create volume in EC2: InvalidParameterValue: An error occurred (InvalidParameterValue) when calling the CreateVolume operation\ncaused by: Volume iops of 4 is too low; minimum is 100."),
			expCreateVolumeErr: awserr.New("InvalidParameterValue", "An error occurred (InvalidParameterValue) when calling the CreateVolume operation", fmt.Errorf("Volume iops of 4 is too low; minimum is 100.")),
		},
```

torredil (commented):
/lgtm

pkg/cloud/cloud_test.go
ConnorJC3: Wouldn't that test the exact same thing as this existing test? https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/190f526d8290b2e7686e9432251a18feea3a34ac/pkg/cloud/cloud_test.go#L461-L471

torredil: Yes, that explains why code coverage remains unchanged - resolving.

torredil (approved):
/lgtm

PR #1894, opened 2024-01-12
Add scalability section to FAQ

AndrewSirenko:
Documentation

In [aws-ebs-csi-driver v1.25.0](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/3981294942b4aa3bdfac95a06d7ab57b43c8130d/CHANGELOG.md?plain=1#L26), we introduced several scalability related improvements to the driver. Some customers have asked for explanations of how this version improves driver scalability, as well as learning how to tune the sidecar parameters to their specific cluster setups.

What testing is done?
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ The EBS CSI Driver is compatible with Kubernetes versions v1.17+ and implements 
 * [Driver Installation](docs/install.md)
 * [Driver Launch Options](docs/options.md)
 * [StorageClass Parameters](docs/parameters.md)
+* [Frequently Asked Questions](docs/faq.md)
 * [Volume Tagging](docs/tagging.md)
 * [Volume Modification](docs/modify-volume.md)
 * [Kubernetes Examples](/examples/kubernetes)
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,5 +1,135 @@
 # Frequently Asked Questions
 
+## Driver performance for large-scale clusters
+
+### Summary of scalability-related changes in v1.25
+
+[Version 1.25 of aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG.md#v1250) featured four improvements to better manage the EBS volume lifecycle for large-scale clusters. 
+
+At a high-level:
+1. Batching EC2 `DescribeVolumes` API Calls across CSI gRPC calls ([#1819](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1819))
+	- This greatly decreases the number of EC2 `DescribeVolumes` calls made by the driver, which significantly reduces your risk of region-level throttling of the ['Describe*' EC2 API Action](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits)
+2. Increasing the default CSI sidecar `worker-threads` values (to 100 for all sidecars) ([#1834](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1834))
+	- E.g. the `external-provisioner` can be simultaneously running 100 `ControllerPublishVolume` operations, the `external-attacher` now has 100 goroutines for processing VolumeAttachments, etc. 
+	- This increases the number of in-flight EBS volume creations / attaches / modifications / deletions managed by the driver (which may increase your risk of region-level ['Mutating action' request throttling by the Amazon EC2 API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits) )
+	- **Note: If you are running multiple clusters within a single AWS account and region and risk hitting your [EC2 API Throttling Limits](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-increase)**. See [Request a limit increase](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits-rate-based) and the below [Fine-tuning the CSI Sidecar worker-threads parameter](faq.md) section
+3. Increasing the default CSI sidecar `kube-api-qps` (to 20) and `kube-api-burst` (to 100) for all sidecars ([#1834](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1834))
+	- Each sidecar can now send a burst of up to 100 [queries](https://kubernetes.io/docs/reference/using-api/api-concepts/#api-verbs) to the Kubernetes API Server before throttling itself. It will then allow up to 20 more requests per second until it stops bursting.   
+	- This keeps Kubernetes objects (`PersistentVolume`, `PersistentVolumeClaim`, and `VolumeAttachment`) more synchronized with the actual state of AWS resources, at the cost of increasing the load on the K8s API Server from `ebs-csi-controller` pods when many volume operations are happening at once. 
+4. Increasing the default CSI sidecar `timeout` values (from 15s to 60s) for all sidecars ([#1824](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1824))   
+    - E.g. the [external-attacher](https://github.com/kubernetes-csi/external-attacher?tab=readme-ov-file#csi-error-and-timeout-handling) will now give the driver up to 60s to report an attachment success/failure before retrying a `ControllerPublishVolume` call. Now the external-attacher won't prematurely time out a `ControllerPublishVolume` call that would've taken 20s before returning a success response.  
+	- This decreases the number of premature timeouts for CSI RPC calls, which reduces the number of replay EC2 API requests made by and waited for by the driver (at the cost of a longer delay during a real driver timeout (e.g. network blip leads to lost `ControllerPublishVolume` response))
+
+### EC2 and K8s CSI Sidecar Throttling Overview
+
+Both the EC2 API and K8s CSI sidecars base their API throttling implementation off of the [token bucket algorithm](https://en.wikipedia.org/wiki/Token_bucket). The Amazon EC2 API Reference provides a thorough explanation and example of how this algorithm is applied: [Request throttling for the Amazon EC2 API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits-rate-based)
+
+Cluster operators can set the CSI Sidecar `--kube-api-burst` (i.e. bucket size) and `--kube-api-qps` (i.e. bucket refill rate) parameters in order to fine-tune how strictly these containers throttle their queries towards the K8s API Server. 
+
+#### Further Reading
+- [Request throttling for the Amazon EC2 API | Amazon Web Services](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits-rate-based)
+- [Managing and monitoring API throttling in your workloads | Amazon Web Services](https://aws.amazon.com/blogs/mt/managing-monitoring-api-throttling-in-workloads/)
+- [Reference: kube-apiserver | Kubernetes](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)
+- [API Priority and Fairness | Kubernetes](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/)
+
+### Fine-tuning CSI sidecar scalability parameters
+
+In [aws-ebs-csi-driver v1.25.0](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG.md#v1250), we changed the following K8s CSI external sidecar parameters to more sensible defaults. See the summary section for an overview of how these parameters affect volume lifecycle management. 
+
+- `--worker-threads` (named `--workers` in [external-resizer](https://github.com/kubernetes-csi/external-resizer))
+- `--kube-api-burst`
+- `--kube-api-qps`
+- `--timeout`
+
+The AWS EBS CSI Driver provides a set of default values intended to balance performance while reducing the risk of reaching the default EC2 rate limits for API calls. 
+
+Cluster operators can increase the `--kube-api-qps` and `--kube-api-burst` of each sidecar to keep the state of Kubernetes objects more in sync with their associated AWS resources. 
+
+Cluster operators that need a greater throughput of volume operations should increase the associated sidecar's `worker-threads` value. When increased, the AWS Account's EC2 API limits may need to be raised to account for the increased rate of API calls.
+
+| Sidecar | Configuration Name | Description | EC2 API Calls Made By Driver                             |
+| ---- | ---- | ---- |----------------------------------------------------------|
+| [external-provisioner](https://github.com/kubernetes-csi/external-provisioner) | provisioner | Watches PersistentVolumeClaim objects and triggers CreateVolume/DeleteVolume | EC2 CreateVolume/DeleteVolume                            |
+| [external-attacher](https://github.com/kubernetes-csi/external-attacher) | attacher | Watches VolumeAttachment objects and triggers ControllerPublish/Unpublish | EC2 AttachVolume/DetachVolume, EC2 DescribeInstances     |
+| [external-resizer](https://github.com/kubernetes-csi/external-resizer) | resizer | Watches PersistentVolumeClaims objects and triggers controller side expansion operation | EC2 ModifyVolume, EC2 DescribeVolumesModifications       |
+| [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter) | snapshotter | Watches Snapshot CRD objects and triggers CreateSnapshot/DeleteSnapshot | EC2 CreateSnapshot/DeleteSnapshot, EC2 DescribeSnapshots |
+
+#### Sidecar Fine-tuning Examples
+
+Create a file named `example-ebs-csi-config-values.yaml` with the following yaml:
+
+```yaml
+sidecars:
+  provisioner:
+    additionalArgs:
+    - "--worker-threads=101"
+    - "--kube-api-burst=200"
+    - "--kube-api-qps=40.0"
+    - "--timeout=61s"
+  resizer:
+    additionalArgs:
+    - "--workers=101"
+```
+
+**Note: The external-resizer uses the `--workers` parameter instead of `--worker-threads`
+
+<details><summary>Self-managed Helm instructions</summary> 
+
+Pass in the configuration-values file when installing/upgrading `aws-ebs-csi-driver`
+
+```sh
+helm upgrade --install aws-ebs-csi-driver \
+--namespace kube-system \
+--values example-ebs-csi-config-values.yaml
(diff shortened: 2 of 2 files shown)


AndrewSirenko:
Should we link this in the changelog for v1.25.0, or is retro-active additions a bad changelog practice?

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+### Fine-tuning CSI sidecar scalability parameters
+
+In [aws-ebs-csi-driver v1.25.0](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/76ae539ae3129b408cd184c5cdfd1ffe476dab1d/CHANGELOG.md?plain=1#L35-L36), we changed the following K8s CSI external sidecar parameters to more sensible defaults. See the summary section for an overview of how these parameters affect volume lifecycle management. 
+
+- `worker-threads` (named `workers` for [external-resizer](https://github.com/kubernetes-csi/external-resizer))
+- `kube-api-burst`
+- `kube-api-qps`
+- `timeout`
+
+If your workloads need a higher throughput of volume operations, or you have higher AWS limits for any of the EC2 APIs in the `EC2 API Calls Made By Driver` column, you may increase consider increasing the associated sidecar's `worker-threads` value.
```
AndrewSirenko, line 55: Do we want to recommend setting `kube-api-burst` and `kube-api-qps` even higher than our new default for large-scale customers, so that the state of K8s objects is more in sync with AWS?

README.md
```diff
@@ -90,6 +90,7 @@ The EBS CSI Driver is compatible with Kubernetes versions v1.17+ and implements
 * [Driver Installation](docs/install.md)
 * [Driver Launch Options](docs/options.md)
 * [StorageClass Parameters](docs/parameters.md)
+* [Troubleshooting and frequently asked questions](docs/faq.md)
```
torredil, line 93: **Astronomical** np: `frequently asked questions` -> `Frequently Asked Questions` for consistency with the surrounding text.

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+ADDON-CONFIG-FILEPATH="./example-addon-config.yaml"
+
+aws eks create-addon \
+  --cluster-name "example-cluster" \
+  --addon-name "aws-ebs-csi-driver" \
+  --service-account-role-arn "arn:aws:iam::123456789012:role/EBSCSIDriverRole" \
+  --configuration-values "file://$ADDON-CONFIG-FILEPATH"
+```
+
+You can confirm that these arguments were set by describing am `ebs-csi-controller` pod and observing the following args under the relevant sidecar container:
```
torredil, line 118: np: `am` -> `a`

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+    additionalArgs:
+    - "--workers=101"
+```
+
+Pass in the add-on configuration-values file when creating your add-on: 
+
+```
+ADDON-CONFIG-FILEPATH="./example-addon-config.yaml"
+
+aws eks create-addon \
```
torredil, line 111: An example `aws eks update-addon` command would also be helpful

docs/faq.md
torredil, line 55: Fix `you may increase consider increasing the associated sidecar's`

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+
+Both the EC2 API and K8s CSI sidecars base their API throttling implementation off of the [token bucket algorithm](https://en.wikipedia.org/wiki/Token_bucket). Both the [Amazon EC2 API Reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits-rate-based) and the [K8s documentation definitions of qps and burst](https://kubernetes.io/docs/reference/config-api/apiserver-eventratelimit.v1alpha1/#eventratelimit-admission-k8s-io-v1alpha1-Limit) describe their throttling systems.  
+
+Let's look at an example from [Request throttling for the Amazon EC2 API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits-rate-based)
+
+<details><summary>Expand for excerpt:</summary>
+
+"With request rate limiting, you are throttled on the number of API requests you make. Each request that you make removes one token from the bucket. For example, the bucket size for _non-mutating_ (`Describe*`) API actions is 100 tokens, so you can make up to 100 `Describe*` requests in one second. If you exceed 100 requests in a second, you are throttled and the remaining requests within that second fail.
```
torredil, line 30: Suggest removing this excerpt from the FAQ section and simply linking to the relevant documentation instead.

As a good rule of thumb, strive to maintain one source of truth - over time, the excerpt above will change and keeping it up to date here is an additional maintainability burden. In the worst case, the excerpt here is not updated and thus the FAQ will have outdated / potentially misleading information - let's prevent that scenario from being possible.

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+	- This decreases the amount of premature timeouts for CSI RPC calls, which reduces the amount of replay EC2 API requests made by and waited for by the driver (at the cost of a longer delay during a real driver timeout (e.g. network blip leads to lost `ControllerPublishVolume` response))
+## EC2 and K8s CSI Sidecar Throttling Overview
+
+Both the EC2 API and K8s CSI sidecars base their API throttling implementation off of the [token bucket algorithm](https://en.wikipedia.org/wiki/Token_bucket). Both the [Amazon EC2 API Reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits-rate-based) and the [K8s documentation definitions of qps and burst](https://kubernetes.io/docs/reference/config-api/apiserver-eventratelimit.v1alpha1/#eventratelimit-admission-k8s-io-v1alpha1-Limit) describe their throttling systems.
```
torredil, line 24: I'm thinking this information may be out of scope for an FAQ doc. In spirit of minimalism and conciseness I would suggest removing it, but would be interested to see other opinions.

torredil (commented):
Few small nps and suggestions, thanks.
/lgtm

docs/faq.md
ConnorJC3: Strongly suggest removing this section entirely. It exists entirely to re-explain part of the AWS docs, and thus risks becoming out of date. Additionally, it makes assumptions about details like when the token refresh actually happens that may not be actually true.

docs/faq.md
ConnorJC3: +1, especially given it links the wrong k8s docs

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+
+You do not need to wait for the bucket to be completely full before you can make API requests. You can use tokens as they are added to the bucket. If you immediately use the refill tokens, the bucket does not reach its maximum capacity. For example, the bucket size for _console non-mutating actions_ is 100 tokens, and the refill rate is 10 tokens per second. If you deplete the bucket by making 100 API requests in a second, you can continue to make 10 API requests per second. The bucket can refill to the maximum capacity only if you make fewer than 10 API requests per second."</details>
+
+Analogous to the above example, cluster operators can set the CSI Sidecar `--kube-api-burst` (i.e. bucket size) and `--kube-api-qps` (i.e. refill rate) parameters in order to fine-tune how strictly these containers throttle their queries towards the K8s API Server. 
+
+#### Further Reading
+- AWS
+	- [Managing and monitoring API throttling in your workloads](https://aws.amazon.com/blogs/mt/managing-monitoring-api-throttling-in-workloads/)
+- K8s
```
ConnorJC3, line 41: Spell out "Kubernetes" here. Also, consider just removing the "AWS" and "K8s" headers.

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+	- [Reference: kube-apiserver | Kubernetes](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)
+	- [Rate Limiting in controller-runtime and client-go](https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/)
```
ConnorJC3, line 43: We should avoid linking to blog posts in our official docs if at all possible, and instead prefer the k8s official documentation.

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+	- [API Priority and Fairness | Kubernetes](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/)
+
+### Fine-tuning CSI sidecar scalability parameters
+
+In [aws-ebs-csi-driver v1.25.0](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/76ae539ae3129b408cd184c5cdfd1ffe476dab1d/CHANGELOG.md?plain=1#L35-L36), we changed the following K8s CSI external sidecar parameters to more sensible defaults. See the summary section for an overview of how these parameters affect volume lifecycle management.
```
ConnorJC3, line 48: Link by header, not line

i.e. use link
```
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG.md#v1250
```

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+- K8s
+	- [Reference: kube-apiserver | Kubernetes](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)
+	- [Rate Limiting in controller-runtime and client-go](https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/)
+	- [API Priority and Fairness | Kubernetes](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/)
+
+### Fine-tuning CSI sidecar scalability parameters
+
+In [aws-ebs-csi-driver v1.25.0](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/76ae539ae3129b408cd184c5cdfd1ffe476dab1d/CHANGELOG.md?plain=1#L35-L36), we changed the following K8s CSI external sidecar parameters to more sensible defaults. See the summary section for an overview of how these parameters affect volume lifecycle management. 
+
+- `worker-threads` (named `workers` for [external-resizer](https://github.com/kubernetes-csi/external-resizer))
```
ConnorJC3, line 50: "for" -> "in"

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+- `kube-api-burst`
+- `kube-api-qps`
+- `timeout`
```
ConnorJC3, line 53: These are CLI parameters, use `--timeout`, `--kube-api-qps`, etc

docs/faq.md
ConnorJC3, line 55: This is passive voice. Suggested rewording:

```
The AWS EBS CSI Driver provides a set of default values intended to balance performance while reducing the risk of reaching the default EC2 rate limits for API calls. Operators of clusters that need a greater throughput of volume operations can increase these values to increase the rate at which the EBS CSI Driver processes volumes. When increased, the AWS Account's EC2 API limits may need to be raised to account for the increased rate of API calls.

File: docs/faq.md [hunk 12]
```diff
@@ -1,5 +1,137 @@
...
+| [external-attacher](https://github.com/kubernetes-csi/external-attacher) | attacher | Watches VolumeAttachment objects and triggers ControllerPublish/Unpublish | EC2 AttachVolume/DetachVolume |
+| [external-resizer](https://github.com/kubernetes-csi/external-resizer) | resizer | Watches PersistentVolumeClaims objects and triggers controller side expansion operation | EC2 ModifyVolume, EC2 DescribeVolumesModifications |
+| [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter) | snapshotter | Watches Snapshot CRD objects and triggers CreateSnapshot/DeleteSnapshot | EC2 CreateSnapshot/DeleteSnapshot, EC2 DescribeSnapshots |
+
+#### Sidecar Fine-tuning Examples
+
+**Note: The external-resizer uses the `--workers` parameter instead of `--worker-threads`
+##### Helm
+
+Make the following diff in the helm chart's default [values.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml) file
```
ConnorJC3, line 69: Do not tell users to modify `values.yaml` directly, that is an anti-pattern and not compatible with most installation methods

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+| Sidecar | Configuration Name | Description | EC2 API Calls Made By Driver |
+| ---- | ---- | ---- | ---- |
+| [external-provisioner](https://github.com/kubernetes-csi/external-provisioner) | provisioner | Watches PersistentVolumeClaim objects and triggers CreateVolume/DeleteVolume | EC2 CreateVolume/DeleteVolume |
+| [external-attacher](https://github.com/kubernetes-csi/external-attacher) | attacher | Watches VolumeAttachment objects and triggers ControllerPublish/Unpublish | EC2 AttachVolume/DetachVolume |
+| [external-resizer](https://github.com/kubernetes-csi/external-resizer) | resizer | Watches PersistentVolumeClaims objects and triggers controller side expansion operation | EC2 ModifyVolume, EC2 DescribeVolumesModifications |
+| [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter) | snapshotter | Watches Snapshot CRD objects and triggers CreateSnapshot/DeleteSnapshot | EC2 CreateSnapshot/DeleteSnapshot, EC2 DescribeSnapshots |
+
+#### Sidecar Fine-tuning Examples
+
+**Note: The external-resizer uses the `--workers` parameter instead of `--worker-threads`
```
ConnorJC3, line 66: This is already noted above, do we really need to include it here again?

docs/faq.md
ConnorJC3: additional np: "You can confirm ..." -> "Confirm that..."

docs/faq.md
```diff
@@ -1,5 +1,137 @@
...
+
+| Sidecar | Configuration Name | Description | EC2 API Calls Made By Driver |
+| ---- | ---- | ---- | ---- |
+| [external-provisioner](https://github.com/kubernetes-csi/external-provisioner) | provisioner | Watches PersistentVolumeClaim objects and triggers CreateVolume/DeleteVolume | EC2 CreateVolume/DeleteVolume |
+| [external-attacher](https://github.com/kubernetes-csi/external-attacher) | attacher | Watches VolumeAttachment objects and triggers ControllerPublish/Unpublish | EC2 AttachVolume/DetachVolume |
+| [external-resizer](https://github.com/kubernetes-csi/external-resizer) | resizer | Watches PersistentVolumeClaims objects and triggers controller side expansion operation | EC2 ModifyVolume, EC2 DescribeVolumesModifications |
+| [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter) | snapshotter | Watches Snapshot CRD objects and triggers CreateSnapshot/DeleteSnapshot | EC2 CreateSnapshot/DeleteSnapshot, EC2 DescribeSnapshots |
```
ConnorJC3, line 62: Some of these use `/`, some use `,`, be consistent

Also, are we not going to include APIs like `DescribeVolumes`? or `DescribeInstances`?

docs/faq.md
```diff
@@ -1,5 +1,137 @@
 # Frequently Asked Questions
 
+## Driver performance for large-scale clusters
+
+### Summary of v1.25 scalability improvements
```
ConnorJC3, line 5: This is such a large "question" should it have a dedicated `scalability.md` instead of being in the FAQ?

README.md
ConnorJC3: additional np: Remove "troubleshooting", as the doc is purely an FAQ

ConnorJC3 (changes requested):

docs/faq.md
AndrewSirenko: Today, because we're adding our FAQ to our README links, I think we should keep this info in our FAQ.

I think this question should be re-raised once our FAQ has enough content (and enough Cxs have discovered it) to warrant splitting parts of it off.

docs/faq.md
AndrewSirenko: > especially given it links the wrong k8s docs

I disagree that these are 'wrong' docs. 

Those k8s docs are the only ones I could find that clearly explain what qps and burst mean in the context of Kubernetes, and therefore a better resource than docs like [kubelet | Kubernetes](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)

AndrewSirenko: > I'm thinking this information may be out of scope for an FAQ doc. In spirit of minimalism and conciseness I would suggest removing it, but would be interested to see other opinions.

I disagree. I think that this section, when not expanded, is small enough to be helpful context rather than out of scope. I think it is required reading for understanding how to tune the sidecar parameters and understand one of the root causes of scalability issues. 

Furthermore, I'd rather have too much info than too little, since anyone looking at an FAQ is in the process of diving deep, and can move onto the next heading if desired. 

However, if either @wmesard, @rdpsin, or Manu agree with @torredil, I can commit to taking this section out.

ConnorJC3: Those docs are about the API Server rate limiting, the client-side throttling on the sidecars happens to work similarly by accident.

docs/faq.md
AndrewSirenko: The reason I included the excerpt is that it was the most helpful explanation of the token bucket system that I've seen, and this system is one that is shared between AWS and K8s. The explanation of the system won't go out of date (But I agree that mentions of specific API names might go out of date and those at a minimum should be removed).

However, I can see an argument for: "We can trust the Cx to figure out how the token bucket algorithm works, especially if we hand them a link to the relevant documentation"

Therefore I'll delete the excerpt.

docs/faq.md
AndrewSirenko: I argue yes, in cx doesn't read both places. I believe the repetition is worth preventing [the commonly made mistake](https://github.com/openshift/csi-operator/pull/72#issuecomment-1821145662).

docs/faq.md
AndrewSirenko: Instead, should I give them the appropriate set commands?

docs/faq.md
AndrewSirenko: What I mean is that those API Server rate limiting docs featured the best explanation of what `burst` and `qps` mean in the context of K8s (because the explanation on the sidecar's documentation had some ambiguity on the relationship between burst and qps)

But I see how linking the documentation for a wholly different part of k8s could also be misleading. Will take out.

docs/faq.md
AndrewSirenko: > Some of these use /, some use ,, be consistent

The `/` denotes symmetrical operations (ie ControllerPublish leads to AttachVolume being called), the `,` denotes multiple different EC2 calls being used in same CSI call. I think this distinction is important, but I'm open to a more intuitive system than `/` vs `,` since it misled you. 

> Also...

I was not going to include DescribeVolumes because we already batch that call (and therefore calls of DescribeVolumes don't scale with calls to CSI CreateVolume) 

I should include DescribeInstances for the Attach/Detach path, yes.

ConnorJC3 (approved):
/lgtm

PR #1895, opened 2024-01-12
Pre-stop LCH enhancements

torredil:

This PR:
- Adds event handlers for add / update events associated with `VolumeAttachment` objects.
- Updates csi-node clusterrole RBAC rules for volume attachments to include `get`, `watch`, and `update`.

see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1665 for context.

**What testing is done?** 

`make test`


charts/aws-ebs-csi-driver/templates/clusterrole-csi-node.yaml
```diff
@@ -11,4 +11,4 @@ rules:
     verbs: ["get", "patch"]
   - apiGroups: ["storage.k8s.io"]
     resources: ["volumeattachments"]
-    verbs: ["list"]
+    verbs: ["get", "list", "watch", "update"]
```
ConnorJC3, line 14: Remove `"update"`, discussed offline and concluded it's not required.

Otherwise lgtm

cmd/hooks/prestop.go
```diff
@@ -28,33 +27,30 @@ No unnecessary delay is added to the termination workflow, as the PreStop hook l
 If the PreStop hook hangs during its execution, the driver node pod will be forcefully terminated after 32 seconds (30s terminationGracePeriodSeconds + 2 second grace period extension from Kubelet).
```
AndrewSirenko, line 27: I don't think this comment will always be accurate post PR, because terminationGracePeriodSeconds might not always be 30s? Maybe change to "terminationGracePeriodSeconds in the pod specification define the upper bound for timeout".

AndrewSirenko (approved):
lgtm other than Connor's change + possibly editing that comment.

cmd/hooks/prestop.go
ConnorJC3: +1, consider adding the words something like "by default" at the end

charts/aws-ebs-csi-driver/templates/clusterrole-csi-node.yaml
torredil: Fixed ✅

cmd/hooks/prestop.go
torredil: Good call, updated comment

AndrewSirenko (approved):

PR #1900, opened 2024-01-19
Bump go/sidecar dependencies ahead of v1.27.0

AndrewSirenko:
Dependencies

Bump go/sidecar dependencies ahead of v1.27.0 (notably volume-modifier-for-k8s sidecar image)

Note: We had to bump container-storage-interface/spec to v1.9.0 due to it being a dependency of volume modifier v0.2.0. Therefore, this PR includes a stub implementation of ControllerModifyVolume in `controller.go`. 

**What testing is done?** 
`make update`, `make verify`


torredil (approved):
Thanks @AndrewSirenko
/lgtm

PR #1902, opened 2024-01-22
Bootstrap buildx builder in prow.sh

ConnorJC3:

Bug fix

`prow.sh` doesn't bootstrap the builder, which causes a race condition when building images in parallel

**What testing is done?** 

Successful prow run: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/aws-ebs-csi-driver-push-images/1749495304584957952


AndrewSirenko (approved):
Thanks for the fix.

Issue #1911, opened 2024-01-25
Allow modifyVolumeRequestHandlerTimeout to be configurable


andrewcharlton:
**Feature Request**

**Issue**
When modifying our volumes, we change both the size and the iops/throughput annotations together so that we only perform a single volume modification call to AWS.  However, we have seen a race condition where the iops/throughput gets modified by itself first, and then the volume is unable to resize because the volume is now in a 6 hour window.

**Proposed Solution**
There is a currently a hard coded 2 second window (the driver's `modifyVolumeRequestHandlerTimeout`) during which both size and iops/throughput modification requests must be received for them to be processed together. It is more important for us that they be processed together (and so avoid the 6 hour windows) than the speed at which they are processed, so we would like to extend this window.

I propose that this becomes an option in the `ControllerOptions` instead, defaulting to 2 seconds, so that users can tune this setting.

If this is reasonable, please let me know and I'll put a PR together.  Thanks!

ConnorJC3:
Hi @andrewcharlton , we talked about this and you're good to go to submit a PR to make this change (note that you, or your employer if you're contributing on their behalf, will need to sign the standard Kubernetes CLA).

We'll get the PR reviewed as soon as time allows, thanks for contributing to the EBS CSI Driver!

AndrewSirenko:
/close 

Due to fix being merged. Thanks for putting the PR together Andrew!

andrewcharlton:
> Due to fix being merged. Thanks for putting the PR together Andrew!

Thank you all for the speedy reply/review!

PR #1915, opened 2024-01-30
Change modifyVolumeRequestHandlerTimeout to a configurable option

andrewcharlton:
New feature

This PR changes the `modifyVolumeRequestHandlerTimeout`, which was previously hardcoded to 2 seconds, into an option that the user can specify. If it is not specified, this should be a complete no-op and default to 2 seconds as before.

The `modifyVolumeRequestHandlerTimeout` determines how long the controller will wait for additional volume modification calls before executing them and modifying the volume. When changing both the size and IOPs/Throughput of a volume simultaneously there are two calls made and they need to be coalesced together within this window for a single call to AWS to be made.  We have seen cases where the 2 seconds is not long enough for this to happen, and only a single one of those changes is made so we would like to extend the window for our use case, where it is more important that the change be atomic than it's latency.

Issue discussing this: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1911

**What testing is done?** 

Added unit test for the new option. Updated coalesce tests to ensure they use the value from the option.  

We have a fork with a larger timeout window running in our environment successfully.


adolsalamanca:
Hello folks,
One important thing to note about this change:

it's very important to make sure that both `resizer` and `modifier` timeout args are higher than the specified `modify-volume-request-handler-timeout`, otherwise users might have contex timeout errors on requests from those to gRPC endpoints in csi-controller.

Timeouts both in `resizer` and `modifier` are by default 10s.

ConnorJC3 (approved):
/lgtm

andrewcharlton:
> Maybe worth adding a comment about this to the code, or to the docs?

Thanks, I've added a comment in both the docs, and the descriptor for the flag.  I added an additional testcase that I'd missed for the flag as well.  Hope that's ok!

torredil (approved):
Thanks @andrewcharlton
/lgtm

ConnorJC3 (approved):
/lgtm

AndrewSirenko (approved):
Thanks for this

PR #1917, opened 2024-02-01
Add option to disable discovery of attached volumes from instance metadata

jsafrane:
/kind bug

Add `--disable-volume-attach-limit-from-metadata` to disable counting of attached volumes from instance metadata.

The option name and its description is not great, I know. Feel free to suggest a better one.

Instance metadata contains block volumes that were attached at the time the instance was started and then it's never updated. In case a node was rebooted with CSI volumes attached, the metadata includes these CSI volumes and the CSI driver "reserves" an attachment slot for them forever, even when they are detached later.

The option disables counting of attached volumes when computing the attach limit, because it's not able to distinguish a CSI volume (should not have a reserved attachment, because scheduler will count it on its own) and a root disk / volume attached by the system admin (should have a reserved attachment).

**What testing is done?** 
Manually I tested that `--disable-volume-attach-limit-from-metadata` ignored a second volume attached to a node.

Fixes: #1808

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -67,6 +67,7 @@ func main() {
 		driver.WithExtraVolumeTags(options.ControllerOptions.ExtraVolumeTags),
 		driver.WithMode(options.DriverMode),
 		driver.WithVolumeAttachLimit(options.NodeOptions.VolumeAttachLimit),
+		driver.WithVolumeAttachLimitFromMetadata(options.NodeOptions.DisableVolumeAttachLimitFromMetadata),
 		driver.WithKubernetesClusterID(options.ControllerOptions.KubernetesClusterID),
 		driver.WithAwsSdkDebugLog(options.ControllerOptions.AwsSdkDebugLog),
 		driver.WithWarnOnInvalidTag(options.ControllerOptions.WarnOnInvalidTag),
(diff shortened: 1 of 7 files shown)


gnufied:
I am just trying to capture and summarize discussions so far.

@zerkms You mentioned does this count ENIs -  https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1843#issuecomment-1832757803 and it does. ENIs are counted separately so I am not sure if https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1843/files#diff-4f4d9c2e0ec5c4d3b4ac2d79abd70ee21bacff65ba05c059396921996dbf6607 broke anything there.


@kon-angelo - is it typical for gardner users to attach same number of additional volumes to all nodes in the cluster or the number of additional volumes is different on each node?

FWIW - the default behaviour of driver is incorrect and causes nodes to report far lower number than actually available on the node. We are kinda working with imperfect tools here, so we are stuck between rock and a hard place. 


I wonder if a configuration option that allows users to configure number of additional volumes they have attached to a node is the right solution, because that would allow other clusters which attach bunch of volumes out-of-band to specify a number and that number will be appropriately deducted.

jsafrane:
An usable alternative would be to explicitly reserve some attachment via something like `--out-of-band-volumes=6`. Still, it's hard for me to describe interaction between `--volume-attach-limit` and `--out-of-band-volumes`.

gnufied:
How about `--attached-volume-count` ? That perhaps explains the interaction more clearly. and we can then describe that, it will subtract from `volumeLimits` ?

jsafrane:
/hold
I created https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1919 with `--reserved-volume-attachments=X`.

torredil:
Merging in #1919
/close

PR #1919, opened 2024-02-02
Add reserved-volume-attachments

jsafrane:
This is an alternative approach to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1917

/kind bug

Introduce `--reserved-volume-attachments` cmd line option that disables loading list of volume attachments from instance metadata.

Instance metadata has volume attachments that were present on instance boot, which may be different than the current nr. of attachments. In addition, it may include CSI volumes that were not drained before reboot, and thus the value may be completely wrong.

`--reserved-volume-attachments` simply reserves given nr. of attachments for any system disks. Typically it should be 1 for the root volume.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1917 introduces just boolean value that would reserve 1 attachment for the root disk, this PR allows users to set nr. of such attachments.

Note that the default value is `-1`, i.e. the current heuristics. Existing CSI driver deployments won't see any change, until they explicitly specify `--reserved-volume-attachments=X`.

**What testing is done?** 
Manually, it indeed ignored multiple attached volumes and reported 26 available attachments with `--reserved-volume-attachments=1`  and 17 with `--reserved-volume-attachments=10` (both on `m6i.xlarge`).


cmd/options/node_options.go
```diff
@@ -31,8 +31,16 @@ type NodeOptions struct {
...
+	ReservedVolumeAttachments int
 }
 
 func (o *NodeOptions) AddFlags(fs *flag.FlagSet) {
-	fs.Int64Var(&o.VolumeAttachLimit, "volume-attach-limit", -1, "Value for the maximum number of volumes attachable per node. If specified, the limit applies to all nodes. If not specified, the value is approximated from the instance type.")
+	fs.Int64Var(&o.VolumeAttachLimit, "volume-attach-limit", -1, "Value for the maximum number of volumes attachable per node. If specified, the limit applies to all nodes and overrides --reserved-volume-attachments. If not specified, the value is approximated from the instance type.")
+	fs.IntVar(&o.ReservedVolumeAttachments, "reserved-volume-attachments", -1, "Number of volume attachments reserved for system use. Not used when --volume-attach-limit is specified. The total amount of volume attachments for a node is computed as: <nr. of attachments for corresponding instance type> - <number of NICs, if relevant to the instance type> - <reserved-volume-attachments value>. When -1, the amount of reserved attachments is loaded from instance metadata that captured state at node boot and may include not only system disks but also CSI volumes.")
```
gnufied, line 45: So you can't specify both. Was there any reason for doing that, other than perhaps both of them being redundant? I mean we could still allow specifying both - in which case limit is determined by `volumeAttachLimit - reservedVolumeAttachment` ?

AndrewSirenko:
Thank you @gnufied and @jsafrane for taking action on this. Apologies for the communication delay, but as of this morning, EBS folks are aligned with this PR's approach and want to merge this short-term fix ahead of v1.28.0.

CC @torredil & @ConnorJC3.

I will personally manually sanity-test this PR by Sunday, but from diff alone it looks great to me. Especially the clarity of documentation in node_options. Thank you again.

cmd/options/node_options.go
ConnorJC3: I think we should consider throwing an error if both are specified - there's no reason to ever use both so the presence of both options strongly indicates an invalid config and/or confused user.

ConnorJC3 (commented):
I think this is the superior approach to #1917 - the configuration is clear and more flexible for usecases like gardener (which now can pass in the exact number of extra volumes they intend to attach instead of relying on our buggy heuristic), and leaves open the option of changing the default behavior in a future release.

/lgtm

AndrewSirenko (approved):

jsafrane:
I added a commit that allows only one of --volume-attach-limit and --reserved-volume-attachments.

AndrewSirenko (approved):

torredil:
Thank you @jsafrane
/lgtm
/approve

PR #1931, opened 2024-02-13
Report zone via well-known topology key in NodeGetInfo

ConnorJC3:

Both?

See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/729#issuecomment-1942026577

**What testing is done?** 

- Manually tested locally by attaching volumes created by both this and a previous version of the driver, and ensuring they both attached
- e2e (see CI)

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -589,7 +589,8 @@ func (d *nodeService) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
 	zone := d.metadata.GetAvailabilityZone()
 
 	segments := map[string]string{
-		TopologyKey: zone,
+		TopologyKey:          zone,
+		WellKnownTopologyKey: zone,
 	}
 
 	outpostArn := d.metadata.GetOutpostArn()
--- a/pkg/driver/node_test.go
+++ b/pkg/driver/node_test.go
@@ -2331,6 +2331,9 @@ func TestNodeGetInfo(t *testing.T) {
 			if at.Segments[TopologyKey] != tc.availabilityZone {
 				t.Fatalf("Expected topology %q, got %q", tc.availabilityZone, at.Segments[TopologyKey])
 			}
+			if at.Segments[WellKnownTopologyKey] != tc.availabilityZone {
+				t.Fatalf("Expected (well-known) topology %q, got %q", tc.availabilityZone, at.Segments[WellKnownTopologyKey])
+			}
 
 			if at.Segments[AwsAccountIDKey] != tc.outpostArn.AccountID {
 				t.Fatalf("Expected AwsAccountId %q, got %q", tc.outpostArn.AccountID, at.Segments[AwsAccountIDKey])


AndrewSirenko (approved):
Thanks for starting to resolve this paper-cut. 

Had a small nit about renaming `WellKnownTopologyKey` but realized it is CO-agnostic and has existed for 3 years already.

torredil (changes requested):
Hey, so I took some time to review the history here and my current understanding is that this PR is not necessary - ill explain.

First, ill start by laying out some context that both the author of this PR and reviewers already know, but will be helpful to future maintainers and external contributors to follow this discussion.

There are 2 places where topology labels come into the picture:

- Controller service -  creates PVs with a topology label.
- Node service - adds a topology label on the node.

Kubernetes or a CO will then use that topology information to ensure that a volume is accessible from a given node when scheduling workloads (wording directly from [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md)).

That is to say, we need to ensure that dynamically provisioned PVs have a label that nodes can satisfy, if the label keys don't match the volumes will not be scheduled.

-----

Back in circa 2021, maintainers of this project tried to switch to the well-known label (which was not so well-known at the time) and made a [mistake](https://github.com/kubernetes/kubernetes/issues/102248#issuecomment-847285023) in adding both the custom label AND the well-known label to provisioned PVs **in the controller side**.  An important detail here as pointed out by [Michelle](https://github.com/kubernetes/kubernetes/issues/102248#issuecomment-847234805) is that the node selector `matchExpression` is ANDed, which I understand to mean that in order to schedule those volumes, both labels (the custom, and the well-known) **must** exist on the node.

Now the reason why this PR is not necessary: CCM already adds the well-known label to all nodes -- it does not make sense for the driver to override that. I have also manually tested and confirmed that to be true.

In short, we can rely on CCM / Kubelet to add the well-known label, and the important bit here over on the Node is that we must continue adding the custom label, otherwise PVs may not be scheduled on nodes which do not have the custom label and `ebs-plugin` node is the only component adding that label on nodes.

------

Given the above, I believe we can safely proceed to switching the label over on the controller to the well-known label if on Kubernetes. To illustrate:

### Node Topology:

- Custom label: `topology.kubernetes.io/zone`-- applied by CSI driver
- Well-known label: `topology.ebs.csi.aws.com/zone` -- applied by CCM / Kubelet

### PV Topology:

`topology.ebs.csi.aws.com/zone` -- applied by CSI driver 

On PVs created by newer driver:

`topology.kubernetes.io/zone`  -- applied by CSI driver if on Kubernetes
`topology.ebs.csi.aws.com/zone` -- applied by CSI driver if not on Kubernetes

That should fix all the issues and enable scaling from 0, which is explained in detail [here](https://github.com/kubernetes/autoscaler/issues/3845). Please let me know if I missed anything and especially any compatibility concerns.

torredil (approved):
Suggest adding the following unit test to [node_test.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b32f203d5d772b9b242c1abdeaf77c9cb69febe3/pkg/driver/node_test.go#L2331)

```
if at.Segments[WellKnownTopologyKey] != tc.availabilityZone {
	t.Fatalf("Expected well-known topology %q, got %q", tc.availabilityZone, at.Segments[WellKnownTopologyKey])
}
```

/lgtm
/approve
/hold

ConnorJC3:
> Suggest adding the following unit test to [node_test.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b32f203d5d772b9b242c1abdeaf77c9cb69febe3/pkg/driver/node_test.go#L2331)

Missed that, thanks! Ready for re-review

AndrewSirenko (approved):

PR #1932, opened 2024-02-13
Add code coverage report

torredil:

This PR makes modifications to `make test` to generate a test coverage report that explicitly excludes all generated mock files, giving contributors and maintainers more insight into code coverage.

In place of using the `grep` tech to filter out mock files, I initially attempted to split out mocked code into its own separate package. However, I quickly discovered that this is not possible due to a circular dependencies issue. To prevent the import cycle we'd have to bring in the tests over to the mock package as well - this is undesirable as a) it would effectively break the code coverage measurement altogether and b) not a standard Go practice.

**What testing is done?** 

```
make test
```
Screenshot 2024-02-13 at 6 24 20 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,9 @@ bin
 # Test binary, build with `go test -c`
 *.test
 
-# Output of the go coverage tool, specifically when used with LiteIDE
+# Output of the go coverage tool
 *.out
+coverage.html
 
 # Vagrant
 .vagrant
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,13 @@ clean:
 test:
 	go test -v -race ./cmd/... ./pkg/...
 
+.PHONY: test/coverage
+test/coverage:
+	go test -coverprofile=cover.out ./cmd/... ./pkg/...
+	grep -v "mock" cover.out > filtered_cover.out
+	go tool cover -html=filtered_cover.out -o coverage.html
+	rm cover.out filtered_cover.out
+
 .PHONY: test-sanity
 test-sanity:
 	go test -v -race ./tests/sanity/...
--- a/docs/makefile.md
+++ b/docs/makefile.md
@@ -88,6 +88,10 @@ Build and push a multi-arch image of the driver based on the OSes in `ALL_OS`, a
 
 Run all unit tests with race condition checking enabled.
 
+### `make test/coverage`
+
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -74,7 +74,10 @@ clean:
 
 .PHONY: test
 test:
```
AndrewSirenko, line 76: I would prefer a separate target (perhaps `make test/coverage`?) for generating the test coverage profile. I believe `make test` should be reserved for just running the unit tests.

AndrewSirenko (changes requested):

Makefile
torredil: Added new target as requested: `make test/coverage`

.gitignore
```diff
@@ -9,8 +9,9 @@ bin
 # Test binary, build with `go test -c`
```
AndrewSirenko, line 9: Top level comment: Consider adding a section to [local-development section of our makefile documentation](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/makefile.md#local-development). Perhaps something like:

```md
### `make test/coverage`

Outputs a filtered version of the each package's unit test coverage profiling via go's coverage tool to a local `coverage.html` file.
```

AndrewSirenko (approved):
Successfully ran on Linux. Left one more comment about adding a section to our makefile documentation. Thanks for this!

Makefile
AndrewSirenko: Oh I just realized while `test/coverage` is in line with what we use to separate sub-commands of other make targets (e.g. `e2e/cluster`, the `test-sanity` target now sticks out. I think this is fine though (if anything we should consider changing `test-sanity` to `test/sanity` in a separate PR).

Makefile
```diff
@@ -76,6 +76,13 @@ clean:
...
 	go test -v -race ./cmd/... ./pkg/...
 
+.PHONY: test/coverage
+test/coverage:
+	go test -v -race -coverprofile=cover.out ./cmd/... ./pkg/...
```
ConnorJC3, line 81: np: could probably ditch the `-v` and `-race` as this is separate from `make test`
/lgtm

.gitignore
torredil: Thanks - added ✅

Makefile
torredil: Thanks - changed ✅

AndrewSirenko (approved):

PR #1933, opened 2024-02-14
Fix Three Papercuts

ConnorJC3:

Bug fix

---

Fixes three small papercuts (each in their own commit):

#### Fix kops clusters becoming stuck when local KUBECONFIG is invalid

When the local kubeconfig represents a cluster that no longer exists, patching (and thus creating) a kops-based cluster will get stuck. This fixes that by explcitly setting the config for patch operations to `/dev/null`

#### Restrain `shfmt` to `hack/` directory

`shfmt` was trying to modify scripts in the `venv` in `bin/`, sometimes causing `make verify/update` to fail locally.

#### Attempt to delete Prow clusters that fail to create

We already delete clusters where the e2e tests fail (as of #1856), but leave behind clusters where creation failed.

---

**What testing is done?** 

Manual/CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ update/gomod:
 
 .PHONY: update/shfmt
 update/shfmt: bin/shfmt
-	./bin/shfmt -w -i 2 -d .
+	./bin/shfmt -w -i 2 -d ./hack/
 
 ## Verifiers
 # Linters and similar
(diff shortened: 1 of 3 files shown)


AndrewSirenko (approved):
Lgtm now, thanks for the papercut fixes!

torredil (approved):
/approve

PR #1939, opened 2024-02-20
[WIP] Release helm chart v2.28.1 with node.reservedVolumeAttachments

AndrewSirenko:
Helm

TODO

**What testing is done?** 
TODO

```
helm upgrade \
  --namespace kube-system \
  ./charts/aws-ebs-csi-driver \
  --set node.reservedVolumeAttachments=2

kubectl describe -n kube-system pod ebs-csi-node-jz4vg
```

```
Controlled By:  DaemonSet/ebs-csi-node
Containers:
  ebs-plugin:
    Container ID:  containerd://33f577a5771588efce2afeaddd145772eba537883f0aa32b9c7a068aaa97115c
    Image:         public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver:v1.28.0
    Image ID:      public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver@sha256:2aa362f4c40746c4318938b2f3afc279aa799fb341025e1ddbd04a5893619d66
    Port:          9808/TCP
    Host Port:     0/TCP
    Args:
      node
      --endpoint=$(CSI_ENDPOINT)
      --reserved-volume-attachments=2
      --logging-format=text
      --v=2
    State:          Running
```


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -377,7 +378,13 @@ node:
   # Enable the linux daemonset creation
   enableLinux: true
   enableWindows: false
-  # The "maximum number of attachable volumes" per node
+  # Number of volume attachments reserved for system use.
+  # Not used when --volume-attach-limit is specified.
```
ConnorJC3, line 382: `--volume-attach-limit` -> `node.volumeAttachLimit`

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -377,7 +378,13 @@ node:
...
+  # The total amount of volume attachments for a node is computed as:
```
ConnorJC3, line 383: Remove this comment and all other comment lines below it, incorrect on non-nitro instances and not necessary

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -377,7 +378,13 @@ node:
   # Enable the linux daemonset creation
   enableLinux: true
   enableWindows: false
-  # The "maximum number of attachable volumes" per node
+  # Number of volume attachments reserved for system use.
```
ConnorJC3, line 381: np: remove periods `.` from end of line, not used elsewhere in `values.yaml`

charts/aws-ebs-csi-driver/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Helm chart
+## v2.28.1
+* Add `reservedVolumeAttachments` to expose ability to override heuristic-determined reserved attachments via  `--reserved-volume-attachments` CLI option from [PR #1919](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1919) through helm ([#1939](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1939), [@AndrewSirenko](https://github.com/AndrewSirenko))
```
ConnorJC3, line 3: np: `helm` -> `Helm`

This is passive voice, reword as
```
Add `reservedVolumeAttachments` parameter that overrides heuristic-determined reserved attachments via ...
```

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -332,6 +332,7 @@ node:
   loggingFormat: text
   logLevel: 2
   priorityClassName:
+  additionalArgs: []
```
torredil, line 335: Question: why expose this parameter, given it introduces the following overhead:

> If i specify a value via additionalArgs and also through its dedicated helm parameter, what takes precedence?

Will chart changes require additional guards for every single parameter, or do we already have a reliable way to determine what the behavior is?

If the only reason for adding this is consistency with controller parameters, or add-on schema, I think removing the invalid add-on exposed parameter should be considered as an alternative to adding this parameter.

Besides the question above, this PR lgtm.

charts/aws-ebs-csi-driver/values.yaml
AndrewSirenko: Would it be helpful to point to cmd/options/node_options.go for a better explanation?

charts/aws-ebs-csi-driver/values.yaml
AndrewSirenko: I think consistency with all other containers is a valid enough reason. Having this parameter would have prevented the need for this re-release, and it is already advertised in add-on schema (which we cannot remove because that is a customer facing change, even if it currently doesn't do anything).

ConnorJC3: +1 to exposing, there's no reason for this container to be special and not allow specifying `additionalArgs` - it's a weird and inconsistent user experience.

charts/aws-ebs-csi-driver/values.yaml
ConnorJC3: No, docs should never point to code. Users do not need to know how the math is done internally, which may change from version to version and instance to instance, only what the parameter does.

AndrewSirenko: Alright, but is there a way we can advertise that setting this value will lead to driver not looking at IMDS for calculating CSI Node allocatable, which is the bigger reasoning behind this option? 

Or would you also define that as an implementation detail. 

Trying to prevent the footgun or setting this parameter without realizing the underlying change in CSI Node allocatable calculation.

ConnorJC3 (approved):
/lgtm

torredil:
I disagree with adding a global passthrough parameter to the node container for the sake of consistency and invalid add-on schema configurations, but commit to proceeding with the changes made in this PR.

/approve

PR #1942, opened 2024-02-21
Fix slow createDisk and batcher unit tests

AndrewSirenko:
Testing

~65s speed up for running `make test` via the following slow unit tests: 
- batcher_test.go: From 15s to 0.250s 
  - Lower timeouts still functionally test code
- Cloud TestCreateDisk: "fail: Volume is not ready to use, volume stuck in creating status and controller context deadline exceeded" subtest: From 60s to 5s 
  - Simulate a controller context deadline in the test. 
  - The whole of TestCreateDisk still needs some work such that unit tests don't sleep serially. Will work on this in a separate PR. 

**What testing is done?** 

Manual testing to make sure batcher and TestCreateDisk unit tests still fail when their implementations break.

`make test`, `make verify`

`go test -race -count=1000` on cloud and batcher packages. 

Measuring `make test` via `time` command:
`make test` pre PR:  2:29.92 
`make test` post PR: 1:25.17

### DIFF
--- a/pkg/batcher/batcher_test.go
+++ b/pkg/batcher/batcher_test.go
@@ -7,6 +7,11 @@ import (
 	"time"
 )
 
+const (
+	defaultMaxDelay = 50 * time.Millisecond
+	slowMaxDelay    = 5 * defaultMaxDelay
+)
+
 func mockExecution(inputs []string) (map[string]string, error) {
 	results := make(map[string]string)
 	for _, input := range inputs {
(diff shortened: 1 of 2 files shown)


torredil (commented):
Pending manual tests to ensure that 50 milliseconds is a sufficiently long delay such that flakes are not a concern, but otherwise the changes in this PR lgtm - thanks for the optimization @AndrewSirenko 👍 

/lgtm

PR #1947, opened 2024-02-26
Batch EC2 DescribeInstances calls

AndrewSirenko:
Feature

Coalesces EC2 DescribeInstances calls across ControllerPublishVolume/ControllerUnpublishVolume RPCs. 

This decreases the likelihood of being exceeding one's [Non-mutating API request token limit](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits) (ie being throttled for making too many`Describe*` calls).

**What testing is done?** 
Manual testing on statefulset to check that DescribeInstances calls were batched across RPCs

CI

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -143,6 +143,8 @@ const (
 const (
 	volumeIDBatcher batcherType = iota
 	volumeTagBatcher
+
+	batchDescribeTimeout = 30 * time.Second
 )
 
 var (
@@ -245,7 +247,9 @@ type batcherType int
 
 // batcherManager maintains a collection of batchers for different types of tasks.
 type batcherManager struct {
-	batchers map[batcherType]*batcher.Batcher[string, *ec2.Volume]
+	volumeIDBatcher   *batcher.Batcher[string, *ec2.Volume]
+	volumeTagBatcher  *batcher.Batcher[string, *ec2.Volume]
+	instanceIDBatcher *batcher.Batcher[string, *ec2.Instance]
 }
 
 type cloud struct {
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -684,6 +680,61 @@ func (c *cloud) DeleteDisk(ctx context.Context, volumeID string) (bool, error) {
...
+	b := c.bm.instanceIDBatcher
+	b.AddTask(task, ch)
+
+	r := <-ch
+
+	if r.Err != nil {
+		return nil, r.Err
+	}
+	if r.Result == nil {
+		return nil, fmt.Errorf("batchDescribeInstances: no instance found %s", task) // TODO Q: Should this be Cloud's ErrNotFound instead?
```
torredil, line 733: > TODO Q: Should this be Cloud's ErrNotFound instead?

Yes +1 to this, that error type is more appropriate

pkg/cloud/cloud.go
```diff
@@ -1190,16 +1237,33 @@ func (c *cloud) getInstance(ctx context.Context, nodeID string) (*ec2.Instance,
...
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO Q should I bring these errors outside of if because they apply to batched and unbatched versions of this?
+		if l := len(instances); l > 1 {
+			return nil, fmt.Errorf("found %d instances with ID %q", l, nodeID)
+		} else if l < 1 {
+			return nil, ErrNotFound
+		}
```
torredil, line 1261: The batcher is designed to execute multiple tasks as a batch and then distribute each individual result back to the respective caller - each task in the batch has its own result channel, and each channel is intended to receive exactly one result corresponding to its task.

We could add the extra check for sanity but I believe the code as written is resilient and is consistent with `describeVolumes`, which is battle tested.

torredil (commented):
/lgtm pending scalability tests before approval

pkg/cloud/cloud.go
AndrewSirenko: I will change `batchDescribeVolumes` as well then to be consistent.

AndrewSirenko:
/hold 

Waiting on scalability tests.

AndrewSirenko:
Settled on 300ms instanceIDBatcher maxDelay by running scalability tests. We found that this value balanced `minimizing amount of DescribeInstances EC2 calls` and worst-case ControllerPublishVolume latency. 

The table below shows total amount DescribeInstances API Calls made when attaching and detaching x volumes in a scalability test. 



| | Unbatched | 150ms | 300ms | 500ms | 
|---|--- |--- |--- |---|
| 100 volumes| 254| 116| 99| 62|
| 2500 volumes |6067|2298|1616|1312 |

torredil (approved):
/lgtm

Issue #1945, opened 2024-02-26
Delay before removing the startup taint from ebs


vivekskrishna:
/kind bug

EBS CSI driver currently supports removing a taints on new ndoes once it has started running sucessfully(implemented via https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1581)

**What you expected to happen?**
I could see that the taint is being removed upon csi driver pod running in that ndoe but current code removes it just before new node service is registered in func newNodeService. But it can be seen that still sometimes we see the issue mentioned in https://github.com/kubernetes/kubernetes/issues/95911 is seen due to which pods are assigned to new ndoes even when volume mount limit will exceed the capacity of new node.

**How to reproduce it (as minimally and precisely as possible)?**
Easy way to reproduce this is to create a pod with 26 ebs mounts and try to provision it on a node which supports say only 25(t3 instance type for example). It can be seen that when this si ried using karpenter dependening on race condition this pod will eventually get scheduled onto a newly spun up t3 instance node.

**Anything else we need to know?**:
This can potentially be fixed if we introduce a sleep initially in removeTaintInBackground before it proceeds to remove the taint in a backed off mode. This might delay the csi node driver removing the taint which will give enough time for csinode limits to be properly registered.

**Environment**
AWS EKS where karpenter + ebs is being used for a node group
- Kubernetes version (use `kubectl version`):
- 1.28 with karpenter 0.31.4
- Driver version:
- ebs csi driver - 1.24.0

If needed I can raise a PR for this

PR #1949, opened 2024-02-27
Ensure CSINode allocatable count is set on node before removing startup taint

torredil:

Bug Fix

Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1945

**What testing is done?** 

manual testing && `make test` && e2e


pkg/driver/node.go
```diff
@@ -103,7 +106,9 @@ func newNodeService(driverOptions *DriverOptions) nodeService {
 
 	// Remove taint from node to indicate driver startup success
 	// This is done at the last possible moment to prevent race conditions or false positive removals
-	go removeTaintInBackground(cloud.DefaultKubernetesAPIClient, removeNotReadyTaint)
+	time.AfterFunc(taintRemovalInitialDelay, func() {
```
ConnorJC3, line 109: Any particular reason to add a 1 second delay? Wasn't mentioned in PR description or original bug

pkg/driver/node_test.go
```diff
@@ -2414,16 +2445,52 @@ func TestRemoveNotReadyTaint(t *testing.T) {
...
+								NodeID: nodeName,
+								Allocatable: &v1.VolumeNodeResources{
+									Count: &count,
+								},
+							},
+						},
+					},
+				}
+
+				csiNodesMock.EXPECT().
```
ConnorJC3, line 2485: I know that I am the one responsible for it, and I'm not asking you to fix what's already broken in this PR.

But in the future, we really need to move from this boilerplate disaster to using the fake client implementation.

pkg/driver/node.go
torredil: Tested various intervals, first call expected failures observed at 900ms, none at 1 second.

AndrewSirenko (approved):

Issue #1946, opened 2024-02-26
Include "kubernetes.io/os" in topology key


davidgiga1993:
**Is your feature request related to a problem?/Why is this needed**
We're operating a k8s cluster with linux and windows nodes. Currently it's not possible to limit a storage class using the `allowed_topologies` to a certain operating system. 

This is required as a safety measure as currently when mounting a pvc that was previously used by a linux container on a windows container (by accident for example) it will corrupt the filesystem.

/feature

**Describe the solution you'd like in detail**
The `kubernetes.io/os` item should be added to the topology key (see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L593)

**Describe alternatives you've considered**
None

I'm happy to contribute this feature but first wanted to clarify it would be in your interest as well.

torredil:
Hi @davidgiga1993 thanks for this request - `kubernetes.io/os` is already reliably populated by Kubelet on the node: https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-os

```
kubectl describe node | grep "kubernetes.io/os"               

beta.kubernetes.io/os=linux
kubernetes.io/os=linux
beta.kubernetes.io/os=linux
kubernetes.io/os=linux
```

there isn't much value in having an additional component (the driver) also manage this label, which is all that adding the label to the topolgy segments in the csi node plugin would do.

davidgiga1993:
Thanks for your reply. I know that this label gets added to the node by the kubelet, however it's not available as topology key which means it's not possible to restrict a storage class to a certain OS.
Example:
```
failed to provision volume with StorageClass "encrypted-standard": error generating accessibility requirements:
topologymap[topology.ebs.csi.aws.com/zone:eu-central-1a topology.kubernetes.io/zone:eu-central-1a] from selected node "ip-100-2-0-184.eu-central-1.compute.internal" is not in requisite: [map[kubernetes.io/os:linux]]   
```

davidgiga1993:
I finally found time to properly evaluate this change, but sadly it doesn't fully work as intended.

Good things first:
- The OS is correctly reported as topology key. This allows me to create 2 different storage classes, one for linux and one for windows.
- When creating a new PVC the topology key is correctly evaluated and a new PV only gets created if the OS of the node matches the OS in the topology key of the storage class

Bad things:
[...]

PR #1950, opened 2024-02-28
Add OS topology key to node segments map

torredil:

This PR introduces the addition of an (OS) topology key to the node segments map returned by `NodeGetInfo` within the node service.

The change enables scoping down a StorageClass using the `allowedTopologies` attribute to an OS type. It is backward compatible and should not impact existing deployments. Nodes will simply start reporting additional topology information post-update.

Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1946

**What testing is done?** 

`make test` && manual testing && CI e2e

Example StorageClass:

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
allowedTopologies:
- matchLabelExpressions:
  - key: kubernetes.io/os
    values:
    - windows
  - key: topology.ebs.csi.aws.com/zone
    values:
    - us-east-2c
```

Without this patch:
```
  Warning  ProvisioningFailed    4s (x4 over 11s)  ebs.csi.aws.com_ebs-csi-controller-56bfc45bc5-5ckcz_d0df82a6-2e08-4622-b0ec-dbe4927673e0  failed to provision volume with StorageClass "ebs-sc": error generating accessibility requirements: topology map[topology.ebs.csi.aws.com/zone:us-east-2c topology.kubernetes.io/zone:us-east-2c] from selected node "ip-192-168-28-75.us-east-2.compute.internal" is not in requisite: [map[kubernetes.io/os:linux topology.ebs.csi.aws.com/zone:us-east-2c]]
```

With this patch:
```
  Normal   ProvisioningSucceeded  17s                ebs.csi.aws.com_ebs-csi-controller-7bf457699-8b7gk_ead7e79a-c4ee-4ad9-a66e-bf02e146c62b   Successfully provisioned volume pvc-8f101bfd-d76a-45a7-88d3-05091814375c
```

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -51,9 +51,10 @@ const (
 	AwsRegionKey    = "topology." + DriverName + "/region"
 	AwsOutpostIDKey = "topology." + DriverName + "/outpost-id"
 
-	WellKnownTopologyKey = "topology.kubernetes.io/zone"
-	// DEPRECATED Use the WellKnownTopologyKey instead
-	TopologyKey = "topology." + DriverName + "/zone"
+	WellKnownZoneTopologyKey = "topology.kubernetes.io/zone"
+	// DEPRECATED Use the WellKnownZoneTopologyKey instead
+	ZoneTopologyKey = "topology." + DriverName + "/zone"
+	OSTopologyKey   = "kubernetes.io/os"
 )
 
 type Driver struct {
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -873,22 +873,22 @@ func pickAvailabilityZone(requirement *csi.TopologyRequirement) string {
 		return ""
 	}
 	for _, topology := range requirement.GetPreferred() {
-		zone, exists := topology.GetSegments()[WellKnownTopologyKey]
+		zone, exists := topology.GetSegments()[WellKnownZoneTopologyKey]
 		if exists {
 			return zone
 		}
 
-		zone, exists = topology.GetSegments()[TopologyKey]
+		zone, exists = topology.GetSegments()[ZoneTopologyKey]
 		if exists {
 			return zone
 		}
 	}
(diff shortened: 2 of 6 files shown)


pkg/driver/driver.go
```diff
@@ -53,7 +53,8 @@ const (
 
 	WellKnownTopologyKey = "topology.kubernetes.io/zone"
 	// DEPRECATED Use the WellKnownTopologyKey instead
-	TopologyKey = "topology." + DriverName + "/zone"
+	TopologyKey   = "topology." + DriverName + "/zone"
```
AndrewSirenko, line 56: Do we want to rename this to `AZTopologyKey`?

AndrewSirenko (approved):
Thanks for this paper-cut alleviation!

pkg/driver/driver.go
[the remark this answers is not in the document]
torredil: auto-formatted

torredil: I can make this change for next rev

pkg/driver/driver.go
```diff
@@ -53,7 +53,8 @@ const (
...
+	OSTopologyKey = "kubernetes.io/os"
```
torredil, line 57: Thoughts on considering "topology." + DriverName + "/os" as an alternative?

ConnorJC3: `kubernetes.io` is a [well-known label](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-os), we should use it - if we wanted we could _also_ specify another label, but I don't think that's necessary

ConnorJC3:
/lgtm

Agree on renaming `TopologyKey`/`WellKnownTopologyKey`, would recommend name of `ZoneTopologyKey` over `AZ`, as "zone" is the terminology generally used by upstream k8s.

PR #1954, opened 2024-03-02
Add helm-tester enabled flag

nunodomingues-td:
Feature

Adding a flag to enable / disable test during pipeline when using chart-testing 

**What testing is done?** 
Manual / pipeline tests

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -466,4 +466,5 @@ useOldCSIDriver: false
 
 # Supply a custom image to the ebs-csi-driver-test pod in helm-tester.yaml
 helmTester:
+  enabled: true
   image: "gcr.io/k8s-staging-test-infra/kubekins-e2e:v20231206-f7b83ffbe6-master"
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -464,7 +464,7 @@ volumeSnapshotClasses: []
 # This parameter should always be false for new installations
 useOldCSIDriver: false
 
-# Supply a custom image to the ebs-csi-driver-test pod in helm-tester.yaml
 helmTester:
   enabled: true
+  # Supply a custom image to the ebs-csi-driver-test pod in helm-tester.yaml
   image: "gcr.io/k8s-staging-test-infra/kubekins-e2e:v20231206-f7b83ffbe6-master"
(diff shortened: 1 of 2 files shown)


ConnorJC3:
Hi @nunodomingues-td , running tests to confirm driver behavior during our `ct` checks is intentional behavior that we want to keep available.

When installing the driver with `helm install` the tests should not be installed. If you need to disable the tests when rendering manifests, you can use `--skip-tests` as documented by @torredil here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1535#issuecomment-1469997855

nunodomingues-td:
Hello @ConnorJC3 

Thanks for your reply, but my issue is not that simple.

At my company we don't install helm chart manually, we're deploying them using a few pipelines and then using a GitOps CD tool to deploy them.

In our pipelines, we have a few tests, and one of them use CT for rendering, testing, and install (dry-run) the helm charts. Our pipelines are being executed and our charts are being deployed (most of them) in graviton instances, which brings me to a few issues:
- We don't want to execute tests again in our pipeline when using third party charts ( we build our own charts and build our tests), because we're _assuming_ that the tests are being executed before the releases are created
- The test is using a intel/amd based image and the node selector is `kubernetes.io/arch: amd64`

That's why the reason of this PR.  All of this to explain that your solution does not work for us and we kindly ask for a review in this.

If you still don't want to move forward, can you give us a even better solution that we can use?

Thanks and best regards

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -466,4 +466,5 @@ useOldCSIDriver: false
 
 # Supply a custom image to the ebs-csi-driver-test pod in helm-tester.yaml
```
AndrewSirenko, line 467: Could you make sure this comment sticks to `helmTester.image`?

AndrewSirenko (changes requested):

charts/aws-ebs-csi-driver/values.yaml
nunodomingues-td: Hello @AndrewSirenko  
Sorry but I'm not understanding your comment. Do you want me to add a description regarding the enable flag?

ConnorJC3:
/lgtm

@nunodomingues-td this looks good to me to go

There was a vulnerability in one of our dependencies that was fixed in #1959, you'll need to rebase your PR on top of the latest master branch to get `govulncheck` to pass.

nunodomingues-td:
> /label tide/merge-method-squash /lgtm
> 
> @nunodomingues-td this looks good go to me, one small thing:
> 
> There was a vulnerability in one of our dependencies that was fixed in #1959, you'll need to rebase your PR on top of the latest master branch to get `govulncheck` to pass.

Hello @ConnorJC3 
My branch is now rebased.  Thanks for the warning

torredil (approved):

AndrewSirenko (approved):

PR #1958, opened 2024-03-06
[WIP] Batch EC2 DescribeSnapshots

AndrewSirenko:
Feature

Coalesces EC2 DescribeSnapshots calls across CreateSnapshot/DeleteSnapshot/ListSnapshots RPCs by Snapshot ID or by tag.

This decreases the likelihood of being exceeding one's [Non-mutating API request token limit](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits) when creating many snapshots at once, which may impact all driver RPCs (ie being throttled for making too many`Describe*` calls). 

**What testing is done?** 

Manual testing on modified [Volume Snapshots example](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/snapshot) to check that calls were batched.

TODO manual testing on ListSnapshots RPC. 

TODO possible snapshot scalability test. We would tune snapshot batcher `maxDelay`s based on these tests. 

CI


AndrewSirenko:
/hold

Holding until we decide on Snapshot scalability tests.

pkg/cloud/cloud_test.go
```diff
@@ -216,16 +215,16 @@ func executeDescribeVolumesTest(t *testing.T, c *cloud, volumeIDs, volumeNames [
 		r[i] = make(chan *ec2.Volume, 1)
 		e[i] = make(chan error, 1)
 
-		go func(req *ec2.DescribeVolumesInput, resultCh chan *ec2.Volume, errCh chan error) {
+		go func(request *ec2.DescribeVolumesInput, resultCh chan *ec2.Volume, errCh chan error) {
```
AndrewSirenko, line 218: Accidental rename of this var, will undo rename when I squash the commits.

pkg/cloud/cloud_test.go
```diff
@@ -322,16 +321,174 @@ func executeDescribeInstancesTest(t *testing.T, c *cloud, instanceIds []string,
 		r[i] = make(chan *ec2.Instance, 1)
 		e[i] = make(chan error, 1)
 
-		go func(req *ec2.DescribeInstancesInput, resultCh chan *ec2.Instance, errCh chan error) {
+		go func(request *ec2.DescribeInstancesInput, resultCh chan *ec2.Instance, errCh chan error) {
```
AndrewSirenko, line 324: Accidental rename of this var, will undo rename when I squash the commits.

pkg/cloud/cloud.go
```diff
@@ -334,14 +342,20 @@ func newBatcherManager(svc ec2iface.EC2API) *batcherManager {
 		volumeTagBatcher: batcher.New(500, 1*time.Second, func(names []string) (map[string]*ec2.Volume, error) {
 			return execBatchDescribeVolumes(svc, names, volumeTagBatcher)
 		}),
-		instanceIDBatcher: batcher.New(50, 300*time.Millisecond, func(names []string) (map[string]*ec2.Instance, error) {
-			return execBatchDescribeInstances(svc, names)
+		instanceIDBatcher: batcher.New(50, 300*time.Millisecond, func(ids []string) (map[string]*ec2.Instance, error) {
+			return execBatchDescribeInstances(svc, ids)
+		}),
+		snapshotIDBatcher: batcher.New(500, 300*time.Millisecond, func(ids []string) (map[string]*ec2.Snapshot, error) {
```
ConnorJC3, line 348: `DescribeSnapshots` can return up to 1000 results before paginating (see `MaxResults` on https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSnapshots.html), instead of the 500 of `DescribeVolumes`, so this should be increased.

Also, why 300ms?

pkg/cloud/cloud.go
```diff
@@ -334,14 +342,20 @@ func newBatcherManager(svc ec2iface.EC2API) *batcherManager {
...
+			return execBatchDescribeSnapshots(svc, ids, snapshotIDBatcher)
+		}),
+		snapshotTagBatcher: batcher.New(500, 300*time.Millisecond, func(names []string) (map[string]*ec2.Snapshot, error) {
```
ConnorJC3, line 351: Same comment as 348

pkg/cloud/cloud_test.go
```diff
@@ -322,16 +321,174 @@ func executeDescribeInstancesTest(t *testing.T, c *cloud, instanceIds []string,
...
+		for _, tag := range snapshot.Tags {
+			if tag.Key != nil && *tag.Key == SnapshotNameTagKey && tag.Value != nil {
+				snapshotNames = append(snapshotNames, *tag.Value)
+			}
+		}
+	}
+	return snapshotIDs, snapshotNames
+}
+
+func TestBatchDescribeSnapshots(t *testing.T) {
```
ConnorJC3, line 388: Should probably add test case for snapshot not found

torredil (approved):
+1 to increasing the max batch value to `1000`, otherwise lgtm!
/lgtm

pkg/cloud/cloud.go
AndrewSirenko: > Why 300ms

- DescribeSnapshots will be called at most ~6 times per second (except if the batch size exceeds 1000 in rare cases) 
- Only 300ms worst case latency added to any snapshot RPC. 

As noted in description, a finer tuning of these values should be done once we prioritize snapshot scalability tests.  Starting with a conservative added latency. 

Am open to bumping this up to 500ms if we're worried 6 calls per second is too much.

PR #1962, opened 2024-03-12
Fix link in Volume Modification example docs

steruck:
bug

Fix link in Volume Modification example docs

**What testing is done?**


AndrewSirenko (approved):
Thanks for the fixup!

PR #1965, opened 2024-03-14
Batch polling of DescribeVolumesModifications

AndrewSirenko:
Feature

Coalesces EC2 DescribeVolumesModifications (DVM) polling across ControllerExpandVolume/ControllerModifyVolume/ModifyVolume RPCs by volumeID.

This decreases the likelihood of being exceeding one's [Non-mutating API request token limit](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits) when modifying many volumes at once, which may impact all driver RPCs (ie being throttled for making too manyDescribe* calls).

Note: We do not batch any DVM call that we are unsure if a volume modification has been made for, due to undocumented behavior of a an error being returned if any volume queried for has never been modified. 

**What testing is done?** 
CI

Rested that VAC based modify volume batches DescribeVolumeModifications at scale. See Data/Impact section below. 

---

**Data/Impact** 

**900 Modified PVCs over 60 seconds across 52 nodes (100Gi increase)**

Note: All examples included 900 non-batchable DescribeVolumesModification calls. 

|  | Unbatched | 300ms | 500ms | 1s | 
|--|--|--|--|--|
| DVM Calls Made (excluding non-batchable 900) | 5500 | 231 | 189 | 114 | 
| Rate-Limiting? | Yes | No | No | No |

---

**200 Modified PVCs over 60 seconds across 12 nodes (100Gi increase)**

Note: Both examples include 200 non-batchable DescribeVolumesModification calls.  

|  | Unbatched | 500ms | 
|--|--|--|
| DVM Calls Made (excluding non-batchable 200) | 624 | 81 |

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -178,6 +178,9 @@ var (
 
 	// ErrInvalidArgument is returned if parameters were rejected by cloud provider
 	ErrInvalidArgument = errors.New("invalid argument")
+
+	// ErrInvalidRequest is returned if parameters were rejected by driver
+	ErrInvalidRequest = errors.New("invalid request")
 )
 
 // Set during build time via -ldflags
@@ -432,7 +435,7 @@ func (c *cloud) batchDescribeVolumes(request *ec2.DescribeVolumesInput) (*types.
 		task = request.Filters[0].Values[0]
 
 	default:
-		return nil, fmt.Errorf("batchDescribeVolumes: invalid request, request: %v", request)
+		return nil, fmt.Errorf("%w: batchDescribeVolumes: request: %v", ErrInvalidRequest, request)
 	}
 
 	ch := make(chan batcher.BatchResult[*types.Volume])
@@ -675,7 +678,7 @@ func (c *cloud) batchDescribeVolumesModifications(request *ec2.DescribeVolumesMo
 	if len(request.VolumeIds) == 1 && request.VolumeIds[0] != "" {
 		task = request.VolumeIds[0]
 	} else {
-		return nil, fmt.Errorf("batchDescribeVolumesModifications: invalid request, request: %v", request)
+		return nil, fmt.Errorf("%w: batchDescribeVolumesModifications: invalid request, request: %v", ErrInvalidRequest, request)
 	}
 
 	ch := make(chan batcher.BatchResult[*types.VolumeModification])
@@ -691,7 +694,7 @@ func (c *cloud) batchDescribeVolumesModifications(request *ec2.DescribeVolumesMo
 	return r.Result, nil
 }
 
-// ResizeOrModifyDisk resizes an EBS volume in GiB increments, rouding up to the next possible allocatable unit, and/or modifies an EBS
+// ResizeOrModifyDisk resizes an EBS volume in GiB increments, rounding up to the next possible allocatable unit, and/or modifies an EBS
 // volume with the parameters in ModifyDiskOptions.
 // The resizing operation is performed only when newSizeBytes != 0.
 // It returns the volume size after this call or an error if the size couldn't be determined or the volume couldn't be modified.
@@ -795,7 +798,7 @@ func (c *cloud) batchDescribeInstances(request *ec2.DescribeInstancesInput) (*ty
 	if len(request.InstanceIds) == 1 && request.InstanceIds[0] != "" {
 		task = request.InstanceIds[0]
 	} else {
-		return nil, fmt.Errorf("batchDescribeInstances: invalid request, request: %v", request)
+		return nil, fmt.Errorf("%w: batchDescribeInstances: request: %v", ErrInvalidRequest, request)
 	}
 
 	ch := make(chan batcher.BatchResult[*types.Instance])
@@ -1161,7 +1164,7 @@ func (c *cloud) batchDescribeSnapshots(request *ec2.DescribeSnapshotsInput) (*ty
 		task = request.Filters[0].Values[0]
 
 	default:
-		return nil, fmt.Errorf("batchDescribeSnapshots: invalid request, request: %v", request)
+		return nil, fmt.Errorf("%w: batchDescribeSnapshots: request: %v", ErrInvalidRequest, request)
 	}
 
 	ch := make(chan batcher.BatchResult[*types.Snapshot])
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -265,11 +265,12 @@ type snapshotBatcherType int
 
 // batcherManager maintains a collection of batchers for different types of tasks.
 type batcherManager struct {
-	volumeIDBatcher    *batcher.Batcher[string, *types.Volume]
-	volumeTagBatcher   *batcher.Batcher[string, *types.Volume]
-	instanceIDBatcher  *batcher.Batcher[string, *types.Instance]
-	snapshotIDBatcher  *batcher.Batcher[string, *types.Snapshot]
-	snapshotTagBatcher *batcher.Batcher[string, *types.Snapshot]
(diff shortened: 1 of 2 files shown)


AndrewSirenko:
/hold 

Holding to avoid merge conflicts for #1963

torredil:
Reviewing after PR is rebased post SDK changes.

AndrewSirenko:
> Reviewing after PR is rebased post SDK changes.

Response to my question about when I should expect a review on this PR.

pkg/cloud/cloud.go
```diff
@@ -638,6 +642,58 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
+}
+
+// batchDescribeVolumesModifications processes a DescribeVolumesModifications request by queuing the task and waiting for the result.
+func (c *cloud) batchDescribeVolumesModifications(request *ec2.DescribeVolumesModificationsInput) (*types.VolumeModification, error) {
+	var task string
+
+	if len(request.VolumeIds) == 1 && request.VolumeIds[0] != "" {
+		task = request.VolumeIds[0]
+	} else {
+		return nil, fmt.Errorf("batchDescribeVolumesModifications: invalid request, request: %v", request)
```
torredil, line 678: Suggesting adding a unit test to cover the code path for an invalid request here.

pkg/cloud/cloud.go
```diff
@@ -638,6 +642,58 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
+	b := c.bm.volumeModificationIDBatcher
+	b.AddTask(task, ch)
+
+	r := <-ch
+
+	if r.Err != nil {
+		return nil, r.Err
+	}
+	if r.Result == nil {
+		return nil, VolumeNotBeingModified
```
torredil, line 692: Suggest adding a unit test for`r.Result == nil`

pkg/cloud/cloud.go
```diff
@@ -1621,25 +1677,55 @@ func (c *cloud) waitForVolumeModification(ctx context.Context, volumeID string)
...
-func (c *cloud) getLatestVolumeModification(ctx context.Context, volumeID string) (*types.VolumeModification, error) {
+func (c *cloud) getLatestVolumeModification(ctx context.Context, volumeID string, isBatched bool) (*types.VolumeModification, error) {
 	request := &ec2.DescribeVolumesModificationsInput{
 		VolumeIds: []string{volumeID},
 	}
-	mod, err := c.ec2.DescribeVolumesModifications(ctx, request)
-	if err != nil {
-		if isAWSErrorModificationNotFound(err) {
+
+	// TODO Q: I see this as the cleanest way to NOT batch certain DVM calls.
```
torredil, line 1709: I think as currently written is a great approach.

pkg/cloud/cloud.go
```diff
@@ -1621,25 +1677,55 @@ func (c *cloud) waitForVolumeModification(ctx context.Context, volumeID string)
...
+		if len(volumeMods) == 0 {
 			return nil, VolumeNotBeingModified
 		}
-		return nil, fmt.Errorf("error describing modifications in volume %q: %w", volumeID, err)
-	}
 
-	volumeMods := mod.VolumesModifications
-	if len(volumeMods) == 0 {
-		return nil, VolumeNotBeingModified
+		return &volumeMods[len(volumeMods)-1], nil // TODO Q Check for nil pointer shenanigan
```
torredil, line 1725: If I understand the question correctly - no need for that check since it is guaranteed that length of volumeMods > 0 at this point.

pkg/cloud/cloud.go
AndrewSirenko: On second thought, this block is not needed because if volume has no modifications an error will be returned. Will remove code block.

pkg/cloud/cloud.go
AndrewSirenko: Great callout!  

Added new commit which adds similar unit test for all four BatchDescribe functions (as well as the `ErrInvalidRequest` err type required for this unit test to be added) so we are consistent.

I'm open to opening a separate PR for this commit because I understand changing other unit tests is out of scope.

pkg/cloud/cloud.go
```diff
@@ -1621,25 +1677,53 @@ func (c *cloud) waitForVolumeModification(ctx context.Context, volumeID string)
...
+			break
+		}
+		request.NextToken = nextToken
+	}
+	return volumeModifications, nil
+}
+
 // getLatestVolumeModification returns the last modification of the volume.
-func (c *cloud) getLatestVolumeModification(ctx context.Context, volumeID string) (*types.VolumeModification, error) {
+func (c *cloud) getLatestVolumeModification(ctx context.Context, volumeID string, isBatched bool) (*types.VolumeModification, error) {
```
ConnorJC3, line 1704: np: rename `isBatched` to something like `batchable`, because it's possible for `isBatched` to be true but for the volume to not be batched (if batching is globally disabled).

pkg/cloud/cloud.go
```diff
@@ -1621,25 +1677,53 @@ func (c *cloud) waitForVolumeModification(ctx context.Context, volumeID string)
 	return nil
 }
 
+func describeVolumesModifications(ctx context.Context, svc EC2API, request *ec2.DescribeVolumesModificationsInput) ([]types.VolumeModification, error) {
+	volumeModifications := []types.VolumeModification{}
+	var nextToken *string
+	for {
+		response, err := svc.DescribeVolumesModifications(ctx, request)
+		if err != nil {
+			if isAWSErrorModificationNotFound(err) {
```
ConnorJC3, line 1686: Is it safe to do this here? What happens if we batch volumes `A` and `B` and only one is being modified, will we return that both are not being modified?

pkg/cloud/cloud.go
```diff
@@ -638,7 +645,56 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
+	defer cancel()
+
+	resp, err := describeVolumesModifications(ctx, svc, request)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]*types.VolumeModification)
+
+	for _, i := range resp {
```
ConnorJC3, line 665: np: `i` is usually used for variables that count up, can we name this `m` or `modification` or something like that?

ConnorJC3 (commented):
mostly lgtm, see below

pkg/cloud/cloud.go
AndrewSirenko: From my understanding, it is safe because for any volume V: 
A. If a DVM call containing V is batched, either
i. There has been atleast 1 volume modification on V before, and therefore DVM would never return `ErrorModificationNotFound`
ii. ModifyVolume would have been called and synchronously succeeded for V, which we know means DVM would return a volume modification with either Pending, Optimizing, Completed, or Failed states (Because ModifyVolume and DVM are consistent with eachother). Never `ErrorModificationNotFound` 

B. Therefore, this block of code only gets hit by pre-modify DVM calls which are never batched. 

---

There are two things we can do to be extra safe if you spot a flaw in my reasoning above:
1. Pass in that `isBatched`/`batchable` boolean into this function and only return `VolumeNotBeingModified` if this is a non-batchable call (if it is a batchable call, return another error such that we don't pre-emptively return true [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/c424cbfe5d9e9d7693b104f9b33857c502af08b7/pkg/cloud/cloud.go#L1603)).
2. Have two separate `describeVolumesModifications` (one for the pre-modification DVM call, one for the post-modify DVM call)

I prefer option 1 to option 2, but I'm not convinced either are necessary.

torredil (approved):
/lgtm

PR #1966, opened 2024-03-14
Add missing instances to instance store volumes table

ConnorJC3:

Bug fix

Adds missing instance store instances to the table
~~Also documents command to generate this table~~
Also adds a script to generate this table

**What testing is done?** 

Confirmed manually against table on AWS docs, as well as previous list to ensure no types were removed/changed


pkg/cloud/volume_limits.go
```diff
@@ -131,215 +131,337 @@ func GetNVMeInstanceStoreVolumesForInstanceType(it string) int {
 	return 0
 }
 
-// / https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes
+// / https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-store-volumes.html
 // / IMDS does not provide NVMe instance store data; we'll just list all instances here
 // / TODO: See if we can get these values from DescribeInstanceTypes API
+// Generated by command, requires aws cli and jq:
+// aws ec2 describe-instance-types --region us-east-1 --filters "Name=instance-storage-supported,Values=true" --query "InstanceTypes[].[InstanceType, InstanceStorageInfo]" | jq -r 'map("\"" + .[0] + "\": " + (.[1].Disks | map(.Count) | add | tostring) + ",") | .[]' | sort
 var nvmeInstanceStoreVolumes = map[string]int{
```
AndrewSirenko, line 139: Is there a reason we don't support [hpc6id.32xlarge instance?](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-store-volumes.html#instance-store-vol-hpc)

4 instance store volumes.

pkg/cloud/volume_limits.go
```diff
@@ -131,215 +131,337 @@ func GetNVMeInstanceStoreVolumesForInstanceType(it string) int {
 	return 0
 }
 
-// / https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes
+// / https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-store-volumes.html
 // / IMDS does not provide NVMe instance store data; we'll just list all instances here
 // / TODO: See if we can get these values from DescribeInstanceTypes API
+// Generated by command, requires aws cli and jq:
+// aws ec2 describe-instance-types --region us-east-1 --filters "Name=instance-storage-supported,Values=true" --query "InstanceTypes[].[InstanceType, InstanceStorageInfo]" | jq -r 'map("\"" + .[0] + "\": " + (.[1].Disks | map(.Count) | add | tostring) + ",") | .[]' | sort
```
AndrewSirenko, line 138: This is great, thank you! Should we consider adding to our release runbook?

AndrewSirenko, line 138: Nit: Shall we consider using `awk` to make sure the sort mostly preserves the original, cleaner to read order? 

`awk -F'[:,]' '{ print $1, $2, $3 }' | awk -F'.' '{ print $1, $2, $3 }' | sort -k1,1 -k2,2n | awk '{ printf "%s.%s: %s\n", $1, $2, $3 }`

`aws ec2 describe-instance-types --region us-east-1 --filters "Name=instance-storage-supported,Values=true" --query "InstanceTypes[].[InstanceType, InstanceStorageInfo]" | jq -r 'map("\"" + .[0] + "\": " + (.[1].Disks | map(.Count) | add | tostring) + ",") | .[]' | awk -F'[:,]' '{ print $1, $2, $3 }' | awk -F'.' '{ print $1, $2, $3 }' | sort -k1,1 -k2,2n | awk '{ printf "%s.%s: %s\n", $1, $2, $3 }'`

The resulting list gets sorted like so, which is a bit nicer to read. 

```
...
"x2gd.large": 1
"x2gd.medium": 1
"x2gd.metal": 2
"x2gd.xlarge": 1
"x2gd.2xlarge": 1
"x2gd.4xlarge": 1
"x2gd.8xlarge": 1
"x2gd.12xlarge": 2
"x2gd.16xlarge": 2
"x2idn.metal": 2
"x2idn.16xlarge": 1
"x2idn.24xlarge": 2
"x2idn.32xlarge": 2
"x2iedn.metal": 2
"x2iedn.xlarge": 1
"x2iedn.2xlarge": 1
"x2iedn.4xlarge": 1
"x2iedn.8xlarge": 1
"x2iedn.16xlarge": 1
"x2iedn.24xlarge": 2
"x2iedn.32xlarge": 2
"z1d.large": 1
"z1d.metal": 2
"z1d.xlarge": 1
"z1d.2xlarge": 1
"z1d.3xlarge": 1
"z1d.6xlarge": 1
"z1d.12xlarge": 2
```

CC @torredil

AndrewSirenko (commented):
Spot checked 20 different instance types. Looks good. A few comments but then will approve.

pkg/cloud/volume_limits.go
ConnorJC3: Given that this is auto-generated, I don't think there's any advantage to spending extra complexity making it human readable - humans aren't intended to read it on any regular basis.

ConnorJC3: Yes, but not relevant to this PR, resolving

pkg/cloud/volume_limits.go
ConnorJC3: Hmm, it's only available in us-east-2 so when I go looking in us-east-1 I don't find it. Be ready for the insane kludge of v2

AndrewSirenko: Wow that's unfortunate. Thanks for letting us know the reason 😅

torredil (approved):
/lgtm

```
$ ./check.sh
Success! no instances were removed
```

```
original="https://raw.githubusercontent.com/kubernetes-sigs/aws-ebs-csi-driver/master/pkg/cloud/volume_limits.go"
updated="https://raw.githubusercontent.com/kubernetes-sigs/aws-ebs-csi-driver/aa15f7650d453e80b262587bd0bcc62242601177/pkg/cloud/volume_limits.go"

extract_table() {
  curl -s "$1" | sed -n '/^var nvmeInstanceStoreVolumes = map\[string\]int{/,/^}/p' | grep -v "^var" | grep -v "^}"
}

original_table=$(extract_table "$original")
updated_table=$(extract_table "$updated")

original_keys=$(echo "$original_table" | grep -o '^[^:]*' | sort)
updated_keys=$(echo "$updated_table" | grep -o '^[^:]*' | sort)

removed_instances=$(diff <(echo "$original_keys") <(echo "$updated_keys") | grep "^<" | cut -d ' ' -f 2)

if [ -n "$removed_instances" ]; then
  echo "Warning!!! the following instances were incorrectly removed from the table"
  echo "$removed_instances"
else
  echo "Success! no instances were removed"
fi
```

AndrewSirenko (approved):
Great work! Thanks for the initiative with the script.

torredil (approved):
/lgtm
/approve

PR #1968, opened 2024-03-14
Randomly generate tag names in e2e tests to prevent flakes due to collisions

AndrewSirenko:
Bug fix e2e

Today we use a fixed tag name in our AWS volume and snapshot tagging single-az e2e tests. This may lead to collisions when multiple tests are done in parallel and/or test resources are not cleaned up properly (e.g. a test is aborted).

This should unblock CI, which is suffering from this problem. 

**What testing is done?** 
CI

### DIFF
--- a/tests/e2e/requires_aws_api.go
+++ b/tests/e2e/requires_aws_api.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"fmt"
+	"github.com/google/uuid"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


tests/e2e/requires_aws_api.go
```diff
@@ -39,6 +40,11 @@ import (
 const testTagName = "testTag"
```
ConnorJC3, line 40: super minor np, but maybe rename this to `testTagNamePrefix` to reduce confusion?

/lgtm

torredil (approved):
Thank you!
/lgtm
/approve

PR #1969, opened 2024-03-15
Validate Karpenter Disruption taints as part of preStop node evaluation

alexandermarston:

Bug Fix

Fix for https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1955

Karpenter introduced a breaking change in https://github.com/kubernetes-sigs/karpenter/pull/508 which means it now uses the `karpenter.sh/disrupting` taint instead of the `node.kubernetes.io/unschedulable` taint when deprovisioning nodes, which means the preStop code will not run.

This PR adds another check for the new Karpenter taint, instead of only looking for `node.kubernetes.io/unschedulable`.

**What testing is done?**


AndrewSirenko:
Thank you for this @alexandermarston /ok-to-test

torredil (approved):
Thank you! @alexandermarston 
/lgtm
/approve

PR #1994, opened 2024-04-03
Fix relationship between `cmd/main` and `pkg/driver`

torredil:

This patch is the first in a series, designed to improve the architecture of the code base with the goal of promoting a more cohesive and understandable design that makes it easier to reason about the driver and make future modifications or extensions more intuitive.

More specifically, this PR addresses a circular dependency between the options package and the driver package by moving the options-related code into the driver package.

#### Summary of Changes

1. Separation of Concerns
    - The `Options` struct has been moved to the `driver` package, consolidating all driver-related configuration options in one place.
    - The `driver` package includes driver options, making it more self-contained and easier to extend.
    - The `cmd` package now focuses solely on the main entry point and setting up the necessary dependencies.

2. Simplified driver creation
    - The `NewDriver` function now accepts a pointer to `Options` instead of a variadic function, making it clearer and more straightforward to create a new driver instance.
    - The `DriverOptions` struct has been removed, and all references to it have been replaced with `Options`, reducing code duplication and improving consistency.

**What testing is done?** 

- `make verify && make test`
- Manual
- CI


torredil:
An upcoming PR adds code coverage for `NewDriver` and removes `NewFakeDriver` altogether by instantiating Cloud as part of driver creation and passing that into the controller service.

torredil:
Pulled changes here into https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1995 to help with the review process.
/close

PR #1999, opened 2024-04-09
Create missing repos in ecr.sh

ConnorJC3:

Bug fix

`ecr.sh` does not create missing ECR repos, this commit adds that

**What testing is done?** 

CI


ConnorJC3:
/retest

flake

PR #2005, opened 2024-04-11
Add explicit `AttachVolume` call in `WaitForAttachmentState`

torredil:

Bug Fix

Today, the driver relies on the `DescribeInstances` API to determine the attachment state of a volume to an EC2 instance. The API may incorrectly report the volume as attached due to EC2's eventual consistency model, as a result the driver will skip making an expected `AttachVolume` API call and proceed to directly poll the attachment state of the volume.

This PR adds an explicit `AttachVolume` API call in the `WaitForAttachmentState` function when there is a mismatch between the expected attachment state and the actual attachment state reported for a volume. If the `AttachVolume` call fails, it is logged for debugging purposes but the overall behavior of returning an error and triggering a retry of the `AttachDisk` operation remains unchanged.

**What testing is done?** 

- `make test && make verify`
- CI

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -1024,6 +1024,15 @@ func (c *cloud) WaitForAttachmentState(ctx context.Context, volumeID, expectedSt
 		// but DescribeVolume told us volume is detached, we will short-circuit this long wait loop and return error
 		// so as AttachDisk can be retried without waiting for 20 minutes.
 		if (expectedState == volumeAttachedState) && alreadyAssigned && (attachmentState != expectedState) {
+			request := &ec2.AttachVolumeInput{
+				Device:     aws.String(expectedDevice),
+				InstanceId: aws.String(expectedInstance),
+				VolumeId:   aws.String(volumeID),
+			}
+			_, err := c.ec2.AttachVolume(ctx, request)
+			if err != nil {
+				return false, fmt.Errorf("WaitForAttachmentState AttachVolume error, expected device but be attached but was %s, volumeID=%q, instanceID=%q, Device=%q, err=%w", attachmentState, volumeID, expectedInstance, expectedDevice, err)
+			}
 			return false, fmt.Errorf("attachment of disk %q failed, expected device to be attached but was %s", volumeID, attachmentState)
 		}
 
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -3029,8 +3029,11 @@ func TestWaitForAttachmentState(t *testing.T) {
 			defer cancel()
 
 			switch tc.name {
-			case "success: detached", "failure: already assigned but wrong state":
+			case "success: detached":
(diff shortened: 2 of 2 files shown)


pkg/cloud/cloud_test.go
```diff
@@ -3031,6 +3031,7 @@ func TestWaitForAttachmentState(t *testing.T) {
 			switch tc.name {
 			case "success: detached", "failure: already assigned but wrong state":
 				mockEC2.EXPECT().DescribeVolumes(gomock.Any(), gomock.Any()).Return(&ec2.DescribeVolumesOutput{Volumes: []types.Volume{detachedVol}}, nil).AnyTimes()
+				mockEC2.EXPECT().AttachVolume(gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()
```
ConnorJC3, line 3034: can we move this into it's own case, and drop the `AnyTimes`? because `AnyTimes` means that _zero_ times is accepted

pkg/cloud/cloud.go
```diff
@@ -1024,6 +1024,15 @@ func (c *cloud) WaitForAttachmentState(ctx context.Context, volumeID, expectedSt
...
 		// so as AttachDisk can be retried without waiting for 20 minutes.
 		if (expectedState == volumeAttachedState) && alreadyAssigned && (attachmentState != expectedState) {
+			request := &ec2.AttachVolumeInput{
+				Device:     aws.String(expectedDevice),
+				InstanceId: aws.String(expectedInstance),
+				VolumeId:   aws.String(volumeID),
+			}
+			_, err := c.ec2.AttachVolume(ctx, request)
+			if err != nil {
+				klog.InfoS("WaitForAttachmentState: AttachVolume call error", "volumeID", volumeID, "instanceID", expectedInstance, "Device", expectedDevice, "err", err)
```
ConnorJC3, line 1034: add a return to this case so the error is passed upstream to the user (this can help them if the error is their fault, such as trying to attach a bx volume to a non-bx instance)

ConnorJC3 (changes requested):

pkg/cloud/cloud_test.go
torredil: Done ✅

pkg/cloud/cloud.go
torredil: Added return

ConnorJC3 (approved):
/lgtm

AndrewSirenko (approved):

stefanc18:
Hi @torredil. We're experiencing frequent attachment failures with the log "expected device to be attached but was detached". By looking on EC2, it looks like the volume is bound to the node, but the volume's state is still "Available", and the attachment status doesn't say "Attached". It looks like your change here would fix this as you're triggering a new `AttachVolume` API call if it gets into this mismatched state, but why still return the error, even if the API call returns no error? Maybe I'm misunderstanding the code, but it doesn't look like it will retry in this case

torredil:
Hey @stefanc18

An error is intentionally returned here to short-circuit the `WaitForAttachmentState` polling.

> but it doesn't look like it will retry in this case

Once `WaitForAttachmentState` returns an error, it will get bubbled up the call stack and returned by `ControllerPublishVolume` which will cause the external attacher sidecar to redrive a request. On the next call, the polling process will proceed as expected since the attach call has already been made.

PR #2006, opened 2024-04-12
Add IMDSv2 support to FAQ

torredil:

Documentation

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -172,6 +172,12 @@ Containers:
       --timeout=61s
 ```
 
+## IMDSv2 Support
+
+The driver supports the use of [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) (Instance Metadata Service Version 2).
+
+To use IMDSv2 with the driver in a containerized environment like Amazon EKS, please ensure that the hop limit for IMDSv2 responses is set to 2 or greater. This is because the default hop limit of 1 is incompatible with containerized applications on Kubernetes that run in a separate network namespace from the instance.
+
 ## CreateVolume (`StorageClass`) Parameters
 
 ### `ext4BigAlloc` and `ext4ClusterSize`


docs/faq.md
```diff
@@ -172,6 +172,12 @@ Containers:
       --timeout=61s
 ```
 
+## IMDSv2 Support
+
+The driver supports the use of [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) (Instance Metadata Service Version 2) for enhanced security when accessing EC2 instance metadata.
```
ConnorJC3, line 177: Drop the words "for enhanced security when accessing EC2 instance metadata", as our next paragraph has you completely eliminate most/all those security benefits.

docs/faq.md
```diff
@@ -172,6 +172,12 @@ Containers:
       --timeout=61s
 ```
 
+## IMDSv2 Support
```
ConnorJC3, line 175: I think we should strongly consider including a line in this section suggesting the use of Kubernetes metadata when possible - IMDS metadata requires configuring IMDS in an insecure manner (by increasing the hop limit).

docs/faq.md
torredil: Dropped!

docs/faq.md
torredil: Hesitant to provide that recommendation given Kubernetes metadata does not have information about attached ENIs ([NumAttachedENIs](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/85563fbdcb31e7234abf9f9a335008ee0c70c5b8/pkg/cloud/metadata_k8s.go#L82) is hard-coded to 1) and thus the allocatable count as reported by `NodeGetInfo` may be incorrect. That said, can add this recommendation in next revision if you strongly recommend it.

ConnorJC3: Fair enough

AndrewSirenko (approved):
This is great-- prevents future users from running into this papercut, thank you!

Issue #1951, opened 2024-02-29
Possible race condition in EBS volume creation


Sinjo:
Hi 👋🏻

I think I've found a race condition that happens very occasionally when creating a PV for a new PVC. The outcome of the race condition is a PVC that is permanently stuck in `Pending` and has to be deleted and recreated.

As a ballpark estimate, based on the number of `CreateVolume` calls per day in CloudTrail and how frequently we see this issue, I would guess that this happens on the order of 1 in 100,000 volume creations.

I've substituted the actual PV and EBS volume names below with placeholders like `pvc-foo` for ease of reading. The identifier of each PVC corresponds directly to the identifier of the EBS volume (i.e. PV `pvc-foo` corresponds to EBS volume `vol-foo`).

# Summary

Based on CloudTrail events, it seems that aws-ebs-csi-driver sometimes creates and then shortly after deletes an EBS volume to provide a PV for a new PVC.

Once that has happened, the controller repeatedly retries the creation of the EBS volume, but it always fails because of how the API request to AWS is formed. Specifically:

  - The name of the PV is generated from the UID of the PVC.
  - The `clientToken`, which is used as an idempotency key in the request to the AWS API, is generated from the name of the PV.
  - This means that `clientToken` will always be the same for a specific PVC, and the retries will fail indefinitely.

From the driver side, every retry results in the following error:

```
E0228 05:00:15.227309       1 driver.go:155] "GRPC error" err="rpc error: code = AlreadyExists desc = Could not create volume \"pvc-foo\": Parameters on this idempotent request are inconsistent with parameters used in previous request(s)"
```

and in the CloudTrail event, the error message is:

```
The client token you have provided is associated with a resource that is already deleted. Please use a different client token.
```

All of the CloudTrail events all have the same `userIdentity` (the service account assumed by the driver) and source IP address (the driver's pod IP).

# Possible EBS volume ID confusion

One really weird thing I saw in the driver logs was this:

```
E0228 05:00:14.031881       1 driver.go:155] "GRPC error" err=<
        rpc error: code = Internal desc = Could not create volume "pvc-foo": failed to get an available volume in EC2: InvalidVolume.NotFound: The volume 'vol-bar' does not exist.
                status code: 400, request id: [redacted]
 >
```

[...]

PR #2009, opened 2024-04-12
Remove DeleteDisk call in CreateDisk path

ConnorJC3:

Bug fix

Today, when we time out waiting for the volume to create, we delete it. This is broken for several reasons:

- It often doesn't even run, because the sidecar will have already closed the context elsewhere
- Whenever it does run, it will trigger #1951 when the CO re-calls `CreateVolume`
- It handles behavior that is the CO's responsibility according to the CSI spec:
```
  //    The CO is responsible for cleaning up volumes it provisioned
  //    that it no longer needs. If the CO is uncertain whether a volume
  //    was provisioned or not when a `CreateVolume` call fails, the CO
  //    MAY call `CreateVolume` again, with the same name, to ensure the
  //    volume exists and to retrieve the volume's `volume_id`
```

Also takes this opportunity to cleanup the extremely unclear error message.

**What testing is done?** 

N/A

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -613,14 +613,7 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	}
 
 	if err := c.waitForVolume(ctx, volumeID); err != nil {
-		// To avoid leaking volume, we should delete the volume just created
-		// TODO: Need to figure out how to handle DeleteDisk failed scenario instead of just log the error
-		if _, error := c.DeleteDisk(ctx, volumeID); error != nil {
-			klog.ErrorS(error, "failed to be deleted, this may cause volume leak", "volumeID", volumeID)
-		} else {
-			klog.V(5).InfoS("[Debug] volume is deleted because it is not in desired state within retry limit", "volumeID", volumeID)
-		}
-		return nil, fmt.Errorf("failed to get an available volume in EC2: %w", err)
+		return nil, fmt.Errorf("timed out waiting for volume to create: %w", err)
 	}
 
 	outpostArn := aws.ToString(response.OutpostArn)
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -667,7 +667,6 @@ func TestCreateDisk(t *testing.T) {
 		volState             string
 		diskOptions          *DiskOptions
 		expDisk              *Disk
-		cleanUpFailedVolume  bool
 		expErr               error
 		expCreateVolumeErr   error
 		expDescVolumeErr     error
@@ -919,9 +918,8 @@ func TestCreateDisk(t *testing.T) {
 				AvailabilityZone: "",
 			},
 			expCreateVolumeInput: &ec2.CreateVolumeInput{},
-			expErr:               fmt.Errorf("failed to get an available volume in EC2: DescribeVolumes generic error"),
+			expErr:               fmt.Errorf("timed out waiting for volume to create: DescribeVolumes generic error"),
 			expDescVolumeErr:     fmt.Errorf("DescribeVolumes generic error"),
-			cleanUpFailedVolume:  true,
 		},
 		{
 			name:       "fail: Volume is not ready to use, volume stuck in creating status and controller context deadline exceeded",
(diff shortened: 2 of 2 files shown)


torredil (approved):
Agree with the changes made in this PR and specifically the point which states that this logic handles behavior that is the CO's responsibility. My understanding is that having a `DeleteDisk` call in the `CreateVolume` path makes is a violation of the CSI spec.

Also worth highlighting that the external provisioner [takes care of](https://github.com/kubernetes-csi/external-provisioner/blob/91e80e508a9b789d833db94196524e6ac02c7ebd/pkg/controller/controller.go#L875) cleaning up orphaned volumes:

```
err = cleanupVolume(ctx, p, delReq, provisionerCredentials)
	if err != nil {
		capErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", capErr, pvName, err)
	}
```

AndrewSirenko:
> Also worth highlighting that the external provisioner [takes care of](https://github.com/kubernetes-csi/external-provisioner/blob/91e80e508a9b789d833db94196524e6ac02c7ebd/pkg/controller/controller.go#L875) cleaning up orphaned volumes:
> 
> ```
> err = cleanupVolume(ctx, p, delReq, provisionerCredentials)
> 	if err != nil {
> 		capErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", capErr, pvName, err)
> 	}
> ```

Today this is not true for EBS CSI Driver, due to kubernetes not knowing the volume-id of the orphaned volume. 

Some CSI Drivers store volume-id in PVC, because they can pass in a volume ID to their CreateVolume equivalent in their backend, which lets Kubernetes know the volume-id so K8s can be sure to delete it if volume was orphaned.

Alas, because Kubernetes and EBS CSI Driver only know volume's Idempotency token (hash of PVC name), and we cannot call EC2 DeleteVolume with an idempotency token _**today**_, we do risk leaking/orphaning a volume (as an edge case).

Can draw out a diagram later.

Either way I agree with the PR because this edge case is NOT solved via the code @ConnorJC3 is deleting, due to the context typically being cancelled by the time we reach this DeleteVolume call. Furthermore this code today can lead to stuck pods IF sidecar timeout is very large, createVolume succeeds, but waitForVolume times out (due to today's 1 min limit). So it is safer to delete this code today and think of something better.

torredil:
@AndrewSirenko Sorry that last bit was poorly phrased and confusing. _Also worth highlighting that the external provisioner takes care of cleaning up volumes_ is a bit clearer.

> Alas, because Kubernetes and EBS CSI Driver only know volume's Idempotency token (hash of PVC name), and we cannot call EC2 DeleteVolume with just an idempotency token today, we do risk leaking/orphaning a volume (as an edge case).

Can you clarify this point - what is the specific scenario in which the driver needs to call EC2's `DeleteVolume` in the `CreateVolume` path (CSI spec violation), where a volume ID is not available? I understand this to mean the RPC is not idempotent.

The correct behavior for cleaning up volumes is as follows:

```
  //    The CO is responsible for cleaning up volumes it provisioned
  //    that it no longer needs. If the CO is uncertain whether a volume
  //    was provisioned or not when a `CreateVolume` call fails, the CO
  //    MAY call `CreateVolume` again, with the same name, to ensure the
  //    volume exists and to retrieve the volume's `volume_id` (unless
  //    otherwise prohibited by "CreateVolume Errors").
```

In practice, this means failing the request and letting the provisioner retry as opposed to attempting to delete the volume, which is what this PR steers us towards - I think we all agree on that point. 

Once Kubernetes (the provisioner) has retrieved the volume ID from a `CreateVolume` call, it can proceed with `DeleteVolume` where the volume ID is passed in via the `DeleteVolumeRequest` created by the provisioner.

> this edge case mentioned above is NOT solved via the code @ConnorJC3 is deleting anyway

Agree 👍 

I would like to learn more about this edge case and look forward to the diagram.

AndrewSirenko (approved):

AndrewSirenko:
> _Also worth highlighting that the external provisioner takes care of cleaning up volumes_ is a bit clearer.

Yep, I agree with this phrasing. 

I only disagreed with the external provisioner cleaning up **orphaned** volumes via the code from the `cleanupVolume` snippet you linked to (which doesn't do anything in the aws-ebs-csi-driver case because EBS requires a volume-id today to delete, which external provisioner won't have access to until a successfully returned retry CreateVolume call). 

> Can you clarify this point - what is the specific scenario in which the driver needs to call EC2's DeleteVolume in the CreateVolume path (CSI spec violation), where a volume ID is not available?

I agree, the driver should not call EC2 DeleteVolume in CreateVolume RPC, I'm not arguing against that. 

I'm saying that until EBS provides a way to call EC2 DeleteVolume with the idempotency token used for EC2 CreateVolume (instead of the volume-id), there may still be an edge case of volume leaks because the `cleanupVolume` snippet can't work. This edge case could occur if a cluster operator deletes a PVC object that hadn't yet led to a CreateVolume RPC success (due to a context timeout), but the volume was created in AWS backend.

Sidenote: Looking more closely at `cleanupVolume`, it only triggers if a volume was created with less than the desired capacity, so I'm not sure it's actually relevant to the discussion even if we had a way of deleting a volume without volume-id.

PR #2010, opened 2024-04-15
Add retry manager to reduce RateLimitExceeded errors

AndrewSirenko:
Feature

**What testing is done?**


AndrewSirenko:
/hold

Until:
- [ ] 5k, 10k scalability tests completed
- [ ] PR Description updated
- [ ] 1k pod scalability test chart posted

pkg/cloud/retry_manager.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws/retry"
+	"time"
+)
+
+const (
+	// retryMaxAttempt sets max number of EC2 API call attempts.
+	// Set high enough to ensure default sidecar timeout will cancel context long before we stop retrying.
+	retryMaxAttempt = 50
+	// retryMaxBackoffAdaptive sets max backoff delay between retry attempts.
+	// Set more aggressive than default of 20 seconds due to adaptive client-side throttling being enabled.
```
ConnorJC3, line 13: I'm not sure I understand the reasoning for this - we're lowering the default delay that AWS picked for the adaptive retryer, correct?

pkg/cloud/cloud_test.go
```diff
@@ -2519,7 +2519,7 @@ func TestGetSnapshotByName(t *testing.T) {
 
 			ctx := context.Background()
 
-			mockEC2.EXPECT().DescribeSnapshots(gomock.Any(), gomock.Any()).Return(&ec2.DescribeSnapshotsOutput{Snapshots: []types.Snapshot{ec2snapshot}}, nil)
+			mockEC2.EXPECT().DescribeSnapshots(gomock.Any(), gomock.Any(), gomock.Any()).Return(&ec2.DescribeSnapshotsOutput{Snapshots: []types.Snapshot{ec2snapshot}}, nil)
```
ConnorJC3, line 2522: Why is this changed?

pkg/cloud/retry_manager.go
```diff
@@ -0,0 +1,54 @@
...
+	modifyVolumeRetryer               *retry.AdaptiveMode
+	createSnapshotRetryer             *retry.AdaptiveMode
+	deleteSnapshotRetryer             *retry.AdaptiveMode
+	enableFastSnapshotRestoresRetryer *retry.AdaptiveMode
+
+	// Restrict unbatchable describe calls when rate-limited so that batched calls are more likely to succeed.
+	unbatchableDescribeVolumesModificationsRetryer *retry.AdaptiveMode
+}
+
+// newRetryManager initializes a new instance of retryManager that will dictate the retry strategies of EC2 API calls.
```
ConnorJC3, line 41: Drop this comment, "newRetryManager initializes a new instance of retryManager" is obvious both from the code and the name of the function

pkg/cloud/retry_manager.go
```diff
@@ -0,0 +1,54 @@
...
+	createVolumeRetryer               *retry.AdaptiveMode
+	deleteVolumeRetryer               *retry.AdaptiveMode
+	attachVolumeRetryer               *retry.AdaptiveMode
+	detachVolumeRetryer               *retry.AdaptiveMode
+	modifyVolumeRetryer               *retry.AdaptiveMode
+	createSnapshotRetryer             *retry.AdaptiveMode
+	deleteSnapshotRetryer             *retry.AdaptiveMode
+	enableFastSnapshotRestoresRetryer *retry.AdaptiveMode
+
+	// Restrict unbatchable describe calls when rate-limited so that batched calls are more likely to succeed.
```
ConnorJC3, line 37: I would suggest dropping this comment as it's entirely speculation

torredil (commented):
Code largely looks good, currently manually testing changes. Will report back with lgtm if no issues surface from manual tests

torredil (commented):
Did not run into any issues manually testing. Will apply lgtm label after Connor's feedback is addressed. I've got one non-blocking idea and question.

---
Idea: worth considering using a helper function to apply the retryer instead of passing the retryer function inline for each EC2 API call, ie:

```
func withRetryer(fn func(ctx context.Context, input interface{}, optFns ...func(*ec2.Options)) (output interface{}, err error), retryer *retry.AdaptiveMode) func(ctx context.Context, input interface{}, optFns ...func(*ec2.Options)) (output interface{}, err error) {
    return func(ctx context.Context, input interface{}, optFns ...func(*ec2.Options)) (output interface{}, err error) {
        optFns = append(optFns, func(o *ec2.Options) {
            o.Retryer = retryer
        })
        return fn(ctx, input, optFns...)
    }
}
```
and then we can use it like this:

```
response, err := withRetryer(c.ec2.CreateVolume, c.rm.createVolumeRetryer)(ctx, requestInput)
```

This way, retryer logic is centralized.

---
Question: 

Currently, each EC2 API call has its own retryer field in the `retryManager` struct - is there a requirement for doing so, or is that level of granularity necessary? (given a subset of API calls may have similar retry requirements)

pkg/cloud/retry_manager.go
AndrewSirenko: The SDK has a default 20 second max-backoff for all services. It is the same for default retryer (standard) and adaptive. I can stick with default 20s.

AndrewSirenko:
> Currently, each EC2 API call has its own retryer field in the `retryManager` struct - is there a requirement for doing so, or is that level of granularity necessary? (given a subset of API calls may have similar retry requirements)

We need each Mutating EC2 API call to have their own retryer field because the sdk throttles on a retryer object level, *not* by API name. While default AWS accounts share request tokens between each EC2 Mutating API, if a customer raises limits for a particular Mutating API (ie CreateVolume) then the token bucket is no longer shared. 

E.g. If we have them share the same retryer, then if a CreateVolume hits RateLimitExceeded, AttachVolume calls will also be throttled by client-side adaptive retryer, even if the two APIs no longer share the same AWS request token bucket (due to raised limits).

Hope that is clearer.

AndrewSirenko:
> Idea: worth considering using a helper function to apply the retryer instead of passing the retryer function inline for each EC2 API call, ie:

Thank you for the suggestion. I personally think that a call-site option is clearer because it matches [the recommended approach of overriding AWS SDK service operations](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/retry#example-package-OverrideSpecificOperation) and avoids a function pointer. Also we may want to add more call-site options in the future.

torredil:
@AndrewSirenko Perfect, that is as clear as can be. Can we add a comment in the `retryManager` to help future maintainers understand the reasoning behind having separate retryer fields for each mutating API?

pkg/cloud/retry_manager.go
```diff
@@ -0,0 +1,50 @@
...
+		})
+	})
+}
+
+// retryManager dictates the retry strategies of EC2 API calls.
+// Each mutating EC2 API has its own retryer because the AWS SDK throttles on a retryer object level, not by API name.
+// While default AWS accounts share request tokens between mutating APIs, users can raise limits for individual APIs.
+// Separate retryers ensures that throttling one API doesn't unintentionally throttle others with separate token buckets.
+type retryManager struct {
+	createVolumeRetryer                            *retry.AdaptiveMode
```
rdpsin, line 27: small np: is it possible to have a map of `<something>:*retry.AdaptiveMode`?

pkg/cloud/retry_manager.go
```diff
@@ -0,0 +1,50 @@
...
+	deleteVolumeRetryer                            *retry.AdaptiveMode
```
rdpsin, line 28: np: we should probably use `aws.Retryer`, instead of `*retry.AdaptiveMode` directly.

pkg/cloud/retry_manager.go
AndrewSirenko: Is there a benefit to using a map here? 

I went for a struct to avoid defining [extra key constants](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/64772f0ffef94daa4b03a0115126bd80c1cfb72f/pkg/driver/constants.go#L35) and for extra type-safety protection from typos. Open to changing though.

ConnorJC3: I thought about this, but I think that all the map does is add additional (unnecessary) checks at each call site for non-nilness.

A map would be useful here if we could dynamically setup the retryer based on the API name but that's not possible.

torredil (approved):
/approve

PR #2021, opened 2024-04-23
Decrease dynamic provisioning time by 1.5 seconds

AndrewSirenko:
Improvement

Cut median dynamic provisioning latency in half (~3.7 to 2 seconds) by polling  EBS volume creation more agressively.  

Today, when dynamically provisioning a volume via our CreateVolume RPC, we wait 3s after an EC2 CreateVolume call to start polling DescribeVolumes. Based on testing in `us-west-2`, `ap-northeast-1`, and `us-east-1`, the p10 & median time between calling EC2 CreateVolume and the volume being created is ~1.5s. 

Polling every 3 seconds means that a typical CreateVolume RPC will take ~3.5 seconds (and a worse case of 6+ seconds once every hundred volumes). In this PR, we use a more aggressive initial delay and polling interval. Finally, we switch to an exponential backoff in order to decrease the likelihood of being rate-limited for DescribeVolumes if volume creation time slows down.  

**What testing is done?** 

Measured seconds between `external-provisioner` CreateVolume RPC first start and first success for 100 PVCs launched across 100 pods on a 30 node cluster. Repeated 3 times for each combination of parameters. 

Final column is what we went for in this PR. 

|                |3s Initial Sleep; 3s poll duration (Today's Performance)|1.5s Initial Sleep; 1s poll duration|1s Initial Sleep; Exponential backoff with .75s initial|1s Initial Sleep; Exponential backoff with .5s initial|1.5s Initial Sleep; Exponential backoff with .5s initial|1.25s Initial Sleep; Exponential backoff with .5s initial|
|-----------------------|--------------------------------------------------------|------------------------------------|-------------------------------------------------------|------------------------------------------------------|--------------------------------------------------------|---------------------------------------------------------|
|p10                    |3.52                                                    |1.90                                |1.76                                                   |1.73                                                  |1.93                                                    |1.76                                                     |
|p50                    |3.76                                                    |2.19                                |2.52                                                   |2.57                                                  |2.19                                                    |2.02                                                     |
|p90                    |3.85                                                    |2.81                                |2.99                                                   |2.86                                                  |2.60                                                    |2.85                                                     |
|p95                    |3.86                                                    |3.25                                |3.11                                                   |3.63                                                  |3.12                                                    |3.09                                                     |
|p99                    |6.66                                                    |5.31                                |4.37                                                   |3.94                                                  |4.40                                                    |4.01                                                     |
|Avg amount of DV calls |27                                                      |41                                  |30                                                     |34                                                    |33                                                      |32                                                       |

NOTE: Tested with 500ms maxDelay (instead of 1s) for describeVolume batcher, because we will decrease that value in a future parameter tuning PR. That maxDelay value mostly affected `p90` `p95` (I presume those affected were the first DescribeVolume calls in every batch)

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -87,17 +87,22 @@ var (
 		VolumeTypeStandard,
 	}
 
+	volumeCreationPollInitialDelay    = 1250 * time.Millisecond
+	volumeCreationPollBackoffDuration = 500 * time.Millisecond
+	volumeCreationPollBackoffFactor   = 1.5
+	volumeCreationPollBackoffSteps    = 25
+	volumeCreationPollBackoffCap      = 3 * time.Second
+
 	volumeModificationDuration   = 1 * time.Second
 	volumeModificationWaitFactor = 1.7
 	volumeModificationWaitSteps  = 10
 
-	volumeAttachmentStatePollSteps = 13
+	volumeAttachmentStatePollSteps  = 13
+	volumeAttachmentStatePollDelay  = 1 * time.Second
+	volumeAttachmentStatePollFactor = 1.8
 )
 
 const (
-	volumeAttachmentStatePollDelay  = 1 * time.Second
-	volumeAttachmentStatePollFactor = 1.8
-
 	volumeDetachedState = "detached"
 	volumeAttachedState = "attached"
 )
@@ -1527,20 +1532,27 @@ func (c *cloud) listSnapshots(ctx context.Context, request *ec2.DescribeSnapshot
 }
 
 // waitForVolume waits for volume to be in the "available" state.
-// On a random AWS account (shared among several developers) it took 4s on average.
 func (c *cloud) waitForVolume(ctx context.Context, volumeID string) error {
+	// Based on our testing in us-west-2 and ap-south-1, the median/p99 time until volume creation is ~1.5/~4 seconds.
+	// We have found that the following parameters are optimal for minimizing provisioning time and DescribeVolumes calls
+	// we queue DescribeVolume calls after [1.25, 0.5, 0.75, 1.125, 1.7, 2.5, 3] seconds.
+	// In total, we wait for ~60 seconds.
 	var (
-		checkInterval = 3 * time.Second
-		// This timeout can be "ovewritten" if the value returned by ctx.Deadline()
-		// comes sooner. That value comes from the external provisioner controller.
-		checkTimeout = 1 * time.Minute
+		backoff = wait.Backoff{
+			Duration: volumeCreationPollBackoffDuration,
+			Factor:   volumeCreationPollBackoffFactor,
+			Steps:    volumeCreationPollBackoffSteps,
+			Cap:      volumeCreationPollBackoffCap,
+		}
 	)
+
+	time.Sleep(volumeCreationPollInitialDelay)
 
 	request := &ec2.DescribeVolumesInput{
 		VolumeIds: []string{volumeID},
 	}
 
-	err := wait.PollUntilContextTimeout(ctx, checkInterval, checkTimeout, false, func(ctx context.Context) (done bool, err error) {
+	err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (done bool, err error) {
 		vol, err := c.getVolume(ctx, request)
 		if err != nil {
 			return true, err


pkg/cloud/cloud.go
```diff
@@ -1527,20 +1531,27 @@ func (c *cloud) listSnapshots(ctx context.Context, request *ec2.DescribeSnapshot
...
-		checkTimeout = 1 * time.Minute
+		initialDelay = volumeCreationPollInitialDelay
+		backoff      = wait.Backoff{
+			Duration: volumeCreationPollBackoffDuration,
+			Factor:   volumeCreationPollBackoffFactor,
+			Steps:    volumeCreationPollBackoffSteps,
+		}
 	)
 
+	time.Sleep(initialDelay)
```
torredil, line 1548: Can we remove this `time.Sleep` altogether? at least for the scenario when batching is enabled?

Ran 1k volume scalability tests today and the data suggests we should do just that.

- v1.30.0 
```
+-------------------------------------------------------------+
|                     EC2 API Calls Table                     |
+-------------------+-------+------+--------------+-----------+
|               API | Count |   OK | Client Error | Throttled |
+-------------------+-------+------+--------------+-----------+
|      CreateVolume |  1000 | 1000 |            0 |         0 |
|   DescribeVolumes |   389 |  389 |            0 |         0 |
+-------------------+-------+------+--------------+-----------+
```

- This PR **without** an initial delay
```
+-------------------------------------------------------------+
|                     EC2 API Calls Table                     |
+-------------------+-------+------+--------------+-----------+
|               API | Count |   OK | Client Error | Throttled |
+-------------------+-------+------+--------------+-----------+
|      CreateVolume |  1000 | 1000 |            0 |         0 |
|   DescribeVolumes |   394 |  394 |            0 |         0 |
+-------------------+-------+------+--------------+-----------+
```



I was scratching my head at these results, and then realized that we batch requests for a total of 1 second before making the DV call. Thus, our current assumption that the first DV call is guaranteed to fail is wrong. 

Besides this request, pr LGTM.

torredil (commented):
Great work! thank you for digging into this and the efficiency improvement, this PR is very significant 💯

pkg/cloud/cloud.go
AndrewSirenko: See the note in the PR

> NOTE: Tested with 500ms maxDelay (instead of 1s) for describeVolume batcher, because we will decrease that value in a future parameter tuning PR. That maxDelay value mostly affected p90 p95 (I presume those affected were the first DescribeVolume calls in every batch)

--- 
> ...we batch requests for a total of 1 second before making the DV call. Thus, our current assumption that the first DV call is guaranteed to fail is wrong.

I don't think this assertion is true. What about all the DV calls that are queued up in batcher within that 1 second? Some of those DV calls might be made 0.1s after createVolume succeeds, and those will likely fail. (Because we know it takes at least 1.4 seconds for a volume to be created). 

MaxDelay in a batcher adds variance to how long we wait before making the DV call, and I don't think we should rely on that as the initialDelay for any polling we do. The larger the maxDelay, the larger the variance. 

Furthermore DV is called in other places such as the ModifyVolume path. A higher maxDelay adds latency for those RPCs. 

Finally, what if we're describing more than 500 volumes within a 1 second interval? Suddenly all the DV calls don't wait the rest of that delay interval, and are more likely to fail & retry.

AndrewSirenko: Although you do bring up a great point of checking if batching is enabled. We should consider a longer delay in that case.

ConnorJC3: > Because we know it takes at least 1.4 seconds for a volume to be created

Do we know this? I was under the impression that testing suggested 1.5 seconds was the _median_ time to create a volume, not minimum. (On that note, assuming I'm correct - do we have a minimum time to create a volume?)

pkg/cloud/cloud.go
```diff
@@ -1527,20 +1531,27 @@ func (c *cloud) listSnapshots(ctx context.Context, request *ec2.DescribeSnapshot
...
+
 	request := &ec2.DescribeVolumesInput{
 		VolumeIds: []string{volumeID},
 	}
 
-	err := wait.PollUntilContextTimeout(ctx, checkInterval, checkTimeout, false, func(ctx context.Context) (done bool, err error) {
+	err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (done bool, err error) {
```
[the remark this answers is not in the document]
ConnorJC3: Suggestion: change the max backoff time to 3 seconds so that once the exponential backoff "ramps up", it behaves similarly to today

pkg/cloud/cloud.go
ConnorJC3: Discussed offline: I am fine with leaving it at 1.25 seconds based on the data we have.

pkg/cloud/cloud.go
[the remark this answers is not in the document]
AndrewSirenko: > Notice a meaningful number of CreateVolume RPCs taking longer than 10 seconds with this change.

Could that be because we succeed sooner on rpcs and then get rate-limited for CreateVolume sooner with this PR?

We don't take longer than 3s for a backoff until the 7th call of createvolume. 
[1.25, 0.5, 0.75, 1.125, 1.7, 2.5, 3.8, 5.7, 8.6, 12.8, 19.2]

But yes agreed we can decrease the max backoff time.

pkg/cloud/cloud.go
```diff
@@ -1527,20 +1532,28 @@ func (c *cloud) listSnapshots(ctx context.Context, request *ec2.DescribeSnapshot
...
+		initialDelay = volumeCreationPollInitialDelay
+		backoff      = wait.Backoff{
+			Duration: volumeCreationPollBackoffDuration,
+			Factor:   volumeCreationPollBackoffFactor,
+			Steps:    volumeCreationPollBackoffSteps,
+			Cap:      volumeCreationPollBackoffCap,
+		}
 	)
 
+	time.Sleep(initialDelay)
```
ConnorJC3, line 1550: Can this just be `time.Sleep(volumeCreationPollInitialDelay)`? Why the extra variable?
Otherwise lgtm

torredil (approved):
/lgtm
/approve

PR #2023, opened 2024-04-26
Add missing Kubernetes license headers

torredil:

This PR adds a new makefile target to generate the Kubernetes license header. The new target is used to add missing headers.

**What testing is done?** 

```
make update
```


Makefile
```diff
@@ -262,6 +262,21 @@ update/gomod:
 update/shfmt: bin/shfmt
 	./bin/shfmt -w -i 2 -d ./hack/
 
+.PHONY: update/add-license-header
+update/add-license-header:
```
AndrewSirenko, line 266: Consider hiding this as a a `/hack` shell script as to not pollute the Makefile as we do for other scripts like `./hack/release-scripts/generate-sidecar-tags` and `./hack/update-kustomize.sh`.

AndrewSirenko:
This will be very helpful, thank you Eddie!

AndrewSirenko (approved):

PR #2028, opened 2024-04-30
`Makefile` & `hack/` Process Improvements

ConnorJC3:

"New Feature"

This PR is split into several much smaller commits:

<ins>Move WINDOWS and WINDOWS_HOSTPROCESS out of Makefile to correct location in config.sh</ins>
Small fix to remove unnecessary variables in the `Makefile`.

<ins>Use Prow BUILD_ID in CI jobs if present to greatly reduce the chance of cluster name collisions in CI</ins>
Prow jobs have a random chance of colliding because `$RANDOM` is only a 4 digit number (You'd think it's 1/10000 but it's much higher than that because of the [Birthday Paradox](https://en.wikipedia.org/wiki/Birthday_problem)).

Use `$BUILD_ID` in CI which is guaranteed to be unique. Had to cut off a little bit of the beginning of the name to make it short enough for `eksctl`.

<ins>Add make cluster/kubeconfig</ins>
Convenience command for local development, used to set the `KUBECONFIG` for the cluster created by `make cluster/create`.

<ins>Add make cluster/image command; Build image and cluster in parallel for CI</ins>
Parallelize building the image and creating the cluster on CI, had to split image build out into a separate command to do this.

**Breaking change for contributors:** `make cluster/e2e` style commands no longer build the image, you must either run `make cluster/image` first or bring your own image.

<ins>Template eksctl cluster file so Windows nodegroup can build in parallel</ins>
Parallelize creating the Linux and Windows nodegroup for `eksctl` clusters by using a cluster file instead of `eksctl` commands.

<ins>Parallelize kubetest2 installation in CI</ins>
`kubetest2` takes surprisingly long to install, might as well do so while building the cluster/image already.

**What testing is done?** 
CI/Manual

### DIFF
--- a/hack/tools/install.sh
+++ b/hack/tools/install.sh
@@ -26,6 +26,8 @@ EKSCTL_VERSION="v0.175.0"
 GINKGO_VERSION="v2.17.1"
 # https://github.com/golangci/golangci-lint
 GOLANGCI_LINT_VERSION="v1.57.2"
+# https://github.com/hairyhenderson/gomplate
+GOMPLATE_VERSION="v3.11.7"
 # https://github.com/helm/helm
 HELM_VERSION="v3.14.4"
 # https://github.com/kubernetes/kops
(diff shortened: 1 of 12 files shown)


hack/tools/install.sh
```diff
@@ -124,6 +126,14 @@ function install_golangci-lint() {
   install_tar_binary "${INSTALL_PATH}" "https://github.com/golangci/golangci-lint/releases/download/${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION:1}-${OS}-${ARCH}.tar.gz" "golangci-lint-${GOLANGCI_LINT_VERSION:1}-${OS}-${ARCH}/golangci-lint"
 }
 
+function install_gomplate() {
+  INSTALL_PATH="${1}"
+
+  # gomplate includes library from no longer existing domain inet.af, and thus cannot be installed via go install
```
AndrewSirenko, line 132: Nit: Can we change comment to `gomplate must be installed via Github releases`

AndrewSirenko (approved):
These are all great quality of life improvements. Thank you for saving hours of time over the next year.

hack/tools/install.sh
ConnorJC3: I'm not sure I understand the request? You want me to drop the context of why it can't be installed?

torredil (approved):
/lgtm

hack/tools/install.sh
AndrewSirenko: Yes because end maintainer doesn't need that deep of context of `inet.af` not existing. But feel free to leave in if you think it's worth it.

PR #2029, opened 2024-05-01
Tune batcher EC2 Describe* delays

AndrewSirenko:
Improvement

Given that no throttling is observed for DescribeVolumes with 7k volume scalability tests, we reduced the current 1 second max batch delay to 500ms, and set that as the standard batcher delay. 

This PR makes sure each RPC takes ~0.25s of batch latency per EC2 Describe call on average (and worst case extra delay of 0.5s).  

This PR also makes sure each batcher will execute twice per second. In the rare case that each batcher is executing at once, the combined 12 requests per second is under the default [EC2 Non-Mutating Action Bucket Refill Rate](https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html) of 20.

**What testing is done?** 
Scalability tests with batcher delays of 1, 0.5, 0.3, and 0.2 seconds for DescribeVolumes and DescribeInstances. 

Final 5000 pod scalability test on default limits account.

|               API | Count |   OK | Client Error | Throttled |
|--|--|--|--|--|
|      CreateVolume |  5676 | 5030 |            0 |       646 |
|      DeleteVolume |  5563 | 4999 |            2 |       562 |
|      AttachVolume |  5646 | 4989 |            0 |       657 |
|      DetachVolume |  5747 | 4996 |          101 |       650 |
|   DescribeVolumes |  4482 | 4482 |            0 |         0 |
| DescribeInstances |  3760 | 3760 |            0 |         0 |

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -151,6 +151,7 @@ const (
 	snapshotTagBatcher
 
 	batchDescribeTimeout = 30 * time.Second
+	batchMaxDelay        = 500 * time.Millisecond // Minimizes RPC latency and EC2 API calls. Tuned via scalability tests.
 )
 
 var (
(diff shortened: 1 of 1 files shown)


torredil (approved):
/lgtm

PR #2037, opened 2024-05-13
Sanitize CSI RPC request logs

torredil:

Security enhancement

The EBS CSI Driver does not support [Token Requests](https://kubernetes-csi.github.io/docs/token-requests.html) , however - out of an abundance of caution - this PR addresses the potential vulnerability of sensitive information being logged inadvertently.

The main change is adding a new utility function `SanitizeRequest` that takes a request object and returns a copy of the request with the "Secrets" field cleared. This function creates a new instance of the same type as the input request, copies all the fields from the original request to the new instance, and sets the "Secrets" field to an empty map if it exists.

**What testing is done?** 

- `make verify && make test`
- CI 
- manual testing

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -77,7 +77,7 @@ func NewControllerService(c cloud.Cloud, o *Options) *ControllerService {
 }
 
 func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
-	klog.V(4).InfoS("CreateVolume: called", "args", *req)
+	klog.V(4).InfoS("CreateVolume: called", "args", util.SanitizeRequest(req))
 	if err := validateCreateVolumeRequest(req); err != nil {
 		return nil, err
 	}
@@ -366,7 +366,7 @@ func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
 }
 
 func (d *ControllerService) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
-	klog.V(4).InfoS("DeleteVolume: called", "args", *req)
+	klog.V(4).InfoS("DeleteVolume: called", "args", util.SanitizeRequest(req))
 	if err := validateDeleteVolumeRequest(req); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/util/util.go
```diff
@@ -144,3 +145,20 @@ func NormalizeWindowsPath(path string) string {
 	}
 	return normalizedPath
 }
+
+// ClearSecrets takes a request object and returns a copy of the request with
+// the "Secrets" field cleared.
+func ClearSecrets(req interface{}) interface{} {
```
AndrewSirenko, line 151: Can we add a simple unit test as we do for [other util functions](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/a8d4dae06a490a21f44e0dad59fb5543e2ab76af/pkg/util/util_test.go#L163)?

torredil: Added ✅

torredil:
Manually tested by following these steps:

Reproduction steps:

1. Set `controller.logLevel` = 4.

2. Give the external provisioner ClusterRole [permissions](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/clusterrole-provisioner.yaml) to get secrets:
```
  - apiGroups: [ "" ]
    resources: [ "secrets" ]
    verbs: [ "get", "list", "watch" ]
```

3. Apply the following secret via `kubectl apply -f `:
```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
  namespace: kube-system
type: Opaque
data:
  username: dXNlcm5hbWU=
  password: cGFzc3dvcmQ=
```

4. Dynamically provision a volume with the following StorageClass:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
parameters:
  csi.storage.k8s.io/provisioner-secret-name: mysecret
  csi.storage.k8s.io/provisioner-secret-namespace: kube-system
``` 

With this change, the secret is not logged.

pkg/driver/controller.go
```diff
@@ -77,7 +77,7 @@ func NewControllerService(c cloud.Cloud, o *Options) *ControllerService {
 }
 
 func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
-	klog.V(4).InfoS("CreateVolume: called", "args", *req)
+	klog.V(4).InfoS("CreateVolume: called", "args", util.ClearSecrets(req))
```
rdpsin, line 80: very minor np: maybe name this to something more generic like `SanitizeRequest`?

torredil: Great suggestion, done ✅

AndrewSirenko (approved):

PR #2040, opened 2024-05-15
Pin kubetest2 dependency

AndrewSirenko:

Bug fix

Audited dependencies for any unpinned versions. `kubetest2` was the only seen unpinned one. By pinning this we make sure no faulty upstream change breaks our CI, as has happened with previous unpinned dependencies.

**What testing is done?** 
`make clean; make tools; kubetest2 --help`


torredil (approved):
/lgtm
/approve

PR #2057, opened 2024-06-10
Increase default provisioner, resizer, snapshotter `retry-interval-max`

AndrewSirenko:
Improvement

Today, the max backoff period for any sidecar RPC is 5 minutes. This is problematic if the user has a misconfigured K8s object (such as invalid IOPS for volume type) because the invalid request will be retried to the EC2 backend every 5 minutes until user resolves the issue. 

Increase default provisioner, resizer, snapshotter `retry-interval-max` to 30 minutes. 

**What testing is done?** 
Deploy driver via helm with default values. 

```
❯ kubectl get pod -n kube-system ebs-csi-controller-9d9969b74-4mcgp -o yaml | grep "retry-interval"
    - --retry-interval-max=30m
    - --retry-interval-max=30m
    - --retry-interval-max=5m
    - --retry-interval-max=30m
```

WIP: Tested by trying to provision volume with volumetype `io3`. Waited 30 minutes and saw that CreateVolume RPCs were being made less often than every 5 min. 

```
│ csi-provisioner I0610 20:12:55.022578       1 event.go:364] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"ebs-claim", UID:"8c6abb77-0228 │
│ csi-provisioner I0610 20:22:14.562941       1 controller.go:1366] provision "default/ebs-claim" class "ebs-sc": started                                                      ...
│ csi-provisioner I0610 20:37:14.563701       1 controller.go:1366] provision "default/ebs-claim" class "ebs-sc": started                                                      
│ csi-provisioner I0610 20:37:14.563816       1 event.go:364] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"ebs-claim", UID:"8c6abb77-0228 │

```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -227,6 +227,9 @@ spec:
             - --kube-api-burst=100
             - --worker-threads=100
             {{- end }}
+            {{- if not (regexMatch "(-retry-interval-max)" (join " " .Values.sidecars.provisioner.additionalArgs)) }}
+            - --retry-interval-max=30m
+            {{- end }}
             {{- range .Values.sidecars.provisioner.additionalArgs }}
             - {{ . }}
             {{- end }}
@@ -280,6 +283,9 @@ spec:
             - --kube-api-burst=100
             - --worker-threads=100
             {{- end }}
+            {{- if not (regexMatch "(-retry-interval-max)" (join " " .Values.sidecars.attacher.additionalArgs)) }}
+            - --retry-interval-max=5m
+            {{- end }}
             {{- range .Values.sidecars.attacher.additionalArgs }}
             - {{ . }}
             {{- end }}
@@ -321,6 +327,9 @@ spec:
             - --kube-api-qps=20
             - --kube-api-burst=100
             - --worker-threads=100
+            {{- end }}
+            {{- if not (regexMatch "(-retry-interval-max)" (join " " .Values.sidecars.snapshotter.additionalArgs)) }}
+            - --retry-interval-max=30m
             {{- end }}
             {{- range .Values.sidecars.snapshotter.additionalArgs }}
             - {{ . }}
@@ -434,6 +443,9 @@ spec:
             - --kube-api-qps=20
             - --kube-api-burst=100
             - --workers=100
+            {{- end }}
+            {{- if not (regexMatch "(-retry-interval-max)" (join " " .Values.sidecars.resizer.additionalArgs)) }}
+            - --retry-interval-max=30m
             {{- end }}
             {{- range .Values.sidecars.resizer.additionalArgs }}
             - {{ . }}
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -279,6 +280,7 @@ spec:
             - --kube-api-qps=20
             - --kube-api-burst=100
             - --worker-threads=100
+            - --retry-interval-max=5m
```
ConnorJC3, line 283: Why 5 minutes for attacher and not 30 like the others? Also, isn't this the default?

ConnorJC3 (changes requested):
Need to run `make update`

charts/aws-ebs-csi-driver/templates/controller.yaml
AndrewSirenko: First this is a 2-way door. In my opinion, customer misconfigurations for volume provisioning, Snapshots, and resize/modify are common enough for volume/snapshot creation and modification, which is why I think we should bump the values for those in this PR today. We can always bump attacher retry-interval-max in a followup PR. 

My worry with attachment is that it has a higher likelihood of failures ocurring for the 10+ minutes to get past the 5 min backoff. An example that comes to mind is karpenter instance detach/termination race delay for Windows taking up to 15 minutes. We would want to retry as close as possible to end of that instance termination to prevent an even longer delay. With today's 5m backoff, we'd in the worst case take 20m to attach. With a higher backoff retry-interval-max that could stretch even longer.

On the other hand there is the common attachment misconfig of attaching too many volumes. But I think in the attach path it is better to err on the side of retrying more often than the other RPCs, even with this common misconfig.

ConnorJC3 (changes requested):
Needs to be moved out of the kube API settings sections on each sidecar, and put into its own section with a guard against `--retry-interval-max` in `additionalArgs`

Otherwise lgtm

AndrewSirenko:
> Needs to be moved out of the kube API settings sections on each sidecar, and put into its own section with a guard against --retry-interval-max in additionalArgs

Good call, @ConnorJC3. Done.

torredil (approved):
```
kubectl describe pod ebs-csi-controller-98ff6995-2rz6x -n kube-system | grep "retry-interval"  
      --retry-interval-max=30m
      --retry-interval-max=5m
      --retry-interval-max=30m
```

```
helm upgrade --install aws-ebs-csi-driver --namespace kube-system ./charts/aws-ebs-csi-driver --values ./charts/aws-ebs-csi-driver/values.yaml --set "sidecars.provisioner.additionalArgs[0]=--retry-interval-max=35m"

kubectl describe pod ebs-csi-controller-74f5c4cf6d-hpqcz -n kube-system | grep "retry-interval"
      --retry-interval-max=35m
      --retry-interval-max=5m
      --retry-interval-max=30m
```

/lgtm

PR #2059, opened 2024-06-11
Use {{ .Release.Namespace }} in ClusterRoleBinding subject namespace

etutuit:
Bug fix

PR is about making subject namespace in ClusterRoleBinding generated based on the namespace release is being deployed to.

With previous hardcoded value of kube-system the test was failing because ServiceAccount didn't have proper permissions assigned, if it was deployed to different namespace than kube-system.

What testing is done?
It was deployed into test cluster and ended up with successful execution after deploying to custom namespace.


AndrewSirenko (approved):
Thanks for the fix!

Issue #1882, opened 2023-12-24
Improve securityContext implementation (Capabilities\Seccomp)


YogevMaty:
What is this PR about? / Why do we need it?

Implements default securityContext Capabilities and Seccomp profiles for the node pod and all containers (including sidecar containers). Also add the option to manage securityContext Capabilities and Seccomp profiles in the helm chart

AndrewSirenko:
Hi @YogevMaty, did you mean to submit this as an issue?

YogevMaty:
yes, I would like that, as the securityContext of - aws-ebs-csi-driver:v1.22.0 , livenessprobe:v2.10.0-eks-1-27-9
 is only 
         privileged: true
        readOnlyRootFilesystem: true
I am not sure if you need all the k8s Capabilities or can limit only to the relevant that are in use 
same for  Seccomp profile - not sure why it is not used

devopserio:
This is definitely needed still to manage PSA restrictions - when will this functionality be added?

PR #2061, opened 2024-06-12
Set RuntimeDefault as default seccompProfile in securityContext

torredil:

Helm parameter

**Note to reviewers:** `seccompProfile` has now been set to `RuntimeDefault`, which uses the default seccomp profile provided by the container runtime. This helps to restrict the system calls the containers can make, enhancing security. Do note that this changes the current default (unspecified, thus `Unconfined`):

```
const (
	// SeccompProfileTypeUnconfined is when no seccomp profile is applied (A.K.A. unconfined).
	SeccompProfileTypeUnconfined SeccompProfileType = "Unconfined"
	// SeccompProfileTypeRuntimeDefault represents the default container runtime seccomp profile.
	SeccompProfileTypeRuntimeDefault SeccompProfileType = "RuntimeDefault"
	// SeccompProfileTypeLocalhost represents custom made profiles stored on the node's disk.
	SeccompProfileTypeLocalhost SeccompProfileType = "Localhost"
)
```

^ https://github.com/kubernetes/kubernetes/blob/6e8e1f53b09328849f655d38eae9bacbbeb3445e/pkg/apis/core/types.go#L3749C1-L3756C2 

For more information, see https://kubernetes.io/docs/tutorials/security/seccomp/.

closes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1882

**What testing is done?** 

```
make verify && make test 
CI
```

### DIFF
--- a/deploy/kubernetes/base/controller.yaml
+++ b/deploy/kubernetes/base/controller.yaml
@@ -128,6 +128,8 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: true
+            seccompProfile:
+              type: RuntimeDefault
         - name: csi-provisioner
           image: public.ecr.aws/eks-distro/kubernetes-csi/external-provisioner:v4.0.1-eks-1-30-4
           imagePullPolicy: IfNotPresent
@@ -157,6 +159,8 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: true
+            seccompProfile:
+              type: RuntimeDefault
         - name: csi-attacher
           image: public.ecr.aws/eks-distro/kubernetes-csi/external-attacher:v4.5.1-eks-1-30-4
           imagePullPolicy: IfNotPresent
@@ -183,6 +187,8 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: true
+            seccompProfile:
+              type: RuntimeDefault
         - name: csi-snapshotter
           image: public.ecr.aws/eks-distro/kubernetes-csi/external-snapshotter/csi-snapshotter:v7.0.2-eks-1-30-4
           imagePullPolicy: IfNotPresent
@@ -208,6 +214,8 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: true
+            seccompProfile:
+              type: RuntimeDefault
         - name: csi-resizer
           image: public.ecr.aws/eks-distro/kubernetes-csi/external-resizer:v1.10.1-eks-1-30-4
           imagePullPolicy: IfNotPresent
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/templates/_node.tpl
```diff
@@ -184,7 +184,17 @@ spec:
...
+              type: {{ .seccompProfile.type }}
+              {{- if eq .seccompProfile.type "Localhost" }}
+              localhostProfile: {{ .seccompProfile.localhostProfile }}
+              {{- end }}
+            {{- if .readOnlyRootFilesystem }}
+            readOnlyRootFilesystem: {{ .readOnlyRootFilesystem }}
+            {{- end }}
+            {{- if .allowPrivilegeEscalation }}
+            allowPrivilegeEscalation: {{ .allowPrivilegeEscalation }}
+            {{- end }}
```
ConnorJC3, line 197: Why are we changing the security context from the existing implementation? This is a breaking change if a user had something like this configured:

```
securityContext:
  runAsUser: 1001
  runAsGroup: 1003
  capabilities:
    add: 'SOME_CAP'
```

(And there's several others missing from the example, like `appArmorProfile` or `procMount`)

I also don't understand why we would even want to be in the business of validating the config here.

deploy/kubernetes/base/controller.yaml
```diff
@@ -126,7 +126,8 @@ spec:
               cpu: 10m
               memory: 40Mi
           securityContext:
-            allowPrivilegeEscalation: false
+            seccompProfile:
+              type: RuntimeDefault
```
ConnorJC3, line 130: Is this guaranteed to exist on all Kubernetes installations? What if the OS the user is using has SELinux disabled?

ConnorJC3 (changes requested):

charts/aws-ebs-csi-driver/templates/_node.tpl
torredil: Good call, based on this feedback I'm inclined to close this PR and keep the defaults as written today. If users wish to configure the `securityContext` beyond the current default values they have the ability to do so.

deploy/kubernetes/base/controller.yaml
torredil: > Is this guaranteed to exist on all Kubernetes installations?

It is available in all supported EKS versions (including extended support). 

> What if the OS the user is using has SELinux disabled?

My understanding is that seccomp and SELinux operate independently. That is to say, if SELinux state is disabled (not enforcing policies), seccomp profiles can still be applied.

charts/aws-ebs-csi-driver/templates/_node.tpl
torredil: Removed template parsing and left the door open there. Scoped down PR to changing the default seccompProfile config.

deploy/kubernetes/base/controller.yaml
ConnorJC3: :facepalm: this is about seccomp not selinux

Discussed offline, new version lgtm

ConnorJC3 (approved):
/lgtm

Discussed offline, this version lgtm

AndrewSirenko (approved):

PR #2082, opened 2024-07-08
Modifying Tags On Existing Volume

mdzraf:

New Feature

This feature allows for EBS CSI Driver customers to Add, Modify, and Delete tags of existing volumes via the official CSI `ControllerModifyVolume` RPC through a `VolumeAttributesClass`.

**What testing is done?** 

Besides running existing unit and e2e tests, this feature was tested manually by doing the following:

1. Make test cluster by running `make cluster/create`
2. Creating and uploading image of driver with changes to ECR Repo
3. Deploying driver to cluster via helm, with the necessary flags. 
4. Creating various testing VolumeAttributesClass documents
5. Applying VACs and verifying that the correct tag operations were performed on the volume by describing the volume and looking at it's tags as well as ensuring that there is proper error handling 

### Notes:
This PR has a few extra VolumeAttributeClass documents that **will be removed before merging,** they are currently there because **the unit tests and e2e tests for this feature are still being worked on  (this is why there is some commented out code in the testing files)** as well as to ensure that I can test any changes made after initial reviews.

### DIFF
--- a/docs/example-iam-policy.json
+++ b/docs/example-iam-policy.json
@@ -25,15 +25,7 @@
       "Resource": [
         "arn:aws:ec2:*:*:volume/*",
         "arn:aws:ec2:*:*:snapshot/*"
-      ],
-      "Condition": {
-        "StringEquals": {
-          "ec2:CreateAction": [
-            "CreateVolume",
-            "CreateSnapshot"
-          ]
-        }
-      }
+      ]
     },
     {
       "Effect": "Allow",
--- a/docs/tagging.md
+++ b/docs/tagging.md
@@ -96,6 +96,36 @@ backup=true
 billingID=ABCDEF
 ```
 
+# Adding, Modifying, and Deleting Tags Of Existing Volumes
+The AWS EBS CSI Driver supports the modifying of tags of existing volumes through `VolumeAttributesClass.parameters` the examples below show the syntax for addition, modification, and deletion of tags within the `VolumeAttributesClass.parameters`. For a walkthrough on how to apply these modifications to a volume follow the [walkthrough for Volume Modification via VolumeAttributeClass](../examples/kubernetes/modify-volume)
+
+**Syntax for Adding or Modifying a Tag**
+
+If a key has the prefix `tagSpecification`, the CSI driver will treat the value as a key-value pair to be added to the existing volume. If there is already an existing tag with the specified key, the CSI driver will overwrite the value of that tag with the new value specified. 
+```
+apiVersion: storage.k8s.io/v1alpha1
+kind: VolumeAttributesClass
+metadata:
+  name: io2-class
+driverName: ebs.csi.aws.com
+parameters:
+  tagSpecification_1: "location=Seattle"
+  tagSpecification_2: "cost-center=" // If the value is left blank, tag is created with an empty value
+```
+**Syntax for Deleting a Tag**
+
+If a key has the prefix `tagDeletion`, the CSI driver will treat the value as a tag key, and the existing tag with that key will be removed from the volume.
+```
+apiVersion: storage.k8s.io/v1alpha1
+kind: VolumeAttributesClass
+metadata:
+  name: io2-class
+driverName: ebs.csi.aws.com
+parameters:
+  tagDeletion_1: "location" // Deletes tag with key "location"
+  tagDeletion_2: "cost-center"
+```
+
 # Snapshot Tagging
 The AWS EBS CSI Driver supports tagging snapshots through `VolumeSnapshotClass.parameters`, similarly to StorageClass tagging.
 
--- a/examples/kubernetes/modify-volume/manifests/volumeattributesclass.yaml
+++ b/examples/kubernetes/modify-volume/manifests/volumeattributesclass.yaml
@@ -21,3 +21,5 @@ driverName: ebs.csi.aws.com
 parameters:
   type: io2
   iops: "10000"
+  tagSpecification_1: "location=Seattle"
+  tagSpecification_2: "cost-center="
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -229,6 +229,12 @@ type ModifyDiskOptions struct {
 	Throughput int32
 }
 
+// ModifyTagsOptions represents parameter to modify the tags of an existing EBS volume
+type ModifyTagsOptions struct {
+	TagsToAdd    map[string]string
+	TagsToDelete []string
+}
+
 // Snapshot represents an EBS volume snapshot
 type Snapshot struct {
 	SnapshotID     string
@@ -744,6 +750,42 @@ func (c *cloud) batchDescribeVolumesModifications(request *ec2.DescribeVolumesMo
 		return nil, r.Err
 	}
 	return r.Result, nil
+}
+
+// ModifyTags adds, updates, and deletes tags for the specified EBS volume.
+func (c *cloud) ModifyTags(ctx context.Context, volumeID string, tagOptions ModifyTagsOptions) error {
+	if len(tagOptions.TagsToDelete) > 0 {
+		deleteTagsInput := &ec2.DeleteTagsInput{
+			Resources: []string{volumeID},
+			Tags:      make([]types.Tag, 0, len(tagOptions.TagsToDelete)),
+		}
+		for _, tagKey := range tagOptions.TagsToDelete {
+			deleteTagsInput.Tags = append(deleteTagsInput.Tags, types.Tag{Key: aws.String(tagKey)})
+		}
+		_, deleteErr := c.ec2.DeleteTags(ctx, deleteTagsInput)
+		if deleteErr != nil {
+			klog.ErrorS(deleteErr, "failed to delete tags", "volumeID", volumeID)
+			return deleteErr
+		}
+	}
+	if len(tagOptions.TagsToAdd) > 0 {
+		createTagsInput := &ec2.CreateTagsInput{
+			Resources: []string{volumeID},
+			Tags:      make([]types.Tag, 0, len(tagOptions.TagsToAdd)),
+		}
+		for k, v := range tagOptions.TagsToAdd {
+			createTagsInput.Tags = append(createTagsInput.Tags, types.Tag{
+				Key:   aws.String(k),
+				Value: aws.String(v),
+			})
+		}
+		_, addErr := c.ec2.CreateTags(ctx, createTagsInput)
+		if addErr != nil {
+			klog.ErrorS(addErr, "failed to create tags", "volumeID", volumeID)
+			return addErr
+		}
+	}
+	return nil
 }
 
 // ResizeOrModifyDisk resizes an EBS volume in GiB increments, rounding up to the next possible allocatable unit, and/or modifies an EBS
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -2509,6 +2509,118 @@ func TestResizeOrModifyDisk(t *testing.T) {
 	}
 }
 
+func TestModifyTags(t *testing.T) {
+	validTagsToAddInput := map[string]string{
+		"key1": "value1",
+		"key2": "value2",
+		"key3": "",
+	}
+
+	validTagsToDeleteInput := []string{
+		"key1",
+		"key2",
+	}
+
+	emptyTagsToAddInput := map[string]string{}
+	emptyTagsToDeleteInput := []string{}
+
+	testCases := []struct {
+		name              string
+		volumeID          string
+		negativeCase      bool
+		modifyTagsOptions ModifyTagsOptions
+		expErr            error
+	}{
+		{
+			name:     "success normal tag addition",
+			volumeID: "mod-tag-test-name",
+			modifyTagsOptions: ModifyTagsOptions{
+				TagsToAdd: validTagsToAddInput,
+			},
+			expErr: nil,
+		},
+		{
+			name:     "success normal tag deletion",
+			volumeID: "mod-tag-test-name",
+			modifyTagsOptions: ModifyTagsOptions{
+				TagsToDelete: validTagsToDeleteInput,
+			},
+			expErr: nil,
+		},
+		{
+			name:     "success normal tag addition and tag deletion",
+			volumeID: "mod-tag-test-name",
+			modifyTagsOptions: ModifyTagsOptions{
+				TagsToAdd:    validTagsToAddInput,
+				TagsToDelete: validTagsToDeleteInput,
+			},
+			expErr: nil,
+		},
+		{
+			name:         "fail: EC2 API generic error TagsToAdd",
+			volumeID:     "mod-tag-test-name",
+			negativeCase: true,
+			expErr:       fmt.Errorf("Generic EC2 API error"),
+			modifyTagsOptions: ModifyTagsOptions{
+				TagsToAdd:    validTagsToAddInput,
+				TagsToDelete: emptyTagsToDeleteInput,
+			},
+		},
+		{
+			name:         "fail: EC2 API generic error TagsToDelete",
+			volumeID:     "mod-tag-test-name",
+			negativeCase: true,
+			expErr:       fmt.Errorf("Generic EC2 API error"),
+			modifyTagsOptions: ModifyTagsOptions{
+				TagsToAdd:    emptyTagsToAddInput,
+				TagsToDelete: validTagsToDeleteInput,
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			mockEC2 := NewMockEC2API(mockCtrl)
+			c := newCloud(mockEC2)
+
+			ctx := context.Background()
+
+			if len(tc.modifyTagsOptions.TagsToAdd) > 0 {
+				if tc.negativeCase {
+					mockEC2.EXPECT().CreateTags(gomock.Any(), gomock.Any()).Return(nil, tc.expErr).Times(1)
+				} else {
+					mockEC2.EXPECT().CreateTags(gomock.Any(), gomock.Any()).Return(&ec2.CreateTagsOutput{}, tc.expErr).Times(1)
+				}
+			}
+			if len(tc.modifyTagsOptions.TagsToDelete) > 0 {
+				if tc.negativeCase {
+					mockEC2.EXPECT().DeleteTags(gomock.Any(), gomock.Any()).Return(nil, tc.expErr).Times(1)
+				} else {
+					mockEC2.EXPECT().DeleteTags(gomock.Any(), gomock.Any()).Return(&ec2.DeleteTagsOutput{}, tc.expErr).Times(1)
+				}
+			}
+
+			err := c.ModifyTags(ctx, tc.volumeID, tc.modifyTagsOptions)
+			if err != nil {
+				if tc.expErr == nil {
+					t.Fatalf("ModifyTags() failed: expected no error, got: %v", err)
+				} else {
+					if !strings.Contains(err.Error(), tc.expErr.Error()) {
+						t.Fatalf("ModifyTags() failed: expected error %v, got: %v", tc.expErr, err)
+					}
+				}
+			} else {
+				if tc.expErr != nil {
+					t.Fatal("ModifyTags() failed: expected error, got nothing")
+				}
+			}
+
+			mockCtrl.Finish()
+		})
+	}
+}
+
 func TestGetSnapshotByName(t *testing.T) {
 	testCases := []struct {
 		name            string
--- a/pkg/driver/controller_modify_volume.go
+++ b/pkg/driver/controller_modify_volume.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/awslabs/volume-modifier-for-k8s/pkg/rpc"
@@ -39,11 +41,16 @@ const (
 	ModificationKeyIOPS = "iops"
 
 	ModificationKeyThroughput = "throughput"
+
+	ModificationAddTag = "tagSpecification"
+
+	ModificationDeleteTag = "tagDeletion"
 )
 
 type modifyVolumeRequest struct {
 	newSize           int64
 	modifyDiskOptions cloud.ModifyDiskOptions
+	modifyTagsOptions cloud.ModifyTagsOptions
 }
 
 func (d *ControllerService) GetCSIDriverModificationCapability(
@@ -68,9 +75,7 @@ func (d *ControllerService) ModifyVolumeProperties(
 		return nil, err
 	}
 
-	_, err = d.modifyVolumeCoalescer.Coalesce(name, modifyVolumeRequest{
-		modifyDiskOptions: *options,
-	})
+	_, err = d.modifyVolumeCoalescer.Coalesce(name, *options)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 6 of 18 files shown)


ConnorJC3 (commented):
Rough initial review, did not review the tests yet and is not super thorough. Will do a more thorough and comprehensive review later.

pkg/driver/controller_modify_volume.go
```diff
@@ -69,7 +75,7 @@ func (d *ControllerService) ModifyVolumeProperties(
 	}
 
 	_, err = d.modifyVolumeCoalescer.Coalesce(name, modifyVolumeRequest{
-		modifyDiskOptions: *options,
+		modifyDiskOptions: options.modifyDiskOptions,
 	})
```
ConnorJC3, line 79: I think we could just change these three lines to
```
_, err = d.modifyVolumeCoalescer.Coalesce(name, options)
```

pkg/cloud/cloud.go
```diff
@@ -746,6 +752,43 @@ func (c *cloud) batchDescribeVolumesModifications(request *ec2.DescribeVolumesMo
 	return r.Result, nil
 }
 
+// ModifyTags adds, updates, and deletes tags for the specified EBS volume.
+func (c *cloud) ModifyTags(ctx context.Context, volumeID string, tagsToAdd map[string]string, tagsToDelete []string) error {
```
ConnorJC3, line 756: `ModifyTags` should probably take a `ModifyTagOptions` instead of the maps separately

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+				}
+				return 0, status.Errorf(codes.Internal, "Could not modify volume %q: %v", volumeID, err)
+			} else {
+				return actualSizeGiB, nil
 			}
-			return 0, status.Errorf(codes.Internal, "Could not modify volume %q: %v", volumeID, err)
-		} else {
-			return actualSizeGiB, nil
 		}
+		//Any error should be caught in one of the conditional statements above, returning 0 here is correct if it doesn't go into the second conditional
```
ConnorJC3, line 166: np: change this comment to something like
```
// No change to the volume was requested, so return an empty result with no error
```

pkg/driver/controller_modify_volume.go
```diff
@@ -39,11 +40,16 @@ const (
...
+
+	ModificationAddTag = "tagSpecification"
+
+	ModificationDeleteTag = "tagDeletion"
 )
 
 type modifyVolumeRequest struct {
 	newSize           int64
 	modifyDiskOptions cloud.ModifyDiskOptions
+	modifyTagsOptions *cloud.ModifyTagsOptions
```
ConnorJC3, line 52: I think now that #2083 is merged this doesn't need to be a pointer anymore (make sure you rebase on top of `master`)

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,127 @@ func TestResizeOrModifyDisk(t *testing.T) {
 	}
 }
 
+func TestModifyTags(t *testing.T) {
+
```
ConnorJC3, line 2513: np: remove this empty line

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,127 @@ func TestResizeOrModifyDisk(t *testing.T) {
...
+			deleteTags:   true,
+			negativeCase: true,
+			expErr:       fmt.Errorf("Generic EC2 API error"),
+			ModifyTagsOptions: &ModifyTagsOptions{
+				TagsToAdd:    emptyTagsToAddInput,
+				TagsToDelete: validTagsToDeleteInput,
+			},
+		},
+	}
+	for _, tc := range testCases {
```
ConnorJC3, line 2590: np: add an empty line between the end of the test cases and the for loop to match the style of other test functions in this file

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
-			options.VolumeType = value
+			options.modifyDiskOptions.VolumeType = value
+		default:
+			if strings.HasPrefix(key, ModificationAddTag) {
+				st := strings.SplitN(value, "=", 2)
+				options.modifyTagsOptions.TagsToAdd[st[0]] = st[1]
+			} else if strings.HasPrefix(key, ModificationDeleteTag) {
+				options.modifyTagsOptions.TagsToDelete = append(options.modifyTagsOptions.TagsToDelete, value)
+			}
+
```
ConnorJC3, line 208: np: remove this empty line

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,127 @@ func TestResizeOrModifyDisk(t *testing.T) {
...
+	emptyTagsToDeleteInput := []string{}
+
+	testCases := []struct {
+		name              string
+		volumeID          string
+		createTags        bool
+		deleteTags        bool
+		negativeCase      bool
+		modifyDiskOptions *ModifyDiskOptions
+		ModifyTagsOptions *ModifyTagsOptions
```
ConnorJC3, line 2535: np: name this `modifyTagsOptions` (starting with lowercase letter)

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,127 @@ func TestResizeOrModifyDisk(t *testing.T) {
...
+		expErr            error
+		mockEC2           *MockEC2API
+		volumes           []types.Volume
```
ConnorJC3, line 2538: I don't believe `tc.mockEC2`, `tc.volumes`, or `tc.modifyDiskOptions` is used in this test, so those parameters should be removed from the struct.

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,127 @@ func TestResizeOrModifyDisk(t *testing.T) {
...
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			mockEC2 := NewMockEC2API(mockCtrl)
+			c := newCloud(mockEC2)
+
+			ctx := context.Background()
+
+			if tc.createTags {
```
ConnorJC3, line 2598: Might be better to do something like `if tc.ModifyTagsOptions.TagsToAdd != nil`, and then you could get rid of `tc.createTags` entirely

(same below with `tc.deleteTags`)

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,127 @@ func TestResizeOrModifyDisk(t *testing.T) {
...
+				if tc.negativeCase {
+					mockEC2.EXPECT().CreateTags(gomock.Any(), gomock.Any()).Return(nil, tc.expErr).Times(1)
```
ConnorJC3, line 2600:
```suggestion
					mockEC2.EXPECT().CreateTags(gomock.Eq(tc.volumeID), gomock.Eq(tc.modifyTagsOptions)).Return(nil, tc.expErr).Times(1)
```

(same suggestion for the other three mock calls in this test)

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,127 @@ func TestResizeOrModifyDisk(t *testing.T) {
...
+		{
+			name:         "fail: EC2 API generic error TagsToDelete",
+			deleteTags:   true,
+			negativeCase: true,
+			expErr:       fmt.Errorf("Generic EC2 API error"),
+			ModifyTagsOptions: &ModifyTagsOptions{
+				TagsToAdd:    emptyTagsToAddInput,
+				TagsToDelete: validTagsToDeleteInput,
+			},
+		},
```
ConnorJC3, line 2588: np: Last two test cases are missing a volume ID

pkg/driver/controller_modify_volume_test.go
```diff
@@ -25,38 +25,53 @@ import (
...
+	validType                   = "gp3"
+	validIops                   = "2000"
+	validIopsInt                = 2000
+	validThroughput             = "500"
+	validThroughputInt          = 500
+	invalidIops                 = "123.546"
+	invalidThroughput           = "one hundred"
+	validTagSpecificationInput  = "key1=tag1"
+	tagSpecificationWithNoValue = "key3="
+	validTagDeletion            = "key2"
```
ConnorJC3, line 37: Can we add an invalid tag specification test as well? For example for a driver-managed tag or something like that?

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
 			klog.InfoS("Key `volumeType` is deprecated, please use `type` instead")
-			options.VolumeType = value
+			options.modifyDiskOptions.VolumeType = value
 		case ModificationKeyVolumeType:
-			options.VolumeType = value
+			options.modifyDiskOptions.VolumeType = value
+		default:
+			if strings.HasPrefix(key, ModificationAddTag) {
+				st := strings.SplitN(value, "=", 2)
+				options.modifyTagsOptions.TagsToAdd[st[0]] = st[1]
```
ConnorJC3, line 204: Add a check here for if `len(st) < 2` (like if the user provides an invalid input without an equal sign), and add a test for that case too

ConnorJC3 (commented):
Mostly lgtm, a few comments and a bunch of nitpicks

pkg/cloud/cloud.go
```diff
@@ -746,6 +752,43 @@ func (c *cloud) batchDescribeVolumesModifications(request *ec2.DescribeVolumesMo
...
+		deleteTagsInput := &ec2.DeleteTagsInput{
+			Resources: []string{volumeID},
+			Tags:      make([]types.Tag, 0, len(tagsToDelete)),
+		}
+		for _, tagKey := range tagsToDelete {
+			deleteTagsInput.Tags = append(deleteTagsInput.Tags, types.Tag{Key: aws.String(tagKey)})
+		}
+		_, deleteErr := c.ec2.DeleteTags(ctx, deleteTagsInput)
+		if deleteErr != nil {
+			klog.ErrorS(deleteErr, "failed to delete tags on volume: "+volumeID)
```
AndrewSirenko, line 767: Nit: When using [klog.ErrorS](https://pkg.go.dev/k8s.io/klog/v2#ErrorS), we typically separate the constant description ("failed to delete tags on volume") from any values that might change depending on the situation (volumeID depends on user).

E.g.

```
klog.ErrorS(deleteErr, "failed to delete tags, "volumeID",  volumeID)
``` 
---

This would match how line 696 is formatted in CreateDisk: 

```
klog.ErrorS(error, "failed to be deleted, this may cause volume leak", "volumeID", volumeID)
```

For more info on what the `keysAndValues...` in the `ErrorS` function signature means, see [Go By Example: Variadic Functions](https://gobyexample.com/variadic-functions)

pkg/cloud/cloud.go
```diff
@@ -746,6 +752,43 @@ func (c *cloud) batchDescribeVolumesModifications(request *ec2.DescribeVolumesMo
...
+		}
+		for k, v := range tagsToAdd {
+			createTagsInput.Tags = append(createTagsInput.Tags, types.Tag{
+				Key:   aws.String(k),
+				Value: aws.String(v),
+			})
+		}
+		_, addErr := c.ec2.CreateTags(ctx, createTagsInput)
+		if addErr != nil {
+			klog.ErrorS(addErr, "failed to create tags on volume: "+volumeID)
```
AndrewSirenko, line 785: Nit: Make usage of `klog.ErrorS` consistent with rest of file like comment above instructs.

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+			}
+		}
+	}
 	return existing, nil
 }
 
+func executeModifyTagsRequest(volumeID string, options modifyVolumeRequest, c cloud.Cloud) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	if options.modifyTagsOptions != nil {
```
torredil, line 133: You can move the `options.modifyTagsOptions != nil` check to the beginning and that will avoid creating an unnecessary context when there are no tags to modify.

Also, curious to know why we are creating a new context here in the first place - is it just how its done elsewhere? should the caller provide an already existing context instead?  injecting the context as opposed to creating it inside a nested function is generally a good idea.

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+		err := c.ModifyTags(ctx, volumeID, options.modifyTagsOptions.TagsToAdd, options.modifyTagsOptions.TagsToDelete)
+		if err != nil {
+			if errors.Is(err, cloud.ErrInvalidArgument) {
+				return status.Errorf(codes.InvalidArgument, "Could not modify volume tags (invalid argument) %q: %v", volumeID, err)
+			}
+			return status.Errorf(codes.Internal, "Could not modify volume tags %q: %v", volumeID, err)
+		}
+	} else {
```
torredil, line 141: This else conditional can be removed.

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+		klog.InfoS("No tags to modify", "volumeID", volumeID)
```
torredil, line 142: Would increasing the V level at which this informational is logged be more appropriate?

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+
 		}
 	}
-
+	if err := validateExtraTags(options.modifyTagsOptions.TagsToAdd, false); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Invalid tag value: %v", err)
+	}
+	if len(options.modifyTagsOptions.TagsToAdd) == 0 && len(options.modifyTagsOptions.TagsToDelete) == 0 {
+		options.modifyTagsOptions = nil
+	}
```
torredil, line 216: Is this conditional necessary? it will already be nil.

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
 }
 
+func executeModifyTagsRequest(volumeID string, options modifyVolumeRequest, c cloud.Cloud) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	if options.modifyTagsOptions != nil {
+		err := c.ModifyTags(ctx, volumeID, options.modifyTagsOptions.TagsToAdd, options.modifyTagsOptions.TagsToDelete)
+		if err != nil {
+			if errors.Is(err, cloud.ErrInvalidArgument) {
+				return status.Errorf(codes.InvalidArgument, "Could not modify volume tags (invalid argument) %q: %v", volumeID, err)
```
AndrewSirenko, line 137: Please use `%w` instead of `%v` for errors in format strings. (Even if other parts of codebase use `%v`)

E.g.

`return status.Errorf(codes.Internal, "Could not modify volume tags %q: %w", volumeID, err)`

This is important because it allows the original error to be "wrapped" by the new error, creating a tree of errors, so that later the error can be "unwrapped" and compared with [errors.Is](https://golang.org/pkg/errors/#Is) and [errors.As](https://golang.org/pkg/errors/#As).

See this StackOverflow answer for more info on preferring %w over %v: https://stackoverflow.com/a/61287626

See [Errors GoDoc](https://pkg.go.dev/errors#pkg-overview) for an example of wrapping and unwrapping errors.

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+			}
+			return status.Errorf(codes.Internal, "Could not modify volume tags %q: %v", volumeID, err)
```
AndrewSirenko, line 139: Use %w instead of %v as instructed in above comment.

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
-			// Kubernetes sidecars treats "Invalid Argument" errors as infeasible and retries less aggressively
-			if errors.Is(err, cloud.ErrInvalidArgument) {
-				return 0, status.Errorf(codes.InvalidArgument, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
+			return 0, err
+		}
+		if (req.modifyDiskOptions.IOPS != 0) || (req.modifyDiskOptions.Throughput != 0) || (req.modifyDiskOptions.VolumeType != "") || (req.newSize != 0) {
+			actualSizeGiB, err := c.ResizeOrModifyDisk(ctx, volumeID, req.newSize, &req.modifyDiskOptions)
+			if err != nil {
+				if errors.Is(err, cloud.ErrInvalidArgument) {
+					return 0, status.Errorf(codes.InvalidArgument, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
```
AndrewSirenko, line 159: Use %w instead of %v

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+				}
+				return 0, status.Errorf(codes.Internal, "Could not modify volume %q: %v", volumeID, err)
```
AndrewSirenko, line 161: Use %w instead of %v

pkg/driver/controller_modify_volume.go
```diff
@@ -107,55 +113,106 @@ func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRe
...
+				options.modifyTagsOptions.TagsToAdd[st[0]] = st[1]
+			} else if strings.HasPrefix(key, ModificationDeleteTag) {
+				options.modifyTagsOptions.TagsToDelete = append(options.modifyTagsOptions.TagsToDelete, value)
+			}
+
 		}
 	}
-
+	if err := validateExtraTags(options.modifyTagsOptions.TagsToAdd, false); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Invalid tag value: %v", err)
```
AndrewSirenko, line 212: Please use %w instead of %v. Perhaps Cmd-F your diff to make sure all instances of this are found.

tests/e2e/modify_volume.go
```diff
@@ -60,15 +60,36 @@ var (
...
+			ShouldResizeVolume:                    false,
+			ShouldTestInvalidModificationRecovery: false,
+			ExternalResizerOnly:                   true,
+		},
+		"with new throughput, and iops annotations": {
 			CreateVolumeParameters: defaultModifyVolumeTestGp3CreateVolumeParameters,
 			ModifyVolumeAnnotations: map[string]string{
 				testsuites.AnnotationIops:       "4000",
 				testsuites.AnnotationThroughput: "150",
+				testsuites.AnnotationsTagSpec:   "key2=test2",
```
AndrewSirenko, line 77: Can we remove this line because this test is not testing tags?

AndrewSirenko (commented):
Great work on this! Left a few comments (mostly on using %w instead of %v). Thanks!

tests/e2e/modify_volume.go
```diff
@@ -123,9 +144,11 @@ var _ = Describe("[ebs-csi-e2e] [single-az] [modify-volume] Modifying a PVC", fu
 	for testName, modifyVolumeTest := range modifyVolumeTests {
 		modifyVolumeTest := modifyVolumeTest
 		Context(testName, func() {
-			It("will modify associated PV and EBS Volume via volume-modifier-for-k8s", func() {
-				modifyVolumeTest.Run(cs, ns, ebsDriver, testsuites.VolumeModifierForK8s)
-			})
+			if !modifyVolumeTest.ExternalResizerOnly {
```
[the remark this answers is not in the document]
ConnorJC3: Skipping implies there are some circumstances under which the test would run - because this test will never run, I would strongly prefer the way it's currently done.

(Skipping also would involve plumbing in the test config or some weirdness like that)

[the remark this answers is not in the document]
AndrewSirenko: If you disagree, then I'm fine with keeping it as is.

But to be fair the subtests themselves will already only run in certain circumstances. When ExternalResizerOnly is false. I believe other projects prefer using Skip() within deepest subtest over this current style for better readability.

And we already have `ExternalResizerOnly` plumbing don't we?

pkg/driver/controller_modify_volume.go
AndrewSirenko: Edit: I was wrong, gRPC status.Errorf does not let you use `%w`. Please ignore this comment in the case of `status.Errorf`. 

Leads to ugly formatting.

tests/e2e/modify_volume.go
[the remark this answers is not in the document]
ConnorJC3: Ah yes, we do, I misunderstood how we were passing the test. In that case I don't care that much, I'll leave the decision up to Rafael.

pkg/driver/controller_modify_volume.go
ConnorJC3: It's initialized up earlier in the function, but once we move `modifyTagOptions` to not be a pointer like my other comment mentions we can drop all these `nil` checks anyways.

docs/example-iam-policy.json
```diff
@@ -25,15 +25,7 @@
...
         "arn:aws:ec2:*:*:snapshot/*"
-      ],
-      "Condition": {
-        "StringEquals": {
-          "ec2:CreateAction": [
-            "CreateVolume",
-            "CreateSnapshot"
-          ]
-        }
-      }
```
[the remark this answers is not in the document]
mdzraf: As far as I am aware, there is no condition to allow CreateTags on ModifyVolume (correct me if I'm wrong), so if the customer was to use the example policy as is, tag modification will not work and they won't be authorized to perform the CreateTags operation on a volume.

pkg/driver/controller_modify_volume.go
mdzraf: I removed it completely as it was a log message that I was using when I was doing testing, there is no log message for any of the other modifications when there are none present(ex. "No IOPS to modify") so it would be inconsistent to leave it as well.

docs/example-iam-policy.json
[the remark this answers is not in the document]
torredil: The change here is removing, not adding any additional permissions.

> so if the customer was to use the example policy as is, tag modification will not work 

The existing managed policy already allows the `ec2:CreateTags` and `ec2:DeleteTags` actions on volume and snapshot resources: https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEBSCSIDriverPolicy.html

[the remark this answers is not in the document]
mdzraf: Yes but for `ec2:CreateTags` it only allows it for the specified conditions `"Condition" : {
        "StringEquals" : {
          "ec2:CreateAction" : [
            "CreateVolume",
            "CreateSnapshot"
          ]
        }
      }` This will not allow it for when we are modifying a volume, which is why I had to remove those conditions

pkg/driver/controller_modify_volume.go
mdzraf: Good catch, thank you, it was redundant to create the context here as it can just be passed in from the one already created in executeModifyVolumeRequest

tests/e2e/modify_volume.go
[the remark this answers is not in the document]
mdzraf: Decided to use the Skip() function, gives better readability

examples/kubernetes/modify-volume/manifests/volumeattributesclass.yaml
```diff
@@ -19,5 +19,7 @@ metadata:
   name: io2-class
 driverName: ebs.csi.aws.com
 parameters:
-  type: io2
-  iops: "10000"
+  tagSpecification_1: "Location=Seaport"
+  tagSpecification_2: "Level=12"
+  tagDeletion_1: "Team"
+  tagSpecification_4: "Food=Pizza"
```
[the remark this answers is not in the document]
mdzraf: Yes, I will clean up these testing VACs before merging.

ConnorJC3 (commented):
lgtm, other than

1. Docs updates
2. Cleanup extra files in `examples/kubernetes/modify-volume/manifests/` and fix the main example
3. Squash commits
4. Rebase to fix `govulncheck` issue (I'll cut a PR to fix later today)

AndrewSirenko (approved):

AndrewSirenko (approved):

docs/tagging.md
```diff
@@ -96,6 +96,37 @@ backup=true
...
+kind: VolumeAttributesClass
+metadata:
+  name: io2-class
+driverName: ebs.csi.aws.com
+parameters:
+  tagDeletion_1: "location" // Deletes tag with key "location"
+  tagDeletion_2: "cost-center"
+```
+
+
```
ConnorJC3, line 129: np: remove one of these blank lines

pkg/cloud/cloud_test.go
ConnorJC3: Monday review: this still needs to be addressed

ConnorJC3 (commented):
Review: 2 final comments, otherwise lgtm

pkg/cloud/cloud_test.go
```diff
@@ -2509,6 +2509,121 @@ func TestResizeOrModifyDisk(t *testing.T) {
...
+
+	testCases := []struct {
+		name              string
+		volumeID          string
+		negativeCase      bool
+		modifyDiskOptions *ModifyDiskOptions
+		modifyTagsOptions ModifyTagsOptions
+		expErr            error
+		mockEC2           *MockEC2API
+		volumes           []types.Volume
```
ConnorJC3, line 2535:
```suggestion
		modifyTagsOptions ModifyTagsOptions
		expErr            error
```

ConnorJC3 (commented):
Changes lgtm, needs squash

torredil (approved):
/lgtm

ConnorJC3 (commented):
/lgtm

PR #2083, opened 2024-07-08
Change coalescer InputType from `comparable` to `any`

ConnorJC3:

Bug fix (internal facing, not customer facing)

`InputType` in the coalescer shouldn't be `comparable` as we don't use it as a key anywhere. Requiring it to be `comparable` breaks coalescing with types that are not comparable such as maps, slices, etc. (for example see PR #2082 where it had to be added as a pointer instead of a map directly).

**What testing is done?** 

CI


AndrewSirenko (approved):
What a gotcha. Thank you.

torredil (approved):
/approve

ConnorJC3:
/retest

is a flake

PR #2086, opened 2024-07-12
Migrate CreateVolume response topology to standard label topology.kubernetes.io/zone

ConnorJC3:
*Is this a bug fix or adding new feature?**

Feature

See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/729#issuecomment-1942026577

**What testing is done?** 

CI/Manual

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -924,7 +924,7 @@ func newCreateVolumeResponse(disk *cloud.Disk, ctx map[string]string) *csi.Creat
 		}
 	}
 
-	segments := map[string]string{ZoneTopologyKey: disk.AvailabilityZone}
+	segments := map[string]string{WellKnownZoneTopologyKey: disk.AvailabilityZone}
 
 	arn, err := arn.Parse(disk.OutpostArn)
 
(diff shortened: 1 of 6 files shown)


ConnorJC3:
the flake experience
/retest

pkg/driver/controller.go
```diff
@@ -924,7 +924,7 @@ func newCreateVolumeResponse(disk *cloud.Disk, ctx map[string]string) *csi.Creat
 		}
```
AndrewSirenko, line 924: Question: Should we also have to change our examples documentation to the well known key like [in storage class example](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ec8718afe87806ed5627c6fec54e54240feb5289/examples/kubernetes/storageclass/manifests/storageclass.yaml#L28)?

pkg/driver/controller.go
```diff
@@ -924,7 +924,7 @@ func newCreateVolumeResponse(disk *cloud.Disk, ctx map[string]string) *csi.Creat
...
 	}
```
AndrewSirenko, line 925: Should we change our e2e tests to use new topology key, as mentioned in this TODO [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ec8718afe87806ed5627c6fec54e54240feb5289/tests/e2e/driver/ebs_csi_driver.go#L54-L55) 
?

torredil (approved):
/lgtm
/approve

PR #2089, opened 2024-07-18
add more unit tests to increase test coverage

ElijahQuinones:
This adds more units tests for more code coverage.

As driver and Cloud coverage are both below 90% coverage this PR is to add more coverage to those areas

**What testing is done?** 
Ran the unit tests to validate they pass and ran make test/coverage to verify that the coverage had increased after each unit test was added

### DIFF
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -82,7 +82,47 @@ func extractVolumeIdentifiers(volumes []types.Volume) (volumeIDs []string, volum
 	}
 	return volumeIDs, volumeNames
 }
+func TestNewCloud(t *testing.T) {
 
+	testCases := []struct {
+		name            string
+		region          string
+		awsSdkDebugLog  bool
+		userAgentExtra  string
+		batchingEnabled bool
+	}{
+		{
+			name:            "success: with awsSdkDebugLog, userAgentExtra, and batchingEnabled",
+			region:          "us-east-1",
+			awsSdkDebugLog:  true,
+			userAgentExtra:  "example_user_agent_extra",
+			batchingEnabled: true,
+		},
+		{
+			name:           "success: with only awsSdkDebugLog, and userAgentExtra",
+			region:         "us-east-1",
+			awsSdkDebugLog: true,
+			userAgentExtra: "example_user_agent_extra",
+		},
+		{
+			name:   "success: with only region",
+			region: "us-east-1",
+		},
+	}
+	for _, tc := range testCases {
+		ec2Cloud, err := NewCloud(tc.region, tc.awsSdkDebugLog, tc.userAgentExtra, tc.batchingEnabled)
+		if err != nil {
+			t.Fatalf("error %v", err)
+		}
+		ec2CloudAscloud := ec2Cloud.(*cloud)
+		assert.Equal(t, ec2CloudAscloud.region, tc.region)
+		if tc.batchingEnabled {
+			assert.NotNil(t, ec2CloudAscloud.bm)
+		} else {
+			assert.Nil(t, ec2CloudAscloud.bm)
+		}
+	}
+}
 func TestBatchDescribeVolumes(t *testing.T) {
 	testCases := []struct {
 		name     string
@@ -549,6 +589,101 @@ func executeDescribeSnapshotsTest(t *testing.T, c *cloud, snapshotIDs, snapshotN
 			}
 		default:
 			t.Errorf("Did not receive a result or an error for a request")
+		}
+	}
+}
+
+func TestCheckDesiredState(t *testing.T) {
+	testCases := []struct {
+		name           string
+		volumeId       string
+		desiredSizeGiB int32
+		options        *ModifyDiskOptions
+		expErr         error
+	}{
+		{
+			name:           "sucess: normal path",
+			volumeId:       "vol-001",
+			desiredSizeGiB: 5,
+			options: &ModifyDiskOptions{
+				VolumeType: VolumeTypeGP2,
+				IOPS:       3000,
+				Throughput: 1000,
+			},
+		},
+		{
+			name:           "failure: volume is still being expanded",
+			volumeId:       "vol-001",
+			desiredSizeGiB: 500,
+			options: &ModifyDiskOptions{
+				VolumeType: VolumeTypeGP2,
+				IOPS:       3000,
+				Throughput: 1000,
+			},
+			expErr: fmt.Errorf("volume \"vol-001\" is still being expanded to 500 size"),
+		},
+		{
+			name:           "failure: volume is still being modified to iops",
+			volumeId:       "vol-001",
+			desiredSizeGiB: 50,
+			options: &ModifyDiskOptions{
+				VolumeType: VolumeTypeGP2,
+				IOPS:       4000,
+				Throughput: 1000,
+			},
+			expErr: fmt.Errorf("volume \"vol-001\" is still being modified to iops 4000"),
+		},
+		{
+			name:           "failure: volume is still being modifed to type",
+			volumeId:       "vol-001",
+			desiredSizeGiB: 50,
+			options: &ModifyDiskOptions{
+				VolumeType: VolumeTypeGP3,
+				IOPS:       3000,
+				Throughput: 1000,
+			},
+			expErr: fmt.Errorf("volume \"vol-001\" is still being modified to type %q", VolumeTypeGP3),
+		},
+		{
+			name:           "failure: volume is still being modified to throughput",
+			volumeId:       "vol-001",
+			desiredSizeGiB: 5,
+			options: &ModifyDiskOptions{
+				VolumeType: VolumeTypeGP2,
+				IOPS:       3000,
+				Throughput: 2000,
+			},
+			expErr: fmt.Errorf("volume \"vol-001\" is still being modified to throughput 2000"),
+		},
+	}
+	for _, tc := range testCases {
+		mockCtrl := gomock.NewController(t)
+		defer mockCtrl.Finish()
+		mockEC2 := NewMockEC2API(mockCtrl)
+		c := newCloud(mockEC2)
+		cloudInstance := c.(*cloud)
+		mockEC2.EXPECT().DescribeVolumes(gomock.Any(), gomock.Any()).Return(&ec2.DescribeVolumesOutput{
+			Volumes: []types.Volume{
+				{
+					VolumeId:   aws.String("vol-001"),
+					Size:       aws.Int32(50),
+					VolumeType: types.VolumeTypeGp2,
+					Iops:       aws.Int32(3000),
+					Throughput: aws.Int32(1000),
+				},
+			},
+		}, nil)
+		_, err := cloudInstance.checkDesiredState(context.Background(), tc.volumeId, tc.desiredSizeGiB, tc.options)
+		if err != nil {
+			if tc.expErr == nil {
+				t.Fatalf("Did not expect to get an error but got %q", err)
+			} else if tc.expErr.Error() != err.Error() {
+				t.Fatalf("checkDesiredState() failed: expected error %q, got: %q", tc.expErr, err)
+			}
+		} else {
+			if tc.expErr != nil {
+				t.Fatalf("checkDesiredState() failed: expected error got nothing")
+			}
 		}
 	}
 }
@@ -1196,6 +1331,21 @@ func TestCreateDisk(t *testing.T) {
 				AvailabilityZone: defaultZone,
 			},
 			expErr: fmt.Errorf("CreateDisk: multi-attach is only supported for io2 volumes"),
+		},
+		{
+			name:       "failure: invalid VolumeType",
+			volumeName: "vol-test-name",
+			diskOptions: &DiskOptions{
+				CapacityBytes: util.GiBToBytes(1),
+				Tags:          map[string]string{VolumeNameTagKey: "vol-test", AwsEbsDriverTagKey: "true"},
+				VolumeType:    "invalidVolumeType",
+			},
+			expDisk: &Disk{
+				VolumeID:         "vol-test",
+				CapacityGiB:      1,
+				AvailabilityZone: defaultZone,
+			},
+			expErr: fmt.Errorf("invalid AWS VolumeType %q", "invalidVolumeType"),
 		},
 	}
 	for _, tc := range testCases {
--- a/pkg/driver/controller_modify_volume_test.go
+++ b/pkg/driver/controller_modify_volume_test.go
@@ -39,6 +39,97 @@ const (
 	invalidTagSpecification     = "aws:test=TEST"
 )
 
+func TestMergeModifyVolumeRequest(t *testing.T) {
+	testCases := []struct {
+		name                        string
+		input                       modifyVolumeRequest
+		existing                    modifyVolumeRequest
+		expectedModifyVolumeRequest modifyVolumeRequest
+		expectError                 bool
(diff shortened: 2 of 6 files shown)


ElijahQuinones:
This initial PR is just to get feedback from the team so the rest of the tests going forward that I write can conform to the same format.

torredil:
/ok-to-test

> This initial PR is just to get feedback from the team so the rest of the tests going forward that I write can conform to the same format.

@ElijahQuinones It's looking good!- inputs/outputs are being parameterized which makes sense and the tests are written in a table-driven manner, which is considered a best practice.

AndrewSirenko:
Looks great to me so far! +1 To eddie's points. Readable tests with no over-reliance on mocking is a great sign as well.

pkg/driver/controller.go
```diff
@@ -130,6 +130,10 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	for key, value := range req.GetParameters() {
 		switch strings.ToLower(key) {
+		// we tell user here that they should use csi.storage.k8s.io/fstype instead but there is no switch cases to handle that
+		// Suggestion to add case FSTypeKey that will set the fstype and in the case "fstype" we print the warning and still set
+		// fstype for them
+
```
ElijahQuinones, line 136: Start of discussion thread

pkg/driver/driver.go
```diff
@@ -84,6 +84,10 @@ func NewDriver(c cloud.Cloud, o *Options, m mounter.Mounter, md metadata.Metadat
 	case AllMode:
 		driver.controller = NewControllerService(c, o)
 		driver.node = NewNodeService(o, md, m, k)
+	/*TODO is the default case even possible to occur if we have a node mode that is not controller driver or all it will return
+	an error on line 71 as ValidateDriverOptions calls to ValidateMode which only allows the contoller node and all modes or it will
+	return an error
+	*/
```
ElijahQuinones, line 90: Start of discussion thread

pkg/driver/validation.go
```diff
@@ -76,6 +76,7 @@ func validateExtraTags(tags map[string]string, warnOnly bool) error {
 		if strings.HasPrefix(k, cloud.AWSTagKeyPrefix) {
 			return fmt.Errorf("Tag key prefix '%s' is reserved", cloud.AWSTagKeyPrefix)
 		}
+		//TODO possibly remove these lines below as failure condition is impossible to encounter
```
ElijahQuinones, line 79: Start of discussion thread

pkg/cloud/cloud_test.go
```diff
@@ -82,7 +82,51 @@ func extractVolumeIdentifiers(volumes []types.Volume) (volumeIDs []string, volum
...
 
+	testCases := []struct {
+		name            string
+		region          string
+		awsSdkDebugLog  bool
+		userAgentExtra  string
+		batchingEnabled bool
+	}{
+		{
+			name:            "sucess with debug,userAgentExtra and batching",
```
ConnorJC3, line 95: np: missing space after comma

pkg/cloud/cloud_test.go
```diff
@@ -82,7 +82,51 @@ func extractVolumeIdentifiers(volumes []types.Volume) (volumeIDs []string, volum
...
+			region:          "us-east-1",
+			awsSdkDebugLog:  true,
+			userAgentExtra:  "example_user_agent_extra",
+			batchingEnabled: true,
+		},
+		{
+			name:            "sucess with only debug and userAgenrExtra",
```
ConnorJC3, line 102: np: typo `userAgenrExtra`

pkg/cloud/cloud_test.go
```diff
@@ -82,7 +82,51 @@ func extractVolumeIdentifiers(volumes []types.Volume) (volumeIDs []string, volum
 	}
 	return volumeIDs, volumeNames
 }
+func TestNewCloud(t *testing.T) {
```
ConnorJC3, line 85: np: `sucess` misspelled in all these test cases

pkg/driver/controller.go
ConnorJC3: `csi.storage.k8s.io/fstype` is a special parameter that causes the `external-provisioner` to do something special, so not handling it here is intentional: https://kubernetes-csi.github.io/docs/external-provisioner.html#storageclass-parameters

We've had this `fstype` warning for so long though that it probably should be removed at some point but I don't think this PR is the appropriate place.

pkg/cloud/cloud_test.go
ConnorJC3: second np: can we put colon after success (like `success: `) for consistency with our test naming elsewhere

pkg/driver/validation_test.go
```diff
@@ -66,6 +66,13 @@ func TestValidateExtraVolumeTags(t *testing.T) {
 			},
 			expErr: fmt.Errorf("Tag key too long (actual: %d, limit: %d)", cloud.MaxTagKeyLength+1, cloud.MaxTagKeyLength),
 		},
+		{
+			name: "invaid tag: key is empty",
+			tags: map[string]string{
+				randomString(cloud.MinTagKeyLength - 1): "extra-tag-value",
```
ConnorJC3, line 72: this becomes `randomString(0)` which is always just `""`

Can we instead do `"": "extra-tag-value",` to make it obvious to readers what this test is doing?

pkg/driver/validation.go
ConnorJC3: Why are they impossible to encounter? Wouldn't something like a `$` (or any other non-allowed character) in the key/value hit the regex? Or is that checked elsewhere?

pkg/driver/driver.go
ConnorJC3: It shouldn't be but we should keep the default case as a defense in depth measure in case of bugs in the parsing code.

pkg/driver/controller_modify_volume_test.go
```diff
@@ -39,6 +39,79 @@ const (
...
 )
 
+func TestMergeModifyVolumeRequest(t *testing.T) {
+	testCases := []struct {
+		name                        string
+		input                       modifyVolumeRequest
+		existing                    modifyVolumeRequest
+		expectedModifyVolumeRequest modifyVolumeRequest
+		expectError                 bool
+	}{
```
ConnorJC3, line 49: Can we add a test case here with `expectError: false`? Something like:

```
		{
			name: "Valid merge of size and iops",
			input: modifyVolumeRequest{
				newSize: 5,
			},
			existing: modifyVolumeRequest{
				modifyDiskOptions: cloud.ModifyDiskOptions{
					IOPS: validIopsInt,
				},
			},
			expectedModifyVolumeRequest: modifyVolumeRequest{
				newSize: 5,
				modifyDiskOptions: cloud.ModifyDiskOptions{
					IOPS: validIopsInt,
				},
			},
			expectError: false,
		},
```

pkg/driver/driver_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			driver, err := NewDriver(mockCloud, tc.o, mockMounter, mockMetadataService, mockKubernetesClient)
+			if tc.hasNode && driver.node == nil {
+				t.Fatalf("Expected driver to have node but driver does not have node")
+			}
+			if tc.hasController && driver.controller == nil {
+				t.Fatalf("Expected driver to have controller but driver does not have controller")
+
```
ConnorJC3, line 88: np: remove empty line

pkg/driver/validation.go
ElijahQuinones: Following up on our in person discussion as in line with [this doc ](https://docs.aws.amazon.com/mediaconnect/latest/ug/tagging-restrictions.html#:~:text=Allowed%20characters%20are%20letters%2C%20numbers%2C%20spaces%20representable%20in%20UTF%2D8%2C%20and%20the%20following%20characters%3A%20.%20%3A%20%2B%20%3D%20%40%20_%20/%20%2D%20(hyphen).%20Amazon%20EC2%20resources%20allow%20any%20characters.) EC2 Resources can have any utf-8 valid tag although currently the regex we are using is bugged as it includes the empty string making any key or value checked count as a match no matter what. I have confirmed that ebs does fall under this category of EC2 resources by creating a volume myself and tagging it with the tag $$$ 


Screenshot 2024-08-05 at 3 43 33 PM


As such I will make a commit to remove the regex and the lines testing for the regex.

pkg/cloud/cloud_test.go
```diff
@@ -553,6 +597,101 @@ func executeDescribeSnapshotsTest(t *testing.T, c *cloud, snapshotIDs, snapshotN
...
+		{
+			name:           "failure: volume is still being expanded",
+			volumeId:       "vol-001",
+			desiredSizeGiB: 500,
+			options: &ModifyDiskOptions{
+				VolumeType: VolumeTypeGP2,
+				IOPS:       3000,
+				Throughput: 1000,
+			},
+			expErr: fmt.Errorf("volume \"vol-001\" is still being expanded to 500 size"),
```
AndrewSirenko, line 627: This hardcoding of error message in test seems like a similar code smell to me as mentioned in our backlog grooming (e.g. we are string matching on error message).

Can we wrap an error in the [modification code](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/aeb6a026f0f509b2fd47b0b7aea0b9649930eafe/pkg/cloud/cloud.go#L1701) similar to [what we do for attachErr](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/aeb6a026f0f509b2fd47b0b7aea0b9649930eafe/pkg/cloud/cloud.go#L955), and then check that with [errors.is()](https://pkg.go.dev/errors#Is) in unit test?

AndrewSirenko (changes requested):

pkg/cloud/cloud_test.go
ConnorJC3: What error would we wrap? This is the original error source.

ConnorJC3 (commented):
Needs squash, otherwise lgtm

pkg/cloud/cloud_test.go
AndrewSirenko: The original source hardcodes custom errors instead of wrapping a ModifyExistingOperationErr as we do for things like attachErr. 

At the source we should create a ModifyExistingOperation err, and wrap that with any variable text (volume-id and parameter size)

Then in the test check for that error, instead of this brittle hardcoding of volume id and exact string matching.

pkg/driver/validation.go
ElijahQuinones: The doc actual mentions ec2 resources can use **any** character not just any utf-8 which I had assumed when reading it at first to further test I found a list of non utf-8 characters and they did in fact work as a tag not sure if there is any niche edge cases that we are missing though .
Screenshot 2024-08-06 at 9 58 03 AM
****

pkg/cloud/cloud_test.go
ConnorJC3: I don't understand why that method is "less brittle" than the current, and it causes the test to be less accurate.

AndrewSirenko: Brittleness here means the unit test would fail if we change something not as important, like the exact wording of the error message.

A less brittle test would make sure a "ModificationInProgress" err is returned here, but not need the error message to match an exact string.  

From one of my favorite resources, [Google Testing Blog: How I Learned To Stop Writing Brittle Tests](https://testing.googleblog.com/2024/04/how-i-learned-to-stop-writing-brittle.html):

"A [“brittle” test](https://abseil.io/resources/swe-book/html/ch12.html#preventing_brittle_tests) is one that fails not for real problems that would break in production, but because the test itself is fragile for innocuous reasons. Error messages, changing the order of metadata headers in a web request, or the order of calls to a heavily-mocked dependency can often cause a brittle test to fail."

ConnorJC3: Changing the error message is breakage though, because the error message is user facing.

For example, if the function suddenly started returning the error `volume \"(nil)\" is still being expanded to -1 size` we would want to know that and fail the test.

AndrewSirenko: If we want to catch "user facing err missing volume-id", what if we have the best of both worlds with a test util function like:

func ValidateErr(testErr, expectedErr errors.error, otherExpectedStrings... string) {
    if !errors.Is(testErr, expectedErr) {
        return false
    }
    for _, s := range otherExpectedStrings {
        if !strings.Contains(testErr, s) { return false }
    }
    return true
}

ConnorJC3: The `errors.Is` there appears to be circular - it's checking the error is a `ModificationInProgress` error, but the only reason that error type was added is for the test. The fact it is a `ModificationInProgress` error does not matter and will not affect any other code - it's a requirement added purely by the test.

(That would actually be a great example of a brittle test, because someone that notices that `ModificationInProgress` is unused and refactors it out of the code will suddenly be faced with a failing test)

pkg/cloud/cloud_test.go
```diff
@@ -82,7 +82,51 @@ func extractVolumeIdentifiers(volumes []types.Volume) (volumeIDs []string, volum
...
 
+	testCases := []struct {
+		name            string
+		region          string
+		awsSdkDebugLog  bool
+		userAgentExtra  string
+		batchingEnabled bool
+	}{
+		{
+			name:            "success: with debug, userAgentExtra, and batching",
```
torredil, line 95: np: Specify the full parameter names as done for `userAgentExtra`
- `debug` -> `awsSdkDebugLog`
- `batching` -> `batchingEnabled`

pkg/cloud/cloud_test.go
```diff
@@ -82,7 +82,51 @@ func extractVolumeIdentifiers(volumes []types.Volume) (volumeIDs []string, volum
...
+			awsSdkDebugLog:  true,
+			userAgentExtra:  "example_user_agent_extra",
+			batchingEnabled: false,
+		},
+		{
+			name:            "success: with only region",
+			region:          "us-east-1",
+			awsSdkDebugLog:  false,
+			userAgentExtra:  "",
+			batchingEnabled: false,
```
torredil, line 113: Lines 111-113 can be removed - in Go, the default value for boolean is `false`, and for strings, it's an empty string `""`.

pkg/cloud/cloud_test.go
AndrewSirenko: Gotcha, would it be helpful to rephrase my desires that the unit test be verifying:

- If volume modification is already in progress, confirm SOME error is returned.
  - Because we want to protect established contract in the least specific way
- Error message has volume-id + relevant mutable parameter somewhere
  - Because it will flag us accidentally screwing up customer facing message

I'm fine with not having a match on `ModificationInProgress`, I just wanted to reduce unit test excessive verification  from "Exact error" to "shared error like attachErr`. I think your point of "let's make sure volume-id is in msg" is an even better way to reduce that excessive verification to just something that matters to customer. 

Would you be aligned with ^^ @ConnorJC3?

AndrewSirenko: Offline thread occurred, it seems like ^^ would not be best option.

I'm resolving this thread and will bring up a clearer version of my thoughts on verification bounds in unit tests through our next team retro!

Thanks @ElijahQuinones @ConnorJC3 for the learnings.

AndrewSirenko (approved):

AndrewSirenko (approved):

PR #2093, opened 2024-07-22
Supporting String Interpolation for Modifying Tags On Existing Volumes Through VAC

mdzraf:

New feature

This will add the support for runtime string interpolation on tag values for a volume upon modification, which allows for customers to specify placeholder values for the PVC namespace, PVC name and PV name, which will then be dynamically computed at runtime.

**What testing is done?** 

Unit tests were ran to ensure that input was parsed properly and manual testing was done after initial changes to `external-resizer` sidecar code were made to pass in the extra metadata necessary to the driver and ensuring that these tags were properly added at runtime.

### DIFF
--- a/pkg/driver/controller_modify_volume.go
+++ b/pkg/driver/controller_modify_volume.go
@@ -28,6 +28,7 @@ import (
 	"github.com/awslabs/volume-modifier-for-k8s/pkg/rpc"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/coalescer"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util/template"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/klog/v2"
@@ -168,6 +169,8 @@ func parseModifyVolumeParameters(params map[string]string) (*modifyVolumeRequest
 			TagsToDelete: make([]string, 0),
 		},
 	}
+	var scTags []string
+	tProps := new(template.PVProps)
 	for key, value := range params {
 		switch key {
 		case ModificationKeyIOPS:
@@ -191,20 +194,29 @@ func parseModifyVolumeParameters(params map[string]string) (*modifyVolumeRequest
 			options.modifyDiskOptions.VolumeType = value
 		case ModificationKeyVolumeType:
 			options.modifyDiskOptions.VolumeType = value
+		case PVCNameKey:
+			tProps.PVCName = value
+		case PVCNamespaceKey:
+			tProps.PVCNamespace = value
+		case PVNameKey:
+			tProps.PVName = value
 		default:
 			if strings.HasPrefix(key, ModificationAddTag) {
-				st := strings.SplitN(value, "=", 2)
-				if len(st) < 2 {
-					return nil, status.Errorf(codes.InvalidArgument, "Invalid tag specification: %v", st)
-				}
-				options.modifyTagsOptions.TagsToAdd[st[0]] = st[1]
+				scTags = append(scTags, value)
 			} else if strings.HasPrefix(key, ModificationDeleteTag) {
 				options.modifyTagsOptions.TagsToDelete = append(options.modifyTagsOptions.TagsToDelete, value)
 			}
 		}
 	}
-	if err := validateExtraTags(options.modifyTagsOptions.TagsToAdd, false); err != nil {
+	addTags, err := template.Evaluate(scTags, tProps, false)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Error interpolating the tag value: %v", err)
+	}
+	if err := validateExtraTags(addTags, false); err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid tag value: %v", err)
+	}
+	for k, v := range addTags {
+		options.modifyTagsOptions.TagsToAdd[k] = v
 	}
 	return &options, nil
 }
--- a/pkg/driver/controller_modify_volume_test.go
+++ b/pkg/driver/controller_modify_volume_test.go
@@ -45,6 +45,7 @@ func TestParseModifyVolumeParameters(t *testing.T) {
 		params          map[string]string
 		expectedOptions *modifyVolumeRequest
 		expectError     bool
+		pvcName         string
 	}{
 		{
 			name:   "blank params",
@@ -63,7 +64,9 @@ func TestParseModifyVolumeParameters(t *testing.T) {
 				ModificationKeyIOPS:       validIops,
 				ModificationKeyThroughput: validThroughput,
 				ModificationAddTag:        validTagSpecificationInput,
+				ModificationAddTag + "_1": "key2={{ .PVCName }}",
 				ModificationDeleteTag:     validTagDeletion,
+				PVCNameKey:                "ebs-claim",
 			},
 			expectedOptions: &modifyVolumeRequest{
 				modifyDiskOptions: cloud.ModifyDiskOptions{
@@ -74,6 +77,7 @@ func TestParseModifyVolumeParameters(t *testing.T) {
 				modifyTagsOptions: cloud.ModifyTagsOptions{
 					TagsToAdd: map[string]string{
 						"key1": "tag1",
+						"key2": "ebs-claim",
 					},
 					TagsToDelete: []string{
 						"key2",


examples/kubernetes/modify-volume/manifests/volumeattributesclass.yaml
```diff
@@ -22,4 +22,4 @@ parameters:
   type: io2
   iops: "10000"
   tagSpecification_1: "location=Seattle"
-  tagSpecification_2: "cost-center="
+  tagSpecification_2: "cost-center="
```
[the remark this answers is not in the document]
AndrewSirenko: Something like

```
git checkout upstream/master -- examples/kubernetes/modify-volume/manifests/volumeattributesclass.yaml
git add examples/kubernetes/modify-volume/manifests/volumeattributesclass.yaml
git commit --amend --no-edit
```

pkg/driver/controller_modify_volume_test.go
```diff
@@ -64,7 +65,9 @@ func TestParseModifyVolumeParameters(t *testing.T) {
 				ModificationKeyIOPS:       validIops,
 				ModificationKeyThroughput: validThroughput,
 				ModificationAddTag:        validTagSpecificationInput,
+				ModificationAddTag + "_1": "key2={{ .PVCName }}",
```
AndrewSirenko, line 68: Is there a reason we are only testing modifyVolume interpolation on PVCName? Not PVCNamespace and PVName? 

Is this because we already trust our unit tests for string interpolation elsewhere? I would consider adding. 

image

pkg/driver/controller_modify_volume.go
```diff
@@ -191,22 +194,31 @@ func parseModifyVolumeParameters(params map[string]string) (*modifyVolumeRequest
...
+	addTags, err := template.Evaluate(scTags, tProps, false)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Error interpolating the tag value: %v", err)
+	}
+	if err := validateExtraTags(addTags, false); err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid tag value: %v", err)
 	}
+	for k, v := range addTags {
+		options.modifyTagsOptions.TagsToAdd[k] = v
+	}
```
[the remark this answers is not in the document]
mdzraf: Good catch, thanks.

torredil (approved):
/lgtm

ConnorJC3 (commented):
lgtm other than squash

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -106,7 +106,7 @@ sidecars:
     image:
       pullPolicy: IfNotPresent
       repository: public.ecr.aws/eks-distro/kubernetes-csi/external-resizer
-      tag: "v1.12.0-eks-1-31-11"
+      tag: "v1.13.1-eks-1-31-13"
```
ElijahQuinones, line 109: Can we use the latest version if we are bumping this anyway 

`v1.13.1-eks-1-32-5`

https://gallery.ecr.aws/eks-distro/kubernetes-csi/external-resizer

deploy/kubernetes/base/controller.yaml
```diff
@@ -224,7 +224,7 @@ spec:
             seccompProfile:
               type: RuntimeDefault
         - name: csi-resizer
-          image: public.ecr.aws/eks-distro/kubernetes-csi/external-resizer:v1.12.0-eks-1-31-11
+          image: public.ecr.aws/eks-distro/kubernetes-csi/external-resizer:v1.13.1-eks-1-31-13
```
ElijahQuinones, line 227: Same as above comment 

can we use `v1.13.1-eks-1-32-5`

deploy/kubernetes/overlays/stable/gcr/kustomization.yaml
```diff
@@ -19,7 +19,7 @@ images:
     newTag: v8.2.0
   - name: public.ecr.aws/eks-distro/kubernetes-csi/external-resizer
     newName: registry.k8s.io/sig-storage/csi-resizer
-    newTag: v1.12.0
+    newTag: v1.13.0
```
ElijahQuinones, line 22: Shouldn't this be `v1.13.1` or does patch version not matter here ?

charts/aws-ebs-csi-driver/values.yaml
mdzraf: Did not know we were bumping to 1-32 as everything else is still 1-31 at the time of this commit, but if that is the case then it would make sense to bump it to 1-32, will do that next rev.

deploy/kubernetes/overlays/stable/gcr/kustomization.yaml
mdzraf: Seems that it does not given that the rest of them in the file only have major versions listed. @torredil or @AndrewSirenko is this correct?

charts/aws-ebs-csi-driver/values.yaml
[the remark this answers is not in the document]
mdzraf: At the time of commit, the version is 1.31, but if we are bumping it up, I can bump it for the next rev, unless there's any reason I shouldn't. Thank you!

deploy/kubernetes/overlays/stable/gcr/kustomization.yaml
mdzraf: Got it, will fix, thank you!

ElijahQuinones:
/lgtm 
Thank you !

AndrewSirenko (approved):

ElijahQuinones (approved):

docs/tagging.md
```diff
@@ -97,8 +97,24 @@ billingID=ABCDEF
...
-
+The AWS EBS CSI Driver supports the modifying of tags of existing volumes through `VolumeAttributesClass.parameters` the examples below show the syntax for addition, modification, and deletion of tags within the `VolumeAttributesClass.parameters`. The driver (in v1.40.0 and later) also supports runtime string interpolation on tag values for a volume upon modification, which allows the specification of placeholder values for the PVC namespace, PVC name, and PV name, which will then be dynamically computed at runtime. **Note: Interpolated tags require the `--extra-modify-metadata` flag to be enabled on the `external-resizer` sidecar. To modify Amazon EBS resource tags through VACs, ensure that you attach the following IAM Policy to the role used by your Amazon EBS CSI driver:**
```
mdzraf, line 100: Format good here?

charts/aws-ebs-csi-driver/values.schema.json
```diff
@@ -261,6 +261,11 @@
           "description": "If set, add pv/pvc metadata to plugin create requests as parameters.",
           "default": true
         },
+        "extraModifyMetadata": {
+          "type": "boolean",
+          "description": "If set, add pv/pvc metadata to plugin modify requests as parameters.",
+          "default": true
+        },
```
torredil, line 268: Thoughts on _not_ introducing this parameter? It seems unlikely that a user would want to enable extra metadata for the provisioner but also disable it for the resizer.

This is not intuitive, but the snapshotter already uses the existing parameter as a way to expose the toggle:

```
- --v={{ .Values.sidecars.snapshotter.logLevel }}
{{- if .Values.controller.extraCreateMetadata }}
- --extra-create-metadata
...
```

so thats one more data point to consider.

torredil: `extraCreateMetadata` was introduced 4 years ago, if we had a time machine I would go back and suggest that we name it `extraMetadata` to future-proof it.

mdzraf: I think it makes sense to use the already existing parameter as well, I added a separate one to give customers the ability to disable it for the resizer only (in the rare case this happens as you mentioned), would like to get thoughts from @ElijahQuinones and @AndrewSirenko on this

docs/tagging.md
torredil: np: For readability, please add a newline before `**Note` section.

charts/aws-ebs-csi-driver/values.schema.json
AndrewSirenko: +1 on not adding a new parameter just for resizer case that we would have to test and maintain. We should re-use extraCreateMetadata (despite the now confusing name). 

I'm surprised we added a parameter with the ability to disable this in the first place. 

Thanks @mdzraf

docs/tagging.md
AndrewSirenko: Honestly I'd prefer moving the string-interpolation create-volume block in lines 41-64 to a separate `## String Interpolation` section. And then cut the "The driver (in v1.40.0 and later) also supports runtime string interpolation" sentence from this paragraph. 

In this new string interpolation section we can mention that this string interpolation feature can be used during dynamic provisioning, volume modification via vac, and for snapshots. That way we can avoid repeating ourselves in three separate places. 

But what do you think? 

---

+1 on separating the IAM policy note to a separate paragraph.

mdzraf: Offline Discussion Decision: Removed version support specification + separated note to newline. We will keep the string interpolation docs/examples separate for snapshots and existing volumes due to their usage examples still being different.

charts/aws-ebs-csi-driver/values.schema.json
mdzraf: Agreed that we will **not** be adding a new parameter.

ElijahQuinones: +1 to not add a new parameter as well.

AndrewSirenko (approved):

PR #2094, opened 2024-07-22
bump dependencies for driver release v1.33.0

ElijahQuinones:
new feature

This PR is to bump the dependencies for the v1.33.0 release 

**What testing is done?** 
`make update && make verify && make test`


pkg/driver/controller.go
```diff
@@ -951,7 +952,7 @@ func newCreateVolumeResponse(disk *cloud.Disk, ctx map[string]string) *csi.Creat
 	}
 }
 
-func newCreateSnapshotResponse(snapshot *cloud.Snapshot) (*csi.CreateSnapshotResponse, error) {
```
AndrewSirenko, line 954: Why are moving the error out of this function and into the return statement of other functions? Does some linter complain?

AndrewSirenko (approved):

pkg/driver/controller.go
ElijahQuinones: make verify was throwing an error as the error from the return is always nil

torredil (approved):
/lgtm

PR #2103, opened 2024-07-31
Ensure ModifyVolume returns InvalidArgument error code if VAC contains invalid parameter

### LINKED ISSUE #2099: ModifyVolume should not modify PVC via VAC containing unsupported parameters
/kind bug

**What happened?**

When attempting [Volume Modification via Volume Attribute Class (VAC)](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/modify-volume#volume-modification-via-volumeattributesclass), modifying my volume with no VAC to have the following VAC led to VolumeModifySuccessful, despite `fakeParam` not being a real mutable parameter for EBS Volumes. 

```
apiVersion: storage.k8s.io/v1alpha1
kind: VolumeAttributesClass
metadata:
  name: io2-class
driverName: ebs.csi.aws.com
parameters:
  type: io2
  iops: "10000"
  fakeParam: "20"
```

[CSI Spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllermodifyvolume-errors) notes that `INVALID_ARGUMENT` gRPC error code must be returned when CO has specified mutable parameters not supported by the volume. 

I would expect that mutable_parameter that EBS CSI Driver does not explicitly support should yield an `INVALID_ARGUMENT` error. This will be extra important to prevent fake volume modifications if EBS releases a elastic volume property that older versions of EBS CSI Driver do not actually use, because it would lead to volumes that were not actually modified. 

Follow [Volume Modification via Volume Attribute Class (VAC)](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/modify-volume#volume-modification-via-volumeattributesclass) example but with an edited volumeattributesclass.yaml

**Environment**
- Kubernetes version (use `kubectl version`): 1.29.6
- Driver version: v1.32.0

mdzraf:
Bug fix

Closes #2099 

**What testing is done?** 
Manual testing was done to verify that volume modification does not succeed and error is returned if non-supported parameter is inputted in VAC


AndrewSirenko (approved):

torredil (approved):
/lgtm

tests/e2e/testsuites/modify_volume_tester.go
```diff
@@ -57,6 +57,15 @@ func (modifyVolumeTest *ModifyVolumeTest) Run(c clientset.Interface, ns *v1.Name
...
 	defer testVolume.Cleanup()
 
+	parameterWithPrefix := func(prefix string, parameters map[string]string) map[string]string {
+		result := make(map[string]string)
+		for key, value := range parameters {
+			result[prefix+key] = value
+		}
+		return result
+	}
+	parametersWithPrefix := parameterWithPrefix("ebs.csi.aws.com/", modifyVolumeTest.ModifyVolumeParameters)
```
torredil, line 67: Would it make more sense to move this logic into `AnnotatePvc`? we can return the result from there and then pass that into `WaitForPvToModify` and so on as needed. We already loop through all the parameters there so I think handling the prefix as well would be easier to reason about and slightly more efficient.

Besides this query, changes lgtm.

AndrewSirenko (commented):
lgtm other than @torredil and @ConnorJC3's comments

I agree that the annotation logic should be moved into a util/helper method in order to make test easier to read. 

Thanks!

ConnorJC3 (commented):
Needs squash otherwise lgtm

torredil (approved):
/lgtm
/approve

PR #2104, opened 2024-07-31
Update to 1.32.0 and add timescale-specific image builds

graveland:
Dockerfile.timescale creates an image that contains the CSI node driver, the liveness probe, and the node driver registrar all in the same image.

I ran `TAG=v1.25.0-dev make timescale-images` and `TAG=v1.32.0-dev make timescale-images` using this branch, and the resulting images have been pushed to our ECR.

If we use this:
- there's only a single pull per node
- we have to update the daemonset for the node driver to use this image, and also start specifying the command to run instead of leaving it up to the container's ENTRYPOINT.

The commands are:
- /livenessprobe
- /csi-node-driver-registrar
- /aws-ebs-csi-driver

I updated the actual code, but I didn't actually use the code to build. Since @andrewcharlton's PR `Change modifyVolumeRequestHandlerTimeout to a configurable option (#1915)` has been merged, we don't really need to actually build this right now. The Dockerfile just imports the binaries from the upstream repositories.


graveland:
Ack, sorry for the noise!

PR #2106, opened 2024-08-09
Add nodeComponentOnly parameter to helm chart

AndrewSirenko:

/feature

Add a nodeComponentOnly parameter to helm chart.

**What testing is done?** 

Here are helm template results with and without setting nodeComponentOnly=true, as well as the diff (as textfiles because gists don't support yaml):

[nodeComponentOnlyTemplate.txt](https://github.com/user-attachments/files/16565770/controllerDisabledTemplate.txt)
[defaultChart.txt](https://github.com/user-attachments/files/16565771/defaultChart.txt)
[diff.txt](https://github.com/user-attachments/files/16565773/diff.txt)

Also deployed the driver with this setting:

```
helm upgrade \
  --namespace kube-system \
  $HOME/workplace/aws-ebs-csi-driver/charts/aws-ebs-csi-driver \
  --values "$HOME/workplace/tools/driver-testing/driver-up/custom_values.yaml" \
  --set nodeComponentOnly=true

Release "aws-ebs-csi-driver" does not exist. Installing it now.
NAME: aws-ebs-csi-driver
LAST DEPLOYED: Fri Aug  9 18:51:28 2024
NAMESPACE: kube-system
STATUS: deployed
REVISION: 1
...

❯ kubectl get pod -n kube-system -l "app.kubernetes.io/name=aws-ebs-csi-driver,app.kubernetes.io/instance=aws-ebs-csi-driver"
NAME                 READY   STATUS    RESTARTS   AGE
ebs-csi-node-4cxvv   3/3     Running   0          12s
ebs-csi-node-57f2r   3/3     Running   0          12s
ebs-csi-node-kss7k   3/3     Running   0          12s
ebs-csi-node-tq9sv   3/3     Running   0          12s
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-ebs-csi-driver/templates/csidriver.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.nodeComponentOnly -}}
 apiVersion: {{ ternary "storage.k8s.io/v1" "storage.k8s.io/v1beta1" (semverCompare ">=1.18.0-0" .Capabilities.KubeVersion.Version) }}
 kind: CSIDriver
 metadata:
@@ -10,3 +11,4 @@ spec:
   {{- if not .Values.useOldCSIDriver }}
   fsGroupPolicy: File
   {{- end }}
+{{- end }}
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -476,6 +476,8 @@ volumeSnapshotClasses: []
 # Intended for use with older clusters that cannot easily replace the CSIDriver object
 # This parameter should always be false for new installations
 useOldCSIDriver: false
+# Deploy EBS CSI Driver without controller and associated resources
+nodeComponentOnly: false
 helmTester:
   enabled: true
   # Supply a custom image to the ebs-csi-driver-test pod in helm-tester.yaml
--- a/charts/aws-ebs-csi-driver/templates/clusterrole-attacher.yaml
+++ b/charts/aws-ebs-csi-driver/templates/clusterrole-attacher.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.nodeComponentOnly -}}
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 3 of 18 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -476,6 +476,8 @@ volumeSnapshotClasses: []
 # Intended for use with older clusters that cannot easily replace the CSIDriver object
 # This parameter should always be false for new installations
 useOldCSIDriver: false
+# Deploy EBS CSI Driver without controller and associated resources
+disableController: false
```
AndrewSirenko, line 480: Anyone got a better name or location for this? I believe we need to have named such that it defaults to false due to internal reasons.

My other thought was top level `NodeDaemonsetOnly` OR a `controller.disabled`

Either way I agree that having it default to false makes all the if not's in each template ugly :(

charts/aws-ebs-csi-driver/templates/csidriver.yaml
```diff
@@ -10,3 +11,4 @@ spec:
   {{- if not .Values.useOldCSIDriver }}
   fsGroupPolicy: File
   {{- end }}
+{{- end }}
```
torredil, line 14: If there is a requirement to skip installing this, `disableController` might not be the appropriate knob for it.

The CSIDriver object should still be deployed even when the controller component is disabled. The developer [docs](https://kubernetes-csi.github.io/docs/csi-driver-object.html#csidriver-object) mention its used to specify how Kubernetes should interact with a driver (not just in the context of the controller plugin).

AndrewSirenko: If the parameter was renamed to `NodeDaemonsetOnly` would that change your answer? 

> The CSIDriver object should still be deployed even when the controller component is disabled.

I'm trying to prevent the case where there is a clash because the CSIDriver resource (along with controller and all other resources) is already deployed.

Consider the following workflow:

1. Install driver (Controller + a few special nodes) 
2. Install driver in NodeOnly mode

If we don't avoid deploying csidriver object in either of the two, there would be a clash, right?

AndrewSirenko:
/hold

For better name. Leaning towards `nodeDriverOnly`

charts/aws-ebs-csi-driver/values.yaml
ElijahQuinones: I like controller.disabled or controllerIsDisabled

AndrewSirenko:
/unhold

Name found

torredil (approved):
/lgtm
/approve

Issue #2105, opened 2024-08-02
Incorrect allocatable volumes count in csinode for AWS vt1*/g4* instance types


mpatlasov:
/kind bug

**What happened?**

`kubectl get csinode <node-name> -o json | jq .spec.drivers` says that allocatable.count is 26 for vt1* instance types and 25 for g4* ones. While actual number of volumes that can be attached to the node is smaller:

type / reported / actual
g4dn.xlarge / 25 / 24
g4ad.xlarge / 25 / 24
vt1.3xlarge / 26 / 24
vt1.6xlarge / 26 / 22

There are many other g4* instance types mentioned [here](https://aws.amazon.com/ec2/instance-types/g4/), but I verified the issue only for g4dn.xlarge and g4ad.xlarge. Reported number for vt1.24xlarge (26) is correct, while numbers for other vt1* types are not.

**What you expected to happen?**

`kubectl get csinode` must report correct max number of volumes to be attached.

**How to reproduce it (as minimally and precisely as possible)?**

Apply the following StatefulSet with 26 replicas:
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: statefulset-vol-limit
spec:
  serviceName: "my-svc"
  replicas: 26
  selector:
    matchLabels:
      app: my-svc
  template:
    metadata:
      labels:
        app: my-svc
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: kubernetes.io/hostname
                operator: In
                values:
                - <node-name>
      containers:
      - name: fedora
        image: registry.fedoraproject.org/fedora-minimal
        command:
        - "sleep"
        - "604800"
        volumeMounts:
        - name: data
          mountPath: /mnt/storage
      tolerations:
        - key: "node-role.kubernetes.io/master"
          effect: "NoSchedule"
  volumeClaimTemplates:
  - metadata:
      name: data
    spec:
      accessModes: [ "ReadWriteOnce" ]
      resources:
        requests:
          storage: 1Gi
```

In a while some pods get stuck at "ContainerCreating" status caused by volumes stuck at attaching status and couldn't be attached to the node. An error for a pod which got stuck looks like that:
```
$ oc describe po statefulset-vol-limit-22
...
[...]

PR #2108, opened 2024-08-12
Fix gpus not being considered when counting allocatables

ElijahQuinones:
bug fix to address part of gh [#2105](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2105) gpus not being considered when counting allocatables.

We need this PR in order to fix the above referenced bug

**What testing is done?** 
Compared manually to [list of gpu instances](https://aws.amazon.com/ec2/instance-types/)
and created unit tests for instances mentioned in ticket and for a gpu instance with 4 gpus.

### DIFF
--- a/hack/generate-gpu-count-table.sh
+++ b/hack/generate-gpu-count-table.sh
@@ -1,0 +1,32 @@
+# Copyright 2024 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+BIN="$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../bin"
+
+function get_gpus_for_region() {
+  REGION="${1}"
+  echo "Getting gpu counts for ${REGION}..." >&2
+  "${BIN}/aws" ec2 describe-instance-types --region "${REGION}" --query "InstanceTypes[?GpuInfo!=null].[InstanceType, GpuInfo]" |
+    jq -r 'map("\"" + .[0] + "\": " + (.[1].Gpus | map(.Count) | add | tostring) + ",") | .[]'
+}
+
+function get_all_gpus() {
+  "${BIN}/aws" account list-regions --max-results 50 | jq -r '.Regions | map(.RegionName) | .[]' | while read REGION; do
+    get_gpus_for_region $REGION
+  done
+}
+
+get_all_gpus | sort | uniq
--- a/hack/generate-gpu-count-table.sh
+++ b/hack/generate-gpu-count-table.sh
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Generates gpu table for `pkg/cloud/volume_limits.go` from the AWS API
+# Ensure you are opted into all opt-in regions before running
+# Ensure your account isn't in any private instance type betas before running
+
 set -euo pipefail
 
 BIN="$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../bin"
@@ -25,7 +29,8 @@ function get_gpus_for_region() {
 
 function get_all_gpus() {
   "${BIN}/aws" account list-regions --max-results 50 | jq -r '.Regions | map(.RegionName) | .[]' | while read REGION; do
-    get_gpus_for_region $REGION
+    sleep 1
+    get_gpus_for_region $REGION &
   done
 }
 
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -138,6 +138,7 @@ func GetDedicatedLimitForInstanceType(it string) int {
 	}
 }
 
+// GetReservedSlotsForInstanceType calculates how many attachment slots are already used up by other devices on shared EBS volume limit instances.
 func GetReservedSlotsForInstanceType(it string) int {
 	total := 0
 	nvmeInstanceStoreVolumes, ok := nvmeInstanceStoreVolumes[it]
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -138,11 +138,17 @@ func GetDedicatedLimitForInstanceType(it string) int {
 	}
 }
 
-func GetNVMeInstanceStoreVolumesForInstanceType(it string) int {
-	if v, ok := nvmeInstanceStoreVolumes[it]; ok {
-		return v
+func GetReservedSlotsForInstanceType(it string) int {
+	total := 0
+	nvmeInstanceStoreVolumes, ok := nvmeInstanceStoreVolumes[it]
+	if ok {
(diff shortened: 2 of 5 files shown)


pkg/driver/node_test.go
```diff
@@ -1183,6 +1183,57 @@ func TestGetVolumesLimit(t *testing.T) {
...
+			metadataMock: func(ctrl *gomock.Controller) *metadata.MockMetadataService {
+				m := metadata.NewMockMetadataService(ctrl)
+				m.EXPECT().GetRegion().Return("us-west-2")
+				m.EXPECT().GetInstanceType().Return("g4ad.xlarge")
+				m.EXPECT().GetNumBlockDeviceMappings().Return(0)
+				m.EXPECT().GetNumAttachedENIs().Return(1)
+				return m
+			},
+		},
+		// 4 gpus
```
AndrewSirenko, line 1220: Nit: Would it be helpful to append this comment to the test name itself `g4dn.12xlarge_volume_attach_limit (4 GPUs, 1 InstanceStoreVolume)`, and adding the number of Instance Store Volumes instead of commenting? 

I'm thinking that this would be helpful if these unit tests fail. Developers can easily see test name for the test's assumptions, instead of stumbling across this comment. Furthermore the inclusion of InstanceStoreVolumes makes clear that GPUs aren't the only thing that uses up slot.

hack/generate-gpu-count-table.sh
```diff
@@ -0,0 +1,32 @@
...
+function get_gpus_for_region() {
+  REGION="${1}"
+  echo "Getting gpu counts for ${REGION}..." >&2
+  "${BIN}/aws" ec2 describe-instance-types --region "${REGION}" --query "InstanceTypes[?GpuInfo!=null].[InstanceType, GpuInfo]" |
+    jq -r 'map("\"" + .[0] + "\": " + (.[1].Gpus | map(.Count) | add | tostring) + ",") | .[]'
+}
+
+function get_all_gpus() {
+  "${BIN}/aws" account list-regions --max-results 50 | jq -r '.Regions | map(.RegionName) | .[]' | while read REGION; do
+    get_gpus_for_region $REGION
```
AndrewSirenko, line 28: Even if we can't combine gpu and InstanceStore queries into one, can we parallelize this inner loop? 

**This script goes from 4.25 minutes to 39 seconds with this change.** 

```
"${BIN}/aws" account list-regions --max-results 50 | jq -r '.Regions | map(.RegionName) | .[]' | while read REGION; do
     sleep 1
     get_gpus_for_region $REGION &
    done
```


The `&` "terminates an asynchronous command" such that it will run in background and run next command (next loop iteration) immediately.

The `sleep 1` sleeps for one second, which prevents the rate-limit issues @ConnorJC3 mentioned during demo. (Or atleast I'm not running into any limits locally) 

Same thing for the instance store generation function


Speedup proof:

```
./hack/generate-gpu-count-table.sh  30.72s user 4.08s system 88% cpu 39.507 total
./hack/generate-gpu-count-table.sh  36.98s user 5.20s system 16% cpu 4:16.95 total
```

I also diffed the results and found no difference.

pkg/driver/node_test.go
ElijahQuinones: agree will add in next rev

rdpsin:
Should we do some manual testing by setting up a cluster with GPU instances?

pkg/cloud/volume_limits.go
```diff
@@ -489,3 +495,58 @@ var nvmeInstanceStoreVolumes = map[string]int{
 	"z1d.metal":       2,
 	"z1d.xlarge":      1,
 }
+
+// / https://aws.amazon.com/ec2/instance-types
+var gpuInstanceGpus = map[string]int{
```
AndrewSirenko, line 500: In a future PR, it would be nice if we could auto-generate these to prevent operator error (perhaps make a new makefile `pre-release` target that also updates sidecars). 

Here is the magic CLI incantation I came up that worked to auto-gen these locally for me if you want to alley-oop it. But I don't mind making the followup PR during the v1.34 release if you're swamped this sprint. 

If you wrap the actual map with `// CODEGEN_GPU START` and ``CODEGEN_GPU STOP` e.g.

```
var gpuInstanceGpus = map[string]int{
// CODEGEN_GPU START
...
// CODEGEN_GPU STOP
}
```

Then the following incantation when run with gnu-sed will update the `volume_limits.go` file in-place

```
ROOT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../../"
BIN="$ROOT_DIR/bin"
VOLUME_LIMITS_FILE="$ROOT_DIR/pkg/cloud/volume_limits.go"

...

export GPU_LIMITS
GPU_LIMITS=$(get_all_gpus | sort | uniq)
sed -n -e "1,/CODEGEN_GPU START/ p" -e "/CODEGEN_GPU STOP/,$ p" "$VOLUME_LIMITS_FILE" -i
sed '/CODEGEN_GPU\ START/a $GPU_LIMITS' "$VOLUME_LIMITS_FILE" -i
envsubst < "$VOLUME_LIMITS_FILE" | sponge "$VOLUME_LIMITS_FILE"
```

^^ does the following:

1. Store the de-duplicated results of `get_all_gpus` in an exported environment variable
2. Deletes all text between the CODEGEN_GPU START & STOP comments
3. Adds the `$GPU_LIMITS` environment between the START and STOP comments which will be...
4. Substituted with the actual list of GPU Limits and written to `volume_limits.go` 

Note that a few things need to be cleaned up before PR (e.g. making sure the limits get pasted with appropriate tabbing) but it's good enough for me to use next release.  

Hope this was helpful

pkg/cloud/volume_limits.go
```diff
@@ -138,11 +138,17 @@ func GetDedicatedLimitForInstanceType(it string) int {
 	}
 }
 
-func GetNVMeInstanceStoreVolumesForInstanceType(it string) int {
-	if v, ok := nvmeInstanceStoreVolumes[it]; ok {
-		return v
+func GetReservedSlotsForInstanceType(it string) int {
```
AndrewSirenko, line 141: Would you add a [Go Doc Comment](https://tip.golang.org/doc/comment) above this function that describes what it's doing? 

Something like:
```
// GetReservedSlotsForInstanceType calculates how many attachment slots are already used up by other devices. These devices include Instance Store Volumes and GPUs
```

It's good practice to provide a go doc comment for any exported (capitalized) name so that those who import the package can easily see what it is used for.

For example, IDEs like VsCode and GoLand are able to parse out these go docs to provide on-hover descriptions of the exported function/const.

ElijahQuinones:
> Should we do some manual testing by setting up a cluster with GPU instances?

I have manually tested the limit outside of Kubernettes entirely by just spinning up a g4ad.xlarge instance and mannualy attaching ebs volumes until I hit the limit where all volumes afterwards are stuck in the attaching state this test concluded that 24 volumes could be attached to the g4ad.xlarge which complies with what the documentation provides.

starts at 28
-1 eni
-1 root volume
-1 gpu 
-1 instance store 
we also ens up with 24 

Additionally the unit tests for a g4ad.xlarge return the expected value of 24 slots.

however if we feel it would cover more bases I can run the test with a g4ad.xlarge cluster just to make sure there is no gotchas.

pkg/cloud/volume_limits.go
ConnorJC3: Suggest we hold off on this for now as this table will likely be dropped in the near future

ElijahQuinones: I agree with Connor here that we may want to hold off for now.However I  will add a ticket on the backlog to do this though as it will be a good improvement

ConnorJC3 (approved):
I've personally done enough manual testing to be confident it works

Needs squash otherwise lgtm

torredil (approved):
Thanks for including additional unit tests.
/lgtm

PR #2109, opened 2024-08-12
Add toggle for PodDisruptionBudget in chart

### LINKED ISSUE #1934: Disable PDB for managed addon
/triage support
Hi, is there a way to delete/change PDB for the add-on via the config ? I don't see the config in the json schema.

AndrewSirenko:

/feature

Fixes #1934 

The EBS CSI Controller Pod's [PodDisruptionBudget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) ensures that 1 controller Pod is always running, even when many nodes are being drained at once. This ensures that PersistentVolumes are always able to be Attached/Detached Created/Deleted, and that stateful pods are not delayed by missing PVs. 

Some customers have requested ability to disable this PodDisruptionBudget  in order to prioritize node draining / Karpenter consolidation over the availability of the EBS CSI Driver. While we don't recommend this, this PR exposes the parameter for those who use EBS only with non-critical applications.  

**What testing is 

See the following rendered helm templates with and without `--set controller.podDisruptionBudget=false`, along with the diff

[change.txt](https://github.com/user-attachments/files/16590119/change.txt)
[default.txt](https://github.com/user-attachments/files/16590120/default.txt)
[diff.txt](https://github.com/user-attachments/files/16590121/diff.txt)
done?** 

Diff:

```
2,22d1
< # Source: aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
< apiVersion: policy/v1
< kind: PodDisruptionBudget
< metadata:
<   name: ebs-csi-controller
<   namespace: default
<   labels:
<     app.kubernetes.io/name: aws-ebs-csi-driver
<     app.kubernetes.io/instance: release-name
<     helm.sh/chart: aws-ebs-csi-driver-2.33.0
<     app.kubernetes.io/version: "1.33.0"
<     app.kubernetes.io/component: csi-driver
<     app.kubernetes.io/managed-by: Helm
< spec:
<   selector:
<     matchLabels:
<       app: ebs-csi-controller
<       app.kubernetes.io/name: aws-ebs-csi-driver
<       app.kubernetes.io/instance: release-name
<   maxUnavailable: 1
< ---

```

Proof in deployed cluster:

```
❯ helm upgrade \
  ... 
  --set controller.podDisruptionBudget=false

❯ kubectl get pdb -A
NAMESPACE     NAME                           MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
kube-system   aws-node-termination-handler   N/A             1                 1                     105m
kube-system   cilium-operator                N/A             1                 1                     105m
kube-system   kube-dns                       N/A             50%               1                     105m
```

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -245,6 +245,10 @@ controller:
   deploymentAnnotations: {}
   podAnnotations: {}
   podLabels: {}
+  podDisruptionBudget:
+    # Warning: Disabling PodDisruptionBudget may lead to delays in stateful workloads starting due to controller
+    # pod restarts or evictions.
+    enabled: true
   priorityClassName: system-cluster-critical
   # AWS region to use. If not specified then the region will be looked up via the AWS EC2 metadata
   # service.
--- a/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.nodeComponentOnly -}}
+{{- if and .Values.controller.podDisruptionBudget.enabled (not .Values.nodeComponentOnly) -}}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
@@ -16,4 +16,4 @@ spec:
   {{- else }}
   minAvailable: 2
   {{- end }}
-{{- end }}
+{{- end -}}


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -245,6 +245,8 @@ controller:
   deploymentAnnotations: {}
   podAnnotations: {}
   podLabels: {}
+  # Warning: Disabling PodDisruptionBudget may lead to stateful workloads not starting due to lack of controller pods
+  podDisruptionBudget: true
```
ConnorJC3, line 249: Would it be worth making this `controller.podDisruptionBudget.enabled`, in case we want to later add something like `contoller.podDisruptionBudget.minAvailable`?

Otherwise lgtm

AndrewSirenko:
/hold 

Wait for #2106 to merge first

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -245,6 +245,9 @@ controller:
   deploymentAnnotations: {}
   podAnnotations: {}
   podLabels: {}
+  podDisruptionBudget:
+    # Warning: Disabling PodDisruptionBudget may lead to stateful workloads not starting due to lack of controller pods
```
ConnorJC3, line 249: Suggest alternative wording that is more clear of the risks ("workloads not starting" makes it sound permanent)

```suggestion
    # Disabling PodDisruptionBudget may lead longer downtime after controller pod restarts/crashes/evictions
```

ConnorJC3 (approved):
/lgtm

PR #2114, opened 2024-08-14
Auto-enable VAC on clusters with beta API version

ConnorJC3:

Feature

Automatically enable VAC on clusters using the beta API version. This improves the experience for beta users and ensures they do not accidentally create a volume with a VAC bound that was not applied due to the feature gate being disabled.

Also updates the VAC documentation to reflect this, and fix a few minor issues.

**What testing is done?** 

Templating:
```
$ helm template charts/aws-ebs-csi-driver | grep VolumeAttributesClass
<blank output>

$ helm template charts/aws-ebs-csi-driver --api-versions "storage.k8s.io/v1beta1/VolumeAttributesClass" | grep VolumeAttributesClass
            - --feature-gates=VolumeAttributesClass=true
            - --feature-gates=VolumeAttributesClass=true
```

Also tested manually on a cluster with and without VAC enabled and observed the desired effect.

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -231,6 +231,9 @@ spec:
             {{- if not (regexMatch "(-retry-interval-max)" (join " " .Values.sidecars.provisioner.additionalArgs)) }}
             - --retry-interval-max=30m
             {{- end }}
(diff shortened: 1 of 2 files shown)


AndrewSirenko (approved):
Great change, thank you!

torredil (approved):
/lgtm
/approve

PR #2115, opened 2024-08-14
Fix accelerators not being considered when counting allocatables

ElijahQuinones:
bug fix fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2105
This PR fixes GH https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2105 as before we where not taking accelerators into consideration when calculating volume limits  
**What testing is done?** 
I spun up and hand added ebs volumes to an instance in each accelerator family to ensure there where no special cases outside of the VT1 family where accelerators take up two slots. 

Additionally I found that the calculation in node.go was incorrect for any instance type in the maxVolumeLimits table the documentation states that for the instance types in that table the calculation of 28 - enis - NVMeInstanceStoreVolumes - gpus/accekerators should be completely  ignored as they are [special cases](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html) . As such I checked with a inf1.xlarge instance type and verified you can in fact attach 26 volumes as it's eni (or anything else for that matter) should not be considered when looking at its limit.

### DIFF
--- a/hack/generate-gpu-count-table.sh
+++ b/hack/generate-gpu-count-table.sh
@@ -15,6 +15,7 @@
 # Generates gpu table for `pkg/cloud/volume_limits.go` from the AWS API
 # Ensure you are opted into all opt-in regions before running
 # Ensure your account isn't in any private instance type betas before running
+# We are exluding the g5.48xlarge instance type as it is a special case that does not comply to regular ebs volume limit calculations
 
 set -euo pipefail
 
@@ -34,4 +35,4 @@ function get_all_gpus() {
   done
 }
 
-get_all_gpus | sort | uniq
+get_all_gpus | sort | uniq | grep -v "g5.48xlarge"
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -19,7 +19,7 @@ import (
 	"strings"
 )
 
-// / https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances
+// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances
 const (
 	highMemoryMetalInstancesMaxVolumes   = 19
 	highMemoryVirtualInstancesMaxVolumes = 27
@@ -51,7 +51,7 @@ func init() {
 
 var dedicatedVolumeLimits = map[string]int{}
 
-// / List of nitro instance types can be found here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances
+// List of nitro instance types can be found here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances
 var nonNitroInstanceFamilies = map[string]struct{}{
 	"t2":  {},
 	"c3":  {},
@@ -67,6 +67,7 @@ var nonNitroInstanceFamilies = map[string]struct{}{
 	"g3":  {},
 	"d2":  {},
 	"h1":  {},
+	"f1":  {},
 }
 
 func IsNitroInstanceType(it string) bool {
@@ -88,8 +89,8 @@ func GetMaxAttachments(nitro bool) int {
 	return nonNitroMaxAttachments
 }
 
-// / Some instance types have a maximum limit of EBS volumes
-// / https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html
+// Some instance types have a maximum limit of EBS volumes
+// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html
 var maxVolumeLimits = map[string]int{
 	"d3.8xlarge":    3,
 	"d3en.12xlarge": 3,
@@ -149,12 +150,16 @@ func GetReservedSlotsForInstanceType(it string) int {
 	if ok {
 		total += gpus
 	}
+	acceleratorSlots, ok := acceleratorSlotsTaken[it]
+	if ok {
+		total += acceleratorSlots
+	}
 	return total
 }
 
-// / https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-store-volumes.html
-// / IMDS does not provide NVMe instance store data; we'll just list all instances here
-// / TODO: See if we can get these values from DescribeInstanceTypes API
+// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-store-volumes.html
+// IMDS does not provide NVMe instance store data; we'll just list all instances here
+// g5.48xlarge is not added to this table as it is in the maxVolumeLimits
 var nvmeInstanceStoreVolumes = map[string]int{
 	"c1.medium":       1,
 	"c1.xlarge":       4,
@@ -242,7 +247,6 @@ var nvmeInstanceStoreVolumes = map[string]int{
 	"g5.16xlarge":     1,
 	"g5.24xlarge":     1,
 	"g5.2xlarge":      1,
-	"g5.48xlarge":     2,
 	"g5.4xlarge":      1,
 	"g5.8xlarge":      1,
 	"g5.xlarge":       1,
@@ -497,7 +501,9 @@ var nvmeInstanceStoreVolumes = map[string]int{
 	"z1d.xlarge":      1,
 }
 
-// / https://aws.amazon.com/ec2/instance-types
+// https://aws.amazon.com/ec2/instance-types
+// Despite the dl1.24xlarge having Gaudi Accelerators describe instance types considers them GPUs as such that instacne type is in this table
+// g5.48xlarge is not added to this table as it is in the maxVolumeLimits
 var gpuInstanceGpus = map[string]int{
 	"dl1.24xlarge":  8,
 	"g3.16xlarge":   4,
@@ -520,7 +526,6 @@ var gpuInstanceGpus = map[string]int{
 	"g5.16xlarge":   1,
 	"g5.24xlarge":   4,
 	"g5.2xlarge":    1,
-	"g5.48xlarge":   8,
 	"g5.4xlarge":    1,
 	"g5.8xlarge":    1,
 	"g5g.16xlarge":  2,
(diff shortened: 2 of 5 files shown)


pkg/driver/node.go
```diff
@@ -791,8 +791,12 @@ func (d *NodeService) getVolumesLimit() int64 {
 		availableAttachments = dedicatedLimit
 	} else if isNitro {
 		enis := d.metadata.GetNumAttachedENIs()
-		nvmeInstanceStoreVolumes := cloud.GetReservedSlotsForInstanceType(instanceType)
-		availableAttachments = availableAttachments - enis - nvmeInstanceStoreVolumes
+		reservedSlots := cloud.GetReservedSlotsForInstanceType(instanceType)
+		if hasMaxvolumeLimit {
+			availableAttachments = availableAttachments - reservedSlots
```
ConnorJC3, line 796: Why would we not want to subtract the ENIs in this case?

pkg/driver/node.go
```diff
@@ -781,8 +781,8 @@ func (d *NodeService) getVolumesLimit() int64 {
 	}
 
 	dedicatedLimit := cloud.GetDedicatedLimitForInstanceType(instanceType)
-	maxEBSAttachments, ok := cloud.GetEBSLimitForInstanceType(instanceType)
-	if ok {
+	maxEBSAttachments, hasMaxvolumeLimit := cloud.GetEBSLimitForInstanceType(instanceType)
```
ConnorJC3, line 784: np: capitalize `volume` in variable name

hack/generate-gpu-count-table.sh
```diff
@@ -34,4 +35,4 @@ function get_all_gpus() {
   done
 }
 
-get_all_gpus | sort | uniq
+get_all_gpus | sort | uniq | grep -vE "(g5.48xlarge)"
```
ConnorJC3, line 38:
```suggestion
get_all_gpus | sort | uniq | grep -v "g5.48xlarge"
```

Is this slightly simpler version possible instead? (ditto with other file)

pkg/driver/node.go
ConnorJC3, line 796:
```suggestion
			availableAttachments = availableAttachments - (enis - 1) - reservedSlots
```

I'm pretty sure this is what we actually should do here, as despite being special cases these devices still are a shared limit.

pkg/cloud/volume_limits.go
```diff
@@ -67,6 +67,7 @@ var nonNitroInstanceFamilies = map[string]struct{}{
 	"g3":  {},
 	"d2":  {},
 	"h1":  {},
+	"f1":  {},
```
AndrewSirenko, line 70: Are we sure f1 is not nitro instance family? 

I see it in the accelerated computing current gen instance list: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances

But perhaps I'm looking in wrong place

pkg/cloud/volume_limits.go
```diff
@@ -551,3 +556,20 @@ var gpuInstanceGpus = map[string]int{
 	"p4de.24xlarge": 8,
 	"p5.48xlarge":   8,
 }
+
+// / Note this table is not a reflection of how many accelerators an instance has but of how many slots their combined acclerators take up
```
AndrewSirenko, line 560: @ConnorJC3 @torredil Is there a reason we are doing `// /` for comments in this file? 

If not, can we delete all the extra `/`'s?

pkg/cloud/volume_limits.go
```diff
@@ -551,3 +556,20 @@ var gpuInstanceGpus = map[string]int{
...
+// / VT instance type accelerators take two slots each with the excpetion of the vt1.24xlarge which takes 0 slots for it's acclerators
+// / inf1 instance types are purposley not added to this table as they are in the maxVolumeLimits table
```
AndrewSirenko, line 562: Typo: purposley

pkg/cloud/volume_limits.go
```diff
@@ -551,3 +556,20 @@ var gpuInstanceGpus = map[string]int{
 	"p4de.24xlarge": 8,
 	"p5.48xlarge":   8,
 }
+
+// / Note this table is not a reflection of how many accelerators an instance has but of how many slots their combined acclerators take up
+// / VT instance type accelerators take two slots each with the excpetion of the vt1.24xlarge which takes 0 slots for it's acclerators
```
AndrewSirenko, line 561: Typo: excpetion

pkg/driver/node.go
ElijahQuinones: Ah yes this is better as if an instance has more than one eni it does count towards the shared limit will fix on next rev thank you !

pkg/cloud/volume_limits.go
ElijahQuinones: if you run `❯ aws ec2 describe-instance-types --instance-types  f1.2xlarge` you will see that `            "Hypervisor": "xen",` and not nitro. As such I manually confirmed this by attaching 40 volumes to the  f1.2xlarge instance type

hack/generate-gpu-count-table.sh
ElijahQuinones:
```
❯ vim test.txt
❯ cat test.txt
"g5.48xlarge": 8,
"g5.43xlarge": 8,
❯ grep -v "(g5.48xlarge)" test.txt
"g5.48xlarge": 8,
"g5.43xlarge": 8,
❯ grep -vE "(g5.48xlarge)" test.txt
"g5.43xlarge": 8,
```

unfortunately no

pkg/cloud/volume_limits.go
ConnorJC3: No special reason I'm aware of, if we get rid of them preferably do the whole file for consistency.

pkg/cloud/volume_limits.go
ConnorJC3: I can confirm F1 definitely is non-nitro

ConnorJC3 (approved):
/lgtm

pkg/cloud/volume_limits.go
AndrewSirenko: Agreed on delete for whole file

pkg/cloud/volume_limits.go
AndrewSirenko: Thanks for the deep dive!

pkg/driver/node_test.go
```diff
@@ -1161,73 +1225,92 @@ func TestGetVolumesLimit(t *testing.T) {
...
-				m.EXPECT().GetInstanceType().Return("mac1.metal")
+				m.EXPECT().GetInstanceType().Return("dl1.24xlarge")
 				m.EXPECT().GetNumBlockDeviceMappings().Return(0)
 				m.EXPECT().GetNumAttachedENIs().Return(1)
 				return m
 			},
 		},
 		{
-			name: "u-12tb1.metal_volume_attach_limit",
+			// will and should fail if g5.48xlarge instance type is in any table other than maxVolumeLimits table
```
AndrewSirenko, line 1235: Hyper-nit: Add space between `limit` and `(` like we do for other unit tests

Also consider rephrasing parenthetical to `(Instance has attached GPUs and NVMe Instance Store volumes but should be ignored for EBS volume limits calculation)` to clarify the different devices/limits at play here for future contributors. 

Same thing for rest of new unit tests.

AndrewSirenko (approved):

ConnorJC3 (approved):
/lgtm

pkg/driver/node_test.go
```diff
@@ -1161,73 +1225,92 @@ func TestGetVolumesLimit(t *testing.T) {
...
-				m.EXPECT().GetInstanceType().Return("u-12tb1.metal")
+				m.EXPECT().GetInstanceType().Return("g5.48xlarge")
 				m.EXPECT().GetNumBlockDeviceMappings().Return(0)
 				m.EXPECT().GetNumAttachedENIs().Return(1)
 				return m
 			},
 		},
 		{
-			name: "g4dn.xlarge_volume_attach_limit (1 GPU 1 InstanceStoreVolume)",
+			// will and should fail if inf1.xlarge instance type is in any table other than maxVolumeLimits table
```
torredil, line 1252: np: remove 'will and'. Similar comment for other instances of this string such as 1269, 1286, and 1303.

pkg/cloud/volume_limits.go
```diff
@@ -551,3 +556,20 @@ var gpuInstanceGpus = map[string]int{
 	"p4de.24xlarge": 8,
 	"p5.48xlarge":   8,
 }
+
+// Note this table is not a reflection of how many accelerators an instance has but of how many slots their combined acclerators take up
+// VT instance type accelerators take two slots each with the exception of the vt1.24xlarge which takes 0 slots for its acclerators
+// inf1 instance types are purposely not added to this table as they are in the maxVolumeLimits table
+var acceleratorSlotsTaken = map[string]int{
```
torredil, line 563: Can we add *where* this information is sourced from in the comment above? it will be helpful to future maintainers.

pkg/cloud/volume_limits.go
```diff
@@ -551,3 +556,20 @@ var gpuInstanceGpus = map[string]int{
 	"p4de.24xlarge": 8,
 	"p5.48xlarge":   8,
 }
+
+// Note this table is not a reflection of how many accelerators an instance has but of how many slots their combined acclerators take up
+// VT instance type accelerators take two slots each with the exception of the vt1.24xlarge which takes 0 slots for its acclerators
```
torredil, line 561: np: `acclerators` -> `accelerators`

torredil (approved):
Largely lgtm, thanks @ElijahQuinones

pkg/cloud/volume_limits.go
AndrewSirenko: +1, especially since AWS Docs are usually really great about re-directing dead pages to their new homes.

pkg/cloud/volume_limits.go
```diff
@@ -551,3 +556,21 @@ var gpuInstanceGpus = map[string]int{
 	"p4de.24xlarge": 8,
 	"p5.48xlarge":   8,
 }
+
+// Note this table is not a reflection of how many accelerators an instance has but of how many slots their combined acclerators take up
+// VT instance type acclerators take two slots each with the exception of the vt1.24xlarge which takes 0 slots for its acclerators
```
ConnorJC3, line 561: Latest rev adds a typo here: "acclerators"

Also, we should squash before merge. Otherwise lgtm.

ConnorJC3 (approved):
/lgtm

PR #2119, opened 2024-08-16
Adding RoleArn option for the driver

sarthakkothari:
New Feature

Adds the ability to provide a role to interact with EC2 APIs.


chu-yik (approved):

wmesard:
This change is adding a fourth method of getting IAM permissions to the driver. So the [relevant section of the documentation](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md#set-up-driver-permissions) should be updated.

sarthakkothari:
@wmesard I've added the details in the [options.md](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2119/files#diff-2e046795c81fb2bcc2667705f88c6421fa988047f147ddb2333ec8da5f638b7bR22) file.

What exactly needs to be added in the installation part of the document? The document doesn't seem to go over any details regarding running the controllers outside of the target cluster.

wmesard:
> What exactly needs to be added in the installation part of the document? The document doesn't seem to go over any details regarding running the controllers outside of the target cluster.

My comment was about authZ, not about where the driver is running. And your addition to that document shouldn't mention anything about where the driver is running either, since that is merely one possible use case for the proposed code change. (In other words, this is not a Tachyon-specific change. Leave the Tachyon use-case out of the discussion.)

The section I linked to says:

> There are several methods to grant the driver IAM permissions:
> * [Using Instance profile]
> * [Using EKS service role]
> * [Using a K8s secret]

I'm saying you should add a fourth bullet. Something like:

> * Specifying a `role-arn`. In certain deployments, the IAM credentials from one of the three methods above do not, themselves, give the driver the required permissions. Rather they give the driver the ability to Assume an IAM Role which *does* have those permissions. See [Driver Options](./options.md).

Or something like that.

ConnorJC3:
/close

Closing this outdated PR for now

PR #2121, opened 2024-08-19
Add node.legacyXFS parameter to fix XFS on nodes with kernel version ≤ 5.4

AndrewSirenko:

Feature

Adds workaround for customers using XFS on nodes with linux kernel version ≤ 5.4.

Warning: This option will be removed in a future release. It is a temporary workaround for users unable to immediately migrate off of older kernel versions.

Adds `legacy-xfs` node driver option and and `node.legacyXFS` helm chart parameter which formats XFS volumes with `bigtime=0,inobtcount=0,reflink=0`, so that they can be mounted onto nodes with linux kernel ≤ 5.4. Volumes formatted with this option may experience issues after 2038, and will be unable to use some XFS features (for example, reflinks).

**What testing is done?** 

Created node with Linux ≤ v5.4, attempted to run stateful pod with and without `legacy-xfs` parameter. Volume only mounted when legacy-xfs=true.


AndrewSirenko:
/hold

Will test on linux 5.4 today. 

But still ready for review.

pkg/driver/node_test.go
```diff
@@ -983,9 +1023,17 @@ func TestNodeStageVolume(t *testing.T) {
 				metadata = tc.metadataMock(ctrl)
 			}
 
+			var options *Options
+			if tc.options != nil {
+				options = tc.options
+			} else {
+				options = &Options{} // Initialize struct to avoid nil pointer dereference
+			}
```
ConnorJC3, line 1031:
```suggestion
			options := tc.options
			if options == nil {
				options = &Options{} // Initialize struct to avoid nil pointer dereference
			}
```

Suggest cleaner alternative

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -392,6 +392,10 @@ node:
   # Enable the linux daemonset creation
   enableLinux: true
   enableWindows: false
+  # Warning: This option will be deprecated in a future version of EBS CSI Driver.
```
ConnorJC3, line 395: Discussion: Should we be more strong about this? Explicitly stating we will remove it might make users less likely to intend to rely on it long term.

Maybe something like:
```
Warning: This option will be removed in a future release. It is intended as a temporary workaround for users unable to immediately migrate off older kernel versions.
```

AndrewSirenko: Agreed. Changed wording in documentation and CLI flag too. (Though we should sync these sources in future)

AndrewSirenko:
/unhold

Customer tested with their custom AMI, driver was functional.

docs/options.md
```diff
@@ -2,20 +2,21 @@
...
+| warn-on-invalid-tag       | true                    | false                                            | To warn on invalid tags, instead of returning an error|
+| reserved-volume-attachments | 2                       | -1                                               | Number of volume attachments reserved for system use. Not used when --volume-attach-limit is specified. When -1, the amount of reserved attachments is loaded from instance metadata that captured state at node boot and may include not only system disks but also CSI volumes.|
+| legacy-xfs                | true                    | false                                            | Warning: This option will be removed in a future release. It is a temporary workaround for users unable to immediately migrate off of older kernel versions. Formats XFS volumes with `bigtime=0,inobtcount=0,reflink=0`, so that they can be mounted onto nodes with linux kernel ≤ v5.4. Note that XFS volumes formatted with this option will only have timestamp records until 2038.|
```
ElijahQuinones, line 22: Not sure if I maybe misunderstood this but, the last line you talk about `Note that XFS volumes formatted with this option will only have timestamp records until 2038` won't this brick the volume after 2038 ? And if so should we mention the implications of it here ?

ConnorJC3 (commented):
lgtm other than Elijah's one comment. Assuming that is addressed (and no further issues are reported in review), I plan to mark this PR as approved (if nobody else does so first) the evening of Tuesday 9/10.

pkg/driver/options.go
```diff
@@ -87,6 +87,9 @@ type Options struct {
 	ReservedVolumeAttachments int
 	// ALPHA: WindowsHostProcess indicates whether the driver is running in a Windows privileged container
 	WindowsHostProcess bool
+	// LegacyXFSProgs enables the use of `` when formatting a volume with XFS. This is necessary if the driver is used
+	// to mount XFS volumes on Linux Nodes with a kernel version ≥ 5.4 due to __.
```
torredil, line 91: This comment needs to be updated (please complete `use of '' ` and  `due to __` statements). Besides this, changes introduced in this PR lgtm.

Issue #2087, opened 2024-07-12
Support Amazon EBS local snapshots on Outposts


smohandass:
/kind bug

**What happened?**

According to [this](https://aws.amazon.com/about-aws/whats-new/2020/10/amazon-ebs-csi-driver-now-supports-aws-outposts/) AWS EBS CSI driver supports creating EBS volumes on worker nodes running in AWS Outposts subnets. 

When creating the snapshot using the volumeSnapshot CRD, the snapshot is created on the AWS region and not on the Outpost. 

According to AWS link [here](https://docs.aws.amazon.com/ebs/latest/userguide/snapshots-outposts.html) , the snapshot can be created on the outpost by passing the outpost ARN. The volumesnapshot CRD however doesn't have a feature to specify the ARN.

**What you expected to happen?**
I expect the EBS CSI driver to provide a way to specify the Outpost ARN.

**How to reproduce it (as minimally and precisely as possible)?**

Create a EKS cluster with worker node on Outpost
Deploy a k8s application with a dynamically provisioned EBS volume running on Outpost. 
Create a snapshot using the `volumeSnapshot` CRD and you can notice the snapshot is created in the region and NOT on Outpost

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):  v1.29.4
- Driver version:  1.32.0

ConnorJC3:
/kind feature
/retitle Support Amazon EBS local snapshots on Outposts

Hi, thanks for the feature request - I've added it to our backlog for evaluation. I can't promise any specific ETA at this time but we'll provide updates on this issue as they are available.

smohandassk10:
@ElijahQuinones  Is the fix available as part of the AWS EBS CSI Driver add on to test?

ElijahQuinones:
@smohandassk10 It is not currently but will be available as part of the September release.

smohandass:
@ElijahQuinones  I do see that a new release v1.35.0 was out today. I tried to test the creation of volumeSnapshot on local outpost and didn't succeeded. Below are the steps, I followed

1. Created a EKS cluster with control plane on region and nodegroup on outpost. 
2. Deployed the latest version (v1.35.0) of EBS CSI driver using helm chart.
3. Deployed the latest version (v8.1.0) of the external snapshotter from https://github.com/kubernetes-csi/external-snapshotter
4. Created a VolumeSnapshotClass and added the outpostarn parameter.
```
[...]

PR #2130, opened 2024-08-26
Add local snapshots on outposts

ElijahQuinones:

fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2087

This PR will allow customers to pass the outpost ARN as a parameter to the VSC.
This will enable customers to have local snapshots saved onto S3 on their outpost and not just in S3 on the region their outpost is in.

**What testing is done?** 

This is currently being tested in outpost.


ElijahQuinones:
/do-not-merge/work-in-progress

pkg/cloud/cloud.go
```diff
@@ -1275,19 +1276,28 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
...
-	request := &ec2.CreateSnapshotInput{
-		VolumeId:          aws.String(volumeID),
-		TagSpecifications: []types.TagSpecification{tagSpec},
-		Description:       aws.String(descriptions),
+	if snapshotOptions.OutpostArn != "" {
+		request = &ec2.CreateSnapshotInput{
+			VolumeId:          aws.String(volumeID),
+			TagSpecifications: []types.TagSpecification{tagSpec},
+			Description:       aws.String(descriptions),
+			OutpostArn:        aws.String(snapshotOptions.OutpostArn),
```
rdpsin, line 1292: Can you check this? I'm assuming in the case when users don't provide `outpostArn`, we end up setting this to `""` instead of `null` so `CreateSnapshot` would fail.

pkg/driver/constants.go
```diff
@@ -102,6 +102,9 @@ const (
 	// TagKeyPrefix contains the prefix of a volume parameter that designates it as
 	// a tag to be attached to the resource
 	TagKeyPrefix = "tagSpecification"
+
+	// OutpostArn represents key for outpost's arn
+	OutpostArn = "outpostarn"
```
rdpsin, line 107: np: `OutpostArnKey` and `"outpostArn"`

pkg/cloud/cloud.go
ElijahQuinones: In the case there is no outpostArn parameter in the VSC than the case in the switch statement in controller.go is never hit so the outpost arn passed to cloud.go is the empty string "" in this case we do not want to give the ec2 api an outpostarn so we drop to the else. Ether way we are creating a createSnapshotInput and passing it to ec2.

pkg/driver/constants.go
ElijahQuinones: Will change in next rev makes it much more clear

pkg/cloud/cloud.go
rdpsin: Ahh, my bad. In that case, can we just create a single request object, and do the check later?

```
req := &ec2.CreateSnapshotInput {
   ...
}

if options.OutpostArn != "" {
       req.OutpostArn = aws.String(options.OutpostArn)
}
```

ElijahQuinones: Ahhh yes we can that. Will add to next rev thank you !

pkg/driver/constants.go
ElijahQuinones: Upon reviewing the code again outpostArn will not work as we are checking for all lower case values but I will change the key in next rev.

pkg/driver/controller.go
```diff
@@ -709,6 +710,8 @@ func (d *ControllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		case FastSnapshotRestoreAvailabilityZones:
 			f := strings.ReplaceAll(value, " ", "")
 			fsrAvailabilityZones = strings.Split(f, ",")
+		case OutpostArnKey:
+			outpostArn = value
```
rdpsin, line 714: Just a thought: should we validate that this is actually an ARN here? The request will fail either way when it reaches EC2, but we can avoid blasting unnecessary requests to EC2.

rdpsin (commented):
Can we also add this option to the docs? I don't see any VSC parameters doc, maybe this a good opportunity to create one.

ElijahQuinones:
Tested In outpost and confirmed we are now able to have local snapshots in outpost.

```
Type    Reason            Age   From                 Message
  ----    ------            ----  ----                 -------
  Normal  CreatingSnapshot  11m   snapshot-controller  Waiting for a snapshot default/ebs-volume-snapshot to be created by the CSI driver.
  Normal  SnapshotCreated   11m   snapshot-controller  Snapshot default/ebs-volume-snapshot was successfully created by the CSI driver.
  Normal  SnapshotReady     9m9s  snapshot-controller  Snapshot default/ebs-volume-snapshot is ready to use.
```
For testing I used the following VSC

```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: csi-aws-vsc
driver: ebs.csi.aws.com
deletionPolicy: Delete
parameters:
  outpostarn: {arn of our outpost}
```

Screenshot 2024-08-29 at 3 38 57 PM

pkg/driver/controller.go
ElijahQuinones: In my opinion I don't see a situation where someone would use anything but their outpost arn as part of the VSC if they are using this param. Validating if this is actually an arn here to me feel like bloat but I am open minded and curious what others think.

rdpsin: > I don't see a situation where someone would use anything but their outpost arn

This is hard to predict. :) 

Nevertheless, I'll defer this to the other team members.

torredil: > Just a thought: should we validate that this is actually an ARN here?

+1, I think we should consider making use of validation that the SDK provides, see [func IsARN(arn string) bool](https://github.com/aws/aws-sdk-go-v2/blob/4ed838eab2a963cb16301501c8b8c3e29dac4c20/aws/arn/arn.go#L80).

ElijahQuinones: sounds good will do thank you guys !

pkg/cloud/cloud_test.go
```diff
@@ -2059,6 +2059,25 @@ func TestCreateSnapshot(t *testing.T) {
 			},
 			expErr: nil,
 		},
+		{
```
AndrewSirenko, line 2062: Once you add in the suggested `isARN` validation, consider adding an extra unit test for it.

Otherwise lgtm, thanks!

ConnorJC3 (commented):
lgtm, would like to test this manually before I approve, but no issues spotted reading the code

pkg/cloud/cloud_test.go
ElijahQuinones: sounds good thank you !

pkg/driver/controller.go
```diff
@@ -711,6 +711,11 @@ func (d *ControllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 			f := strings.ReplaceAll(value, " ", "")
 			fsrAvailabilityZones = strings.Split(f, ",")
 		case OutpostArnKey:
+			if arn.IsARN(value) {
+				outpostArn = value
+			} else {
+				return nil, status.Errorf(codes.InvalidArgument, "Invalid parameter value %s is not a valid arn", value)
+			}
 			outpostArn = value
```
ElijahQuinones, line 719: removing now

docs/snapshot.md
```diff
@@ -1,3 +1,14 @@
+## Supported Parameters
+| Parameter                          | Description of value                                      |
+|------------------------------------|-----------------------------------------------------------|
+| fastSnapshotRestoreAvaliblityZones | Comma separated list of availability zones                |
```
rdpsin, line 4: np: `AvailabilityZones`

docs/snapshot.md
```diff
@@ -34,6 +41,26 @@ parameters:
...
+
+
+# Amazon EBS Local Snapshots on Outposts
+
+The EBS CSI Driver provides support for [Amazon EBS local snapshots on Outposts](https://docs.aws.amazon.com/ebs/latest/userguide/snapshots-outposts.html) via `VolumeSnapshotClass.parameters.outpostArn`.
+
+By default, snapshots of EBS volumes on an Outpost are stored in Amazon S3 in the Region of the Outpost. You can also use Amazon EBS local snapshots on Outposts to store snapshots of volumes on an Outpost locally in Amazon S3 on the Outpost itself. This ensures that the snapshot data resides on the Outpost, and on your premises. In addition, you can use AWS Identity and Access Management (IAM) policies and permissions to set up data residency enforcement policies to ensue that snapshot data does not leave the Outpost. This is especially useful if you reside in a country or region that is not yet served by an AWS Region and that has data residency requirements.
```
rdpsin, line 53: np: Suggest removing this. This is reiterating what's already in the doc.

ElijahQuinones: We do the same for fast snapshot. The # Fast Snapshot Restores section in the first paragraph of the linked doc. Would you suggest we remove this from both places and just have a link to the doc ?

pkg/driver/controller.go
```diff
@@ -709,6 +710,12 @@ func (d *ControllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		case FastSnapshotRestoreAvailabilityZones:
 			f := strings.ReplaceAll(value, " ", "")
 			fsrAvailabilityZones = strings.Split(f, ",")
+		case OutpostArnKey:
```
AndrewSirenko, line 713: Nit: Would the validateCreateSnapshotRequest helper function be a better place for this arn check? (Due to having one place in the code where most of the InvalidArgument errors are created) 

The function definition is on line 790

ConnorJC3: In my opinion errors from parameters should be in the switch statement, that is consistent with the code and reduces duplicate parsing.

docs/snapshot.md
ConnorJC3: Agree - if there is information verbatim copied from the EBS docs, it should be removed (and replaced with a link).

pkg/driver/controller.go
AndrewSirenko: Where else in the switch statement are errors returned? I don't see any such code consistency. 

Seems to me that switch statement is for parsing, not for validation. 

And wouldn't we want to not make a needless call to describeSnapshots (I understand that we batch, but still).

AndrewSirenko: > In my opinion errors from parameters should be in the switch statement, that is consistent with the code and

The only error from parameters that is in the switch statement is the error one receives for using a non-accepted parameter. I think it's cleaner to have the function named `validateCreateSnapshotRequest` validate as much of the request as possible before any AWS API calls are wastefully made (e.g. DescribeSnapshots), than to have yet another place an error can be returned. 

But given that this is a nit, I'm fine with this conversation being resolved.

ConnorJC3 (approved):
Needs squash otherwise lgtm

Tested and confirmed I can create a snapshot bound to a specific outpost:
image

PR #2132, opened 2024-08-30
Make govulncheck only run on release branches

AndrewSirenko:

N/A

This makes the govulncheck github action only run on PRs merging into release branches:
- New vulnerability blocks the merge of a new contributor, requiring a rebase. This is not a great contributor experience and has led to delays due to PR churn.

The risk here is that we suddenly find out about a vulnerability for our release PR, instead of in any pr, which may delay our release by 1-3 hours (due to waiting for CI to re-run). But that should be rare since we are bumping dependencies right before a release anyway.

**What testing is done?** 

CI

### DIFF
--- a/.github/workflows/govulncheck.yaml
+++ b/.github/workflows/govulncheck.yaml
@@ -13,8 +13,10 @@
 # limitations under the License.
 
 name: 'govulncheck'
-on: [pull_request]
-
+on:
+  pull_request:
+    branches:
+      - 'release-*'
 jobs:
   govulncheck:
     runs-on: ubuntu-latest


ConnorJC3 (changes requested):
From what I can tell reading the GitHub docs, this will make it impossible for the job to fail at all, which means that we will not get the feedback unless someone goes out of their way to read the job results.

It's possible that I am wrong, but please test this in a personal fork and link example PRs here so that we can see what the result looks like before merging.

AndrewSirenko:
Thank you for your concern @ConnorJC3, but please hold off on review until it is no longer marked WIP because I have yet to do the testing that I described in PR comment.

What I saw from the docs and relevant stack overflow posts was that the job will fail with a red X, the failure will not block other CI actions for the PR. So there will be a failure shown without blocking merge, but this must be validated.

ConnorJC3:
My suggestion would be to run it only on the release branch. If we make it optional on Prow, we are going to eventually accidentally merge a release with a known CVE.

There's no circumstance where we would want to intentionally merge a release PR containing known vulnerabilities, so the only potential downside of this solution is false positives. False positives are very rare (I think we've had one over the past year?) and the only reason we're talking about them is recency bias. Additionally, it's trivial to disable the job if that ever happens during a release.

ConnorJC3 (approved):
/lgtm

torredil (approved):
/approve

PR #2138, opened 2024-09-09
use protobuf content type instead of json for k8s client

bhavi-koduru:

This MR is a part of effort to elevate single eks cluster performance by migrating the EKS components to use protobuf instead of json.

Modify kubeconfig type to use content type application/vnd.kubernetes.protobuf instead of json for performance gain.

This change essentially lets the ebs driver (client) to talk to apiserver using protobuf and falls back to json if protobuf isn't supported. It will only affect the wire format between ebs driver and apiserver.

**What testing is done?**

### DIFF
--- a/pkg/cloud/metadata/k8s.go
+++ b/pkg/cloud/metadata/k8s.go
@@ -84,6 +84,8 @@ func DefaultKubernetesAPIClient(kubeconfig string) KubernetesAPIClient {
 				}
 			}
 		}
+		config.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
+		config.ContentType = "application/vnd.kubernetes.protobuf"
 		// creates the clientset
 		clientset, err = kubernetes.NewForConfig(config)
 		if err != nil {


AndrewSirenko:
FYI @ConnorJC3 @ElijahQuinones @torredil Seems like this is documented in [Kubernetes API Concepts | Alternate Representations of resources](https://kubernetes.io/docs/reference/using-api/api-concepts/#alternate-representations-of-resources)

@bhavi-koduru mentioned that this support was enabled in k8s 1.13 so we are safe for all supported K8s versions. (x-ref: https://github.com/kubernetes/client-go/issues/76).

However, some CRDs don't support protobuf, but it seems that snapshot CRDs do support (see [external-snapshotter/client/apis/...types.go](https://github.com/kubernetes-csi/external-snapshotter/blob/d5c03dbde7ca6664707c05870eae3de4255766fb/client/apis/volumesnapshot/v1/types.go#L249).  I may do some research to ensure no noticeable performance regression before approving this PR, but our e2e tests are passing. 

Seems like there have been some (old) cases of setting ContentType to protobuf [breaking things](https://github.com/kubernetes/ingress-gce/issues/1049), but that might have been because those projects did not set the fallback to be JSON via `config.AcceptContentTypes`.

ConnorJC3:
If protobuf is so much better than JSON is there a reason it isn't enabled by default in `client-go`? Are we opting into potentially lest tested or less stable code here?

Also, is it even worth taking this risk for the driver binary itself? It rarely communicates with the API server (off the top of my head I can only think of the startup taint removal and pre-stop hooks).

AndrewSirenko:
> If protobuf is so much better than JSON is there a reason it isn't enabled by default in client-go? 

Yeah I had same question, turns out was not enabled by default because not all CRDs support protobuf, and there was concern about protobuf encodings being less stable than JSON encodings back in 2017. At-least that is what this issue rabbit hole mentions: https://github.com/kubernetes/client-go/issues/76

mengqiy:
CRDs currently only support json. That's why this change has a fallback after protobuf.
Protobuf encoding is as reliable as (if not more) json encoding. Protobuf is used for all core types between apiserver <-> etcd and apiserver <-> kube-controller-manager, scheduler, kubelet and more core components.

Controller-runtime which is most popular SDK to build controller has adopted protobuf in 2020: https://github.com/kubernetes-sigs/controller-runtime/pull/1149

There is an attempt https://github.com/kubernetes/kubernetes/pull/125314 in upstream to change the default. But it will take some time.

bhavi-koduru:
> is there a world where PRs on the [kubernetes-csi](https://github.com/kubernetes-csi) sidecars (e.g. [external-attacher](https://github.com/kubernetes-csi/external-attacher), resizer, provisioner, snapshotter) makes more sense?

The scope of this task is to get eks/aws components using protobuf. Hence, we are not focusing on upstream components currently.

torredil:
I think @ConnorJC3 and @AndrewSirenko raise valid concerns here. That said, I'm more inclined to accept this change if https://github.com/kubernetes-csi/external-attacher/pull/585 is accepted, especially given that standard components default to protobuf.

hakuna-matatah:
>If protobuf is so much better than JSON is there a reason it isn't enabled by default in client-go? Are we opting into potentially less tested or less stable code here?

Main reason why it wasn't defaulted is because of this today FMU https://github.com/kubernetes/kubernetes/issues/26130

If it wasn't exposed, defaulting would have happened already.

torredil (approved):
Thank you for the contribution!
/lgtm

AndrewSirenko (approved):

ConnorJC3:
/approve

I did some deep digging, and based on what I found this looks to be safe in the versions we support. I would still encourage the submitter to consider contributing to other upstream projects like the sidecars if they're looking for a big improvement howerver.

ConnorJC3:
/retest

likely flake

PR #2143, opened 2024-09-12
Add permissions to all GitHub actions

ConnorJC3:

"Bug fix"

The `kubernetes-sigs` GitHub org was moved to a GitHub enterprise account, changing the default permissions for actions. Explicitly define the permissions we need in each action so they continue to run successfully.

**What testing is done?** 

CI


torredil (approved):
/lgtm

AndrewSirenko (commented):
Suggestion: Could we follow the best-practice of commenting why you need each permission as seen [in the kubernetes-sigs/karpenter project](https://github.com/kubernetes-sigs/karpenter/blob/8e40c0c922242922545d1037da9843ee318b437a/.github/workflows/release.yaml#L11-L13)?

ConnorJC3:
> Suggestion: Could we follow the best-practice of commenting why you need each permission as seen [in the kubernetes-sigs/karpenter project](https://github.com/kubernetes-sigs/karpenter/blob/8e40c0c922242922545d1037da9843ee318b437a/.github/workflows/release.yaml#L11-L13)?

Added comments to all the permissions that aren't `read: contents`. I can add them there if people want, but as that permissions is necessary to checkout the repo it would just be `# Checkout repo` on every action.

PR #2160, opened 2024-09-27
Add K8s 1.31 to CI by pinning kops commit instead of version

AndrewSirenko:
CI

This PR should not be merged until we fix any rate-limit issues in our CI (likely by submitting a PR to kops for different retry policies) 

We want our CI to test changes to the EBS CSI Driver against the latest stable version of Kubernetes to proactively spot issues and guarantee that we pass the Kubernetes Storage conformance tests.

However, EKS and recently kOps do not release support for the latest stable Kubernetes version until 1-2 months afterwards, which blocks us from upgrading our CI to the latest Kubernetes version.

Many of upstream e2e tests run against newer versions of Kubernetes without the latest kops version (see [test-infra jobs](https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes/kops)). This is because there are usually a few unreleased fixes on the main branch that allow kops to be used for newer Kubernetes versions when combined with exporting `KOPS_RUN_TOO_NEW_VERSION=true`. This PR pins our CI to a Kops commit instead of a release so that we can do the same with our custom cluster testing system. 

**What testing is done?** 

Local + CI

### DIFF
--- a/hack/e2e/kops/kops.sh
+++ b/hack/e2e/kops/kops.sh
@@ -32,6 +32,7 @@ function kops_create_cluster() {
   KOPS_PATCH_FILE=${10}
   KOPS_PATCH_NODE_FILE=${11}
   KOPS_STATE_FILE=${12}
+  declare -x KOPS_TOO_NEW_VERSION=true
 
   if kops_cluster_exists "${CLUSTER_NAME}" "${KOPS_BIN}" "${KOPS_STATE_FILE}"; then
     loudecho "Replacing cluster $CLUSTER_NAME with $CLUSTER_FILE"
--- a/hack/tools/install.sh
+++ b/hack/tools/install.sh
@@ -33,8 +33,8 @@ GOMPLATE_VERSION="v4.1.0"
 # https://github.com/helm/helm
 HELM_VERSION="v3.16.2"
 # https://github.com/kubernetes/kops
-# NOTE: Keep at v1.29.0 until ELB usage bug fixed
-KOPS_VERSION="v1.29.0"
+# NOTE: We pin kops to a commit instead of a release to support newer versions of k8s earlier
+KOPS_COMMIT="aaa35cc5304f9b191ca9828b552e62bddc5b263a"
 # https://pkg.go.dev/sigs.k8s.io/kubetest2?tab=versions
 KUBETEST2_VERSION="v0.0.0-20240905095256-f6e8664cd2b1"
 # https://github.com/golang/mock
@@ -153,9 +153,11 @@ function install_helm() {
 }
 
 function install_kops() {
+  # Build from source so we can test latest Kubernetes version earlier.
   INSTALL_PATH="${1}"
 
-  install_binary "${INSTALL_PATH}" "https://github.com/kubernetes/kops/releases/download/${KOPS_VERSION}/kops-${OS}-${ARCH}" "kops"
+  # Lower max processes to avoid oom-killed
+  GOMAXPROCS=1 install_go "${INSTALL_PATH}" "k8s.io/kops/cmd/kops@${KOPS_COMMIT}"
 }
 
 function install_kubetest2() {
--- a/hack/e2e/config.sh
+++ b/hack/e2e/config.sh
@@ -35,7 +35,7 @@ OUTPOST_INSTANCE_TYPE=${OUTPOST_INSTANCE_TYPE:-${INSTANCE_TYPE}}
 # kops: must include patch version (e.g. 1.19.1)
 # eksctl: mustn't include patch version (e.g. 1.19)
 # NOTE: Keep KOPS at v1.29.x until ELB usage bug fixed
-K8S_VERSION_KOPS=${K8S_VERSION_KOPS:-1.29.6}
+K8S_VERSION_KOPS=${K8S_VERSION_KOPS:-1.31.1}
 K8S_VERSION_EKSCTL=${K8S_VERSION_EKSCTL:-1.31}
 
 EBS_INSTALL_SNAPSHOT=${EBS_INSTALL_SNAPSHOT:-"true"}


AndrewSirenko:
Building kops from a commit works now! Now off to submitting a kops PR to have a smoother aws sdk retry policy (which likely changed when moving aws sdk go v1 -> v2). 

```
Error: error running tasks: deadline exceeded executing task TargetGroup/tcp-e2e-18419240520678154-93ufpp. Example error: listing ELB TargetGroups: operation error Elastic Load Balancing v2: DescribeTargetGroups, exceeded maximum number of attempts, 3, https response error StatusCode: 400, RequestID: 7854c0be-1273-4d5f-a4c2-069f20a20912, api error Throttling: Rate exceeded
```

Drafting again for now.

AndrewSirenko:
/retest

Seems like e2e flake, NOT cluster up flake:

```
  [FAIL] External Storage [Driver: ebs.csi.aws.com] [Testpattern: Dynamic PV (default fs)] fsgroupchangepolicy [It] (OnRootMismatch)[LinuxOnly], pod created with an initial fsgroup, new pod fsgroup applied to volume contents
```

hack/tools/install.sh
```diff
@@ -33,8 +33,8 @@ GOMPLATE_VERSION="v4.1.0"
 # https://github.com/helm/helm
 HELM_VERSION="v3.15.4"
 # https://github.com/kubernetes/kops
-# NOTE: Keep at v1.29.0 until ELB usage bug fixed
-KOPS_VERSION="v1.29.0"
+# NOTE: We pin kops to a commit instead of a release to support newer versions of k8s earlier
+KOPS_COMMIT="3a8da3d7fc1d632ef5ffe4a6ef7e5e9fc048e0c7"
```
ElijahQuinones, line 37: How often do we plan on updating this to go to the latest commit of KOPS?

hack/prow-e2e.sh
```diff
@@ -71,6 +71,8 @@ export CLUSTER_NAME="e2e-${BUILD_ID:-${RANDOM}}.k8s.local"
 export KOPS_BUCKET=${KOPS_BUCKET:-"k8s-kops-csi-shared-e2e"}
 # Always use us-west-2 in CI, no matter where the local client is
 export AWS_REGION=us-west-2
+# Allow deploying newer K8s clusters before kops releases catch up
+export KOPS_RUN_TOO_NEW_VERSION=true
```
ConnorJC3, line 75: Drop this diff - it's already done in kops.sh

hack/e2e/kops/kops.sh
```diff
@@ -19,6 +19,9 @@
 
 set -euo pipefail
 
+# Allow deploying newer K8s clusters before kops releases catch up
+export KOPS_RUN_TOO_NEW_VERSION=true
```
ConnorJC3, line 23: This file is not run directly, this should be done in `kops_create_cluster` (as well as the other functions if it is needed there)

hack/tools/install.sh
```diff
@@ -153,9 +153,11 @@ function install_helm() {
 }
 
 function install_kops() {
+  # Build from source so we can test latest Kubernetes version earlier.
   INSTALL_PATH="${1}"
 
-  install_binary "${INSTALL_PATH}" "https://github.com/kubernetes/kops/releases/download/${KOPS_VERSION}/kops-${OS}-${ARCH}" "kops"
+  # Lower max processes to avoid oom-killed
+  GOMAXPROCS=1 install_go "${INSTALL_PATH}" "k8s.io/kops/cmd/kops@${KOPS_COMMIT}"
 }
```
torredil, line 161: Would it be best to only install kOps using a commit if is explicitly defined, otherwise fall back to the existing logic?

```
install_binary "${INSTALL_PATH}" "https://github.com/kubernetes/kops/releases/download/${KOPS_VERSION}/kops-${OS}-${ARCH}" "kops"
```

ConnorJC3: I would suggest against this change, as it would bloat the file and I'm not really sure what the usecase would be

torredil: Seems relatively useful to be able to test against specific release tags, as we have been doing, unless there are extenuating circumstances for which we'd want to override and use a specific commit instead. Anyways, we could effectively accomplish the same by specifying the hash associated with the relevant tag - its only a minor inconvenience to cross check that, so I'm aligned with removing the previous code path.

hack/tools/install.sh
AndrewSirenko: As often as the other dependency bumps, monthly.

ElijahQuinones (approved):

hack/e2e/kops/kops.sh
```diff
@@ -32,6 +32,7 @@ function kops_create_cluster() {
   KOPS_PATCH_FILE=${10}
   KOPS_PATCH_NODE_FILE=${11}
   KOPS_STATE_FILE=${12}
+  export KOPS_RUN_TOO_NEW_VERSION=true
```
torredil, line 35: If necessary, can we export this only once at the top of the file? 

Even better, if this parameter is only relevant during cluster bootstrap then lets export it in `kops_create_cluster()` and remove the other instances.

AndrewSirenko: This used to be once at top of file, and was set when the other script sourced kops.shm but @ConnorJC3 preferred in each function: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2160#discussion_r1799892385

This parameter is relevant to any operation involving a too-new k8s cluster, which is why it's not just in `kops_create_cluster`.

For the record, I too would prefer it just once at top of file, but don't think it's worth holding up the PR.

AndrewSirenko: Changed to function scoped `declare -x KOPS_TOO_NEW_VERSION=true` that @ConnorJC3 found as compromise.  (Though it is bash voodoo)

torredil (approved):

hack/e2e/kops/kops.sh
```diff
@@ -99,7 +100,6 @@ function kops_patch_cluster_file() {
   KOPS_PATCH_FILE=${2} # input must be yaml
   KIND=${3}            # must be either Cluster or InstanceGroup
   ROLE=${4}            # must be either Master or Node
-
```
ConnorJC3, line 102: Drop this unnecessary diff, otherwise lgtm

ConnorJC3 (approved):
/lgtm

AndrewSirenko:
/retest

Windows e2e flake unrelated to kops

```

Kubernetes e2e suite: [It] External Storage [Driver: ebs.csi.aws.com] [Testpattern: Dynamic PV (default fs)(allowExpansion)] volume-expand Verify if offline PVC expansion works expand_more
--
```

I've seen this flake a few times in the past two weeks. making backlog item to look into it.

PR #2161, opened 2024-09-27
Upgrade to eks 1.31

AndrewSirenko:
CI

Upgrade our CI to test against  eks 1.31 cluster

Note: Not exporting the VAC-modify-volume test in this PR for these eksctl jobs because we likely are upgrading kops jobs to use k8s 1.31 soon anyway, which will allow us to run those tests everywhere again. But open to it if we want that test coverage sooner. 

**What testing is done?** 
CI


ConnorJC3 (approved):
/lgtm
/approve

PR #2196, opened 2024-10-24
Add middleware to log server errors

### LINKED ISSUE #2154: insufficient capacity API possibly errors not propagating into logs and events
On 2024-08-30 starting at roughly 1900 UTC and ending many hours later the `use1-az2` AZ was unable to provision new `io2` volumes. The csi driver logs and events posted back to pending pods was generic, eg:

> failed to provision volume with StorageClass "aws-ebs-io2-020000i": rpc error: code = Internal desc = Could not create volume "pvc-307a77e9-0446-4bd0-8a49-8cf87d31d18f": could not create volume in EC2: operation error EC2: CreateVolume, failed to get rate limit token, retry quota exceeded, 0 available, 5 requested

It was not clear what the error was from these logs. This looked fairly generic to our responding engineers. It was not until someone thought to try creating a volume manually via the web console saw the error message `"There is currently insufficient capacity to serve the request. Try again at a later time or in a different availability zone"`

CleanShot 2024-09-18 at 12 02 46

I don't know if this error was being returned from the AWS API to the CSI driver or not. If so, it would be ideal to have this error propagated back to the pod/pvc events.

ConnorJC3:

Both? Kinda?

Fixes #2154

When the SDK hits a retryable error (for example, most server errors or network issues), it does not log the error in the log, preventing users and maintainers from discovering the nature of the issue. There is a built in logger for retries, but it would be sub-optimal to use that because it would flood the logs any time `RequestLimitExceeded` errors occurred. Thus, this PR instead introduces a simple logging middleware that allows us to customize the verbosity of the error message based on whether the error is a RLE/throttle error.

**What testing is done?** 

Manually tested (We don't have a way to test log messages atm, and maybe shouldn't even be doing so)

### DIFF
--- a/pkg/cloud/handlers.go
+++ b/pkg/cloud/handlers.go
@@ -22,13 +22,12 @@ import (
 	"time"
 
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
+	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/middleware"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/metrics"
 	"k8s.io/klog/v2"
 )
-
-const requestLimitExceededErrorCode = "RequestLimitExceeded"
 
 // RecordRequestsHandler is added to the Complete chain; called after any request
 func RecordRequestsMiddleware() func(*middleware.Stack) error {
@@ -40,13 +39,12 @@ func RecordRequestsMiddleware() func(*middleware.Stack) error {
 			if err != nil {
 				var apiErr smithy.APIError
 				if errors.As(err, &apiErr) {
-					if apiErr.ErrorCode() == requestLimitExceededErrorCode {
+					if _, isThrottleError := retry.DefaultThrottleErrorCodes[apiErr.ErrorCode()]; isThrottleError {
 						operationName := awsmiddleware.GetOperationName(ctx)
 						labels = map[string]string{
 							"operation_name": operationName,
 						}
 						metrics.Recorder().IncreaseCount("cloudprovider_aws_api_throttled_requests_total", labels)
-						klog.InfoS("Got RequestLimitExceeded error on AWS request", "request", operationName)
 					} else {
 						metrics.Recorder().IncreaseCount("cloudprovider_aws_api_request_errors", labels)
 					}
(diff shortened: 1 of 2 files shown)


pkg/cloud/handlers.go
```diff
@@ -60,6 +61,32 @@ func RecordRequestsMiddleware() func(*middleware.Stack) error {
 	}
 }
 
+// LogServerErrorsMiddleware is a middleware that logs server errors received when attempting to contact the AWS API
+// A specialized middleware is used instead of the SDK's built-in retry logging to allow for customizing the verbosity
+// of throttle errors vs server/unknown errors, to prevent flooding the logs with throttle errors
+func LogServerErrorsMiddleware() func(*middleware.Stack) error {
```
torredil, line 67: Instead of creating a new middleware function, could we use the existing [RecordRequestsMiddleware](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/e8df5a884b866cebd4e09a85354fd851189cde63/pkg/cloud/handlers.go#L51) (which already logs throttled errors) to handle the logging?

For example, non-throttled errors could be logged within this conditional: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/e8df5a884b866cebd4e09a85354fd851189cde63/pkg/cloud/handlers.go#L50-L52

ConnorJC3: Personally, I think it's better to leave these separate as even if it costs us a few lines of code, it separates the concerns and makes the logic much easier to understand compared to mixing the two.

That said, it's obviously possible to combine them so I'll leave this open to gather consensus on the best approach.

torredil: Should we decide to proceed with this change then would it make sense to at least migrate that log statement from `RecordRequestsMiddleware` to this new function?

ConnorJC3: Yes - 100%. Whichever decision, there should only be one place for logging the errors.

AndrewSirenko: +1 on separating concerns, and making RecordRequests not have the logging side effect.

pkg/cloud/handlers.go
```diff
@@ -60,6 +61,32 @@ func RecordRequestsMiddleware() func(*middleware.Stack) error {
...
+	return func(stack *middleware.Stack) error {
+		return stack.Finalize.Add(middleware.FinalizeMiddlewareFunc("LogServerErrorsMiddleware", func(ctx context.Context, input middleware.FinalizeInput, next middleware.FinalizeHandler) (output middleware.FinalizeOutput, metadata middleware.Metadata, err error) {
+			output, metadata, err = next.HandleFinalize(ctx, input)
+			if err != nil {
+				var apiErr smithy.APIError
+				if errors.As(err, &apiErr) {
+					if _, isThrottleError := retry.DefaultThrottleErrorCodes[apiErr.ErrorCode()]; isThrottleError {
+						// Only log throttle errors under a high verbosity as we expect to see many of them
```
AndrewSirenko, line 75: I would rather not change our current customer-facing log level for throttle errors in this PR, if the RecordRequest middleware has already been logging those requests at default verbosity. 

Customer may want to increase rate-limits instead of EBS CSI Driver silently taking longer to manage volumes. 

(On the other hand, customer can already alarm on rate-limit metric)

pkg/cloud/handlers.go
torredil: It seems odd duplicating an entire middleware function for a log statement but I support the team's decision, resolving.

pkg/cloud/handlers.go
ConnorJC3: The problem with this is that it will flood the error logs on bursty workloads with entirely benign error messages. Treating a retryable throttling response the same as a "VolumeCapacityExceeded" or "ServerInternal" error greatly reduces the signal to noise ratio of the logs.

torredil: I agree Connor that the existing logging for throttle errors is too aggressive and support increasing the default log level. 

Drew brings up a good point that we might want to split that out into another PR to make it more explicit. I don't have a strong preference, we can do it here as well by splitting it into a separate commit.

AndrewSirenko: > greatly reduces the signal to noise ratio of the logs.

I agree with this.

 My comment was mostly to call out that it's a customer facing change, so we should proceed with some caution and not sneak it in another PR. 

Separate commit is fine and/or a small note in the changelog. Resolving!

AndrewSirenko (approved):

ElijahQuinones (approved):

PR #2200, opened 2024-10-29
Update external-snapshotter to have same timeout as other sidecars

ElijahQuinones:
Nether
This PR updates the timeout of the snapshot-controller side car to be 60s to conform to the timeouts of the other sidecars set in [PR 1824](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1824)
**What testing is done?** 
CI


ElijahQuinones:
> Can we just use the default value? back in the day, we [decided](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1824#issuecomment-1805955674) not to set this value because the external snapshotter already has a [60s timeout value](https://github.com/kubernetes-csi/external-snapshotter/blob/33b4e6120ccc6427cf1ccc87b2314768ee37d754/cmd/csi-snapshotter/main.go#L71).
> 
> The default timeout value for other sidecars is much lower:
> 
> * volume-modifier: [10s](https://github.com/awslabs/volume-modifier-for-k8s/blob/25bb74de6d8a24145cda799bc8d99b51601519ab/cmd/main.go#L33)
> * resizer: [10s](https://github.com/kubernetes-csi/external-resizer/blob/f387899b5a5af4e601a93f1c963aa646f4e54c1b/cmd/csi-resizer/main.go#L67)
> * attacher: [15s](https://github.com/kubernetes-csi/external-attacher/blob/c8baf257d680036756ae7ef0acdfd3dddb06e77c/cmd/csi-attacher/main.go#L62)

PR #2204, opened 2024-11-01
Enable recommended golang-ci linters

AndrewSirenko:
Refactor

We rely on golang-ci's suite of linters to analyze our code for potential errors or style issues. We have had most linters disabled for a while, instead relying on PR reviewers to catch common errors. 

This PR re-enables golang-ci's linters, and refactors the code base to pass those linters.

Please see .golangci.yml for a list of enabled/disabled linters and individual commit messages for the code changes required to enable each of them. 

It was requested that this be a single PR, though if it takes too long to review I will split it into two PRs: 
- one for zero-risk changes (style)
- and one for slightly risky changes (refactoring) 

**What testing is done?** 
Manual testing, CI, make verify


cmd/hooks/prestop.go
```diff
@@ -59,18 +60,19 @@ func PreStop(clientset kubernetes.Interface) error {
 
 	nodeName := os.Getenv("CSI_NODE_NAME")
 	if nodeName == "" {
-		return fmt.Errorf("PreStop: CSI_NODE_NAME missing")
+		return errors.New("PreStop: CSI_NODE_NAME missing")
```
AndrewSirenko, line 63: Changing fmt.Errorf -> errors.New if Errorf is not necessary is one of the many performance improvements  `perfsprint` linter suggests.

I'm open to ignoring this specific rule if it is too annoying to future maintainers, but it's not too big of a deal ¯\_(ツ)_/¯

.golangci.yml
```diff
@@ -12,371 +12,135 @@
...
+    - maintidx # Maintainability index
+    - mnd # Magic Number Detection. Many false positives
+    - nestif # Don't allow too many nested if statements
+    - nlreturn # Always have empty line before return
+    - nonamedreturns
+    - testpackage # Require separate test package to catch leaky unexported dependencies
+    - varnamelen # Long var names happen
+    - wsl # Too strict of a whitespace linter
+    # Consider adding in future
+    - err113 # Do not create errors dynamically from scratch. Instead, wrap static (package-level) error.
```
AndrewSirenko, line 142: This linter is a 2-3 point task on its own to add meaningful error wraps for external package errors.

.golangci.yml
```diff
@@ -12,371 +12,135 @@
...
+    - wrapcheck # Same as err113
+    - gochecknoglobals  # Do not allow global variables
+    - godox # Do not allow TODOs
+    - paralleltest # There are many tests that aren't parallelized
```
AndrewSirenko, line 146: Took a stab at this but there are many tests that would require nolint because they are not written to be parallelized.

.golangci.yml
```diff
@@ -12,371 +12,135 @@
...
+    - nestif # Don't allow too many nested if statements
+    - nlreturn # Always have empty line before return
+    - nonamedreturns
+    - testpackage # Require separate test package to catch leaky unexported dependencies
+    - varnamelen # Long var names happen
+    - wsl # Too strict of a whitespace linter
+    # Consider adding in future
+    - err113 # Do not create errors dynamically from scratch. Instead, wrap static (package-level) error.
+    - wrapcheck # Same as err113
+    - gochecknoglobals  # Do not allow global variables
```
AndrewSirenko, line 144: Many many globals, 2+ point risky task on its own.

.golangci.yml
```diff
@@ -12,371 +12,135 @@
...
+      - "github.com/onsi/ginkgo/v2"
 linters:
   enable:
-    - megacheck
-    - govet
-  disable:
-    - maligned
-    - prealloc
-    - scopelint
+    - asasalint
```
ConnorJC3, line 39: Probably better to use `enable-all` (and the list of disables below) rather than listing every enable _and_ disable?

.golangci.yml
```diff
@@ -12,371 +12,135 @@
...
+    - gomnd # Magic Number Detection. Many false positives.
+    - gomoddirectives # We need `replace` in `go.mod`
+    - interfacebloat # No more than 10 interface methods
+    - ireturn # Accept interfaces return concrete types
+    - lll # Limit line length
+    - maintidx # Maintainability index
+    - mnd # Magic Number Detection. Many false positives
+    - nestif # Don't allow too many nested if statements
+    - nlreturn # Always have empty line before return
+    - nonamedreturns
```
ConnorJC3, line 137: This linter is missing a comment why we're disabling it. Also, isn't this a pretty bad anti-pattern - maybe it should be in the "Consider adding" section below?

cmd/main.go
```diff
@@ -108,7 +107,7 @@ func main() {
 			klog.ErrorS(err, "failed to get version")
 			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 		}
-		fmt.Println(versionInfo)
+		klog.Info(versionInfo)
```
ConnorJC3, line 110: This one we should probably put back in and `nolint`, it's normal to print version info entirely by itself, it shouldn't come with a timestamp like a log entry.

.golangci.yml
AndrewSirenko: I thought about this but then every time golang-ci updates we risk blocking CI because of new linters being added. 

Also there are some warnings because deprecated linters are still enabled in enable-all. 

IMO better to explicitly opt in to avoid this annoyance. I see no significant drawback to being more explicit about the linters we allow, we can always add more linters later.

.golangci.yml
ConnorJC3: Given that it's used in a bunch of places I don't think it would be reasonable at this point to enable it in this PR (unless you really want to volunteer to do the extra work). but I would suggest we move it down to the "maybe enable later" section so that we can make that decision in the future.

.golangci.yml
AndrewSirenko: > The primary drawback is that we have to search for and explicitly enable any new linter

Yep totally agree that we will miss out on new linters until someone explicitly adds them. (Though is that a bad thing)

I thought about adding a dependency upgrades runbook step of "If upgrading golang-ci, run with `--enable-all` to spot any new linters" but that is even more annoying than upgrading + manually disabling...

> deprecated linters is unfortunate

Yeah I couldn't find anything either. I'll consider raising an issue? But to be fair deprecated linter warnings are a small annoyance. 

---

Thanks for your thoughts, I don't have a strong opinion either way, it's a two way door that boils down to: 

"Do we want the person doing dependency upgrades to enable/disable new linters as they are added, or should we batch those decisions every once in a while"

I'll default to enabling all linters as you originally suggested unless @ElijahQuinones or @torredil have a preference on ^^

pkg/driver/controller.go
```diff
@@ -55,9 +55,10 @@ var (
 	}
 )
 
-const isManagedByDriver = "true"
+const trueStr = "true"
+const isManagedByDriver = trueStr
```
torredil, line 59: Could we remove `trueStr` and simplify this to 
```
const isManagedByDriver = "true"
```

pkg/cloud/volume_limits.go
```diff
@@ -28,6 +28,7 @@ const (
 	nitroMaxAttachments                  = 28
 )
 
+//nolint:gochecknoinits // TODO Refactor to avoid using init function to prevent side-effects
```
torredil, line 31: Unless its already in the backlog, could we create an issue about this so as to not forget it?

cmd/main.go
```diff
@@ -108,6 +108,7 @@ func main() {
 			klog.ErrorS(err, "failed to get version")
 			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 		}
+		//nolint:forbidigo
```
torredil, line 111: It's not clear why this comment is added unless you read through the PR discussion. Suggest adding that clarity directly in the comment so that future maintainers don't accidentally change this.

torredil (commented):
Largely lgtm 👍 
Left a few non-blocking comments.

/approve

pkg/driver/controller.go
```diff
@@ -489,7 +484,8 @@ func validateControllerUnpublishVolumeRequest(req *csi.ControllerUnpublishVolume
 
 func (d *ControllerService) ControllerGetCapabilities(ctx context.Context, req *csi.ControllerGetCapabilitiesRequest) (*csi.ControllerGetCapabilitiesResponse, error) {
 	klog.V(4).InfoS("ControllerGetCapabilities: called", "args", req)
-	var caps []*csi.ControllerServiceCapability
+
+	caps := make([]*csi.ControllerServiceCapability, 0, len(controllerCaps))
```
ElijahQuinones, line 488: If we are changing cap to capability should we change caps to capabilities ?

pkg/driver/controller.go
ConnorJC3: `trueStr` is used in other places (where "is managed by driver" doesn't make sense) so I don't think we can - if we want to remove one of the two it should be `isManagedByDriver`

pkg/driver/controller_modify_volume.go
```diff
@@ -90,31 +90,31 @@ func newModifyVolumeCoalescer(c cloud.Cloud, o *Options) coalescer.Coalescer[mod
 func mergeModifyVolumeRequest(input modifyVolumeRequest, existing modifyVolumeRequest) (modifyVolumeRequest, error) {
 	if input.newSize != 0 {
 		if existing.newSize != 0 && input.newSize != existing.newSize {
-			return existing, fmt.Errorf("Different size was requested by a previous request. Current: %d, Requested: %d", existing.newSize, input.newSize)
+			return existing, fmt.Errorf("different size was requested by a previous request. Current: %d, Requested: %d", existing.newSize, input.newSize)
```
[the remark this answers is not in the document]
torredil: https://staticcheck.dev/docs/checks#ST1005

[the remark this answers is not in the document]
ElijahQuinones: Thank you Eddie resolving comment !

pkg/driver/request_coalescing_test.go
```diff
@@ -218,6 +218,7 @@ func testRequestFail(t *testing.T, executor modifyVolumeExecutor) {
 // 3) Change volume type to NewVolumeType2
 // The expected result is the resizing request succeeds and one of the volume-type requests fails.
 func testPartialFail(t *testing.T, executor modifyVolumeExecutor) {
+	t.Helper()
 	const NewVolumeType1 = "gp3"
```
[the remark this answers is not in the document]
AndrewSirenko: This is unrelated to this PR

I'm impartial either way for the constant name.

pkg/driver/controller.go
AndrewSirenko: I was changing cap to capability to avoid colliding with built-in `cap` function, which this linter spotted.

Open to changing controllerCaps to controllerCapabilities if others want that in this PR.

pkg/driver/controller.go
AndrewSirenko: I left `isManagedByDriver` in because it was used in quite a few places to enhance readability. Can replace if we want.

I do not think we should be using isManagedByDriver to signify "true" in other places in the file.

pkg/driver/controller.go
ConnorJC3: Honestly I think that `caps` is clear enough in the context (and especially in `controllerCaps` keeps the variable size within reason), but don't have a strong opinion either way.

AndrewSirenko: Sounds good, resolving then.

ConnorJC3 (commented):
/lgtm (other than needing a squash - please don't approve until those fixup commits are squashed)

ConnorJC3 (approved):
/lgtm

PR #2207, opened 2024-11-05
fix write no permission

lucming:
bugfix

The storage plugin used by the application is changed from lvm to ebs, and an error is reported that there is no write permission. logs as below:
image-20241105
We found out that the mount point for the lvm plugin on the host machine had 777 permissions,  code: https://github.com/metal-stack/csi-driver-lvm/blob/master/pkg/lvm/lvm.go#L214 ,  while the ebs plugin's mount point had 755 permissions. After manually executing chmod 777,  the service running normally.

**What testing is done?**

### DIFF
--- a/pkg/mounter/mount_linux.go
+++ b/pkg/mounter/mount_linux.go
@@ -269,7 +269,7 @@ func (m *NodeMounter) MakeFile(path string) error {
 // This function is mirrored in ./sanity_test.go to make sure sanity test covered this block of code
 // Please mirror the change to func MakeFile in ./sanity_test.go.
 func (m *NodeMounter) MakeDir(path string) error {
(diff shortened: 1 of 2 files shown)


lucming:
> /close
> 
> Hi @lucming - unfortunately we will be unable to accept this PR. Creating mount directories with a permission set of `755` is an intentional security feature. Creating the directory with a permission of `777` would potentially allow any user on the host to control or modify the container's volume under the right conditions, as `777` grants all users write permissions.
> 
> If you need assistance with deploying the EBS CSI Driver (or would like to request a feature for your usecase), you can ask for support by creating an issue on this GitHub repository.

Does lvm not need to consider this issue?

PR #2212, opened 2024-11-06
Set image.pullPolicy to Always for Makefile installations

ConnorJC3:

Makefile/CI change

Because the tag is based on the cluster name, not setting `image.pullPolicy` in dev can lead to stale images. Set it to `Always` so that a fresh image is guaranteed.

**What testing is done?** 

Manually tested + CI


AndrewSirenko (approved):
/approve

ElijahQuinones (approved):

PR #2213, opened 2024-11-06
Update patch-cluster.yaml to v1beta1 for Vac

AndrewSirenko:

/kind cleanup

Update patch-cluster.yaml to v1beta1 for Vac so we can use our own VAC example in CI for Kops clusters.  

**What testing is done?**


ElijahQuinones (approved):
/lgtm

PR #2216, opened 2024-11-07
Expose NVMe Volume Metrics

torredil:

New feature.

This PR adds support for collecting detailed performance statistics for CSI managed devices. The log page for each device is retrieved at gather time, transformed into a prometheus-compatible format, and exposed at a `/metrics` HTTP endpoint.

**What testing is done?** 

```
make verify && make test

ginkgo run --focus='should create a stateful pod and read NVMe metrics from node plugin' -v
Test Suite Passed
```

```
curl 127.0.0.1:3302/metrics

...
# HELP volume_queue_length Current volume queue length
# TYPE volume_queue_length gauge
volume_queue_length{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e"} 0
# HELP write_io_latency_histogram Histogram of write I/O latencies (in microseconds)
# TYPE write_io_latency_histogram histogram
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="1"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="2"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="4"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="8"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="16"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="32"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="64"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="128"} 0
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="256"} 58
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="512"} 14260
write_io_latency_histogram_bucket{instance_id="i-0725b88fb342cd1c9",volume_id="vol-0aa37802a7c970c8e",le="1024"} 83115
...
```

### DIFF
--- a/pkg/driver/options.go
+++ b/pkg/driver/options.go
@@ -89,6 +89,8 @@ type Options struct {
 	WindowsHostProcess bool
 	// LegacyXFSProgs formats XFS volumes with `bigtime=0,inobtcount=0,reflink=0`, so that they can be mounted onto nodes with linux kernel ≤ v5.4. Volumes formatted with this option may experience issues after 2038, and will be unable to use some XFS features (for example, reflinks).
 	LegacyXFSProgs bool
+	// CsiMountPointPath is the path where CSI volumes are expected to be mounted on the node.
+	CsiMountPointPath string
 }
 
 func (o *Options) AddFlags(f *flag.FlagSet) {
@@ -118,6 +120,7 @@ func (o *Options) AddFlags(f *flag.FlagSet) {
 		f.IntVar(&o.ReservedVolumeAttachments, "reserved-volume-attachments", -1, "Number of volume attachments reserved for system use. Not used when --volume-attach-limit is specified. The total amount of volume attachments for a node is computed as: <nr. of attachments for corresponding instance type> - <number of NICs, if relevant to the instance type> - <reserved-volume-attachments value>. When -1, the amount of reserved attachments is loaded from instance metadata that captured state at node boot and may include not only system disks but also CSI volumes.")
 		f.BoolVar(&o.WindowsHostProcess, "windows-host-process", false, "ALPHA: Indicates whether the driver is running in a Windows privileged container")
 		f.BoolVar(&o.LegacyXFSProgs, "legacy-xfs", false, "Warning: This option will be removed in a future version of EBS CSI Driver. Formats XFS volumes with `bigtime=0,inobtcount=0,reflink=0`, so that they can be mounted onto nodes with linux kernel ≤ v5.4. Volumes formatted with this option may experience issues after 2038, and will be unable to use some XFS features (for example, reflinks).")
+		f.StringVar(&o.CsiMountPointPath, "csi-mount-point-prefix", "", "A prefix of the mountpoints of all CSI-managed volumes. If this value is non-empty, all volumes mounted to a path beginning with the provided value are assumed to be CSI volumes owned by the EBS CSI Driver and safe to treat as such (for example, by exposing volume metrics).")
 	}
 }
 
--- a/pkg/metrics/nvme.go
+++ b/pkg/metrics/nvme.go
@@ -1,0 +1,413 @@
+//go:build linux
+// +build linux
+
+// Copyright 2024 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the 'License');
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"bytes"
+	"encoding/binary"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+	"unsafe"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sys/unix"
+	"k8s.io/klog/v2"
+)
+
+// EBSMetrics represents the parsed metrics from the NVMe log page.
+type EBSMetrics struct {
+	EBSMagic              uint64
+	ReadOps               uint64
+	WriteOps              uint64
+	ReadBytes             uint64
+	WriteBytes            uint64
+	TotalReadTime         uint64
+	TotalWriteTime        uint64
+	EBSIOPSExceeded       uint64
+	EBSThroughputExceeded uint64
+	EC2IOPSExceeded       uint64
+	EC2ThroughputExceeded uint64
+	QueueLength           uint64
+	ReservedArea          [416]byte
+	ReadLatency           Histogram
+	WriteLatency          Histogram
+}
+
+type Histogram struct {
+	BinCount uint64
+	Bins     [64]HistogramBin
+}
+
+type HistogramBin struct {
+	Lower uint64
+	Upper uint64
+	Count uint64
+}
+
+// As defined in <linux/nvme_ioctl.h>.
+type nvmePassthruCommand struct {
+	opcode      uint8
+	flags       uint8
+	rsvd1       uint16
+	nsid        uint32
+	cdw2        uint32
+	cdw3        uint32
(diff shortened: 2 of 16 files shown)


tests/e2e/nvme_metrics.go
```diff
@@ -0,0 +1,166 @@
...
+
+		By("Verifying NVMe metrics")
+		expectedMetrics := []string{
+			"total_read_ops",
+			"total_write_ops",
+			"total_read_bytes",
+			"total_write_bytes",
+			"read_io_latency_histogram",
+			"write_io_latency_histogram",
+		}
```
ConnorJC3, line 125: Why less metrics here than in the unit test? Are we not expecting to also see metrics like `volume_queue_length` and `read_io_latency_histogram`?

tests/e2e/nvme_metrics.go
```diff
@@ -0,0 +1,166 @@
...
+				break
+			}
+		}
+		Expect(csiNodePod).NotTo(BeNil(), "No CSI node pod found")
+
+		By("Setting up port-forwarding")
+		args := []string{"port-forward",
+			"-n", "kube-system",
+			csiNodePod.Name,
+			"3302:3302"}
```
ConnorJC3, line 106: Unless this is being forced by a linter/gofmt, it's probably easier to read if these arguments are on the same line.

tests/e2e/nvme_metrics.go
torredil: Included the full list in latest rev.

tests/e2e/nvme_metrics.go
torredil: Yep, updated PR to include these arguments on the same line.

ConnorJC3 (approved):
/lgtm

pkg/metrics/nvme_test.go
```diff
@@ -0,0 +1,330 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := parseLogPage(tt.input)
+
+			if tt.wantErr != "" {
+				if err == nil {
+					t.Errorf("parseLogPage() error = nil, wantErr %v", tt.wantErr)
+					return
+				}
+				if !strings.Contains(err.Error(), tt.wantErr) {
```
AndrewSirenko, line 183: Is there a reason we prefer string checking string errors instead of using `errors.Is(err.Error(), tt.wantErr)`? 

This causes brittle unit tests due to duplication because we have to update error strings in tests as well as pkg/metrics/nvme.go

Instead we can wrap errors in the nvme.go and re-use those errors in unit tests.

pkg/driver/options.go
```diff
@@ -118,6 +120,7 @@ func (o *Options) AddFlags(f *flag.FlagSet) {
...
 		f.BoolVar(&o.WindowsHostProcess, "windows-host-process", false, "ALPHA: Indicates whether the driver is running in a Windows privileged container")
 		f.BoolVar(&o.LegacyXFSProgs, "legacy-xfs", false, "Warning: This option will be removed in a future version of EBS CSI Driver. Formats XFS volumes with `bigtime=0,inobtcount=0,reflink=0`, so that they can be mounted onto nodes with linux kernel ≤ v5.4. Volumes formatted with this option may experience issues after 2038, and will be unable to use some XFS features (for example, reflinks).")
+		f.StringVar(&o.CsiMountPointPath, "csi-mount-point-prefix", "", "A prefix of the mountpoints of all CSI-managed volumes. If this value is non-empty, all volumes mounted to a path beginning with the provided value are assumed to be CSI volumes owned by the EBS CSI Driver and safe to treat as such (for example, by exposing volume metrics).")
```
AndrewSirenko, line 123: We should probably add a unit test case for this in [pkg/driver/options_test.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/7bdd15e4848d625c3763b5eb5ac1b0038a536bdd/pkg/driver/options_test.go#L24)

torredil: Added unit test.

pkg/metrics/nvme_test.go
torredil: Sure, wrapped errors in nvme.go so that we can re-use them in unit tests.

pkg/metrics/nvme.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+
+		// Check mount source (field 3) for directly mounted NVMe devices
+		m := fields[3]
+		if strings.HasPrefix(m, "/nvme") {
+			device := "/dev" + m
+			deviceMap[device] = true
+		}
+
+		// Check root (field 9) for block devices
```
AndrewSirenko, line 349: Looking at the test coverage profile of nvme.go, it seems that we don't have unit tests for Collect and getCSIManagedDevices. 

I understand not unit testing some of the functions that just wrap binaries (eg executeLsblk and fetchDevicePathToVolumeIDMapping), especially because we have e2e tests, but are there really no unit tests we can do for ^^?

Especially the business logic in these lines is something we can protect with a unit test (and either fake the os.ReadFile, or extract this mountinfo manipulation into a separate, unit-testable function?)

What do you think?

---

Source: 

```
go test -v -coverprofile cover.out ./pkg/...
go tool cover -html cover.out -o cover.html
open cover.html
```

AndrewSirenko (approved):
Looks great, left a few non-blocking unit test comments/questions.

pkg/metrics/nvme.go
torredil: `Collect` and `getCSIManagedDevices` are great examples of code paths where end to end / integration tests are more appropriate. 

I'm not inclined to make these changes given that we don't gain any meaningful code coverage writing those mocks, it requires a significant amount of refactoring, and the code paths are covered by the end to end test.

AndrewSirenko:
/hold 

For @ElijahQuinones's review

ConnorJC3 (commented):
/lgtm

ElijahQuinones (approved):
/unhold
/approve

LGTM thank you for also adding the service in this PR :) !

SuperQ:
Hey, thanks for adding these metrics.

Unfortunately, there are a few things about this change that don't follow Prometheus best practices.

https://prometheus.io/docs/practices/naming/

Would you be willing to accept changes to adapt the data to meet best practices?

torredil:
Hi @SuperQ, thank you for reaching out about this and engaging with the community on Prometheus best practices.

We are aware of that documentation and its one of the things our team carefully evaluated when implementing this feature. While we generally strive to follow best practices, in this specific case we had a requirement and intentionally decided to maintain consistency with the [underlying EBS metrics output](https://docs.aws.amazon.com/ebs/latest/userguide/nvme-detailed-performance-stats.html#nvme-stats) to allow for direct mapping with the AWS documentation and CloudWatch metrics.

That said, could you share an example of the change you'd propose? I'd like to discuss this with the team if there is an opportunity to incorporate best practices for naming conventions while maintaining the consistency requirement.

SuperQ:
There's basically no way to make the underlying metric match best practices without breaking changes.

My changes would be:
* Prefix the metrics with `aws_ebs_csi...` to namespace the metrics as specific to this driver.
* Map the unit from microseconds to seconds to match unit best practices 
* Rename the metric to match best practices.

I also reviewed the code and found that the upstream data is missing the histogram cumulative sum data. ☹️

torredil:
That makes sense @SuperQ, I'll bring up this feedback for discussion during our next weekly sync and provide a response here shortly after. We also need to consider [existing metrics](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/metrics.md) exported by the controller plugin when adding/changing prefixes if the team decides to move forward with the proposal.

PR #2221, opened 2024-11-08
Finetune sidecar retry-interval-start

ElijahQuinones:
This PR changes the retry-interval-start of the provisioner, resizer, snapshotter and attacher sidecars
These changes where tested with scalability tests as well as testing an infeasible state within 30 minutes.

The scalability tests for the provisioner revealed that the overall time of the scalability tests was not negatively impacted by the increase in the retry-Interval-start,

The infeasible error tests revealed the following. 

```
2 second retry-interval-start -> 1 gRPC  saved 
4 second retry-interval-start -> 2 gRPCs saved 
8 second retry-interval-start -> 3 gRPCs saved 
```
(The above infeasible is the case for the resizer and snapshotter as well )

As the scalability tests revealed that there is no negative time impact from increasing the retry-interval-start I have oped to select `4 seconds `  this both saves us 2 gRPCs and is not a staggering increases in the case that there is something that the tests run did not catch. 

For the attacher side car the scalability tests revealed that the number of errors is not negatively impacted from moving from 1 second to 500 milliseconds.

The benefit of this  is we can not manually induces another  controllerPublishVolume gRPC and as such moving down to .5 seconds will allow us to have one more "instant" gRPC before we reach the 5 minute max. 

```release-note
Finetune provisioner, resizer, snapshotter, and attacher sidecars retry-interval-start value 
```


charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -444,6 +453,9 @@ spec:
             - --kube-api-qps=20
             - --kube-api-burst=100
             - --workers=100
+            {{- end }}
+             {{- if not (regexMatch "(-retry-interval-start)" (join " " .Values.sidecars.resizer.additionalArgs)) }}
```
ConnorJC3, line 457: nitpick: remove one extra space in the indent here

ConnorJC3 (commented):
lgtm other than one small nitpick

charts/aws-ebs-csi-driver/templates/controller.yaml
ElijahQuinones: Ahhh man was practicing my vim skills and must have missed it thank you will fix in rev 2

torredil (approved):

AndrewSirenko (approved):

deploy/kubernetes/base/controller.yaml
```diff
@@ -232,6 +235,7 @@ spec:
             - --kube-api-qps=20
             - --kube-api-burst=100
             - --workers=100
+            - --retry-interval-start=4s
```
AndrewSirenko, line 238: Do we still want to increase this value, given that the resizer sidecar will soon retry any InvalidArgument RPC errors at retry-interval-max? That will already prevent misconfigured K8s resources from spamming EC2 backend, without making other types of retries take longer.  

What types of errors other than InvalidArgument would we still want our customers to have a delayed retry for?

AndrewSirenko: Resolving because of the EV modification cooldown being a large source of non-InvalidArgument RPC errors.

PR #2230, opened 2024-11-14
Fix Trivy TOOMANYREQUESTS errors

ConnorJC3:
#### What type of PR is this?

/kind cleanup

This should eliminate or greatly reduce the number of `TOOMANYREQUESTS` errors hit by the Trivy action via:
- Only running the PR scanner on PRs to release branches (which is where we care about avoiding vulnerabilities in our dependencies)
- Adds ECR Public as a fallback to both Trivy actions

N/A

```release-note
NONE
```


torredil (approved):
/lgtm

AndrewSirenko (approved):

PR #2242, opened 2024-11-25
Update docs/design.md

AndrewSirenko:
#### What type of PR is this?

/kind documentation

This PR updates [docs/design.md](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/design.md) so that it is no longer out of date. It also provides a high-level overview of both components of the EBS CSI Driver and adds a few sequence diagrams to help visualize the most common driver workflows.   

n/a

```release-note
Update docs/design.md and add high-level overview. 
```

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -1,78 +1,116 @@
 # AWS EBS CSI Driver
-## Problems with current in-tree cloud provider
-### Cache of used / free device names
 
-On AWS, it&#39;s the client who [must assign device names](https://aws.amazon.com/premiumsupport/knowledge-center/ebs-stuck-attaching/) to volumes when calling AWS.AttachVolume. At the same time, AWS [imposes some restrictions on the device names](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html).
+## EBS CSI Driver on Kubernetes
 
-Therefore Kubernetes AWS volume plugin maintains cache of used / free device names for each node. This cache is lost when controller-manager process restarts. We try to populate the cache during startup, however there are some corner cases when this fails. TODO: exact flow how we can get wrong cache.
+### High Level Summary
 
-It would be great if either AWS itself assigned the device names, or there would be robust way how to restore the cache after restart, e.g. using some persistent database. Kubernetes should not care about the device names at all.
+The EBS CSI Driver is used by Container Orchestrators to manage the lifecycle of AWS Elastic Block Storage Volumes. It is compliant to the [Container Storage Interface Specification](https://github.com/container-storage-interface/spec/blob/master/spec.md).
 
-### DescribeVolumes quota
+On Kubernetes (K8s), the driver is split into two components: the ebs-csi-controller Deployment and ebs-csi-node DaemonSet. 
 
-In order to attach/detach volume to/from a node, current AWS cloud provider issues AWS.AttachVolume/DetachVolume call and then it polls DescribeVolume until the volume is attached or detached. The frequency of DescribeVolume is quite high to minimize delay between AWS finishing attachment of the volume and Kubernetes discovering that. Sometimes we even hit API quota for these calls.
+The ebs-csi-controller watches K8s storage resources for changes, acts on those changes by making AWS EC2 API calls, and then updates those K8s resources once all associated EBS volumes reflect those changes. Inside the ebs-csi-controller pod are several CSI sidecar containers that interact with Kubernetes resources, as well as the ebs-plugin container which manages EBS Volumes via AWS API calls. The sidecars trigger these EBS volume operations by making CSI Remote Procedure Calls (RPCs) against the ebs-plugin's controller service. 
 
-It would be better if CSI driver could get reliable and fast event from AWS when a volume has become attached / detached.
+The ebs-csi-node DaemonSet ensures the Kubelet can manage the attached EBS storage devices via the ebs-plugin container. The Kubelet ensures these EBS volumes appear as block devices or mounted filesystem directories inside the container workloads, by triggering CSI RPCs against the ebs-plugin's node service.   
 
-Or the driver could batch the calls and issue one big DescribeVolume call with every volume that&#39;s being attached/detached in it.
+If you remember one thing from this document, remember that:
+- ebs-csi-controller interacts with K8s storage resources, calls AWS EC2 APIs
+- ebs-csi-node runs on every Node, used by Kubelet to perform privileged operations on attached storage devices
 
-### AWS API weirdness
+To illustrate these relationships between Kubernetes, the EBS CSI Driver, and AWS, we can look at what happens when you dynamically provision a volume for a given stateful workload:
 
-AWS API is quite different to all other clouds.
+```mermaid
+sequenceDiagram
+    actor o as Operator
+    participant k as Kubernetes
+    participant c as ebs-csi-controller
+    participant a as AWS EC2 API
+    participant n as ebs-csi-node
+    participant os as Node OS
 
-- AWS.AttachVolume/DetachVolume can take ages (days, weeks) to complete. For example, when Kubernetes tries to detach a volume that&#39;s still mounted on a node, it will be Detaching until the volume is unmounted or Kubernetes issues force-detach. All other clouds return sane error relatively quickly (e.g. &quot;volume is in use&quot;) or force-detach the volume.
-- AWS.DetachVolume with force-detach is useless. Documentation says: Forced detachment of a stuck volume can cause damage to the file system or the data it contains or an inability to attach a new volume using the same device name, unless you reboot the instance.
-  - We cannot reboot instance after each force-detach nor we can afford to &quot;loose&quot; a device name. AWS supports only 40 volumes per node and even that is quite low number already.
-- AWS.CreateVolume is not idempotent. There is no way to create a volume with ID provided by user. Such call would then fail when such volume already exists.
-- AWS.CreateVolume does not return errors when creating an encrypted volume using either non-existing or non-available KMS key (e.g. with wrong permission). It returns success instead and it even returns volumeID of some volume! This volume exists for a short while and it&#39;s deleted in couple of seconds.
+    o->>k: Create PVC + Pod
+    activate k
 
-### Errors with slow kubelet
+    note over k: Pod 'Pending' <br/> PVC 'Pending'
+    k->>c: Notifies about PVC
+    activate c
+    c->>a: Ensure volume created
+    a-->>c: 
+    c-->>k: Create PV handle=<EBS volumeID> <br/> PVC and PV 'bound'
+    deactivate c
+    
+    note over k: Pod 'Scheduled' <br/> VolumeAttachment (VA) created
+    k->>c: Notify about VolumeAttachment (VA)
+    activate c
+    c->>a: Ensure volume attached
+    a-->>c: 
+    c-->>k: Update VA: <br/> Attached=true <br/> DevicePath='/dev/xvdaa'
+    deactivate c
+    
+    note over k: Pod 'ContainerCreating'
+    k->>n: Kubelet Triggers NodeStageVolume
+    activate n
+    opt If Filesystem 
+    n->>os: Format/fsck device
+    os-->>n: 
+    end
+    n->>os: Mount to '/var/lib/kubelet/<pod-id>/volumes/'
+    os-->>n: 
+    n-->>k: Kubelet update Node.volumesAttached
+    deactivate n
+    
+    k->>n: Kubelet Triggers NodePublishVolume
+    activate n
+    n->>os: Bind-mount to container's 'mountPath'
+    os-->>n: 
+    n-->>k: Kubelet update Node.volumesInUse
+    deactivate n
+    
+    note over k: Pod 'Running'
+    k-->>o: Stateful workload running
+    deactivate k
+```
 
-Very rarely a node gets too busy and kubelet starves for CPU. It does not unmount a volume when it should and Kubernetes initiates detach of the volume.
-
-## Requirements
+## CSI Driver Design Requirements
 
 ### Idempotency
 
-All CSI driver calls should be idempotent. A CSI method call with the same parameters must always return the same result. It&#39;s task of CSI driver to ensure that. Examples:
+All CSI calls should be idempotent. The CSI plugin must ensure that a CSI call with the same parameters will always return the same result. Examples:
 
 - CreateVolume call must first check that the requested EBS volume has been already provisioned and return it if so. It should create a new volume only when such volume does not exist.
 - ControllerPublish (=i.e. attach) does not do anything and returns &quot;success&quot; when given volume is already attached to requested node.
 - DeleteVolume does not do anything and returns success when given volume is already deleted (i.e. it does not exist, we don&#39;t need to check that it had existed and someone really deleted it)
 
-Note that it&#39;s task of the CSI driver to make these calls idempotent if related AWS API call is not.
+Note that it&#39;s task of the ebs-plugin to make these calls idempotent even if the related AWS API call is not.
 
 ### Timeouts
 
-gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (=CSI driver) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
+gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (i.e. ebs-plugin) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
 
-Kubernetes will retry failed calls, usually after some exponential backoff. Kubernetes heavily relies on idempotency here - i.e. when the driver finished an operation after the client timed out, the driver will get the same call again and it should return success/error based on success/failure of the previous operation.
+Kubernetes sidecars will retry failed calls, usually after some exponential backoff. These sidecars rely on idempotency here - i.e. when the plugin finished an operation after the client timed out, the plugin will get the same call again, and it should return success/error based on success/failure of the previous operation.
 
 Example:
 
-1. Kubernetes calls ControllerPublishVolume(vol1, nodeA) ), i.e. &quot;attach vol1 to nodeA&quot;.
-2. The CSI driver checks vol1 and sees it&#39;s not attached to nodeA yet. It calls AttachVolume(vol1, nodeA).
-3. The attachment takes a long time, Kubernetes times out.
-4. Kubernetes sleeps for some time.
+1. csi-attacher calls ControllerPublishVolume(vol1, nodeA) ), i.e. &quot;attach vol1 to nodeA&quot;.
+2. ebs-plugin checks vol1 and sees it&#39;s not attached to nodeA yet. It calls EC2 AttachVolume(vol1, nodeA).
+3. The attachment takes a long time, RPC times out.
+4. csi-attacher sleeps for some time.
 5. AWS finishes attaching of the volume.
-6. Kubernetes re-issues ControllerPublishVolume(vol1, nodeA) again.
-7. The CSI driver checks vol1 and sees it is attached to nodeA and returns success immediately.
+6. csi-attacher re-issues ControllerPublishVolume(vol1, nodeA) again.
+7. ebs-plugin checks vol1 and sees it is attached to nodeA and returns success immediately.
 
 Note that there are some issues:
 
-- Kubernetes can change its mind at any time. E.g. a user that wanted to run a pod on the node in the example got impatient so he deleted the pod at step 4. In this case Kubernetes will call ControllerUnpublishVolume(vol1, nodeA) to &quot;cancel&quot; the attachment request. It&#39;s up to the driver to do the right thing - e.g. wait until the volume is attached and then issue detach() and wait until the volume is detached and \*then\* return from
-- Note that Kubernetes may time out waiting for ControllerUnpublishVolume too. In this case, it will keep calling it until it gets confirmation from the driver that the volume has been detached (i.e. until the driver returns either success or non-timeout error) or it needs the volume attached to the node again (and it will call ControllerPublishVolume in that case).
+- Kubernetes can change its mind at any time. E.g. a user that wanted to run a pod on the node in the example got impatient so he deleted the pod at step 4. In this case csi-attacher will call ControllerUnpublishVolume(vol1, nodeA) to &quot;cancel&quot; the attachment request. It&#39;s up to the driver to do the right thing - e.g. wait until the volume is attached and then issue detach() and wait until the volume is detached and \*then\* return from
+- Note that Kubernetes may time out waiting for ControllerUnpublishVolume too. In this case, it will keep calling it until it gets confirmation from the driver that the volume has been detached (i.e. until ebs-plugin returns either success or non-timeout error) or it needs the volume attached to the node again (and it will call ControllerPublishVolume in that case).
 - The same applies to NodeStage and NodePublish calls (&quot;mount device, mount volume&quot;). These are typically much faster than attach/detach, still they must be idempotent when it comes to timeouts.
 
-It looks complicated, but it should be actually simple - always check that if the required operation has been already done
+In summary, always check that if the required operation has already been done
 
 ### Restarts
 
 The CSI driver should survive its own crashes or reboots of the node where it runs. For the controller service, Kubernetes will either start a new driver on a different node or re-elect a new leader of stand-by drivers. For the node service, Kubernetes will start a new driver shortly.
 
-The perfect CSI driver should be stateless. After start, it should recover its state by observing the actual status of AWS (i.e. describe instances / volumes). Current cloud provider follows this approach, however there are some corner cases around restarts when Kubernetes can try to attach two volumes to the same device on a node.
-
-When the stateless driver is not possible, it can use some persistent storage outside of the driver. Since the driver should support multiple Container Orchestrators (like Mesos), it must not use Kubernetes APIs. It should use AWS APIs instead to persist its state if needed (like AWS DynamoDB). We assume that costs of using such db will be negligible compared to rest of Kubernetes.
+The ideal CSI driver is stateless. After start, it should recover its state by observing the actual status of AWS (i.e. describe instances / volumes).
 
 ### No credentials on nodes
 
@@ -80,7 +118,15 @@ General security requirements we follow in Kubernetes is &quot;if a node gets co
 
 There should be a way how to run the CSI driver (=container) in &quot;node mode&quot; only. Such driver would then respond only to node service RPCs and it would not have any credentials to AWS (or very limited credentials, e.g. only to Describe things). Paranoid people would deploy CSI driver in &quot;node only&quot; mode on all nodes where Kubernetes runs user containers.
 
+### Cache of used / free device names
+
+On AWS, it&#39;s the client who [must assign device names](https://aws.amazon.com/premiumsupport/knowledge-center/ebs-stuck-attaching/) to volumes when calling AWS.AttachVolume. At the same time, AWS [imposes some restrictions on the device names](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html).
+
+Therefore, the ebs-plugin maintains a cache of used / free device names for each node. This cache is lost when the container restarts. We try to populate the cache during startup, however there are some corner cases when this fails. TODO: exact flow how we can get wrong cache.
+
 ## High level overview of CSI calls
+
+TODO Question: Thoughts on turning this into a Markdown table? 
 
 ### Identity Service RPC
 
@@ -90,7 +136,7 @@ Blindly return:
 
 ```
   Name: ebs.csi.aws.com
-  VendorVersion: 0.x.y
+  VendorVersion: 1.x.y
 ```
 
 #### GetPluginCapabilities
@@ -101,12 +147,13 @@ Blindly return:
    Capabilities:
      - CONTROLLER_SERVICE
      - ACCESSIBILITY_CONSTRAINTS
+     - ...
 ```
 
 #### Probe
 
-- Check that the driver is configured and it can do simple AWS operations, e.g. describe volumes or so.
-- This call is used by Kubernetes liveness probe to check that the driver is healthy. It&#39;s called every ~10 seconds, so it should not do anything &quot;expensive&quot; or time consuming.  (10 seconds are configurable, we can recommend higher values).
+- Check that the driver is configured can do simple AWS operations, e.g. describe volumes or so.
+- This call is used by Kubernetes liveness probe to check that the driver is healthy. It&#39;s called every ~10 seconds, so it should not do anything &quot;expensive&quot; or time-consuming.  (10 seconds are configurable, we can recommend higher values).
 
 ### Controller Service RPC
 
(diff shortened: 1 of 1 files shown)


docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+
+The ebs-csi-node DaemonSet ensures the Kubelet can manage the attached EBS storage devices via the ebs-plugin container. The Kubelet ensures these EBS volumes appear as block devices or mounted filesystem directories inside the container workloads, by triggering CSI RPCs against the ebs-plugin's node service.   
+
+If you remember one thing from this document, remember that:
+- ebs-csi-controller interacts with K8s storage resources, calls AWS EC2 APIs
+- ebs-csi-node runs on every Node, used by Kubelet to perform privileged operations on attached storage devices
+
+To illustrate these relationships between Kubernetes, the EBS CSI Driver, and AWS, we can look at what happens when you dynamically provision a volume for a given stateful workload:
+
+```mermaid
```
AndrewSirenko, line 21: Instead of having in-line diagrams, we can embed a `.mermaid` file from an attachments directory instead.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
+### Cache of used / free device names
+
+On AWS, it&#39;s the client who [must assign device names](https://aws.amazon.com/premiumsupport/knowledge-center/ebs-stuck-attaching/) to volumes when calling AWS.AttachVolume. At the same time, AWS [imposes some restrictions on the device names](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html).
+
+Therefore, the ebs-plugin maintains a cache of used / free device names for each node. This cache is lost when the container restarts. We try to populate the cache during startup, however there are some corner cases when this fails. TODO: exact flow how we can get wrong cache.
+
 ## High level overview of CSI calls
 
+TODO Question: Thoughts on turning this into a Markdown table?
```
AndrewSirenko, line 129: I lean towards having the RPCs listed in a table, so it's cleaner to look at. 

More detailed explanations can be done in an appendix. 

If we are in favor of this I will add a commit that turns it into a table.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
-Very rarely a node gets too busy and kubelet starves for CPU. It does not unmount a volume when it should and Kubernetes initiates detach of the volume.
+## EBS CSI Driver on Kubernetes
+
+### High Level Summary
+
+The EBS CSI Driver is used by Container Orchestrators to manage the lifecycle of AWS Elastic Block Storage Volumes. It is compliant to the [Container Storage Interface Specification](https://github.com/container-storage-interface/spec/blob/master/spec.md).
+
+On Kubernetes (K8s), the driver is split into two components: the ebs-csi-controller Deployment and ebs-csi-node DaemonSet. 
+
+The ebs-csi-controller watches K8s storage resources for changes, acts on those changes by making AWS EC2 API calls, and then updates those K8s resources once all associated EBS volumes reflect those changes. Inside the ebs-csi-controller pod are several CSI sidecar containers that interact with Kubernetes resources, as well as the ebs-plugin container which manages EBS Volumes via AWS API calls. The sidecars trigger these EBS volume operations by making CSI Remote Procedure Calls (RPCs) against the ebs-plugin's controller service.
```
ElijahQuinones, line 11: nit: gRPC not RPC

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+
+On Kubernetes (K8s), the driver is split into two components: the ebs-csi-controller Deployment and ebs-csi-node DaemonSet. 
+
+The ebs-csi-controller watches K8s storage resources for changes, acts on those changes by making AWS EC2 API calls, and then updates those K8s resources once all associated EBS volumes reflect those changes. Inside the ebs-csi-controller pod are several CSI sidecar containers that interact with Kubernetes resources, as well as the ebs-plugin container which manages EBS Volumes via AWS API calls. The sidecars trigger these EBS volume operations by making CSI Remote Procedure Calls (RPCs) against the ebs-plugin's controller service. 
+
+The ebs-csi-node DaemonSet ensures the Kubelet can manage the attached EBS storage devices via the ebs-plugin container. The Kubelet ensures these EBS volumes appear as block devices or mounted filesystem directories inside the container workloads, by triggering CSI RPCs against the ebs-plugin's node service.
```
ElijahQuinones, line 13: nit gRPC not RPC

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
-gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (=CSI driver) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
+gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (i.e. ebs-plugin) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
 
-Kubernetes will retry failed calls, usually after some exponential backoff. Kubernetes heavily relies on idempotency here - i.e. when the driver finished an operation after the client timed out, the driver will get the same call again and it should return success/error based on success/failure of the previous operation.
+Kubernetes sidecars will retry failed calls, usually after some exponential backoff. These sidecars rely on idempotency here - i.e. when the plugin finished an operation after the client timed out, the plugin will get the same call again, and it should return success/error based on success/failure of the previous operation.
```
ElijahQuinones, line 89: nit: All sidecars use an exponential backoff can we remove usually.

ElijahQuinones, line 89: nit: change plugin to csi plugin to be in line with rest of doc.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
 Example:
 
-1. Kubernetes calls ControllerPublishVolume(vol1, nodeA) ), i.e. &quot;attach vol1 to nodeA&quot;.
-2. The CSI driver checks vol1 and sees it&#39;s not attached to nodeA yet. It calls AttachVolume(vol1, nodeA).
-3. The attachment takes a long time, Kubernetes times out.
-4. Kubernetes sleeps for some time.
+1. csi-attacher calls ControllerPublishVolume(vol1, nodeA) ), i.e. &quot;attach vol1 to nodeA&quot;.
+2. ebs-plugin checks vol1 and sees it&#39;s not attached to nodeA yet. It calls EC2 AttachVolume(vol1, nodeA).
+3. The attachment takes a long time, RPC times out.
```
ElijahQuinones, line 95: nit: RPC -> gRPC

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
-- Note that Kubernetes may time out waiting for ControllerUnpublishVolume too. In this case, it will keep calling it until it gets confirmation from the driver that the volume has been detached (i.e. until the driver returns either success or non-timeout error) or it needs the volume attached to the node again (and it will call ControllerPublishVolume in that case).
+- Kubernetes can change its mind at any time. E.g. a user that wanted to run a pod on the node in the example got impatient so he deleted the pod at step 4. In this case csi-attacher will call ControllerUnpublishVolume(vol1, nodeA) to &quot;cancel&quot; the attachment request. It&#39;s up to the driver to do the right thing - e.g. wait until the volume is attached and then issue detach() and wait until the volume is detached and \*then\* return from
```
ElijahQuinones, line 103: Suggestion: change driver here to be ebs-plugin to fit updated example.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+- Note that Kubernetes may time out waiting for ControllerUnpublishVolume too. In this case, it will keep calling it until it gets confirmation from the driver that the volume has been detached (i.e. until ebs-plugin returns either success or non-timeout error) or it needs the volume attached to the node again (and it will call ControllerPublishVolume in that case).
 - The same applies to NodeStage and NodePublish calls (&quot;mount device, mount volume&quot;). These are typically much faster than attach/detach, still they must be idempotent when it comes to timeouts.
 
-It looks complicated, but it should be actually simple - always check that if the required operation has been already done
+In summary, always check that if the required operation has already been done
```
ElijahQuinones, line 107: nit: Add a period at end of sentence.

docs/design.md
ElijahQuinones: I really like this idea im game for it.

docs/design.md
```diff
@@ -157,19 +266,26 @@ Blindly return:
   rpc:
```
ElijahQuinones, line 266: Can we change this to gRPC

docs/design.md
```diff
@@ -339,6 +339,66 @@ Blindly return:
     - STAGE\_UNSTAGE\_VOLUME
 ```
 
+## Coalescing ControllerExpandVolume & ControllerModifyVolume
+
+### EC2 ModifyVolume and Request Coalescing
+
+AWS exposes one unified ModifyVolume API to change the size, volume-type, IOPS, or throughput of your volume. AWS imposes a 6-hour cooldown after a successful volume modification.
+
+However, the CSI Specification exposes two separate RPCs that rely on ebs-plugin calling this EC2 ModifyVolume API: ControllerExpandVolume, for increasing volume size, and ControllerModifyVolume, for all other volume modifications. To avoid the 6-hour cooldown, we coalesce these separate expansion and modification requests by waiting for up to two seconds, and then perform one merged EC2 ModifyVolume API Call.
```
ElijahQuinones, line 348: nit: RPCs -> gRPCs

docs/design.md
AndrewSirenko: IMO gRPC is more of an implementation detail of what kind of RPC is made, but can change if folks think it is helpful.

docs/design.md
AndrewSirenko: I'm pretty sure this is what the driver actually returns.

See [CSI Spec ControllerGetCapabilities response struct](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllergetcapabilities)

ElijahQuinones: gotcha will close accordingly

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+The ideal CSI driver is stateless. After start, it should recover its state by observing the actual status of AWS (i.e. describe instances / volumes).
 
 ### No credentials on nodes
 
 General security requirements we follow in Kubernetes is &quot;if a node gets compromised then the damage is limited to the node&quot;. Paranoid people typically dedicate handful of nodes in Kubernetes cluster as &quot;infrastructure nodes&quot; and dedicate these nodes to run &quot;infrastructure pods&quot; only. Regular users can&#39;t run their pods there. CSI attacher and provisioner is an example of such &quot;infrastructure pod&quot; - it need permission to create/delete any PV in Kubernetes and CSI driver running there needs credentials to create/delete volumes in AWS.
 
 There should be a way how to run the CSI driver (=container) in &quot;node mode&quot; only. Such driver would then respond only to node service RPCs and it would not have any credentials to AWS (or very limited credentials, e.g. only to Describe things). Paranoid people would deploy CSI driver in &quot;node only&quot; mode on all nodes where Kubernetes runs user containers.
```
torredil, line 119: Thoughts on also removing the "No credentials on nodes" section? (which I understand has been in here for several years). It adds an arbitrary restriction, its not actually a design requirement. At the very least,  please replace "Paranoid people" with "security conscious users".

torredil (approved):
Largely lgtm. Especially the sequence diagrams, this revision significantly elevates the doc.

/lgtm!

ElijahQuinones (approved):

docs/design.md
AndrewSirenko: Hmm, I'm inclined to keep this section because I believe it lays out a few great points on why node mode container should have limited credentials.

Will replace 'paranoid' wording though. Thanks!

youwalther65: I am still a bit confused about the second paragraph of this chapter. Why should people run an EBS CSI driver in "node only" mode which is not able to provision any EBS volume (because it only has Describe* permissions)

AndrewSirenko: Both the EBS CSI Driver Controller Deployment and Node Daemonset pods have the same `ebs-plugin` container. The difference is that in the controller pod we set the plugin's mode to `controller`, and in the node pods we set the plugin's mode to `node`. For example, this node mode is set on Daemonset's ebs-plugin container in the helm template [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/39822b5ceb4d8565e15039715dfdc622c14fa6e5/charts/aws-ebs-csi-driver/templates/_node.tpl#L72)

Because the node pods set the ebs-plugin container to node mode, it won't respond to Controller Service RPCs (ControllerCreateVolume). 

Today, we also don't give any of the daemonset's pods access to the EBS CSI Controller IAM Role, because the Node Service RPCs like NodeStageVolume don't need to talk to the EC2 API. 

Finally, we ensure only the EBS CSI Controller pods have the higher-risk K8s RBAC for actions like patching PV and VolumeAttachment  resources. 

Because of the 3 points above, a security conscious customer can ensure that the EBS CSI Controller pods are only scheduled on extra hardened nodes. If an intruder gains access to any of the other nodes on the cluster they wouldn't be able to create/delete EBS volumes via the `ebs-csi-node` pod, or mess with the cluster's storage resources. 

Did I understand your question correctly @youwalther65?

youwalther65: Exactly. Great, I'd like to see this answer in the doc itself!

docs/design.md
```diff
@@ -101,28 +147,47 @@ Blindly return:
...
      - CONTROLLER_SERVICE
      - ACCESSIBILITY_CONSTRAINTS
+     - ...
 ```
 
 #### Probe
 
-- Check that the driver is configured and it can do simple AWS operations, e.g. describe volumes or so.
-- This call is used by Kubernetes liveness probe to check that the driver is healthy. It&#39;s called every ~10 seconds, so it should not do anything &quot;expensive&quot; or time consuming.  (10 seconds are configurable, we can recommend higher values).
+- Check that the driver is configured can do simple AWS operations, e.g. describe volumes or so.
```
ConnorJC3, line 155: This isn't actually true (and I don't think ever was?) and should be dropped

docs/design.md
```diff
@@ -133,21 +198,65 @@ Checks if the required volume exists and is &quot;available&quot; (not attached
...
 
 #### ListVolumes
 
-Not implemented in the initial release, Kubernetes does not need it.
+Not implemented, Kubernetes does not need it.
 
 #### GetCapacity
 
-Not implemented in the initial release, Kubernetes does not need it.
+Not implemented, Kubernetes does not need it.
```
ConnorJC3, line 259: Same as above, drop "Kubernetes does not need it"

ConnorJC3: Actually, there are so many RPCs that we don't implement (e.g. anything related to group snapshot) we should probably drop any not implemented RPCs from the doc rather than trying to list them all.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+
+The ebs-csi-controller watches K8s storage resources for changes, acts on those changes by making AWS EC2 API calls, and then updates those K8s resources once all associated EBS volumes reflect those changes. Inside the ebs-csi-controller pod are several CSI sidecar containers that interact with Kubernetes resources, as well as the ebs-plugin container which manages EBS Volumes via AWS API calls. The sidecars trigger these EBS volume operations by making CSI Remote Procedure Calls (RPCs) against the ebs-plugin's controller service. 
+
+The ebs-csi-node DaemonSet ensures the Kubelet can manage the attached EBS storage devices via the ebs-plugin container. The Kubelet ensures these EBS volumes appear as block devices or mounted filesystem directories inside the container workloads, by triggering CSI RPCs against the ebs-plugin's node service.   
+
+If you remember one thing from this document, remember that:
+- ebs-csi-controller interacts with K8s storage resources, calls AWS EC2 APIs
+- ebs-csi-node runs on every Node, used by Kubelet to perform privileged operations on attached storage devices
```
ConnorJC3, line 17: Drop this paragraph entirely, it is just a duplicate of information already listed in a "high level summary".

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
-All CSI driver calls should be idempotent. A CSI method call with the same parameters must always return the same result. It&#39;s task of CSI driver to ensure that. Examples:
+All CSI calls should be idempotent. The CSI plugin must ensure that a CSI call with the same parameters will always return the same result. Examples:
 
 - CreateVolume call must first check that the requested EBS volume has been already provisioned and return it if so. It should create a new volume only when such volume does not exist.
 - ControllerPublish (=i.e. attach) does not do anything and returns &quot;success&quot; when given volume is already attached to requested node.
 - DeleteVolume does not do anything and returns success when given volume is already deleted (i.e. it does not exist, we don&#39;t need to check that it had existed and someone really deleted it)
 
-Note that it&#39;s task of the CSI driver to make these calls idempotent if related AWS API call is not.
+Note that it&#39;s task of the ebs-plugin to make these calls idempotent even if the related AWS API call is not.
```
ConnorJC3, line 83: Drop the first half of this diff, `ebs-plugin` is a confusing term and means the same thing as "CSI driver". We should be using what is essentially an internal term with no meaning, when "CSI Driver" is the official and well understood term.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
 ### Timeouts
 
-gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (=CSI driver) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
+gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (i.e. ebs-plugin) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
```
ConnorJC3, line 87: Ditto above, drop this diff.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
-Kubernetes will retry failed calls, usually after some exponential backoff. Kubernetes heavily relies on idempotency here - i.e. when the driver finished an operation after the client timed out, the driver will get the same call again and it should return success/error based on success/failure of the previous operation.
+Kubernetes sidecars will retry failed calls after exponential backoff. These sidecars rely on idempotency here - i.e. when ebs-plugin finished an operation after the client timed out, the ebs-plugin will get the same call again, and it should return success/error based on success/failure of the previous operation.
 
 Example:
 
-1. Kubernetes calls ControllerPublishVolume(vol1, nodeA) ), i.e. &quot;attach vol1 to nodeA&quot;.
-2. The CSI driver checks vol1 and sees it&#39;s not attached to nodeA yet. It calls AttachVolume(vol1, nodeA).
-3. The attachment takes a long time, Kubernetes times out.
-4. Kubernetes sleeps for some time.
+1. csi-attacher calls ControllerPublishVolume(vol1, nodeA) ), i.e. &quot;attach vol1 to nodeA&quot;.
```
ConnorJC3, line 93: Keep the wording Kubernetes, if you want to call out which sidecar makes the request you could do `Kubernetes (via the csi-attacher sidecar) calls ...`

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+2. ebs-plugin checks vol1 and sees it&#39;s not attached to nodeA yet. It calls EC2 AttachVolume(vol1, nodeA).
```
ConnorJC3, line 94: Ditto above, drop this diff.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+3. The attachment takes a long time, RPC times out.
+4. csi-attacher sleeps for some time.
 5. AWS finishes attaching of the volume.
-6. Kubernetes re-issues ControllerPublishVolume(vol1, nodeA) again.
-7. The CSI driver checks vol1 and sees it is attached to nodeA and returns success immediately.
+6. csi-attacher re-issues ControllerPublishVolume(vol1, nodeA) again.
```
ConnorJC3, line 98: Ditto above, keep the wording "Kubernetes".

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+7. ebs-plugin checks vol1 and sees it is attached to nodeA and returns success immediately.
```
ConnorJC3, line 99: Ditto above, drop this diff.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
-- Note that Kubernetes may time out waiting for ControllerUnpublishVolume too. In this case, it will keep calling it until it gets confirmation from the driver that the volume has been detached (i.e. until the driver returns either success or non-timeout error) or it needs the volume attached to the node again (and it will call ControllerPublishVolume in that case).
+- Kubernetes can change its mind at any time. E.g. a user that wanted to run a pod on the node in the example got impatient so he deleted the pod at step 4. In this case csi-attacher will call ControllerUnpublishVolume(vol1, nodeA) to &quot;cancel&quot; the attachment request. It&#39;s up to the ebs-plugin to do the right thing - e.g. wait until the volume is attached and then issue detach() and wait until the volume is detached and \*then\* return from
```
ConnorJC3, line 103: Ditto above, keep the wording "Kubernetes".

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
+- Note that Kubernetes may time out waiting for ControllerUnpublishVolume too. In this case, it will keep calling it until it gets confirmation from the driver that the volume has been detached (i.e. until ebs-plugin returns either success or non-timeout error) or it needs the volume attached to the node again (and it will call ControllerPublishVolume in that case).
```
ConnorJC3, line 104: Ditto above, drop this diff.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 - The same applies to NodeStage and NodePublish calls (&quot;mount device, mount volume&quot;). These are typically much faster than attach/detach, still they must be idempotent when it comes to timeouts.
 
-It looks complicated, but it should be actually simple - always check that if the required operation has been already done
+In summary, always check that if the required operation has already been done.
```
ConnorJC3, line 107: Suggest wording "is already completed" instead of "has already been done".

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
 ### Restarts
 
 The CSI driver should survive its own crashes or reboots of the node where it runs. For the controller service, Kubernetes will either start a new driver on a different node or re-elect a new leader of stand-by drivers. For the node service, Kubernetes will start a new driver shortly.
 
-The perfect CSI driver should be stateless. After start, it should recover its state by observing the actual status of AWS (i.e. describe instances / volumes). Current cloud provider follows this approach, however there are some corner cases around restarts when Kubernetes can try to attach two volumes to the same device on a node.
-
-When the stateless driver is not possible, it can use some persistent storage outside of the driver. Since the driver should support multiple Container Orchestrators (like Mesos), it must not use Kubernetes APIs. It should use AWS APIs instead to persist its state if needed (like AWS DynamoDB). We assume that costs of using such db will be negligible compared to rest of Kubernetes.
+The ideal CSI driver is stateless. After start, it should recover its state by observing the actual status of AWS (i.e. describe instances / volumes).
```
ConnorJC3, line 113: Worth including language about non-stateless parts we have here? e.g. "likely bad device names" map?

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
 There should be a way how to run the CSI driver (=container) in &quot;node mode&quot; only. Such driver would then respond only to node service RPCs and it would not have any credentials to AWS (or very limited credentials, e.g. only to Describe things). Paranoid people would deploy CSI driver in &quot;node only&quot; mode on all nodes where Kubernetes runs user containers.
 
+### Cache of used / free device names
+
+On AWS, it&#39;s the client who [must assign device names](https://aws.amazon.com/premiumsupport/knowledge-center/ebs-stuck-attaching/) to volumes when calling AWS.AttachVolume. At the same time, AWS [imposes some restrictions on the device names](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html).
+
+Therefore, the ebs-plugin maintains a cache of used / free device names for each node. This cache is lost when the container restarts. We try to populate the cache during startup, however there are some corner cases when this fails. TODO: exact flow how we can get wrong cache.
```
ConnorJC3, line 125: This isn't true - no such cache exists. We call `DescribeInstances` to look at the block device mappings for every single attach call.

docs/design.md
```diff
@@ -133,21 +198,65 @@ Checks if the required volume exists and is &quot;available&quot; (not attached
...
+
 #### ControllerUnpublishVolume
 
 Checks that given volume is not attached to given node. Returns success if so. Issues AWS.DetachVolume and marks the detached device name as free (more on that below). TODO: this has complicated idempotency expectations. It cancels previously called ControllerPublishVolume (i.e.AWS is still attaching the volume and Kubernetes now wants the volume to be detached).
 
+#### ControllerExpandVolume
+
+Checks that given volume is not expanded yet, calls EC2 ModifyVolume and ensures the modification enters the 'optimizing' state. 
+
+Note: If ControllerModifyVolume is triggered within 2 seconds of ControllerExpandVolume, they will share an EC2 ModifyVolume call.
```
ConnorJC3, line 239: Suggest saying `approximately 2 seconds` here as the way we do this isn't exact.

docs/design.md
```diff
@@ -133,21 +198,65 @@ Checks if the required volume exists and is &quot;available&quot; (not attached
...
+
+Note: If ControllerModifyVolume is triggered within 2 seconds of ControllerExpandVolume, they will share an EC2 ModifyVolume call. 
+
+#### ControllerModifyVolume
+
+Checks if volume needs modification, calls EC2 ModifyVolume, and ensures modification enters 'optimizing' state.
+
+- If tags need to be created/modified, call EC2 CreateTags. 
+
+Note: If ControllerExpandVolume is triggered within 2 seconds of ControllerModifyVolume, they will share an EC2 ModifyVolume call.
```
ConnorJC3, line 247: Ditto above recommendation on `approximately 2 seconds`.

docs/design.md
```diff
@@ -183,18 +299,24 @@ Not implemented yet.
...
-
-Just unmount the volume.
-
 #### NodePublishVolume
 
 Just bind-mount the volume.
 
+#### NodeUnstageVolume
+
+Just unmount the volume.
```
ConnorJC3, line 308: I understand you didn't add it, but suggest dropping "Just" from all these sections.

docs/design.md
```diff
@@ -183,18 +299,24 @@ Not implemented yet.
...
 #### NodeUnpublishVolume
 
 Just unmount the volume.
 
+#### NodeExpandVolume
+
+If the attached volume has been formatted with a filesystem, resize the filesystem.
+
+#### NodeGetVolumeStats
+
```
ConnorJC3, line 319: Empty section?

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
 
 ### Timeouts
 
-gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (=CSI driver) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
+gRPC always passes a timeout together with a request. After this timeout, the gRPC client call actually returns. The server (i.e. ebs-plugin) can continue processing the call and finish the operation, however it has no means how to inform the client about the result.
 
-Kubernetes will retry failed calls, usually after some exponential backoff. Kubernetes heavily relies on idempotency here - i.e. when the driver finished an operation after the client timed out, the driver will get the same call again and it should return success/error based on success/failure of the previous operation.
+Kubernetes sidecars will retry failed calls after exponential backoff. These sidecars rely on idempotency here - i.e. when ebs-plugin finished an operation after the client timed out, the ebs-plugin will get the same call again, and it should return success/error based on success/failure of the previous operation.
```
ConnorJC3, line 89: Ditto above, drop the "driver" -> "ebs-plugin" part of this diff.

docs/design.md
ConnorJC3: I suggest we stick with RPC except in the specific few sentences where it specifically matters that the protocal is gRPC.

As a guideline - if CSI were to switch to HTTP tomorrow, and all you would do for a sentence is replace "gRPC" with "HTTP RPC", then it should be "RPC" as the transport mechanism does not matter.

For the sections where it really does matter that it is gRPC, such as the sections explaining timeout handling, it should say gRPC.

docs/design.md
```diff
@@ -1,87 +1,133 @@
...
-  - We cannot reboot instance after each force-detach nor we can afford to &quot;loose&quot; a device name. AWS supports only 40 volumes per node and even that is quite low number already.
-- AWS.CreateVolume is not idempotent. There is no way to create a volume with ID provided by user. Such call would then fail when such volume already exists.
-- AWS.CreateVolume does not return errors when creating an encrypted volume using either non-existing or non-available KMS key (e.g. with wrong permission). It returns success instead and it even returns volumeID of some volume! This volume exists for a short while and it&#39;s deleted in couple of seconds.
 
-### Errors with slow kubelet
-
-Very rarely a node gets too busy and kubelet starves for CPU. It does not unmount a volume when it should and Kubernetes initiates detach of the volume.
+## EBS CSI Driver on Kubernetes
+
+### High Level Summary
```
ConnorJC3, line 5: This section is dangerously close to documenting CSI and/or CSI sidecar design rather than EBS CSI design. For example, if the information here can be found in the CSI spec's architecture section (https://github.com/container-storage-interface/spec/blob/master/spec.md#architecture) we are better off linking there than making our own version.

docs/design.md
```diff
@@ -217,6 +339,66 @@ Blindly return:
...
+
+    o->>s: Updated PVC VACName + PVC capacity
+    activate o
+    activate s
+    
+    s->>p: ControllerExpandVolume RPC
+    activate p
+    note over p: Wait up to 2s for other RPC
+    s->>p:  ControllerModifyVolume RPC
+
```
AndrewSirenko, line 370: Need to remember to add a describevolumes call here.

docs/design.md
AndrewSirenko: I politely initially disagree with this feedback. I think the phrasing 'ebs-plugin' is clearer because that is what the container is called within our driver. Also the upstream csi driver documentation refers to this component (that gets called by sidecars and interacts with storage provider) as driver-plugin, no?

Calling both the whole deployment, and the container itself the EBS CSI Driver confuses many operators new to our driver, including me when I first started. 

Open to hear more thoughts though.

AndrewSirenko: Well, sometimes CSI documentation does use plugin to refer to sidecars and company

"Node Plugin

The node component should be deployed on every node in the cluster through a DaemonSet. It consists of the CSI driver that implements the CSI Node service and the node-driver-registrar sidecar container"

Is there a better term we can use to differentiate?

ConnorJC3: > Calling both the whole deployment, and the container itself the EBS CSI Driver confuses many operators new to our driver, including me when I first started.

I would agree this could be confusing, but it is the former ("the whole deployment"), not the latter ("the container itself") that is inaccurate and causing the confusion. I don't think this document describes the `Deployment` or `DaemonSet` anywhere as "the EBS CSI Driver" with no qualifications, but if it does that language should be altered.

Using the term `ebs-plugin` causes confusion because it is not the standard terminology. It is important to get the terminology right here because the term "CSI Driver" is used this way extensively throughout the community and upstream code/documentation (e.g. https://kubernetes-csi.github.io/docs).

Depending on how the EBS CSI Driver is deployed, it may not even have an `ebs-plugin` container. For example, here is how the semi-popular Gardener tool deploys the driver, they use the name `csi-driver` for the driver container: https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml#L43

AndrewSirenko: Let me include a short glossary/terminology ala https://github.com/awslabs/soci-snapshotter/blob/main/docs/glossary.md in a future PR where we can make this distinction crisp. 

> Using the term ebs-plugin causes confusion because it is not the standard terminology.

On the other hand the term 'ebs-plugin container' here can be helpful because it lets operators know exactly which container to look at for troubleshooting. In the glossary section, we can mention that as of v1.37.0, our csi driver is deployed as the `ebs-plugin` container by default. 

What's confusing to newcomers here is that  the whole package of Deployment + Daemonset is often referred to as the EBS CSI Driver add-on. Perhaps a glossary can remedy this. 

> Depending on how the EBS CSI Driver is deployed, it may not even have an ebs-plugin container. For example Gardener.

Do they even deploy the EBS CSI Driver Controller? I consider installation methods not explicitly approved by this repository as out of scope for this document.

docs/design.md
AndrewSirenko: It's a short summary that we want to re-iterate because of how essential it is. When ramping up folks on the EBS CSI Driver, I've had to re-iterate this point several times because it is not obvious.

I can rename the section to high-level overview if that helps, but I value keeping this here.

docs/design.md
AndrewSirenko: Forgot to leave a TODO here. 

Might add the following: "Returns the amount of available/total/used bytes and inodes for a given volume.

docs/design.md
youwalther65: The wording might be misleading a bit and some folks may think these are containers . Probably something like:
EBS CSI controller component runs as a K8s Deployment and interacts ...
EBS CSI node component runs as a K8s DaemonSet on every node, ...

docs/design.md
AndrewSirenko: And that modification state could be either optimizing OR completed

docs/design.md
AndrewSirenko: We already reference/link the CSI Spec in this section. But I will retitle the section to make it clear I'm talking about EBS volume management on Kubernetes, not just EBS CSI Driver. 

The CSI Spec's architecture section lists four different possible architectures for CSI drivers and it is not immediately clear which one applies to the EBS CSI installation. Figure one mentions:

```
a centralized Controller Plugin is available on the CO master host and the Node Plugin is available on all of the CO Nodes.
```

For EKS customers, the controller plugin is on SOME node, not `the CO master host`.

Secondly, this section explicitly explains the difference between `ebs-csi-controller` and `ebs-csi-node` Pods, which prevents common mistakes many folks make when ramping up on EBS CSI Driver (e.g. I got the "ebs csi driver logs", but it didn't include any Create/AttachVolume calls, only mount operations). 

Most importantly, multiple CSEs/customers/TAMs have explicitly said that this summary is more helpful than linking multiple different sources of Kubernetes/CSI/EBS documentation for understanding and troubleshooting EBS-backed volumes on their cluster.

AndrewSirenko: Discussed offline, totally agree on putting more emphasis on ebs csi DRIVER instead of the kubernetes components. Will do some rework/cutting.

PR #2244, opened 2024-11-26
Add FIPS image and Helm option

ConnorJC3:
#### What type of PR is this?

/kind feature

This PR adds a FIPS image and helm chart option `fips` that enables that image as well as deplying with the environment variable needed to instruct the AWS SDK to use FIPS endpoints.

Manually/CI

```release-note
Build additional FIPS image and add Helm FIPS parameter
```

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -11,6 +11,12 @@ image:
 customLabels: {}
 # k8s-app: aws-ebs-csi-driver
 
+# Instruct the AWS SDK to use AWS FIPS endpoints, and deploy container built with BoringCrypto (a FIPS-validated cryptographic library) instead of the Go default
+#
+# The EBS CSI Driver FIPS images have not undergone FIPS certification, and no official guarnatee is made about the compliance of these images under the FIPS standard
+# Users relying on these images for FIPS compliance should perform their own independent evaluation
+fips: false
+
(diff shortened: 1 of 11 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -11,6 +11,12 @@ image:
 customLabels: {}
 # k8s-app: aws-ebs-csi-driver
 
+# Instruct the AWS SDK to use AWS FIPS endpoints, and deploy container built with BoringCrypto (a FIPS-validated cryptographic library) instead of the Go default
+#
+# The EBS CSI Driver FIPS images have not undergone FIPS certification, and no official guarnatee is made about the compliance of these images under the FIPS standard
+# Users relying on these images for FIPS compliance should perform their own independent evaluation
+fips: false
```
ElijahQuinones, line 18: nit: can we change it to enableFips

ConnorJC3: I thought about this, but it's not really "enabling" FIPS so the name would be misleading. Maybe `fipsMode` or `enableFipsMode` although those sound kinda janky.

AndrewSirenko (approved):

torredil (approved):
Manually tested changes and found no issues.

/lgtm
/approve

AndrewSirenko:
/retest

looks like flake

/approve

PR #2247, opened 2024-11-27
Fix supported filesystem types

AndrewSirenko:
#### What type of PR is this?

/kind cleanup

This PR fixes our list of supported filesystems by removing any instance of `ext2` from our project. EBS CSI Driver has not been able to manage volumes formatted with `ext2` for many versions. 

Our driver image has not supported the `ext2` binary since we've moved to a minimal base image, and resize doesn't work on `ext2` which presents a footgun for those attempting to edit the size of their PVC.

CI

```release-note
Remove ext2 from list of supported filesystem types.
```


torredil (commented):
np: suggest explicitly mentioning that `ext2` is being removed from the list of supported fs types in the release notes.

ie, Fix list of supported filesystem types -> Remove `ext2` from list of supported filesystem types.

ElijahQuinones (approved):
LGTM double checked unit tests and confirmed we have no cases where we use ext2

PR #2248, opened 2024-11-27
Update metrics to align with Prometheus best practices

torredil:
#### What type of PR is this?

/kind cleanup

This PR updates the metrics registered by the node plugin to align with Prometheus best practices by:
- Adding a `aws_ebs_csi` namespace prefix.
- Following Prometheus conventions such as adding `_total` suffix for counter metrics.
- Using base units (seconds) instead of microseconds.

It also introduces a new parameter `controller.deprecatedMetrics` to disable registering the existing controller metrics that use an outdated namespace, such as:

```
cloudprovider_aws_api_request_duration_seconds_bucket{request="AttachVolume",le="0.005"} 0
```

Metrics using the new namespace will be registered by default if `controller.enableMetrics` is configured:
```
aws_ebs_csi_api_request_duration_seconds_bucket{request="AttachVolume",le="0.005"} 0
```

- `make verify && make test`
- Manually
- CI

```release-note
BREAKING CHANGE: Node plugin metrics have been renamed to follow Prometheus best practices:
- Added "aws_ebs_csi_" prefix.
- Added "_total" suffix for counters.
- Changed time units from microseconds to seconds for all counters.

Controller plugin metrics now use the "aws_ebs_csi_" prefix instead of "cloudprovider_aws_".
Old metric names can be disabled via the controller.deprecatedMetrics parameter.
```

#### Note for reviewers:

Please note that time units have been changed from microseconds to seconds for all counters. The read/write IO latency histograms is observed in microseconds (the buckets for these histograms are retrieved directly from the log page which record the data in microseconds).

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -8,7 +8,7 @@ $ helm repo add prometheus-community https://prometheus-community.github.io/helm
 $ helm repo update
 $ helm install prometheus prometheus-community/kube-prometheus-stack
 ```
-2. Enable metrics by setting `enableMetrics: true` in [values.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml).
+2. Enable metrics by configuring `controller.enableMetrics` and `node.enableMetrics`.
 
 3. Deploy EBS CSI Driver:
 ```sh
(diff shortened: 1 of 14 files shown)


torredil:
cc: @SuperQ

docs/metrics.md
```diff
@@ -8,7 +8,7 @@ $ helm repo add prometheus-community https://prometheus-community.github.io/helm
 $ helm repo update
 $ helm install prometheus prometheus-community/kube-prometheus-stack
 ```
-2. Enable metrics by setting `enableMetrics: true` in [values.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml).
+2. Enable metrics by configuring `controller.enableMetrics` or `node.enableMetrics`.
```
AndrewSirenko, line 11: Consider replacing 'or' with 'and' to make it clearer that you need to separately enable both.

Or add a sentence about how controller  metrics measure x, and node metrics measure y.

torredil: Fixed ✅

AndrewSirenko (approved):

ElijahQuinones (approved):

AndrewSirenko (approved):

ElijahQuinones (approved):
/LGTM all metrics in seconds

PR #2250, opened 2024-12-01
enabling node metrics causes helm apply failure on malformed yaml

mindw:
#### What type of PR is this?

/kind bug
Enabling node metrics causes helm apply failure on malformed yaml. 
The flag for enabling the it is missing from `values.yaml`.

By installing and upgrading the chart - 
```
helm -n kube-system diff upgrade aws-ebs-csi-driver helm-charts/aws-ebs-csi-driver --reset-then-reuse-values --set node.enableMetrics=true
```

```release-note
Helm chart failed to install/upgrade if `node.enableMetrics=true` was set.
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/metrics.yaml
+++ b/charts/aws-ebs-csi-driver/templates/metrics.yaml
@@ -41,7 +41,7 @@ spec:
 {{- end }}
 {{- end }}
 ---
-{{- if .Values.node.enableMetrics -}}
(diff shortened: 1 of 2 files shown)


AndrewSirenko (approved):

mindw:
Hi, anything missing on my end to have this PR merged?

ConnorJC3 (approved):
/lgtm

PR #2253, opened 2024-12-04
Selinux Mount Option

ConnorJC3:
#### What type of PR is this?

/kind feature

Adds a new helm parameter, `node.selinux`, that mounts the relevant directories to allow the selinux-aware `mount` binary to perform selinux-specific optimizations. (This parameter can also be used in the future for any other selinux optimizations on the node).

This must be a parameter that is default-disabled, because `/sys/fs/selinux` will not exist unless SELinux is actually enabled on the nodes.

CI/Manually

```release-note
Add Helm parameter `node.selinux` to enable SELinux-specific mounts on the node DaemonSet
```

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -342,6 +342,9 @@ controller:
   # Enable dnsConfig for the controller and node pods
   dnsConfig: {}
 node:
+  # Enable SELinux-only optimizations on the EBS CSI Driver node pods
+  # Must only be set true if all linux nodes in the DaemonSet have SELinux enabled
+  selinux: false
   env: []
   envFrom: []
   kubeletPath: /var/lib/kubelet
--- a/charts/aws-ebs-csi-driver/templates/_node.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_node.tpl
@@ -127,6 +127,12 @@ spec:
               mountPath: /csi
             - name: device-dir
               mountPath: /dev
+            {{- if .Values.node.selinux }}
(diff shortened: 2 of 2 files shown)


ConnorJC3:
flake

/retest

ConnorJC3:
/retest

lots of flakes today

ElijahQuinones (approved):

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -342,6 +342,9 @@ controller:
   # Enable dnsConfig for the controller and node pods
   dnsConfig: {}
 node:
+  # Enable SELinux-only optimizations on the EBS CSI Driver node pods
+  # Must only be set true if all linux nodes in the DaemonSet have SELinux enabled
```
AndrewSirenko, line 346: Non-blocking comment: Do we want to recommend  [additional daemonsets feature](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/d99da6d023b7523a7998c0b1e21a8f741e21349a/docs/additional-daemonsets.md?plain=1) in-line here. 

An extra sentence like: "Please rely on additional daemonsets feature if this is not the case."

AndrewSirenko (approved):

charts/aws-ebs-csi-driver/values.yaml
ConnorJC3: That's an advanced feature the user should think about before enabling so I'd honestly rather not give an affirmative suggestion for this relatively minor feature.

torredil (approved):
/lgtm
/approve

PR #2270, opened 2024-12-19
Fixup FIPS Makefile/E2E

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Fixes two issues currently flaking our CI left and right:

- Use a different tag when pushing the FIPS image in `sub-push-fips`
- Switch `make cluster/image` from `all-push` to `sub-push` so it doesn't waste time building the a1compat and FIPS images

Manual/CI

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,7 @@ sub-push: all-image-registry push-manifest
 
 .PHONY: sub-push-fips
 sub-push-fips:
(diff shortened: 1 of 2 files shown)


AndrewSirenko (approved):
Great call on not wasting time with a1 images too. Thanks.

PR #2279, opened 2025-01-03
Change make test to also run sanity tests

ElijahQuinones:
#### What type of PR is this?
/kind feature

This PR will change make test to also run sanity tests so the CI job will run both unit and sanity tests on all PRs.
Ran make test and verified both unit tests and sanity tests run and pass.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ clean:
 
 .PHONY: test
 test:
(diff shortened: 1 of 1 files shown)


torredil (approved):
/lgtm

PR #2280, opened 2025-01-06
Fix cloudbuild.sh target

ConnorJC3:
#### What type of PR is this?

/kind cleanup

`hack/cloudbuild.sh` references invalid makefile target `all-push-for-release`, it should be `all-push` - this causes all of our commits to `master` to fail to build (and would break building images for release as well).

Manually

```release-note
NONE
```


ElijahQuinones (approved):
/lgtm

PR #2286, opened 2025-01-09
Add values.schema.json

ElijahQuinones:
#### What type of PR is this?

/kind feature
This PR adds the values.schema.json file. Allowing us to make sure changes to helm chart parameters also appear in the eks add-on version of the driver as long as it makes sense to do so. 
```
❯ helm lint .
==> Linting .
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
```

```release-note
Add values.schema.json to validate changes in values.yaml make it to the add-on schema.
```

### DIFF
--- a/charts/aws-ebs-csi-driver/values.schema.json
+++ b/charts/aws-ebs-csi-driver/values.schema.json
@@ -1,0 +1,1342 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Configurable parameters of the AWS EBS CSI Driver",
+  "properties": {
+    "a1CompatibilityDaemonSet": {
+      "type": "boolean",
+      "description": "Enable compatibility for the A1 instance family via use of an AL2-based image in a separate DaemonSet",
+      "default": false
+    },
+    "additionalDaemonSets": {
+      "type": ["object", "null"],
+      "description": "Additional DaemonSets of the node pod",
+      "default": null,
+      "patternProperties": {
+        "^.*$": {
+          "type": "object",
+          "$ref": "#/properties/node"
+        }
+      }
+    },
+    "defaultStorageClass": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "description": "Creates a default StorageClass.",
+          "default": false
+        }
+      }
+    },
+    "helmTester": {
+      "type": "object",
+      "description": "Supply a custom image to the ebs-csi-driver-test pod in helm-tester.yaml",
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "default": true
+        },
+        "image": {
+          "type": "string"
+        }
+      }
+    },
+    "imagePullSecrets": {
+      "type": "array",
+      "default": []
+    },
+    "volumeSnapshotClasses": {
+      "type": "array",
+      "description": "Add VolumeSnapshotClass resources",
+      "default": []
+    },
+    "awsAccessSecret": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string",
+          "default": "aws-secret"
+        },
+        "keyId": {
+          "type": "string",
+          "default": "key_id"
+        },
+        "accessKey": {
+          "type": "string",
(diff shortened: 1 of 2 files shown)


charts/aws-ebs-csi-driver/values.schema.json
```diff
@@ -0,0 +1,1499 @@
...
+          }
+        },
+        "priorityClassName": {
+          "description": "Priority class for the Node Daemonset",
+          "type": [
+            "string",
+            "null"
+          ],
+          "default": ""
+        },
```
torredil, line 743: I was wondering why this diff exists (for the add-on, this section is defined as follows):

```
"priorityClassName": {
      "description": "Priority class for the Node Daemonset",
       "type": "string",
       "default": "system-node-critical"
},
```

Which is correct, because that default value is defined in the node template: 

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b28e70ded7863ce6382671b601c3722848ab92f2/charts/aws-ebs-csi-driver/templates/_node.tpl#L48

but is not currently set in `values.yaml`:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b28e70ded7863ce6382671b601c3722848ab92f2/charts/aws-ebs-csi-driver/values.yaml#L360

I think this would be a good opportunity to fix that discrepancy.

torredil (commented):
One suggested improvement, otherwise PR largely lgtm.

charts/aws-ebs-csi-driver/values.schema.json
ElijahQuinones: will do !

AndrewSirenko (commented):
/lgtm

Largely looks good to me, thank you! 

The real test will be going through release process and merging with add-on.

charts/aws-ebs-csi-driver/values.schema.json
```diff
@@ -0,0 +1,1499 @@
...
+      "default": false
+    },
+    "storageClasses": {
+      "type": "array",
+      "description": "Add StorageClass resources",
+      "default": []
+    },
+    "fips": {
+      "type": "boolean",
+      "description": "Instruct the AWS SDK to use AWS FIPS endpoints, and deploy container built with BoringCrypto (a FIPS-validated cryptographic library) instead of the Go default. The EBS CSI Driver FIPS images have not undergone FIPS certification, and no official guarnatee is made about the compliance of these images under the FIPS standard. Users relying on these images for FIPS compliance should perform their own independent evaluation",
```
AndrewSirenko, line 114: There is a typo here for guarantee. Did we run a spellcheck on this file? Your IDE should do this for you (or have a plugin).

AndrewSirenko (changes requested):
Can we run this file through a spellcheck? (Your IDE should do this for you)

charts/aws-ebs-csi-driver/values.schema.json
ElijahQuinones: Typos here are also present in values.yaml as these descriptions where direct ported from there. Will run spell check on IDE and fix in both places on next rev thank you !

charts/aws-ebs-csi-driver/values.schema.json
```diff
@@ -0,0 +1,1499 @@
...
+            "image": {
+              "type": "object",
+              "properties": {
+                "pullPolicy": {
+                  "type": "string",
+                  "default": "IfNotPresent"
+                },
+                "tag": {
+                  "type": "string",
+                  "default": "v0.5.0"
```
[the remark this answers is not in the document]
ElijahQuinones: We could however the default here is not actually required maybe to reduce the maintenance burden of this we could omit the default here unless you feel strongly about having the default here.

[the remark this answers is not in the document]
AndrewSirenko: I think that's a great idea, thank you.

charts/aws-ebs-csi-driver/values.schema.json
AndrewSirenko: Thanks for fixing!

torredil (approved):
/lgtm

AndrewSirenko (approved):

PR #2288, opened 2025-01-14
[DO NOT MERGE] Test new upstream sidecars

ConnorJC3:
Something weird going on with the external storage tests, this should test in isolation if the sidecars are the issue
/hold


AndrewSirenko (commented):
context: https://kubernetes.slack.com/archives/C09QZFCE5/p1736798710626099

PR #2292, opened 2025-01-16
Create EBS CSI Driver scale-test tool

AndrewSirenko:
#### What type of PR is this?

/kind documentation

Validate that the EBS CSI Driver is capable of managing volumes at the scale of your largest EKS clusters with our `scale-test` tool. See `/hack/ebs-scale-test/README.md` for more info.

Karpenter cluster-type, more scalability test-types, and more scalability test observability features coming soon.

Scope of this PR: One-step run pre-allocated scale-sts scalability tests.

Running `./scale-test create && ./scale-test run && ./scale-test clean` with no extra environment variables set will:

1. Create an EKS cluster
2. Install EBS CSI Driver with metrics enabled
3. Deploy StorageClass and StatefulSet, then scale that sts to 1000 replicas (which will provision and publish 1000 volumes)
4. Delete sts (unpublishing and deleting volumes)
5. Export metrics, ebs-plugin logs, driver deployment + daemonset yaml, and other test artifacts to a local directory and an S3 bucket
6. Clean up all scalability test resources from your AWS account

Revision 2 edit: Ran through scale-test create, run, and clean loop again. Also manually tested extra features, the results of which I included in the review comments of those features. Thanks. 

Please follow the README.md to run these scale tests yourself. 

Timing:

Cluster Setup: ~18 min
Run: ~8 min for 1000 replica test
Cluster Cleanup: ~10 min

```
❯ ./scale-test setup && ./scale-test run && ./scale-test clean
Deploying EKS cluster. See configuration in /tmp/tmp.FEdux6Eoal/cluster-config.yaml
2025-01-16 14:17:05 [ℹ]  eksctl version 0.194.0-dev+02ef28ee3.2024-10-22T18:42:34Z
...
Applying /workplace/andsirey/aws-ebs-csi-driver/hack/ebs-scale-test/helpers/scale-test/scale-sts-test/scale-sts.yaml. Exported to /tmp/tmp.FGNUXsiFTA/scale-manifest.yaml
...
partitioned roll out complete: 101 new pods have been updated...
Deleting StatefulSet
statefulset.apps "ebs-scale-test" deleted
storageclass.storage.k8s.io "ebs-scale-test" deleted
Waiting for all PVs to be deleted
101 PVs still exist, waiting...
...
Exporting everything in /tmp/tmp.FGNUXsiFTA to S3
Metrics exported to s3://ebs-scale-tests/ebs-scale-pre-allocated-scale-sts-101-2025-01-16T14:35UTC/
2025-01-16 14:44:57 [ℹ]  deleting EKS cluster "ebs-scale-pre-allocated"
2025-01-16 14:52:04 [✔]  all cluster resources were deleted
```

```release-note
Add `scale-test` tool for running EBS CSI Driver scalability tests
```


hack/ebs-scale-test/README.md
```diff
@@ -0,0 +1,76 @@
+# EBS CSI Driver Scalability Tests
+
+EBS uses EBS CSI Driver scalability tests to validate that each release of our driver can manage EBS volume lifecycle for large-scale clusters. 
+
+Setup and run an EBS CSI Driver scalability test with our `scale-test` tool:
```
AndrewSirenko, line 5: Shall I add wording about results being exported to local dir + S3 bucket here? Or is that implicit.

hack/ebs-scale-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+# Run a scalability test and export results to S3.
+./scale-test run
+
+# Cleanup all AWS resources related to scalability cluster. 
+./scale-test cleanup
+```
+
+## Pre-requisites
+
+REVIEWER NOTE: I'm open to relying on `make tools` /bin dependencies. But that might be confusing to those just wanting to run scale tests.
```
AndrewSirenko, line 25: Discuss!

hack/ebs-scale-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+# Find default values at top of `scale-test` script. 
+```
+
+## Types of scalability tests
+
+Set the `CLUSTER_TYPE` and `TEST_TYPE` environment variables to set up and run different scalability tests. 
+
+- `CLUSTER_TYPE` dictates what type of scalability cluster `scale-test` creates and which nodes are used during a scalability test run. Options include: 
+  - 'pre-allocated': Additional worker nodes are created during cluster setup. By default, we pre-allocate 1 `m7a.48xlarge` EC2 instance for every 100 StatefulSet replicas. 
+
```
AndrewSirenko, line 61: Extra newline will fix in later revision

hack/ebs-scale-test/helpers/cluster-setup/scale-cluster-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+iam:
+  withOIDC: true
+  podIdentityAssociations:
+    - namespace: kube-system
+      serviceAccountName: ebs-csi-controller-sa
+      wellKnownPolicies:
+        ebsCSIController: true
+
+managedNodeGroups:
+{{- if eq ( getenv "CLUSTER_TYPE" ) "pre-allocated" }}
```
AndrewSirenko, line 33: We have to use templating here because for Karpenter cluster-type, we'll need a different nodegroup.

hack/ebs-scale-test/helpers/scale-test/scale-sts-test/scale-sts.yaml
```diff
@@ -0,0 +1,68 @@
...
+        accessModes: [ "ReadWriteOnce" ]
+        storageClassName: "ebs-scale-test"
+        resources:
+          requests:
+            storage: 1Gi
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: Delete
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
```
AndrewSirenko, line 63: Thoughts on adding "ebs-scale-test: $CLUSTER_NAME" tags to each volume? And then when we cleanup resources we can check for any leaked volumes (which doesn't happen in my testing but better safe than sorry).

hack/ebs-scale-test/scale-test
```diff
@@ -0,0 +1,108 @@
...
+export CLUSTER_NAME
+export EXPORT_DIR
+export S3_BUCKET
+export SCALABILITY_TEST_RUN_NAME
+
+CLUSTER_TYPE=${CLUSTER_TYPE:="pre-allocated"}
+TEST_TYPE=${TEST_TYPE:="scale-sts"}
+REPLICAS=${REPLICAS:=1000}
+DRIVER_VALUES_FILEPATH=${DRIVER_VALUES_FILEPATH:="$BASE_DIR/helpers/cluster-setup/scale-driver-values.yaml"}
+# TODO Q: Does anyone need an $OVERRIDE_KUBECONFIG? Let's discuss.
```
AndrewSirenko, line 35: Discuss! I'd prefer a wrapper script to run simultaneous scalability tests in different regions.

hack/ebs-scale-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+```sh
+# Affect test
+CLUSTER_TYPE              # Type of scalability cluster to create.
+TEST_TYPE                 # Type of scale test to run.
+REPLICAS                  # Number of StatefulSet replicas to create.
+DRIVER_VALUES_FILEPATH    # Custom values file passed to EBS CSI Driver Helm chart.
+
+# Names
+CLUSTER_NAME              # Base name used by `eksctl` to create AWS resources.
+EXPORT_DIR                # Where to export scale test metrics/logs locally.
```
AndrewSirenko, line 48: Should add .gitignore for default export_dir path

hack/ebs-scale-test/helpers/cluster-setup/scale-cluster-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - instanceType: m7a.48xlarge
+    amiFamily: AmazonLinux2
+    name: pre-allocated-ng
+    desiredCapacity: {{ .Env.PRE_ALLOCATED_NODES }}
+{{- end }}
+
+addons:
+  - name: eks-pod-identity-agent
+  - name: snapshot-controller
```
AndrewSirenko, line 42: I'm considering pinning these dependencies to a specific version... 

And add as a step to our dependancy upgrade runbook.

hack/ebs-scale-test/helpers/cluster-setup/scale-driver-values.yaml
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Default values.yaml for ebs-scale-test installation of aws-ebs-csi-driver
+image:
+  pullPolicy: Always
+controller:
+  logLevel: 7
+  replicaCount: 1
```
torredil, line 20: This is a very pragmatic solution for solving the issue of different controller pods having different leaders for the sidecars - can we add a comment explaining that? its not immediately obvious.

hack/ebs-scale-test/scale-test
```diff
@@ -0,0 +1,108 @@
...
+## Overridable environment variables. See README.md
+export CLUSTER_TYPE
+export TEST_TYPE
+export REPLICAS
+export DRIVER_VALUES_FILEPATH
+
+export CLUSTER_NAME
+export EXPORT_DIR
+export S3_BUCKET
+export SCALABILITY_TEST_RUN_NAME
```
torredil, line 29: Suggest auto-generating this variable and making it an internal implementation detail, unless there is a real need to override it.

AndrewSirenko: By default we auto generate this run name (using a mix of cluster name, test type, replica count, and timestamp)

I'd prefer to still expose this as overridable because I have found it helpful to have my own custom run naming scheme when repeatedly running scale tests. E.g. Tuning the optimal value for a parameter like max_batch_weight. But I'm open to reevaluating that if we think the extra param is distracting in the docs.

hack/ebs-scale-test/README.md
rdpsin: I would prefer us to be explicit. 

Also, should we add a note about the permissions one needs to successfully run this test?

hack/ebs-scale-test/scale-test
```diff
@@ -0,0 +1,108 @@
...
+CLUSTER_TYPE=${CLUSTER_TYPE:="pre-allocated"}
+TEST_TYPE=${TEST_TYPE:="scale-sts"}
+REPLICAS=${REPLICAS:=1000}
+DRIVER_VALUES_FILEPATH=${DRIVER_VALUES_FILEPATH:="$BASE_DIR/helpers/cluster-setup/scale-driver-values.yaml"}
+# TODO Q: Does anyone need an $OVERRIDE_KUBECONFIG? Let's discuss.
+
+CLUSTER_NAME=${CLUSTER_NAME:="ebs-scale-$CLUSTER_TYPE"}
+EXPORT_DIR=${EXPORT_DIR:=$(mktemp -d)}
+S3_BUCKET=${S3_BUCKET:="ebs-scale-tests"}
+SCALABILITY_TEST_RUN_NAME=${SCALABILITY_TEST_RUN_NAME:="$CLUSTER_NAME-$TEST_TYPE-$REPLICAS-$(date -u +%Y-%m-%dT%H:%M%Z)"}
```
rdpsin, line 40: Is it worth adding a timestamp even when provided with the test name? As it stands, wouldn't running the test multiple times with the same name overwrite the bucket contents of the previous runs?

hack/ebs-scale-test/scale-test
```diff
@@ -0,0 +1,108 @@
...
+
+CLUSTER_TYPE=${CLUSTER_TYPE:="pre-allocated"}
+TEST_TYPE=${TEST_TYPE:="scale-sts"}
+REPLICAS=${REPLICAS:=1000}
+DRIVER_VALUES_FILEPATH=${DRIVER_VALUES_FILEPATH:="$BASE_DIR/helpers/cluster-setup/scale-driver-values.yaml"}
+# TODO Q: Does anyone need an $OVERRIDE_KUBECONFIG? Let's discuss.
+
+CLUSTER_NAME=${CLUSTER_NAME:="ebs-scale-$CLUSTER_TYPE"}
+EXPORT_DIR=${EXPORT_DIR:=$(mktemp -d)}
+S3_BUCKET=${S3_BUCKET:="ebs-scale-tests"}
```
rdpsin, line 39: Consider removing the default here and throwing error if not provided? Considering S3 buckets name as globally unique.

hack/ebs-scale-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Types of scalability tests
+
+Set the `CLUSTER_TYPE` and `TEST_TYPE` environment variables to set up and run different scalability tests. 
+
+- `CLUSTER_TYPE` dictates what type of scalability cluster `scale-test` creates and which nodes are used during a scalability test run. Options include: 
+  - 'pre-allocated': Additional worker nodes are created during cluster setup. By default, we pre-allocate 1 `m7a.48xlarge` EC2 instance for every 100 StatefulSet replicas. 
+
+
+- `TEST_TYPE` dictates what type of scalability test we want to run. Options include: 
+  - 'scale-sts': Scales a StatefulSet to `$REPLICAS`. Waits for all pods to be ready. Delete Sts. Waits for all PVs to be deleted. Exercises
```
rdpsin, line 64: 'Exercises '?

hack/ebs-scale-test/scale-test
AndrewSirenko: Hmm, I was thinking that those who override test name would want complete control over (and could add their own preferred timestamp format or timezone themselves).

But you bring up the good point of accidentally overwriting repeat runs. I will make the change.

hack/ebs-scale-test/scale-test
AndrewSirenko: I prefer setting a reasonable default instead of putting the burden of setting S3_BUCKET for each run on every user. Though I'm open to making S3_BUCKET a required parameter if other reviewers agree. 

Would choosing a rarer default bucket name satisfy your concern? Or clearer documentation? Something like `ebs-scale-tests-default-bucket-name` is even more unlikely to already be a bucket in user's account.

hack/ebs-scale-test/README.md
AndrewSirenko: Great catch. Was thinking about adding the sentence: "Exercises the complete dynamic provisioning lifecycle for block volumes."

hack/ebs-scale-test/scale-test
AndrewSirenko: Thinking a bit more on this, would an even better change be erroring if the run_name already exists within the S3 bucket? Or would you still prefer force-appending our default timestamp format on each run. @rdpsin

rdpsin: Both sound like decent options to me. Whatever the team decides. :)

hack/ebs-scale-test/README.md
ElijahQuinones: I think it might be best to rely on these so that we can ensure the smallest level of variance based on personal setups. It makes it easier to help someone if they are having trouble running the tests as we know what dependencies they are using.

hack/ebs-scale-test/README.md
ElijahQuinones: IMO yes it will make making changes to scale tests after running them easier.

hack/ebs-scale-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+You can mix and match `CLUSTER_TYPE` and `TEST_TYPE`.
+
+## Contributing scalability tests
+
+`scale-test` parses arguments and wraps scripts and configuration files in the `helpers` directory. These helper scripts manage the scalability cluster and test runs. 
+
+We rely on [gomplate](https://github.com/hairyhenderson/gomplate) to render configuration files based on environment variables. 
+
+The `helpers` directory includes:
+- `/helpers/cluster-setup`: Holds scripts and configuration for cluster and add-on setup/cleanup.
```
ElijahQuinones, line 75: np: add-on here suggests that it is the eks addon version of the ebs-csi-driver that we are installing but looking at manage-cluster.sh below we are installing via helm.

hack/ebs-scale-test/helpers/scale-test/scale-sts-test/scale-sts.yaml
ElijahQuinones: If there was for some reason an interruption in the test like someone forgot to use nohup and disconnected from the network :) having the tags would also allow for easy manual deletion of leaked resources.

hack/ebs-scale-test/scale-test
ElijahQuinones: Don't need it now but if it is not too much effort on your end may be worth doing in case in the future we do need to do so for any reason.

hack/ebs-scale-test/scale-test
ElijahQuinones: what about `ebs-csi-driver-scale-test-bucket` in case there are other "ebs-scale-tests" buckets out there

hack/ebs-scale-test/scale-test
ElijahQuinones: Could we print a warning message of something like `warning run_name already exists if you continue you will overwrite the contents of run_name are you sure you want to continue` followed by a y/n prompt.

hack/ebs-scale-test/scale-test
```diff
@@ -0,0 +1,108 @@
...
+  create_cluster
+  deploy_ebs_csi_driver
+}
+
+# Functions sourced from helpers/scale-test/...
+run-scale() {
+  # TODO Q: Is it worth restarting EBS CSI Controller pod to ensure clean metrics/logs? Or move EBS CSI Driver install/cleanup to 'run' instead of 'setup'? Let's discuss.
+  aws eks update-kubeconfig --name "$CLUSTER_NAME"
+  sts_scale_test
+  export_to_s3
```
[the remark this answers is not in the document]
ElijahQuinones: Yeah im okay with that coming during that time instead of in this PR

hack/ebs-scale-test/scale-test
AndrewSirenko: @rdpsin I just realized your point. Yes lets remove the default and throw an error if not provided. 

Alternatively we can append something like account-id, but I think you are right it's cleanest just to error out. 

@ElijahQuinones there can only be one bucket named `ebs-csi-driver-scale-test-bucket` across all AWS accounts. Unless we generate a unique name by default on behalf of user, they will get err `BucketAlreadyExists` when we attempt to create bucket.

hack/ebs-scale-test/README.md
AndrewSirenko: Add-on refers to eks-pod-identity and snapshot-controller addons. 

I can remove `and add-on` from the sentence though if that context is not needed.

hack/ebs-scale-test/scale-test
AndrewSirenko: I prefer avoiding `y/n` prompts in scripts that may be used in automated processes (like a wrapper script that manages scalability clusters/runs simultaneously) because then the caller has to rely on `yes` or `expect` to work around that forced interaction.

Because I foresee us wrapping `ebs-scale-test`, I prefer that we make a consistent choice on behalf of caller (fast-failing would be my preference).

hack/ebs-scale-test/scale-test
torredil: I think a pragmatic solution here would be following what we do for the kOps bucket used in CI: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b28e70ded7863ce6382671b601c3722848ab92f2/hack/e2e/config.sh#L45

Prefixing the bucket name with an account ID practically guarantees that it is unique.

rdpsin: Sounds good, either of the options.

hack/ebs-scale-test/README.md
AndrewSirenko: On second thought, our makefile targets and `ebs-scale-test` are fully separate. We may decide to couple them in the future, but for now it's least complicated to fully de-couple them.

hack/ebs-scale-test/scale-test
AndrewSirenko: Closing because this can be added at the time the use-case is needed.

hack/ebs-scale-test/README.md
AndrewSirenko: By default EXPORT_DIR  is a `mktemp` directory in /tmp/, therefore no .gitignore addition is needed.

hack/ebs-scale-test/helpers/cluster-setup/scale-cluster-config.yaml
AndrewSirenko: Will remove snapshot-controller as dependency until we have snapshot scale tests. Probably best we stick with default eks-pod-identity-agent

hack/ebs-scale-test/scale-test
```diff
@@ -0,0 +1,108 @@
...
+
+# Functions sourced from helpers/cluster-setup/manage-cluster.sh
+setup-scale() {
+  create_cluster
+  deploy_ebs_csi_driver
+}
+
+# Functions sourced from helpers/scale-test/...
+run-scale() {
+  # TODO Q: Is it worth restarting EBS CSI Controller pod to ensure clean metrics/logs? Or move EBS CSI Driver install/cleanup to 'run' instead of 'setup'? Let's discuss.
```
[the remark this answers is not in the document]
AndrewSirenko: Decided to just restart ebs csi controller ahead of run. Will document. 

We don't want to cleanup driver within run because folks may want to investigate logs/metrics afterwards. 

Folks may want to uninstall and reinstall different ebs csi driver self-managed/add-on themselves after cluster creation or between scalability tests.

hack/ebs-scale-test/scale-test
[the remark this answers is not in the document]
AndrewSirenko: Added, proof: 

```
❯ cat /tmp/ebs-scale-test/ebs-scale-pre-allocated-scale-sts-100-2025-01-28T20:20UTC/cleaned_data.txt
aws_ebs_csi_api_request_duration_seconds_bucket{request="AttachVolume",le="+Inf"} 100
aws_ebs_csi_api_request_duration_seconds_bucket{request="CreateVolume",le="+Inf"} 100
aws_ebs_csi_api_request_duration_seconds_bucket{request="DeleteVolume",le="+Inf"} 160
aws_ebs_csi_api_request_duration_seconds_bucket{request="DescribeInstances",le="+Inf"} 57
aws_ebs_csi_api_request_duration_seconds_bucket{request="DescribeVolumes",le="+Inf"} 114
aws_ebs_csi_api_request_duration_seconds_bucket{request="DetachVolume",le="+Inf"} 124
aws_ebs_csi_api_request_errors_total{request="DetachVolume"} 57
cardinality_enforcement_unexpected_categorizations_total 0
controller_persistentvolume_delete_duration_seconds_bucket{class="ebs-scale-test",le="+Inf"} 100
```

hack/ebs-scale-test/scale-test
AndrewSirenko: Solved by fast failing

```
❯ SCALABILITY_TEST_RUN_NAME="ebs-scale-pre-allocated-scale-sts-100-2025-01-28T20:20UTC" ./scale-test run
                           PRE ebs-scale-pre-allocated-scale-sts-100-2025-01-28T20:20UTC/
                           PRE ebs-scale-pre-allocated-scale-sts-1000-2025-01-23T20:30UTC/
ERROR: Your S3 bucket already contains directory with name SCALABILITY_TEST_RUN_NAME: 's3://ebs-scale-tests-208536326202/ebs-scale-pre-allocated-scale-sts-100-2025-01-28T20:20UTC'. Please pick a unique SCALABILITY_TEST_RUN_NAME.
```

hack/ebs-scale-test/scale-test
AndrewSirenko: Addressed in next revision.

hack/ebs-scale-test/README.md
AndrewSirenko: Addressed later in README in latest revision.

hack/ebs-scale-test/helpers/cluster-setup/scale-cluster-config.yaml
ElijahQuinones: I see you where going to remove the snapshot-controller but it still remains did you change your mind here ?

hack/ebs-scale-test/README.md
```diff
@@ -13,19 +13,23 @@ export REPLICAS="1000"
...
+# Run a scalability test and export results.
 ./scale-test run
 
 # Cleanup all AWS resources related to scalability cluster. 
 ./scale-test cleanup
 ```
 
+Results will be exported to a local directory (`$EXPORT_DIR`) and an S3 Bucket in your AWS account (`$S3_BUCKET`).
+
+Note: The `ebs-csi-controller` pods will be restarted at the beginning of every scale run to clear metrics/logs.
```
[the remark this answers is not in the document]
AndrewSirenko: Customer can override replica count. The command I use will restart all `ebs-csi-controller` pods.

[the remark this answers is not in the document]
AndrewSirenko: Changed language to "Any `ebs-csi-controller` pod(s)..."

AndrewSirenko:
/retest

Flake: `Error: validation failed: wait time exceeded during validation`

ElijahQuinones:
/approve 

Ran through scale tests locally and confirmed they run as intended send metrics and logs to S3 and clean up all resources when done. 

```
./scale-test create && ./scale-test run && ./scale-test clean;
...
2025-01-31 16:11:46 [✔]  saved kubeconfig as "/home/elijahlq/.kube/config"
...
2025-01-31 16:11:46 [✔]  all EKS cluster resources for "ebs-scale-pre-allocated" have been created
...
2025-01-31 16:33:15 [✔]  all cluster resources were deleted
```

PR #2303, opened 2025-01-22
Fix wait for volume ready ~60s

### LINKED ISSUE #2302: Wait for volume ready backoff stops after ~7 seconds
/kind bug

**What happened?**
The volume readiness wait time is approximately 7 seconds, contrary to the expected ~60 seconds mentioned in the comments.

The volume readiness wait time should extend up to ~60 seconds.

Our cloud environment is AWS-compatible (from an API perspective), and we utilize the CSI driver. However, in our setup, volume creation can take longer than 7 seconds.

**Environment**
	•	Kubernetes version (use kubectl version): 1.30.4
	•	Driver version: 1.33.0

khizunov:
Fix for #2302 

```release-note
Fix backoff when waiting for volume creation
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -281,8 +281,7 @@ var (
 		creationBackoff: wait.Backoff{
 			Duration: 500 * time.Millisecond,
 			Factor:   1.5,
-			Steps:    25,
-			Cap:      3 * time.Second,
+			Steps:    11,
 		},
 
 		// Most attach/detach operations on AWS finish within 1-4 seconds.


AndrewSirenko (approved):
Thank you for this fix @khizunov 

According to aws-sdk-go-v2 backoff duration reaching `Cap` seconds will forcefully set remaining backoff steps to 0, which will stop future backoffs:

```
// A limit on revised values of the duration parameter. If a
	// multiplication by the factor parameter would make the duration
	// exceed the cap then the duration is set to the cap and the
	// steps parameter is set to zero.
	Cap time.Duration
```

Deleting Cap will also align logic between volume creation/attachment/modification backoffs.

torredil (approved):
/lgtm

PR #2313, opened 2025-01-26
Fix: modify-volume doc typo

Phaow:
#### What type of PR is this?
/kind cleanup

- A small typo fix which is visible in the modify-volume doc(`-volume-ids` -> `--volume-ids`).
- Reduce customer cost the volume-modifier example pvc from `100Gi`  to `10Gi` to keep consistent with VAC example.

### DIFF
--- a/docs/modify-volume.md
+++ b/docs/modify-volume.md
@@ -66,7 +66,7 @@ spec:
   storageClassName: ebs-sc
   resources:
     requests:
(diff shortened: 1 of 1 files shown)


AndrewSirenko (approved):
Thank you for this!

ElijahQuinones (approved):

PR #2334, opened 2025-02-10
Fix raw pointer log in `EnableFastSnapshotRestores`

torredil:
#### What type of PR is this?

/kind bug

Currently, there is a logging issue in [cloud.EnableFastSnapshotRestores](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/4f4c7b7ec9617e5887a8cee4e1834c1e60fd9b59/pkg/cloud/cloud.go#L1421) which surfaces because the existing code is printing out raw pointers from the error structures rather than their underlying string values. With this cc, we explicitly extract the error code and error message from each failed FSR operation and consolidate them into a single, readable error message.

Exceed the regional FSR limit to induce the error, which can be reviewed by looking at the controller logs.

Without this patch:

```
E0210 16:42:42.944417       1 driver.go:108] "GRPC error" err="rpc error: code = Internal desc = Failed to create Fast Snapshot Restores for snapshot ID \"snapshot-1c0d40ba-7170-40e9-aac0-ccf9b8cd004d\": failed to create fast snapshot restores for snapshot snap-08e035467a65daf9f: [{[{0xc00080f650 0xc00092b668 {}}] 0xc00080f680 {}}]"
```

With this patch:

```
E0210 18:32:56.934997       1 driver.go:108] "GRPC error" err="rpc error: code = Internal desc = Failed to create Fast Snapshot Restores for snapshot ID \"snapshot-1c0d40ba-7170-40e9-aac0-ccf9b8cd004d\": Error Code: FastSnapshotRestoreLimitExceeded, Error Message: Your account's regional limit would be exceeded by your request to enable fast snapshot restore on snapshot 'snap-02946ca3646445835' in Availability Zone 'us-east-1f'"
```

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -1431,7 +1431,13 @@ func (c *cloud) EnableFastSnapshotRestores(ctx context.Context, availabilityZone
 		return nil, err
 	}
 	if len(response.Unsuccessful) > 0 {
-		return response, fmt.Errorf("failed to create fast snapshot restores for snapshot %s: %v", snapshotID, response.Unsuccessful)
+		var errDetails []string
+		for _, r := range response.Unsuccessful {
+			for _, e := range r.FastSnapshotRestoreStateErrors {
+				errDetails = append(errDetails, fmt.Sprintf("Error Code: %s, Error Message: %s", aws.ToString(e.Error.Code), aws.ToString(e.Error.Message)))
+			}
+		}
+		return nil, errors.New(strings.Join(errDetails, "; "))
 	}
 	return response, nil
 }


pkg/cloud/cloud.go
```diff
@@ -1431,7 +1431,13 @@ func (c *cloud) EnableFastSnapshotRestores(ctx context.Context, availabilityZone
...
 	}
 	if len(response.Unsuccessful) > 0 {
-		return response, fmt.Errorf("failed to create fast snapshot restores for snapshot %s: %v", snapshotID, response.Unsuccessful)
+		var errDetails []string
+		for _, r := range response.Unsuccessful {
+			for _, e := range r.FastSnapshotRestoreStateErrors {
+				errDetails = append(errDetails, fmt.Sprintf("Error Code: %s, Error Message: %s", aws.ToString(e.Error.Code), aws.ToString(e.Error.Message)))
+			}
+		}
+		return nil, errors.New(strings.Join(errDetails, "; "))
```
AndrewSirenko, line 1440: Question: Are we confident an `unsuccessful` response will always return at least 1 FastSnapshotRestoreStateErrors? 

In the case that there are no FastSnapshotRestoreStateErrors, `strings.Join(errDetails, "; ")` will return the empty string.

AndrewSirenko: On second thought this is fine. An err with empty message is still an err: https://go.dev/play/p/vqB3y2Z0Cq9

AndrewSirenko (approved):

PR #2335, opened 2025-02-11
Reconcile some differences between helm chart and values.schema.json

ElijahQuinones:
#### What type of PR is this?

/kind cleanup

Fix some of the differences between the helm chart and our values.schema.json file

Manually auditing both  and then running `make update && make verify && make test ` as well as `helm lint`

```release-note
Reconcile some differences between helm chart and values.schema.json
```

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -405,7 +405,7 @@ node:
     automountServiceAccountToken: true
   # Enable the linux daemonset creation
   enableLinux: true
-  enableWindows: false
+  enableWindows: true
   # Warning: This option will be removed in a future release. It is a temporary workaround for users unable to immediately migrate off of older kernel versions.
   # Formats XFS volumes with bigtime=0,inobtcount=0,reflink=0, for mounting onto nodes with linux kernel version <= 5.4.
   # Note that XFS volumes formatted with this option will only have timestamp records until 2038.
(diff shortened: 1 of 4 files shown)


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -405,7 +405,7 @@ node:
     automountServiceAccountToken: true
   # Enable the linux daemonset creation
   enableLinux: true
-  enableWindows: false
+  enableWindows: true
```
AndrewSirenko, line 408: Lets remember to mark this as an important item in the Helm changelog.

deploy/kubernetes/base/node-windows.yaml
```diff
@@ -0,0 +1,179 @@
...
+  labels:
+    app.kubernetes.io/name: aws-ebs-csi-driver
+spec:
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      app: ebs-csi-node
+      app.kubernetes.io/name: aws-ebs-csi-driver
+  updateStrategy:
+
```
AndrewSirenko, line 16: Extra newline?

torredil (approved):
/lgtm

deploy/kubernetes/base/node-windows.yaml
ElijahQuinones: When I remove it make update automatically adds it back.

ElijahQuinones: Fixed underlying issue in node-windows.yaml

PR #2342, opened 2025-02-14
Skip IMDS call if AWS_EC2_METADATA_DISABLED=true

AndrewSirenko:
#### What type of PR is this?

/kind bug

Today we attempt to call IMDS even if we know it is unavailable (via `AWS_EC2_METADATA_DISABLED=true` env var). 

/hold

Unit tests, CI

Will test on instance with and without EC2 metadata disabled shortly. 

```release-note
Skip IMDS call if AWS_EC2_METADATA_DISABLED=true
```

### DIFF
--- a/pkg/cloud/metadata/metadata.go
+++ b/pkg/cloud/metadata/metadata.go
@@ -42,17 +42,26 @@ type MetadataServiceConfig struct {
 
 var _ MetadataService = &Metadata{}
 
+// NewMetadataService retrieves instance Metadata from one of the client in MetadataServiceConfig.
+// It prefers EC2MetadataClient (IMDS) in order to get an accurate number of attached devices.
 func NewMetadataService(cfg MetadataServiceConfig, region string) (MetadataService, error) {
-	metadata, err := retrieveEC2Metadata(cfg.EC2MetadataClient, region)
-	if err == nil {
-		klog.InfoS("Retrieved metadata from IMDS")
-		return metadata.overrideRegion(region), nil
+	// Don't make an IMDS call if we know it's disabled
+	if os.Getenv("AWS_EC2_METADATA_DISABLED") == "true" {
+		klog.V(2).InfoS("Environment variable AWS_EC2_METADATA_DISABLED set to 'true'. Will not rely on IMDS for instance metadata")
+	} else {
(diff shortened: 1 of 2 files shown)


pkg/cloud/metadata/metadata.go
```diff
@@ -42,17 +42,27 @@ type MetadataServiceConfig struct {
...
+// NewMetadataService retrieves instance Metadata from one of the client in MetadataServiceConfig.
+// It prefers EC2MetadataClient (IMDS) in order to get an accurate number of attached devices.
 func NewMetadataService(cfg MetadataServiceConfig, region string) (MetadataService, error) {
-	metadata, err := retrieveEC2Metadata(cfg.EC2MetadataClient, region)
-	if err == nil {
-		klog.InfoS("Retrieved metadata from IMDS")
-		return metadata.overrideRegion(region), nil
+	// Don't make an IMDS call if we know it's disabled
+	imdsDisabled := os.Getenv("AWS_EC2_METADATA_DISABLED")
+	if imdsDisabled == "true" {
```
ElijahQuinones, line 50: nit: do we need a variable here or can we just do 
```
if os.Getenv("AWS_EC2_METADATA_DISABLED"){
}
```

pkg/cloud/metadata/metadata.go
```diff
@@ -42,17 +42,27 @@ type MetadataServiceConfig struct {
...
+		klog.ErrorS(err, "Retrieving IMDS metadata failed, falling back to Kubernetes metadata")
 	}
-	klog.ErrorS(err, "Retrieving IMDS metadata failed, falling back to Kubernetes metadata")
 
-	metadata, err = retrieveK8sMetadata(cfg.K8sAPIClient)
+	klog.V(4).InfoS("Attempting to retrieve instance metadata from Kubernetes API")
+	metadata, err := retrieveK8sMetadata(cfg.K8sAPIClient)
 	if err == nil {
-		klog.InfoS("Retrieved metadata from Kubernetes")
+		klog.V(4).InfoS("Retrieved metadata from Kubernetes")
```
ElijahQuinones, line 65: Just curious why are we changing to v 4 here

torredil (approved):
/lgtm

Issue #2301, opened 2025-01-22
Volume attachment limits for p4d.24xlarge are too low?


j-vizcaino:
/kind bug

**What happened?**

`csinode` for `p4d.24xlarge` reports 6 allocatable EBS volumes but can support more.

According to the [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html) these instance types should support up to 11 EBS volumes. 
As our `p4d.24xlarge` instances include 4 EFA/ENI devices, this brings down the number to 7. Taking into account the root EBS volume, this brings us to 6.

BUT, those instances, with EFA, support at least 8 EBS (+1 for root) volumes (see below)

**How to reproduce it (as minimally and precisely as possible)?**

* create a `pd4.24xlarge` instance, with EFA enabled
* describe the associated `csinode` resource (or look for the `ebs-csi-node` pod log line): the allocatable volume count is 6
* attach 6 EBS volumes, using pods
* create a 7th pod: pod stays in pending due to insufficient allocatable EBS capacity (expected behaviour)
* update the `ebs-csi-node` daemonset and force the number of allocatable EBS volumes to 11, adding `--volume-attachment-limits=11`
* watch the 7th pod start, with an additional EBS volume
* (bonus) adding a 8th pod works as well

**Anything else we need to know?**:

It's unclear if the issue is related to how the available EBS volumes computation is performed in the ebs-csi code, or if it's an AWS issue, with the EC2 metadata endpoint not reporting numbers correctly, but it's clear that those instances support more EBS attachments than what the driver reports.

**Environment**

- Kubernetes version (use `kubectl version`): `v1.29.12-eks-2d5f260`
- Driver version: `v1.38.1-eksbuild.2` (EKS addon)

j-vizcaino:
Output of `lspci` on a `p4d.24xlarge` node with both EFA and 9 EBS (including root volume) attachments (in case this helps)

```
# lspci -tv
-+-[0000:a0]-+-01.0  Amazon.com, Inc. Elastic Network Adapter (ENA)
 |           +-1b.0  Amazon.com, Inc. Elastic Fabric Adapter (EFA)
 |           +-1c.0  NVIDIA Corporation GA100 [A100 SXM4 40GB]
 |           +-1d.0  NVIDIA Corporation GA100 [A100 SXM4 40GB]
 |           +-1e.0  Amazon.com, Inc. NVMe SSD Controller
[...]

PR #2353, opened 2025-02-20
Fix p4d.24xlarge reporting wrong volume limit

ElijahQuinones:
#### What type of PR is this?

/kind bug 
fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2301
Eddie manually attached 28 volumes to a p4d.24xlarge and the public docs have been updated to show it can have 28 ebs volumes attached  see https://docs.aws.amazon.com/ec2/latest/instancetypes/ac.html#ac_storage-ebs

```release-note
Fix p4d.24xlarge reporting wrong volume limit
```

### DIFF
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -34,6 +34,7 @@ func init() {
 	// The limit is not shared with other device attachments: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#nitro-system-limits
 	instanceFamilies := []string{"m8g", "m7i", "m7i-flex", "m7a", "c8g", "c7i", "c7i-flex", "c7a", "r7a", "r7i", "r7iz", "r8g", "x8g", "u7i", "u7inh", "g6", "g6e", "gr6", "i7ie", "i8g", "p5", "p5e", "p5en"}
 	commonInstanceSizes := []string{"medium", "large", "xlarge", "2xlarge", "4xlarge", "8xlarge", "12xlarge"}
+	dedicatedVolumeLimits["p4d.24xlarge"] = 28
 
 	for _, family := range instanceFamilies {
 		for _, size := range commonInstanceSizes {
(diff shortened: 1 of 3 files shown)


pkg/cloud/volume_limits.go
```diff
@@ -532,7 +533,7 @@ var nvmeInstanceStoreVolumes = map[string]int{
 
 // https://aws.amazon.com/ec2/instance-types
 // Despite the dl1.24xlarge having Gaudi Accelerators describe instance types considers them GPUs as such that instance type is in this table
-// g5.48xlarge is not added to this table as it is in the maxVolumeLimits.
+// g5.48xlarge and p4d.24xlarge are not added to this table as it is in the maxVolumeLimits.
```
torredil, line 536: We'll want to also update [hack/generate-gpu-count-table.sh](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/dba6d7090f33e5ccd1452214691e0d1433d7881a/hack/generate-gpu-count-table.sh#L38) and [generate-instance-store-table.sh](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/dba6d7090f33e5ccd1452214691e0d1433d7881a/hack/generate-instance-store-table.sh#L40) otherwise running those scripts will re-add the values.

pkg/cloud/volume_limits.go
```diff
@@ -532,7 +532,7 @@ var nvmeInstanceStoreVolumes = map[string]int{
 
 // https://aws.amazon.com/ec2/instance-types
 // Despite the dl1.24xlarge having Gaudi Accelerators describe instance types considers them GPUs as such that instance type is in this table
-// g5.48xlarge is not added to this table as it is in the maxVolumeLimits.
+// g5.48xlarge and p4d.24xlarge are not added to this table as they are in the maxVolumeLimits.
```
AndrewSirenko, line 535: Probably out of scope of this PR, but did we ever check p4de.24xlarge instance type? I wonder if that one is also mis-marked... Feel free to create backlog spike but I'd be surprised if it behaves differently from p4d.24xlarge. 

Otherwise PR approach looks good to me.

AndrewSirenko (approved):

pkg/cloud/volume_limits.go
ElijahQuinones: Yeah I would love to investigate that as well but separately from this PR

ConnorJC3 (approved):
/lgtm

`p4de.24xlarge` probably suffers from the same defect, but until we're sure we shouldn't change it as that could cause stalled all mounts, and the docs _claim_ it's a shared limit

torredil (approved):
/lgtm

PR #2356, opened 2025-02-21
Fix incorrect properties validation in Helm schema

ConnorJC3:
#### What type of PR is this?

/kind bug

Fixes schema validation in several places:
- Adds missing `additionalProperties: false` in several places
- Adds missing properties validation to `otelTracing`
- Removes properties validation in several places where the data is directly passed through into Kubernetes (We should not validate Kubernetes objects and instead just directly pass them through to k8s, as what is valid changes between k8s versions and is already validated by Kubernetes itself on installation)

Manually

```release-note
NONE
```

### DIFF
--- a/charts/aws-ebs-csi-driver/values.schema.json
+++ b/charts/aws-ebs-csi-driver/values.schema.json
@@ -16,6 +16,7 @@
     },
     "additionalDaemonSets": {
       "type": ["object", "null"],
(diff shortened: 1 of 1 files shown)


torredil (approved):
/lgtm

AndrewSirenko (approved):

ElijahQuinones (approved):

PR #2357, opened 2025-02-21
Prevent null deref when enableWindows and otelTracing enabled on node

ConnorJC3:
#### What type of PR is this?

/kind bug

Order of Helm `end` is incorrect in Windows, causing a null pointer deref when windows and otel tracing are enabled at the same time:

```
Error: template: aws-ebs-csi-driver/templates/node-windows.yaml:4:4: executing "aws-ebs-csi-driver/templates/node-windows.yaml" at <include "node-windows" (deepCopy $ | mustMerge $defaultArgs)>: error calling include: template: aws-ebs-csi-driver/templates/_node-windows.tpl:114:26: executing "node-windows" at <.Values.fips>: nil pointer evaluating interface {}.fips
```

Manually/CI

```release-note
Prevent nil pointer deref in Helm chart when `node.enableWindows` and `node.otelTracing` are both set
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/_node-windows.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_node-windows.tpl
@@ -111,10 +111,10 @@ spec:
               value: {{ .otelServiceName }}
             - name: OTEL_EXPORTER_OTLP_ENDPOINT
               value: {{ .otelExporterEndpoint }}
+            {{- end }}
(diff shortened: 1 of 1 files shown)


AndrewSirenko (approved):

torredil (approved):
/approve

ElijahQuinones (approved):

PR #2363, opened 2025-02-25
Add Prometheus Annotations to the Node Service

mdzraf:
#### What type of PR is this?

/kind feature

Customer requested to be able to scrape the metrics from the drivers into CloudWatch Agent. 

This PR Adds Prometheus annotations to the Node Service, this way metrics can be automatically discovered and scraped by Prometheus, enabling monitoring and observability of the node-level metrics through the CloudWatch agent.

Deploying driver with changes and ensuring that annotations were set by running `kubectl describe service ebs-csi-node -n kube-system`

```
NONE
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/metrics.yaml
+++ b/charts/aws-ebs-csi-driver/templates/metrics.yaml
@@ -47,6 +47,9 @@ kind: Service
 metadata:
   name: ebs-csi-node
   namespace: {{ .Release.Namespace }}
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "3302"
   labels:
     app: ebs-csi-node
 spec:


charts/aws-ebs-csi-driver/templates/metrics.yaml
```diff
@@ -47,6 +47,9 @@ kind: Service
 metadata:
   name: ebs-csi-node
   namespace: {{ .Release.Namespace }}
+  annotations:
```
AndrewSirenko, line 50: Is there a reason we only are adding these standard prometheus annotations to node service and not controller service? 

Out of PR scope aside: I'm worried that the customer experience differs for controller and node services (i.e. for controller we create a ServiceMonitor on behalf of customer, but for node service customers should have their scrapers auto-discover via annotations. For short term should we add annotations to both services and longterm decoupling serviceMonitor creation from controller.enableMetrics?)

charts/aws-ebs-csi-driver/templates/metrics.yaml
```diff
@@ -47,6 +47,9 @@ kind: Service
...
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "3302"
```
AndrewSirenko, line 52: Out-of-PR-scope aside: I'm worried that the customer experience differs for controller and node services (i.e. for controller we create a ServiceMonitor on behalf of customer, but for node service customers should have their scrapers auto-discover via annotations. For short term should we add annotations to both services and longterm decoupling serviceMonitor creation from controller.enableMetrics?)

rdpsin, line 52: Should these be configurable?

charts/aws-ebs-csi-driver/templates/metrics.yaml
AndrewSirenko: Discussed offline, we will add that later when we re-think customer metrics enabling experience.

charts/aws-ebs-csi-driver/templates/metrics.yaml
mdzraf: Discussed offline with rest of team, we decided to leave this as is for now. We have added a spike to our backlog to consider what we want the exact customer metrics enabling experience to be in the future.

ConnorJC3 (approved):
/approve

PR #2390, opened 2025-03-10
Switching deprecatedMetrics Default Value to False

mdzraf:
#### What type of PR is this?
/kind cleanup

This PR switches the default value of deprecatedMetrics to false. Customers that still have the need to use the deprecated metrics can set it to true. 

N/A

```
Sets `deprecatedMetrics` to false by default
```


AndrewSirenko (commented):
How do folks opt-in enable deprecatedMetrics via helm after this PR? I don't see anything in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml?

Did we mean to add a helm chart parameter? If we want folks to use controller.additionalArgs can we add some documentation somewhere like the Changelog comment?

mdzraf:
> How do folks opt-in enable deprecatedMetrics via helm after this PR? I don't see anything in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml?
> 
> Did we mean to add a helm chart parameter? If we want folks to use controller.additionalArgs can we add some documentation somewhere like the Changelog comment?

Discussed offline, decided that because we do not want to have to drop it from the schema in a future release, we will go with option 2 of adding documentation in the Changelog on how to set it to true by using controller.additionalArgs

AndrewSirenko (approved):

ElijahQuinones (approved):
/lgtm

PR #2411, opened 2025-03-19
Add binary-based kops installation, retain commit as override

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Today, several CI runs are failing because of goproxy issues with kops. We originally decided to always install via commit for simplicity/consistency, but given this impact on CI, instead install via binary when possible with commit as override.

Manually/CI

```release-note
NONE
```


AndrewSirenko (approved):
/approve

PR #2414, opened 2025-03-19
Ignore ECR login failure in make cluster/image

ConnorJC3:
#### What type of PR is this?

/kind cleanup

In some cases, logging in to ECR doesn't work but isn't necessary, such as if the user has a login helper installed. Instead of failing out on this step, explicitly ignore failures as any issues will show up shortly after in the push anyways.

Manually

```release-note
NONE
```


AndrewSirenko (approved):
/lgtm
/approve

PR #2416, opened 2025-03-19
chore: remove commented assertion

AndrewSirenko:

We can't assert this due to batcher failures

```release-note
NONE
```


ConnorJC3 (approved):
/lgtm

PR #2424, opened 2025-03-28
Output junit report for `tests/e2e/...` E2E jobs

ConnorJC3:
#### What type of PR is this?

/kind feature

Adds a junit report for the "local" E2E jobs (those in `tests/e2e/...`). Automatically outputs this report to the `ARTIFACTS` environment variable if present (this is how Prow tells the job where to put artifacts). Also removes an unused and incorrect `-report-dir` CLI flag that appears to be a previous attempt at this.

Click single-az / multi-az jobs down below after they finish running, see that they now show which tests passed/failed/were skipped

```release-note
NONE
```


ConnorJC3:
/retest

flake

AndrewSirenko (approved):

PR #2456, opened 2025-04-30
Add snapshot-volume-scale to scalability test suite

torredil:

This PR adds a scale test for the CreateSnapshot / restore volume from snapshot code paths.

Manually:

```
export TEST_TYPE="snapshot-volume-scale"
export REPLICAS=1000
./scale-test run
```

https://gist.github.com/torredil/f1cd9806d013e56115f1ea93f018384d

```release-note
NONE
```

### DIFF
--- a/hack/ebs-scale-test/helpers/cluster-setup/manage-cluster.sh
+++ b/hack/ebs-scale-test/helpers/cluster-setup/manage-cluster.sh
@@ -41,11 +41,20 @@ cleanup_cluster() {
 
 check_lingering_volumes() {
   lingering_vol_count=$(aws ec2 describe-volumes \
-    --filters "Name=tag-key,Values=ebs-scale-test" \
+    --filters "Name=tag:ebs-scale-test,Values=${SCALABILITY_TEST_RUN_NAME}" \
     --query 'length(Volumes[*])' \
     --output text)
 
   [[ lingering_vol_count -ne 0 ]] && echo "WARNING: detected $lingering_vol_count lingering ebs-scale-test EBS volumes in $AWS_ACCOUNT_ID. Please run \`aws ec2 describe-volumes --filters 'Name=tag-key,Values=ebs-scale-test'\` and audit their AWS resource tags. Note these volumes may belong to a different scalability run than $SCALABILITY_TEST_RUN_NAME"
+}
+
+check_lingering_snapshots() {
+  lingering_snap_count=$(aws ec2 describe-snapshots \
+    --filters "Name=tag:ebs-scale-test,Values=${SCALABILITY_TEST_RUN_NAME}" \
+    --query 'length(Snapshots[*])' \
+    --output text)
+
+  [[ lingering_snap_count -ne 0 ]] && echo "WARNING: detected $lingering_snap_count lingering ebs-scale-test EBS snapshots from run ${SCALABILITY_TEST_RUN_NAME} in $AWS_ACCOUNT_ID. Please run \`aws ec2 describe-snapshots --filters 'Name=tag:ebs-scale-test,Values=${SCALABILITY_TEST_RUN_NAME}'\` and audit their AWS resource tags."
 }
 
 ## EBS CSI Driver
@@ -67,3 +76,12 @@ deploy_ebs_csi_driver() {
   echo "This script is not meant to be run directly, only sourced as a helper!"
   exit 1
 )
+
+deploy_snapshot_controller() {
+  echo "Deploying snapshot controller and CRDs"
+  kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/"${EBS_INSTALL_SNAPSHOT_VERSION}"/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+  kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/"${EBS_INSTALL_SNAPSHOT_VERSION}"/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
+  kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/"${EBS_INSTALL_SNAPSHOT_VERSION}"/client/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+  kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/"${EBS_INSTALL_SNAPSHOT_VERSION}"/client/config/crd/snapshot.storage.k8s.io_volumesnapshotcontents.yaml
+  kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/"${EBS_INSTALL_SNAPSHOT_VERSION}"/client/config/crd/snapshot.storage.k8s.io_volumesnapshots.yaml
+}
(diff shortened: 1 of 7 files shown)


hack/ebs-scale-test/helpers/cluster-setup/manage-cluster.sh
```diff
@@ -67,3 +76,12 @@ deploy_ebs_csi_driver() {
   echo "This script is not meant to be run directly, only sourced as a helper!"
   exit 1
 )
+
+deploy_external_snapshotter() {
```
AndrewSirenko, line 80: Is it worth explicitly tying snapshot controller/crds version to the version of `external-snapshotter` we use?

We can add this somewhere like `hack/e2e/config.sh` so we can track potential regressions to snapshotter releases?

hack/ebs-scale-test/helpers/scale-test/snapshot-volume-scale-test/snapshot-volume-scale.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: ebs-snapshot-test
+provisioner: ebs.csi.aws.com
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+parameters:
+  type: gp2
```
AndrewSirenko, line 23: Any reason why gp2? We want to run these scale tests in regions that don't support gp3?

hack/ebs-scale-test/scale-test
```diff
@@ -80,6 +81,7 @@ check_dependencies_helper() {
 # Functions sourced from helpers/cluster-setup/manage-cluster.sh
 setup_scale() {
   create_cluster
+  deploy_external_snapshotter
```
AndrewSirenko, line 84: Nit: I thought external-snapshotter refers to the sidecar within the EBS CSI Driver? And that here we are actually deploying snapshot_controller and snapshot_crds.

Can we rename this to something like `deploy_snapshot_controller` to prevent confusion?

AndrewSirenko (approved):
Looks great! Left a few non-blocking comments/questions.

hack/ebs-scale-test/helpers/cluster-setup/manage-cluster.sh
torredil: Sure, parameterizing the version sounds like a good idea - done in latest rev.

hack/ebs-scale-test/scale-test
torredil: Renamed to `deploy_snapshot_controller`.

hack/ebs-scale-test/helpers/scale-test/snapshot-volume-scale-test/snapshot-volume-scale.yaml
AndrewSirenko: Ah for context that expand and modify manifest needs to be gp2 because we use VAC to convert to gp3. Looking back, a comment there would have been helpful.

The scale-sts test uses gp3. But it shouldn't matter either way.

hack/ebs-scale-test/helpers/scale-test/snapshot-volume-scale-test/snapshot-volume-scale.sh
```diff
@@ -0,0 +1,193 @@
...
+  echo "Deleting PVCs"
+  delete_pvcs
+
+  echo "Waiting for PVs to be deleted"
+  wait_for_pvs_deleted
+
+  echo "Cleaning up snapshot resources"
+  cleanup_snapshot_resources
+
+  trap - EXIT
```
ConnorJC3, line 62: What's the point of this?

hack/ebs-scale-test/helpers/scale-test/snapshot-volume-scale-test/snapshot-volume-scale.yaml
```diff
@@ -0,0 +1,50 @@
...
+    app: snapshot-scale-test
+spec:
+  accessModes:
+    - ReadWriteOnce
+  storageClassName: ebs-snapshot-test
+  resources:
+    requests:
+      storage: 1Gi
+---
+{{- end }}
```
ConnorJC3, line 50: np: missing newline end of file

hack/ebs-scale-test/helpers/cluster-setup/scale-driver-values.yaml
```diff
@@ -27,3 +27,6 @@ sidecars:
     additionalArgs: ["--http-endpoint=:8082"]
   attacher:
     additionalArgs: ["--http-endpoint=:8084"]
+  snapshotter:
+    additionalArgs: ["--http-endpoint=:8085"]
+
```
ConnorJC3, line 32: np: extra newline

hack/ebs-scale-test/helpers/cluster-setup/manage-cluster.sh
```diff
@@ -48,6 +48,15 @@ check_lingering_volumes() {
   [[ lingering_vol_count -ne 0 ]] && echo "WARNING: detected $lingering_vol_count lingering ebs-scale-test EBS volumes in $AWS_ACCOUNT_ID. Please run \`aws ec2 describe-volumes --filters 'Name=tag-key,Values=ebs-scale-test'\` and audit their AWS resource tags. Note these volumes may belong to a different scalability run than $SCALABILITY_TEST_RUN_NAME"
 }
 
+check_lingering_snapshots() {
+  lingering_snap_count=$(aws ec2 describe-snapshots \
+    --filters "Name=tag-key,Values=ebs-scale-test" \
+    --query 'length(Snapshots[*])' \
+    --output text)
+
+  [[ lingering_snap_count -ne 0 ]] && echo "WARNING: detected $lingering_snap_count lingering ebs-scale-test EBS snapshots in $AWS_ACCOUNT_ID. Please run \`aws ec2 describe-snapshots --filters 'Name=tag-key,Values=ebs-scale-test'\` and audit their AWS resource tags. Note these snapshots may belong to a different scalability run than $SCALABILITY_TEST_RUN_NAME"
```
ConnorJC3, line 57: Could we tag by run name to affirmatively know this?

hack/ebs-scale-test/helpers/scale-test/snapshot-volume-scale-test/snapshot-volume-scale.sh
```diff
@@ -0,0 +1,193 @@
...
+    else
+      echo "Only $bound_pvc_count PVCs are bound, waiting for a total of $REPLICAS..."
+      sleep 5
+    fi
+  done
+}
+
+create_snapshots() {
+  for i in $(seq 0 $(($REPLICAS - 1))); do
+    for j in $(seq 1 $SNAPSHOTS_PER_VOLUME); do
```
ConnorJC3, line 80: Because this is bash, it can be simplified to `for i in {1..$(($REPLICAS-1))}` and `for j in {1..$SNAPSHOTS_PER_VOLUME}` (and thus not rely on the `seq` binary)

hack/ebs-scale-test/helpers/scale-test/snapshot-volume-scale-test/snapshot-volume-scale.sh
torredil: This clears the previously set trap:
```
trap 'echo "Test interrupted! Deleting test resources to prevent leak"; cleanup_snapshot_resources' EXIT
```

It's a common pattern in cleanup scripts to ensure resources are properly cleaned up even if the script fails, the trap is removed once everything has completed successfully.

hack/ebs-scale-test/helpers/scale-test/snapshot-volume-scale-test/snapshot-volume-scale.sh
torredil: Done ✅

ConnorJC3 (approved):
/lgtm

Latest revision with changes lgtm, did not test manually yet though

Issue #2439, opened 2025-04-10
support templating of --extra-tags


aaron-trout:
**Is your feature request related to a problem?/Why is this needed**

I would like to enforce some particular tags for cost tracking on all volumes created by the EBS CSI driver, therefore I want to use `--extra-tags` rather than StorageClass params. Additionally, I want to inject dynamic info in the tag values like PVC namespace.

A [comment from @ConnorJC3 in #1664](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1664#issuecomment-1613638034) suggested that this is possible, however that does not appear to be the case. In [controller.go#L301](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/c859322416e46447ec6cafbcf78eb160f2583930/pkg/driver/controller.go#L301) we can see that the templating is only ran against the StorageClass tags, not the tags from `--extra-tags`

/feature

**Describe the solution you'd like in detail**

Also run templating on the tags provided via `--extra-tags`

**Describe alternatives you've considered**

Alternatives appear to be adding tags via StorageClass params or VolumeAttributesClass but both of these can be inadvertently bypassed if users create their own StorageClass (something we see a lot due to requiring different encryption keys / other reasons).

**Additional context**

I tried it out before reviewing the code:

<img width="715" alt="Image" src="https://github.com/user-attachments/assets/bb4c0e3f-6423-4718-a4aa-993557e20808" />

AndrewSirenko:
Hi @aaron-trout, I just reproduced your issue. 

```
controller:
   additionalArgs: ['--extra-tags=VolumeNamespace={{ .PVCNamespace }},extrakey=extraval']
```

<img width="489" alt="Image" src="https://github.com/user-attachments/assets/97d4ba7a-b5b0-43e1-8959-9830325f243a" />

You are indeed correct, templating is not run against these extra-tags (need to also double check that tags in VAC at provision time are templated). 

We'll address and better document this use-case. Thank you for the clear issue!

aaron-trout:
Thanks @AndrewSirenko / @mdzraf! This will be really helpful for enforcing particular tags at the EBS CSI Driver level rather than having to manage it on all StorageClass objects.

PR #2470, opened 2025-05-13
Adding Proper Handling of VAC and Controller Tags On Volume Creation

### LINKED ISSUE #2442: VAC tags should appear on volumes created with VAC

Today, tags specified by a VolumeAttributesClass will only appear on EBS Volume associated with PVC if the VAC was added to PVC *after* volume creation [as shown in our example](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/modify-volume)

A PVC that was created with a specified volumeAttributesClassName should have the tags that the VAC specifies (this is what happens with other parameters like volume type)

Customers can specify tags in their storage class or the `--extra-tags` driver parameter. 

**Additional context**

The following manifests when created should yield an EBS Volume with location and cost-center tags. With EBS CSI Driver v1.41.0 it creates an io2 volume with no additional tags. 

```
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
parameters:
  type: gp3
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  volumeAttributesClassName: io2-class
  resources:
    requests:
      storage: 10Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: app
spec:
  containers:
  - name: app
    image: public.ecr.aws/amazonlinux/amazonlinux
    command: ["/bin/sh"]
    args: ["-c", "while true; do echo $(date -u) | tee /data/out.txt; sleep 5; done"]
    volumeMounts:
    - name: persistent-storage
      mountPath: /data
  volumes:
  - name: persistent-storage
    persistentVolumeClaim:
      claimName: ebs-claim
---
apiVersion: storage.k8s.io/v1beta1
kind: VolumeAttributesClass
metadata:
  name: io2-class
driverName: ebs.csi.aws.com
parameters:
  type: io2
  iops: "3000"
  tagSpecification_1: "location=Seattle"
  tagSpecification_2: "cost-center="

```

mdzraf:
#### What type of PR is this?

kind feature

This PR adds proper handling of volumeAttributesClass tags so that the tags specified in the VAC are templated, validated, and added to the volume on volume creation. It also handles the templating and validation of any `--extra-tags` on the controller on volume creation.

Closes #2442 #2439

1. Ran make test 
```
> make test 
Ran 76 of 92 Specs in 0.442 seconds
SUCCESS! -- 76 Passed | 0 Failed | 1 Pending | 15 Skipped
```

2. Used/deployed the following resource configurations and then ran `aws ec2 describe-volumes --volume-ids vol-<ID fo provisioned volume>` to make sure that the tags specified in the VAC were added and properly interpolated. 
```
---
apiVersion: storage.k8s.io/v1beta1
kind: VolumeAttributesClass
metadata:
  name: io2-class
driverName: ebs.csi.aws.com
parameters:
  type: io2
  iops: "3000"
  tagSpecification_1: "pvcnamespaceTEST={{ .PVCNamespace }}"
  tagSpecification_2: "pvcnameTEST={{ .PVCName }}"
  tagSpecification_3: "pvnameTEST={{ .PVName }}"
  tagSpecification_4: "location=Seattle"
  tagSpecification_5: "cost-center="
  tagSpecification_6: 'backup={{ .PVCNamespace | contains "prod" }}'
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  volumeAttributesClassName: io2-class
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 4Gi
 ---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
---
apiVersion: v1
kind: Pod
metadata:
  name: app
spec:
  containers:
  - name: app
    image: public.ecr.aws/amazonlinux/amazonlinux
    command: ["/bin/sh"]
    args: ["-c", "while true; do echo $(date -u) >> /data/out.txt; sleep 5; done"]
    volumeMounts:
    - name: persistent-storage
      mountPath: /data
  volumes:
  - name: persistent-storage
    persistentVolumeClaim:
      claimName: ebs-claim
```

## **Result of describe call**
```
{
            "Iops": 2000,
            "Tags": [
                {
                    "Key": "Name",
                    "Value": "ebs-csi-e2e.k8s.local-dynamic-pvc-cffd7dc2-3d86-49a2-b787-e64861fc1f7c"
                },
                {
                    "Key": "cost-center",
                    "Value": ""
                },
                {
                    "Key": "ebs.csi.aws.com/cluster",
                    "Value": "true"
                },
                {
                    "Key": "CSIVolumeName",
                    "Value": "pvc-cffd7dc2-3d86-49a2-b787-e64861fc1f7c"
                },
                {
                    "Key": "extrakey",
                    "Value": "extraval"
                },
                {
                    "Key": "backup",
                    "Value": "false"
                },
                {
                    "Key": "kubernetes.io/cluster/ebs-csi-e2e.k8s.local",
                    "Value": "owned"
                },
                {
                    "Key": "kubernetes.io/created-for/pvc/name",
                    "Value": "ebs-claim"
                },
                {
                    "Key": "KubernetesCluster",
                    "Value": "ebs-csi-e2e.k8s.local"
                },
                {
                    "Key": "location",
                    "Value": "Seattle"
                },
                {
                    "Key": "kubernetes.io/created-for/pvc/namespace",
                    "Value": "default"
                },
                {
                    "Key": "pvcnamespaceTEST",
                    "Value": "default"
                },
                {
                    "Key": "VolumeNamespace",
                    "Value": "default"
                },
                {
                    "Key": "pvcnameTEST",
                    "Value": "ebs-claim"
                },
                {
                    "Key": "pvnameTEST",
                    "Value": "pvc-cffd7dc2-3d86-49a2-b787-e64861fc1f7c"
                },
                {
                    "Key": "kubernetes.io/created-for/pv/name",
                    "Value": "pvc-cffd7dc2-3d86-49a2-b787-e64861fc1f7c"
                }
            ],
            "VolumeType": "io2"...
```

```release-note
Added Proper Handling To Ensure Tags Added in a VAC Which a PVC Points To Get Added to the Provisioned Volume. Also Added Proper Handling To Ensure that extra-tags on the Controller Also Get Added to the Provisioned Volume.
```

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -213,7 +213,7 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}
 
-	modifyOptions, err := parseModifyVolumeParameters(req.GetMutableParameters())
+	modifyOptions, err := parseCreateRequestMutableParameters(req.GetMutableParameters(), d.options.ExtraTags, *tProps)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid mutable parameter: %v", err)
 	}
@@ -301,9 +301,6 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		volumeTags[NameTag] = d.options.KubernetesClusterID + "-dynamic-" + volName
 		volumeTags[KubernetesClusterTag] = d.options.KubernetesClusterID
 	}
-	for k, v := range d.options.ExtraTags {
-		volumeTags[k] = v
-	}
 
 	addTags, err := template.Evaluate(scTags, tProps, d.options.WarnOnInvalidTag)
 	if err != nil {
@@ -315,6 +312,10 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	for k, v := range addTags {
+		volumeTags[k] = v
+	}
+
+	for k, v := range modifyOptions.modifyTagsOptions.TagsToAdd {
 		volumeTags[k] = v
 	}
 
@@ -350,6 +351,63 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Errorf(errCode, "Could not create volume %q: %v", volName, err)
 	}
 	return newCreateVolumeResponse(disk, responseCtx), nil
+}
+
+func parseCreateRequestMutableParameters(params map[string]string, extraTags map[string]string, tProps template.PVProps) (*modifyVolumeRequest, error) {
+	options := modifyVolumeRequest{
+		modifyTagsOptions: cloud.ModifyTagsOptions{
+			TagsToAdd:    make(map[string]string),
+			TagsToDelete: make([]string, 0),
+		},
+	}
+
+	rawTagsToAdd := []string{}
+	for key, value := range params {
+		switch key {
+		case ModificationKeyIOPS:
+			iops, err := strconv.ParseInt(value, 10, 32)
+			if err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Could not parse IOPS: %q", value)
+			}
+			options.modifyDiskOptions.IOPS = int32(iops)
+		case ModificationKeyThroughput:
+			throughput, err := strconv.ParseInt(value, 10, 32)
+			if err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Could not parse throughput: %q", value)
+			}
+			options.modifyDiskOptions.Throughput = int32(throughput)
+		case DeprecatedModificationKeyVolumeType:
+			if _, ok := params[ModificationKeyVolumeType]; ok {
+				klog.Infof("Ignoring deprecated key `volumeType` because preferred key `type` is present")
+				continue
+			}
+			klog.InfoS("Key `volumeType` is deprecated, please use `type` instead")
+			options.modifyDiskOptions.VolumeType = value
+		case ModificationKeyVolumeType:
+			options.modifyDiskOptions.VolumeType = value
+		default:
+			switch {
+			case strings.HasPrefix(key, ModificationAddTag):
+				rawTagsToAdd = append(rawTagsToAdd, value)
+			default:
+				return nil, status.Errorf(codes.InvalidArgument, "Invalid mutable parameter key: %s", key)
+			}
+		}
+	}
+
+	for k, v := range extraTags {
+		rawTagsToAdd = append(rawTagsToAdd, k+"="+v)
+	}
+
+	addTags, err := template.Evaluate(rawTagsToAdd, tProps, false)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Error interpolating the tag value: %v", err)
+	}
+	if err := validateExtraTags(addTags, false); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Invalid tag value: %v", err)
+	}
+	options.modifyTagsOptions.TagsToAdd = addTags
+	return &options, nil
 }
 
 func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {


pkg/driver/controller.go
```diff
@@ -352,6 +353,63 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+		modifyTagsOptions: cloud.ModifyTagsOptions{
+			TagsToAdd:    make(map[string]string),
+			TagsToDelete: make([]string, 0),
+		},
+	}
+
+	rawTagsToAdd := []string{}
+	for key, value := range params {
+		switch key {
+		case ModificationKeyIOPS:
```
AndrewSirenko, line 367: I'm a bit worried at the amount [code duplication](https://www.codewithjason.com/duplication/#:~:text=The%20main%20reason%20duplication%20is,deviating%20behavior%20is%20necessarily%20wrong.) we have for validating parameters that can be specified in either the StorageClass or the equivalent mutable_parameters from VolumeAttributesClass. Can we consolidate the logic?

Before this PR there are 2 separate functions that validate parameters shared between SC and VAC like IOPS:
- controller.go: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/6c1235db741c76ad9e886feda5a1822f7e24fc13/pkg/driver/controller.go#L147
- controller_modify_volume.go: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/881426113a05ea934399147792c15a2ab89e28df/pkg/driver/controller_modify_volume.go#L178

This PR adds a *third* place we copy paste similar code.

--- 

For the create path, can we instead overwrite any SC parameters with values from VAC mutable_parameters BEFORE validating them? They share the same parameter key anyway, and can be validated within the loop [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/6c1235db741c76ad9e886feda5a1822f7e24fc13/pkg/driver/controller.go#L133) (in the long run we should consider refactoring that loop + switch) 

This is in line with CSI Spec: 

```Values specified in mutable_parameters MUST take precedence over the values from parameters.```

File: pkg/driver/controller.go [hunk 2]
```diff
@@ -318,6 +315,10 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		volumeTags[k] = v
 	}
 
+	for k, v := range modifyOptions.modifyTagsOptions.TagsToAdd {
```
AndrewSirenko, line 318: To me, it isn't intuitive that tags specified via [extra-tags option](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/options.md) would wind up in a struct called `modifyOptions.modifyTagOptions.TagsToAdd`. 

Have you considered instead:
1. Collating all validated scTags, extraTags, and vacTags into an array called `tagsToEvaluate`
2.  Put them through `template.Evaluate` and `validateExtraTags`
3. Append them to list of volumeTags that get passed in to cloud.DiskOptions

That way all tags, no matter their source, go through a clear set of steps? Bonus points if these steps are located near each other so it's easy to see how all tags are processed.

pkg/driver/controller_test.go
```diff
@@ -1497,6 +1497,81 @@ func TestCreateVolume(t *testing.T) {
 				}
 			},
 		},
+		{
```
ConnorJC3, line 1500: Do we need a test case for interpolation of CLI flags?

pkg/driver/controller.go
```diff
@@ -206,28 +206,66 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "Invalid mutable parameter: %v", err)
-	}
+	mutableParameters := req.GetMutableParameters()
 
 	// "Values specified in mutable_parameters MUST take precedence over the values from parameters."
 	// https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume
-	if modifyOptions.modifyDiskOptions.VolumeType != "" {
-		volumeType = modifyOptions.modifyDiskOptions.VolumeType
+	// Iterating mutableParameters separately to not incorrectly set default values on creation in case of parameters not having these set.
```
ConnorJC3, line 220: I'm not sure I understand what this comment is trying to say - can you clarify?

pkg/driver/controller.go
```diff
@@ -206,28 +206,66 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+
+	for key, value := range d.options.ExtraTags {
+		tagsToEvaluate = append(tagsToEvaluate, key+"="+value)
 	}
-	if modifyOptions.modifyDiskOptions.IOPS != 0 {
-		iops = modifyOptions.modifyDiskOptions.IOPS
+
+	addTags, err := template.Evaluate(tagsToEvaluate, tProps, d.options.WarnOnInvalidTag)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Error interpolating the tag value: %v", err)
```
ConnorJC3, line 260: np: remove "the" (should be "Error interpolating tag value: %v")

pkg/driver/controller.go
```diff
@@ -206,28 +206,66 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+			volumeType = value
+		default:
+			switch {
+			case strings.HasPrefix(key, ModificationAddTag):
+				tagsToEvaluate = append(tagsToEvaluate, value)
+			default:
+				return nil, status.Errorf(codes.InvalidArgument, "Invalid mutable parameter key: %s", key)
+			}
+		}
+	}
```
ConnorJC3, line 252: I'm not going to ask you to fix it in this PR, but this parameter handling in our driver is getting out of control and needs to be unified. We now have 3 (or maybe 4?) different versions of essentially the same code with minor tweaks that have to be simultaneously maintained.

pkg/driver/controller.go
```diff
@@ -206,28 +206,66 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+				continue
+			}
+			klog.InfoS("Key `volumeType` is deprecated, please use `type` instead")
+			volumeType = value
+		case VolumeTypeKey:
+			volumeType = value
+		default:
+			switch {
+			case strings.HasPrefix(key, ModificationAddTag):
+				tagsToEvaluate = append(tagsToEvaluate, value)
```
ConnorJC3, line 247: How does this handle a tag value being specified in both the SC and VAC parameters?

pkg/driver/controller.go
mdzraf: Next rev will be fixed, also removed it from the modify path as well.

pkg/driver/controller.go
mdzraf: Agreed, we should look into unifying these.

pkg/driver/controller_test.go
mdzraf: When you say CLI flags do you mean like if an user tried adding tags through the cli? I.e `aws ec2 create-tags --resources vol-12345689  --tags Key=pvname,Value={{ .PVName }}`

AndrewSirenko (approved):
Conditional lgtm other than Connor's comments. Thanks for cleaning up our code and adding some tests!

pkg/driver/controller_test.go
AndrewSirenko: I believe what Connor meant was adding a test for:

If someone added tags with the driver [extra-tags](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/6af42f368864be9f4ea38d1d7092d2325192719c/docs/options.md?plain=1#L12) parameter, would they be interpolated?

You can add this to your current test by modifying controller_test.go line 1562 from

```
awsDriver := ControllerService{
					cloud:    mockCloud,
					inFlight: internal.NewInFlight(),
					options:  &Options{},
				}
```

To something like 

```
awsDriver := ControllerService{
					cloud:    mockCloud,
					inFlight: internal.NewInFlight(),
					options:  &Options{ExtraTags: map[string]string{"SOME-KEY": "SOME-TEMPLATABLE-VAL"}},
				}
```

And then checking that the created volume had the interpolated tag!

mdzraf: Ah, thanks for clearing that up, added in next rev!

ConnorJC3 (approved):
/lgtm

torredil (approved):
This is good to go but please fix the commit history before approving.

/lgtm

ElijahQuinones (approved):

PR #2471, opened 2025-05-13
Separate go.mod for E2E tests (don't import k/k in driver binary)

AndrewSirenko:
#### What type of PR is this?

/kind cleanup

This PR is based off of and supercedes #2284 

Kubernetes projects should avoid adding `k8s.io/kubernetes` to their dependency graph to reduce dependency bloat reduce the surface area for vulnerabilities. 

Most of these imports are only used in our e2e tests. Therefore we have created a new `go.mod` for our e2e tests.

We also need to replace the helper function `volume.NewMetricsStatFS(req.GetVolumePath())` with our own Linux and Windows implementations. 

`make test`

Manually testing that the filesystem stats are accurately reported to kubelet on Linux and Windows nodes. 

Screenshot 2025-05-12 at 4 23 03 PM

Screenshot 2025-05-12 at 3 53 55 PM

```release-note
NONE
```


pkg/mounter/mount_linux.go
```diff
@@ -303,3 +304,73 @@ func (m *NodeMounter) Unstage(path string) error {
...
+func (m *NodeMounter) GetVolumeStats(volumePath string) (VolumeStats, error) {
+	stats := VolumeStats{}
+
+	statfs := &unix.Statfs_t{}
+	err := unix.Statfs(volumePath, statfs)
+	if err != nil {
+		return stats, err
+	}
+
+	// Get byte stats (safely)
```
AndrewSirenko, line 318: Our `gosec` linter complained on the initial version of this PR due to G115 (casting without bounds checks)

Happy to just ignore the linter for this function `//nolint:gosec` because it is highly unlikely we would overflow/underflow because max EBS volume is 64Ti, and int64 max value can accommodate `8388608` Terabytes and we also don't support any filesystem that can accommodate this high a size

Thoughts?

ConnorJC3: No strong opinion, but my gut reaction is to just disable that check here - as you mention it should be practically impossible for this to overflow - _and even if it does_, it's not a security issue just some invalid metrics will be vended.

pkg/mounter/mount_windows.go
```diff
@@ -282,3 +285,43 @@ func (m *NodeMounter) Unstage(target string) error {
...
+	ret, _, err := syscall.Syscall6(
+		procGetDiskFreeSpaceEx.Addr(),
+		4,
+		uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(volumePath))),
+		uintptr(unsafe.Pointer(&stats.AvailableBytes)),
+		uintptr(unsafe.Pointer(&stats.TotalBytes)),
+		uintptr(unsafe.Pointer(&totalFreeBytes)),
+		0,
+		0,
+	)
```
ConnorJC3, line 319: Would it not make more sense to use `GetDiskFreeSpaceEx` in `golang.org/x/sys/windows` instead of `Syscall6`? That would also avoid all the weird global variable DLL shuffling I think.

go.mod
```diff
@@ -34,17 +30,12 @@ require (
 	k8s.io/client-go v1.5.2
 	k8s.io/component-base v0.33.0
 	k8s.io/klog/v2 v2.130.1
-	k8s.io/kubernetes v1.33.0
```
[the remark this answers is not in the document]
AndrewSirenko: We cannot remove *all* of them (tried that earlier)

But on further look we can remove many of them. Thanks!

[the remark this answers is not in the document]
AndrewSirenko: Edit: Misread comment, we can delete all the k/k replace directives in top level go.mod.

We can also delete a bunch of the replace directives for the e2e go.mod as well.

pkg/mounter/mount_windows.go
ConnorJC3: My suggestion would be to use `x/sys/windows` unless there is a strong reason not to because (1) it will likely result in cleaner code and (2) Golang discourages use of the `syscall` package ("Most of the functions, types, and constants defined in this package are also available in the [golang.org/x/sys](https://pkg.go.dev/golang.org/x/sys) package. That package has more system call support than this one, and most new code should prefer that package where possible." in https://pkg.go.dev/syscall)

pkg/mounter/mount_windows.go
```diff
@@ -282,3 +285,43 @@ func (m *NodeMounter) Unstage(target string) error {
...
+	if ret == 0 {
```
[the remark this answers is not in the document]
ConnorJC3: (Note: probably redundant if we switch to x/sys/windows

pkg/mounter/mount_windows.go
AndrewSirenko: Yep agreed, switched in latest commit.

AndrewSirenko:
Weird, external windows resize tests failing despite not modifying any Controller/NodeExpandVolume code:  

```
FailedMount: MountVolume.Setup failed while expanding volume for volume "pvc-3f792f99-1bc0-4e3f-86e9-89c562b009f7" : Expander.NodeExpand failed to expand the volume : rpc error: code = Internal desc = Could not resize volume "vol-0893721da3dc84e97" ("2"): rpc error: code = Unknown desc = error getting sizemin,sizemax from mount. cmd: Get-Volume -UniqueId "$Env:volumeID" | Get-partition | Get-PartitionSupportedSize | Select SizeMax | ConvertTo-Json, output: Get-PartitionSupportedSize : Failed
  Activity ID: {0cc5c101-44f7-4d4c-8e9b-a154a36c42fe}
  At line:1 char:56
  + ...  "$Env:volumeID" | Get-partition | Get-PartitionSupportedSize | Selec ...
  +                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~
      + CategoryInfo          : NotSpecified: (StorageWMI:ROOT/Microsoft/.../MSFT_Partition) [Get-PartitionSupportedSize 
     ], CimException
      + FullyQualifiedErrorId : StorageWMI 4,Get-PartitionSupportedSize
   
  {
      "SizeMax":  null
  }
```

/retest

ElijahQuinones (approved):
/lgtm 

make test and CI e2e tests pass

torredil:
np: Regarding the release note, this patch is an internal implementation detail of the driver and the auto-generated CHANGELOG will already include the full list of removed / modified dependencies so I suggest removing it.

/lgtm

PR #2475, opened 2025-05-14
Update io2MaxIOPSPerGb from 500 to 1000

AndrewSirenko:
#### What type of PR is this?

/kind bug

EC2 updated io2 max IOPS to size ratio from 500 to 1000: 

```
❯ aws ec2 create-volume --availability-zone us-west-2a --size 6 --iops 10000 --volume-type io2

An error occurred (InvalidParameterCombination) when calling the CreateVolume operation: io2 volumes configured with greater than 64 TiB or 256K IOPS or 1000:1 IOPS:GB ratio are not supported.

❯ aws ec2 create-volume --availability-zone us-west-2a --size 6 --iops 6000 --volume-type io2
{
    "Iops": 6000,
    "Tags": [],
    "VolumeType": "io2",
    "MultiAttachEnabled": false,
    "VolumeId": "vol-0a8c747261f4e3a15",
    "Size": 6,
    "SnapshotId": "",
    "AvailabilityZone": "us-west-2a",
    "State": "creating",
    "CreateTime": "2025-05-14T14:55:30+00:00",
    "Encrypted": true
}
```

We should ensure driver reflects true value. In the longterm, we should avoid massaging/pre-validating user's IOPS parameter to avoid situations like this. 

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 6Gi

apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
parameters:
  csi.storage.k8s.io/fstype: ext4
  type: io2
  iopsPerGB: "999"
  encrypted: "true"
```

Before PR:

Volume incorrectly capped at 3k IOPS (6 * 500)

image

After:

Volume created with 5994 IOPS!

image

```release-note
Increase io2 volume maximum iops:size ratio from 500:1 to 1000:1 to match EC2 behavior
```

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -29,12 +29,12 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
 * If the requested IOPS (either directly from `iops` or from `iopsPerGB` multiplied by the volume's capacity) produces a value above the maximum IOPS allowed for the [volume type](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html), the IOPS will be capped at the maximum value allowed. If the value is lower than the minimal supported IOPS value per volume, either an error is returned (the default behavior), or the value is increased to fit into the supported range when `allowautoiopspergbincrease` is `"true"`.
 * You may specify either the "iops" or "iopsPerGb" parameters, not both. Specifying both parameters will result in an invalid StorageClass.
 
-| Volume Type                | Min total IOPS | Max total IOPS | Max IOPS per GB   |
-|----------------------------|----------------|---------------|-------------------|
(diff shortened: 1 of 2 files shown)


torredil (approved):
Super important update. For the upcoming release, we should consider modifying the auto-generated CHANGELOG to include an additional note at the top (similar to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG.md#v1400) discussing provisioned volume performance implications for versions not including this patch.

/lgtm

PR #2478, opened 2025-05-15
Add security Makefile target

AndrewSirenko:
#### What type of PR is this?

/kind cleanup

Add security Makefile target that currently runs govulncheck against only our ebs-csi-driver module. That way we can move our govulncheck GH action to Prow (so we can override it). 

Spotted a vuln on release-1.39 branch!

```
❯ git checkout release-1.39
Your branch is ahead of 'upstream/release-1.39' by 1 commit.

❯ make clean && make security
rm -rf bin/
/home/linuxbrew/.linuxbrew/bin/python3
./hack/tools/check-security.sh
=== Symbol Results ===

Vulnerability #1: GO-2025-3547
    Kubernetes kube-apiserver Vulnerable to Race Condition in k8s.io/kubernetes
  More info: https://pkg.go.dev/vuln/GO-2025-3547
  Module: k8s.io/kubernetes
    Found in: k8s.io/kubernetes@v1.32.2
    Fixed in: N/A
    Example traces found:
      #1: pkg/driver/node.go:43:2: driver.init calls volume.init, which eventually calls common.init
      #2: pkg/cloud/metadata/k8s.go:105:45: metadata.KubernetesAPIInstanceInfo calls 
...
Your code is affected by 2 vulnerabilities from 1 module.
This scan also found 0 vulnerabilities in packages you import and 3
vulnerabilities in modules you require, but your code doesn't appear to call
these vulnerabilities.
Use '-show verbose' for more details.
make: *** [security] Error 3
❯ echo $?
2
```

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,6 +212,10 @@ update-sidecar-dependencies: update-truth-sidecars generate-sidecar-tags update/
 update-image-dependencies: update-sidecar-dependencies
 	./hack/release-scripts/update-e2e-images
 
+.PHONY: security
+security: bin/govulncheck
+	./hack/tools/check-security.sh
+
 ## CI aliases
 # Targets intended to be executed mostly or only by CI jobs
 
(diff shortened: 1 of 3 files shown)


torredil (approved):
/lgtm

PR #2479, opened 2025-05-16
Fix windows example

AndrewSirenko:
#### What type of PR is this?

/kind documentation

Windows example is broken.

Smaller [nanoserver image](https://hub.docker.com/r/microsoft/windows-nanoserver) wasn't working, so for now using `ltsc2022`. 

Run through windows example after creating windows cluster

```release-note
NONE
```


ConnorJC3 (commented):
/lgtm

Diff lgtm, suggest changing release notes to `NONE` as this does not introduce a user-visible change in what we release

ConnorJC3:
Oh, I forgot to mention but I did a manual `kubectl apply -f manifests` on a Windows cluster and it worked, pod came up all good.

PR #2480, opened 2025-05-16
Update tests/e2e go.mod in Dependabot

ConnorJC3:
#### What type of PR is this?

/kind cleanup

We split out a separate `go.mod` in #2471, but now Dependabot PRs are failing because it only tries to update the main `go.mod`. This PR fixes that issue.

It wasn't

```release-note
NONE
```


ElijahQuinones (approved):
/lgtm
/approve

Issue #2447, opened 2025-04-18
Driver fails to remove `ebs.csi.aws.com/agent-not-ready` taint because `eks:node-manager` adds it later


miguelhar:
/kind bug

**What happened?**
```
I0418 20:03:39.807826       1 main.go:153] "Initializing metadata"
I0418 20:03:39.812678       1 ec2.go:93] "Number of attached ENIs" attachedENIs=3
I0418 20:03:39.814895       1 metadata.go:48] "Retrieved metadata from IMDS"
I0418 20:03:39.815106       1 mount_linux.go:326] Cannot create temp dir to detect safe 'not mounted' behavior: mkdir /tmp/kubelet-detect-safe-umount3866425026: read-only file system
I0418 20:03:39.815511       1 driver.go:69] "Driver Information" Driver="ebs.csi.aws.com" Version="v1.39.0"
I0418 20:03:39.815681       1 driver.go:138] "Listening for connections" address="/csi/csi.sock"
E0418 20:03:40.826459       1 node.go:855] "Unexpected failure when attempting to remove node taint(s)" err="isAllocatableSet: driver not found on node ip-10-0-124-248.us-east-2.compute.internal"
E0418 20:03:41.334123       1 node.go:855] "Unexpected failure when attempting to remove node taint(s)" err="isAllocatableSet: driver not found on node ip-10-0-124-248.us-east-2.compute.internal"
I0418 20:03:41.365892       1 node.go:574] "NodeGetInfo: called" args=""
I0418 20:03:42.343460       1 node.go:935] "CSINode Allocatable value is set" nodeName="ip-10-0-124-248.us-east-2.compute.internal" count=24
I0418 20:03:42.343483       1 node.go:896] "No taints to remove on node, skipping taint removal"
^C%
```
The driver does not find the `'ebs.csi.aws.com/agent-not-ready=true'` taint, hence its not removed.
`eks:node-manager` taints the node  afterwards, causing the node to be unusable(There is nothing that removes the taint at this point). Restarting the pod resolves the issue because the driver can now find the taint.

There is a race condition that is currently being mitigated (most of the time with) a 1 second delay, set [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L75). 

**What you expected to happen?**

The driver attempts to remove the taint after `eks:node-manager` had added it.

**How to reproduce it (as minimally and precisely as possible)?**

[...]

PR #2492, opened 2025-05-19
Reliably remove driver startup taint

torredil:
#### What type of PR is this?

Fixes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2447

/kind bug

The current implementation uses a 1 s sleep to look for the `ebs.csi.aws.com/agent‑not‑ready` taint.  If `eks:node‑manager` (or any other controller) applies the taint *after* that initial probe, the driver never retries and the node stays unschedulable until the pod is restarted.

This patch replaces that logic with a short‑lived, event‑driven Node informer filtered to the current node as a better mechanism for reliably removing the startup taint.

Note: Some minor refactoring was done in the existing unit tests because we no longer need to mock the call to retrieve the node object in `removeNotReadyTaint` -  now, the node object gets passed in directly into `removeNotReadyTaint` from the informer's event handling.

- `make verify && make test`.
- Manually, a scenario where where the taint is present before node pod starts, and a scenario where the taint is added after node pod starts.

Driver logs:
```
kubectl logs ebs-csi-node-hjq2g -n kube-system                                               ✔  torredil-Isengard@dev.us-east-1.eksctl.io ○  torredil@dev-dsk-torredil-1d-e1003bcc
Defaulted container "ebs-plugin" out of: ebs-plugin, node-driver-registrar, liveness-probe
I0519 13:43:52.611344       1 main.go:153] "Initializing metadata"
I0519 13:43:52.611404       1 metadata.go:54] "Attempting to retrieve instance metadata from IMDS"
I0519 13:43:52.615709       1 metadata.go:57] "Retrieved metadata from IMDS"
I0519 13:43:52.616754       1 envvar.go:172] "Feature gate default state" feature="InformerResourceVersion" enabled=false
I0519 13:43:52.616768       1 envvar.go:172] "Feature gate default state" feature="InOrderInformers" enabled=true
I0519 13:43:52.616772       1 envvar.go:172] "Feature gate default state" feature="WatchListClient" enabled=false
I0519 13:43:52.616776       1 envvar.go:172] "Feature gate default state" feature="ClientsAllowCBOR" enabled=false
I0519 13:43:52.616780       1 envvar.go:172] "Feature gate default state" feature="ClientsPreferCBOR" enabled=false
I0519 13:43:52.617071       1 driver.go:69] "Driver Information" Driver="ebs.csi.aws.com" Version="v1.43.0"
I0519 13:43:52.629161       1 reflector.go:430] "Caches populated" type="*v1.Node" reflector="/gomodcache/k8s.io/client-go@v0.33.0/tools/cache/reflector.go:285"
I0519 13:43:54.167774       1 node.go:1009] "CSINode Allocatable value is set" nodeName="ip-192-168-15-218.ec2.internal" count=26
I0519 13:43:54.184703       1 node.go:996] "Removed taint(s) from local node" node="ip-192-168-15-218.ec2.internal"
I0519 13:43:54.184719       1 node.go:877] "Successfully removed agent‑not‑ready taint; stopping watcher"
```

```
kubectl describe node ip-192-168-15-218.ec2.internal | grep "Taints"            
Taints:             <none>
```

```release-note
Reliably remove the `ebs.csi.aws.com/agent-not-ready` taint at start‑up using a short‑lived informer, eliminating a race condition that could leave nodes unschedulable.
```

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -38,7 +38,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 )
 
@@ -66,15 +68,11 @@ var (
 		csi.NodeServiceCapability_RPC_EXPAND_VOLUME,
 		csi.NodeServiceCapability_RPC_GET_VOLUME_STATS,
 	}
+)
 
-	// taintRemovalInitialDelay is the initial delay for node taint removal.
-	taintRemovalInitialDelay = 1 * time.Second
-	// taintRemovalBackoff is the exponential backoff configuration for node taint removal.
-	taintRemovalBackoff = wait.Backoff{
-		Duration: 500 * time.Millisecond,
-		Factor:   2,
-		Steps:    10, // Max delay = 0.5 * 2^9 = ~4 minutes
-	}
+const (
+	// taintWatcherDuration is the maximum duration for the not-ready taint watcher to run.
+	taintWatcherDuration = 1 * time.Minute
 )
 
 // NodeService represents the node service of CSI driver.
@@ -89,11 +87,9 @@ type NodeService struct {
 // NewNodeService creates a new node service.
 func NewNodeService(o *Options, md metadata.MetadataService, m mounter.Mounter, k kubernetes.Interface) *NodeService {
 	if k != nil {
-		// Remove taint from node to indicate driver startup success
-		// This is done at the last possible moment to prevent race conditions or false positive removals
-		time.AfterFunc(taintRemovalInitialDelay, func() {
-			removeTaintInBackground(k, taintRemovalBackoff, removeNotReadyTaint)
-		})
+		// Watch for the agent‑not‑ready taint for up to one minute and remove it
+		// as soon as allocatable is available.
+		startNotReadyTaintWatcher(k, taintWatcherDuration)
 	}
 
 	return &NodeService{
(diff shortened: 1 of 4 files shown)


pkg/driver/node.go
```diff
@@ -848,29 +839,104 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+		return
+	}
+
+	stopCh := make(chan struct{})
+	var closeOnce sync.Once
+	closeWatch := func() { closeOnce.Do(func() { close(stopCh) }) }
+
+	factory := informers.NewSharedInformerFactoryWithOptions(
+		clientset,
+		0,
```
ConnorJC3, line 859:
```suggestion
		0, // No resync
```

I had to look up what the `0` does here, maybe add a comment like this to indicate?

pkg/driver/node.go
```diff
@@ -848,29 +839,104 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+
+	attemptTaintRemoval := func(n *corev1.Node) {
+		if !hasNotReadyTaint(n) {
+			return
+		}
+		if err := checkAllocatable(clientset, nodeName); err != nil {
+			return
+		}
+		if err := removeNotReadyTaint(clientset); err != nil {
+			klog.ErrorS(err, "failed to remove agent‑not‑ready taint")
```
ConnorJC3, line 874: np: start log message with a capital letter

pkg/driver/node.go
```diff
@@ -848,29 +839,104 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+				attemptTaintRemoval(n)
+			}
+		},
+		UpdateFunc: func(_, newObj interface{}) {
+			if n, ok := newObj.(*corev1.Node); ok {
+				attemptTaintRemoval(n)
+			}
+		},
+	}); err != nil {
+		klog.ErrorS(err, "taint‑watcher: failed to add event handler")
```
ConnorJC3, line 893: np: start log message with a capital letter (failed -> Failed)

Same np several times in this file later

pkg/driver/node.go
```diff
@@ -848,29 +839,104 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+		}
+		if err := checkAllocatable(clientset, nodeName); err != nil {
+			return
+		}
+		if err := removeNotReadyTaint(clientset); err != nil {
+			klog.ErrorS(err, "failed to remove agent‑not‑ready taint")
+			return
+		}
+		klog.V(2).InfoS("Successfully removed agent‑not‑ready taint; stopping watcher")
+		closeWatch()
```
ConnorJC3, line 878: Why are we stopping here? Isn't the point of this bugfix that the taint gets re-added and thus needs to be re-removed?

pkg/driver/node.go
```diff
@@ -848,29 +839,104 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+func startNotReadyTaintWatcher(clientset kubernetes.Interface, maxWatchDuration time.Duration) {
+	nodeName := os.Getenv("CSI_NODE_NAME")
+	if nodeName == "" {
+		klog.V(4).InfoS("CSI_NODE_NAME missing, skipping taint watcher")
+		return
+	}
+
+	stopCh := make(chan struct{})
+	var closeOnce sync.Once
+	closeWatch := func() { closeOnce.Do(func() { close(stopCh) }) }
```
ConnorJC3, line 855: Given that we don't want to stop after removal (see my below comment), I think this channel can be removed entirely

pkg/driver/node.go
torredil: Done ✅

pkg/driver/node.go
torredil: Done ✅

pkg/driver/node.go
```diff
@@ -89,11 +82,9 @@ type NodeService struct {
...
 func NewNodeService(o *Options, md metadata.MetadataService, m mounter.Mounter, k kubernetes.Interface) *NodeService {
 	if k != nil {
-		// Remove taint from node to indicate driver startup success
-		// This is done at the last possible moment to prevent race conditions or false positive removals
-		time.AfterFunc(taintRemovalInitialDelay, func() {
-			removeTaintInBackground(k, taintRemovalBackoff, removeNotReadyTaint)
-		})
+		// Watch for the agent‑not‑ready taint for up to one minute and remove it
+		// as soon as allocatable is available.
+		startNotReadyTaintWatcher(k, 1*time.Minute)
```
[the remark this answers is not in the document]
torredil: Done ✅

pkg/driver/node.go
torredil: It's not that the taint gets re-added, its more that there are no taints to remove when the check is performed, so the taint removal is skipped, see logs in the original bug report:

```
...
I0418 20:03:42.343483       1 node.go:896] "No taints to remove on node, skipping taint removal"
```

I think not stopping the informer for the full `taintWatcherDuration` makes sense though to handle the scenario you mentioned (a controller re-applying the taint after it is removed), especially given its short lifespan.

pkg/driver/node.go
torredil: Done ✅

ConnorJC3 (approved):
/lgtm

Manually tested by adding a taint to the node twice about 30 seconds after startup, and it was removed both times.

ElijahQuinones (approved):
/lgtm 
/approve

PR #2500, opened 2025-05-21
Return `ResourceExhausted` gRPC error code in `ControllerPublishVolume` when `AttachmentLimitExceeded`

torredil:
#### What type of PR is this?

/kind cleanup

This PR updates the Attach code path to return the `ResourceExhausted` gRPC error code in `ControllerPublishVolume` when the attachment limit is exceeded for the instance. In newer versions of Kubernetes, CSI drivers can opt in to a feature that allows for updating the `CSINode.Allocatables.Count` value when this gRPC error code is reported in the VolumeAttachment object.

```
make verify && make test
```

Manually, by deploying the driver with `volumeAttachLimit: 200` and scaling up StatefulSet to induce this failure mode. See driver log:

```
E0521 18:02:25.853256       1 driver.go:108] "GRPC error" err="rpc error: code = ResourceExhausted desc = Attachment limit exceeded for volume \"vol-0d8d2bab6fc5dd31b\" on node \"i-00f897f741bb9326a\": attachment limit exceeded: operation error EC2: AttachVolume, https response error StatusCode: 400, RequestID: c56fd5ce-f4d5-4857-b1dc-908c935a705a, api error AttachmentLimitExceeded: You may attach up to 27 devices for this instance type. Please use a larger instance size to attach more volumes. For a detailed list of limits please see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html"
```

```release-note
NONE
```

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -422,6 +422,9 @@ func (d *ControllerService) ControllerPublishVolume(ctx context.Context, req *cs
 			klog.InfoS("ControllerPublishVolume: volume not found", "volumeID", volumeID, "nodeID", nodeID)
 			return nil, status.Errorf(codes.NotFound, "Volume %q not found", volumeID)
 		}
+		if errors.Is(err, cloud.ErrAttachmentLimitExceeded) {
+			return nil, status.Errorf(codes.ResourceExhausted, "Attachment limit exceeded for volume %q on node %q: %v", volumeID, nodeID, err)
+		}
 		return nil, status.Errorf(codes.Internal, "Could not attach volume %q to node %q: %v", volumeID, nodeID, err)
 	}
 	klog.InfoS("ControllerPublishVolume: attached", "volumeID", volumeID, "nodeID", nodeID, "devicePath", devicePath)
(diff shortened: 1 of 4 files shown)


ElijahQuinones (approved):
/lgtm

pkg/driver/controller.go
```diff
@@ -422,6 +422,9 @@ func (d *ControllerService) ControllerPublishVolume(ctx context.Context, req *cs
 			klog.InfoS("ControllerPublishVolume: volume not found", "volumeID", volumeID, "nodeID", nodeID)
 			return nil, status.Errorf(codes.NotFound, "Volume %q not found", volumeID)
 		}
+		if errors.Is(err, cloud.ErrAttachmentLimitExceeded) {
+			return nil, status.Errorf(codes.ResourceExhausted, "Attachment limit exceeded for volume %q on node %q: %v", volumeID, nodeID, err)
+		}
```
AndrewSirenko, line 427: I see that for other error cases or happy cases we log a message. (e.g. ErrNotFound case above)

Is there a reason we don't want to log a message in this case? Something like "ControllerPublishVolume: Attachment limit exceeded for volume %q on node %q: %v"?

AndrewSirenko (approved):
Left one question, but lgtm either way. Thanks!

pkg/driver/controller.go
AndrewSirenko: Thanks for the answer! Made a PR removing that duplicate log: [Remove duplicate ErrNotFound log](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2505)

PR #2311, opened 2025-01-24
Fix node.selinux configuration parameter & release Helm Chart v2.39.1

AndrewSirenko:
#### What type of PR is this?

/kind bug

#### What is this PR about? / Why do we need it?

Fix node.selinux configuration parameter because Pod spec.volumes.hostpath does not have a `readOnly`. That belongs in pod.spec.containers.volumeMounts.

When you install via helm this readOnly is dropped with a warning. However this breaks add-on. 

```
❯ helm upgrade --install aws-ebs-csi-driver \
    --namespace kube-system --set "node.selinux=true" \
    aws-ebs-csi-driver/aws-ebs-csi-driver
Release "aws-ebs-csi-driver" does not exist. Installing it now.
W0124 19:27:54.986662   29648 warnings.go:70] unknown field "spec.template.spec.volumes[4].hostPath.readOnly"
W0124 19:27:54.986683   29648 warnings.go:70] unknown field "spec.template.spec.volumes[5].hostPath.readOnly"
```

#### How was this change tested?

```
❯ helm template aws-ebs-csi-driver ./charts/aws-ebs-csi-driver > pre.yml
helm template aws-ebs-csi-driver ./charts/aws-ebs-csi-driver --set node.selinux=true > post.yml

diff pre.yml post.yml -C 5
*** pre.yml	2025-01-24 21:04:48.509855690 +0000
--- post.yml	2025-01-24 21:04:48.599854884 +0000
***************
*** 480,489 ****
--- 480,495 ----
                mountPropagation: "Bidirectional"
              - name: plugin-dir
                mountPath: /csi
              - name: device-dir
                mountPath: /dev
+             - name: selinux-sysfs
+               mountPath: /sys/fs/selinux
+               readOnly: true
+             - name: selinux-config
+               mountPath: /etc/selinux/config
+               readOnly: true
            ports:
              - name: healthz
                containerPort: 9808
                protocol: TCP
            livenessProbe:
***************
*** 576,585 ****
--- 582,599 ----
              type: Directory
          - name: device-dir
            hostPath:
              path: /dev
              type: Directory
+         - name: selinux-sysfs
+           hostPath:
+             path: /sys/fs/selinux
+             type: Directory
+         - name: selinux-config
+           hostPath:
+             path: /etc/selinux/config
+             type: File
          - name: probe-dir
            emptyDir: {}
  ---
  # Source: aws-ebs-csi-driver/templates/controller.yaml
  # Controller Service
```

```
[...]

Issue #2228, opened 2024-11-14
Add support for SELinux mount options 


gbucknel:
**Is your feature request related to a problem? Please describe.**
I'd like to run Jenkins on Bottlerocket. Jenkins uses an EBS volume and has millions of small files. 
It runs as a stateful set. On Bottlerocket (which is selinux enabled ) when the pod moves to a new node it needs to relabel every single file and so the pod can stay down for half an hour while the relabelling runs. 

This selinux relabelling feature out last year seems to address this : 

https://kubernetes.io/blog/2023/04/18/kubernetes-1-27-efficient-selinux-relabeling-beta/

The crux of it is that node will mount the volume with the same set of configured selinux labels and will then not need to  relabel anything, and no downtime will happen when moving to a new node. 

From the blog post, 5 different conditions need to be met for this to work and the 5th is : 

_The volume plugin or the CSI driver responsible for the volume supports mounting with SELinux mount options._

I understand this isn't the case, given this closed issue -> https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1544 . Am opening this issue to start the conversation again. 

**Describe the solution you'd like in detail**
Please add selinux support to this driver. 

**Describe alternatives you've considered**
I currently run Jenkins on Amazon Linux 2023 which has selinux disabled . Would be keen to get it to run on bottlerocket for uniformity (and the other advantages of bottlerocket).

ConnorJC3:
/type feature

Hi, thanks for this feature request. We haven't explicitly tested it, but as a manual workaround you likely could use the feature if you meet all the other requirements and manually add `seLinuxMount: true` to our `CSIDriver` object.

I'll keep this open as a feature request to officially test/support this - I can't provide a specific ETA at this time but the team will provide any updates on this issue.

kenny-monster:
Hey @ConnorJC3. I'm one of @gbucknel's colleagues.

I had a go at your suggestion of updating the object (`kubectl edit csidriver ebs.csi.aws.com`) so that `seLinuxMount: true` is set. I then drained the Amazon Linux node the Jenkins instance was sitting on and made sure that it scheduled on a Bottlerocket node. The pod exhibited the slow startup times as we'd observed in the past when we determined it was an issue with SELinux relabelling.

ConnorJC3:
[...]

PR #2507, opened 2025-05-29
Set seLinuxMount on CSIDriver when appropriate

paulgrocholske:
#### What type of PR is this?
/kind bug

In conjunction with https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2253 and https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2311, addresses https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2228

Currently the `.node.selinux` configuration value mounts directories which are required for selinux-aware mounting. However, it does not set the `seLinuxMount` field in the CSIDriver spec which is required for Kubernetes to pass SELinux content mount options to the driver. This PR fixes that.

```release-note
NONE
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-ebs-csi-driver/templates/csidriver.yaml
@@ -11,4 +11,7 @@ spec:
   {{- if not .Values.useOldCSIDriver }}
   fsGroupPolicy: File
   {{- end }}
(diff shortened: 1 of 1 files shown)


AndrewSirenko:
Thanks for the PR! 

/ok-to-test

AndrewSirenko (approved):
/approve

PR #2511, opened 2025-06-04
Locked NodeExpandVolume by Volume ID

mdzraf:
#### What type of PR is this?

/kind feature

This PR ensures that if NodeExpandVolume is called more than once on the same volume while a previous expand request is in progress, only the current in-flight call succeeds and there are no attempts to perform multiple resizes on the same volume. This is done by locking NodeExpandVolume based on the volume ID. 

Added a unit test that test for this specific mechanism working on NodeExpandVolume tests. 

```release-note
Added Lock To Prevent Multiple Resize Attempts On The Same Volume If An In-flight Resize Request Already Exists.
```


ConnorJC3 (approved):
/lgtm

PR #2516, opened 2025-06-11
Fix dependency upgrades

AndrewSirenko:
#### What type of PR is this?

/kind cleanup

/area dependencies

See #2515 for list of dependencies bumped. 

Supercedes #2515

Need to also pin tests/go.mod grpc/otelgrpc to v0.58.0 due to breaking change it causes in other dep k8s.io/cri-client.

Also splits dependabot PRs for main go.mod versus tests/e2e/go.mod so we can catch what breaks more easily in future. 

```release-note
NONE
```


.github/dependabot.yaml
```diff
@@ -41,6 +41,19 @@ updates:
...
+      - dependency-type: "all"
+    schedule:
+      interval: weekly
+      day: "wednesday"
+      time: "06:00"
+      timezone: "America/New_York"
+    labels:
+      - "area/dependency"
+      - "ok-to-test"
+      - "release-note-none"
```
ConnorJC3: I don't believe these can be safely split because `tests/e2e` depends on the main package, which means if the main package is updated `tests/e2e` will need a go.mod/go.sum update (and thus break the build if that isn't done in the same PR)

tests/e2e/go.mod
```diff
@@ -155,6 +155,7 @@ require (
 replace (
 	github.com/containerd/errdefs => github.com/containerd/errdefs v1.0.0 // Mitigate errdefs breaking change (Remove after all depdendencies upgrade to v1.0.0)
 	github.com/kubernetes-sigs/aws-ebs-csi-driver => ../../
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc => go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.58.0 // TODO: Remove once k8s.io/cri-client includes v0.61+ of this dependancy
```
ConnorJC3: The problem wasn't/isn't which dependency is being used, it's the removed methods, so the TODO is inaccurate

Suggest wording along the lines of
```
TODO: Remove after https://github.com/kubernetes/kubernetes/pull/131951 is released in k8s.io/cri-client
```

PR #2517, opened 2025-06-13
Add `--metadata-sources` parameter to opt-out or choose fallback ordering of metadata sources

### LINKED ISSUE #2262: Occasional retrieving IMDS metadata failed on AL2023
/kind bug

We currently have AL2 nodes and have never had a problem with this.

When switching to AL2023 nodes, occasionally the ebs-csi-node will fail to retrieve metadata from IMDS. This only appears to happen at node startup time, if we restart the ebs-csi-node daemonset, it is able to retrieve metadata from IMDS reliably.

It does appear to successfully fallback to getting metadata from Kubernetes, but we think IMDS should not be failing like this.

**What happened?**
```
I1211 20:07:09.634316       1 main.go:157] "Initializing metadata"
E1211 20:07:14.635517       1 metadata.go:51] "Retrieving IMDS metadata failed, falling back to Kubernetes metadata" err="could not get EC2 instance identity metadata: operation error ec2imds: GetInstanceIdentityDocument, request canceled, context deadline exceeded"
I1211 20:07:14.645753       1 metadata.go:55] "Retrieved metadata from Kubernetes"
I1211 20:07:14.646110       1 driver.go:69] "Driver Information" Driver="ebs.csi.aws.com" Version="v1.34.0"
I1211 20:07:16.167040       1 node.go:941] "CSINode Allocatable value is set" nodeName="ip-100-64-153-121.ec2.internal" count=31
```

```
I1211 20:24:41.226237       1 main.go:157] "Initializing metadata"
I1211 20:24:42.479940       1 metadata.go:48] "Retrieved metadata from IMDS"
I1211 20:24:42.480783       1 driver.go:69] "Driver Information" Driver="ebs.csi.aws.com" Version="v1.34.0"
I1211 20:24:43.497952       1 node.go:941] "CSINode Allocatable value is set" nodeName="ip-100-64-251-153.ec2.internal" count=31
```

Our launch template looks like:
```
  NodeLaunchTemplate2023:
    Type: AWS::EC2::LaunchTemplate
    Condition: CreateManagedNodegroup2023
    DependsOn:
    - Cluster
    Properties:
      LaunchTemplateData:
        BlockDeviceMappings:
        - DeviceName: /dev/xvda
          Ebs:
            DeleteOnTermination: true
            Encrypted: true
            VolumeSize: !Ref WorkerVolumeSize
            VolumeType: gp3
        MetadataOptions:
          HttpEndpoint: enabled
          HttpPutResponseHopLimit: 2
          HttpTokens: required
          InstanceMetadataTags: disabled
        NetworkInterfaces:
        - DeviceIndex: 0
          Groups:
          - !GetAtt Cluster.ClusterSecurityGroupId
```

And our managed nodegroup looks like:
```
  ManagedNodegroup2023a:
    Type: AWS::EKS::Nodegroup
    Condition: CreateManagedNodegroup2023
    DependsOn:
    - Cluster
    - NodeInstanceRole
    - NodeLaunchTemplate2023
    Properties:
      AmiType: AL2023_x86_64_STANDARD
      CapacityType: ON_DEMAND
      ClusterName: !Ref Cluster
      InstanceTypes:
      - !Ref WorkerInstanceType
      LaunchTemplate:
        Id: !Ref NodeLaunchTemplate2023
        Version: !GetAtt NodeLaunchTemplate2023.LatestVersionNumber
      NodeRole: !GetAtt NodeInstanceRole.Arn
      ScalingConfig:
        DesiredSize: !Ref NodegroupSizeDesired
        MaxSize: !Ref NodegroupSizeMaximum
        MinSize: !Ref NodegroupSizeMinimum
      Subnets:
      - Fn::ImportValue:
          !Sub "${VpcName}-private-a"
      UpdateConfig:
        MaxUnavailable: 1
```
**Environment**
- Kubernetes version (use `kubectl version`): v1.30.6-eks-7f9249a
- Driver version: v1.34.0

AndrewSirenko:
#### What type of PR is this?

/kind feature

Dictates which sources are used to retrieve instance metadata. The driver will attempt to rely on each source order until one succeeds. 

Fixes #2262 if you override value to 'imds'

Furthermore, faster node readiness for clusters that disable pod access to IMDS if one sets value to 'kubernetes' (because no need to wait 10s for IMDS calls to timeout) 

This PR also renames anything that refers EC2Metadata as IMDS in the codebase. 

Manually:
- Default case
- the invalid source case
- the case where I disable IMDS for certain nodes so that we see fallback  to k8s metadata working, 
- case where I set value to just 'imds' and disable IMDS so that I see pods continously restarting (instead of using k8s metadata)
- Case where I set value to just 'kubernetes' and check that IMDS was skipped

Helm: 

```
❯ helm template aws-ebs-csi-driver ./charts/aws-ebs-csi-driver --set node.metadataSources="" | grep "metadata-s" -C 5
NOTHING
❯ helm template aws-ebs-csi-driver ./charts/aws-ebs-csi-driver --set node.metadataSources="imds" | grep "metadata-s" -C 5
          image: public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver:v1.44.0
          imagePullPolicy: IfNotPresent
          args:
            - node
            - --endpoint=$(CSI_ENDPOINT)
            - --metadata-sources=imds
            - --logging-format=text
            - --v=2
          env:
            - name: CSI_ENDPOINT
              value: unix:/csi/csi.sock
--
          imagePullPolicy: IfNotPresent
          args:
            - node
            - --endpoint=$(CSI_ENDPOINT)
            - --csi-mount-point-prefix=/var/lib/kubelet/plugins/kubernetes.io/csi/ebs.csi.aws.com/
            - --metadata-sources=imds
            - --logging-format=text
            - --v=2
          env:
            - name: CSI_ENDPOINT
              value: unix:/csi/csi.sock

```

```release-note
Add `--metadata-sources` which dictates which sources are used to retrieve instance metadata. The driver will attempt to rely on each source order until one succeeds. Valid options currently include 'imds' and 'kubernetes'.
```


charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -406,6 +406,7 @@ node:
   # Enable the linux daemonset creation
   enableLinux: true
   enableWindows: true
+  metadataSources: "imds,kubernetes"
```
ConnorJC3: Does it instead make sense to default this to unset, and use the driver default? Because otherwise we have two levels of defaults

pkg/cloud/metadata/imds.go
```diff
@@ -50,10 +50,10 @@ var DefaultEC2MetadataClient = func() (EC2Metadata, error) {
 	return svc, nil
 }
 
-func EC2MetadataInstanceInfo(svc EC2Metadata) (*Metadata, error) {
+func IMDSInstanceInfo(svc IMDS) (*Metadata, error) {
 	docOutput, err := svc.GetInstanceIdentityDocument(context.Background(), &imds.GetInstanceIdentityDocumentInput{})
 	if err != nil {
-		return nil, fmt.Errorf("could not get EC2 instance identity metadata: %w", err)
+		return nil, fmt.Errorf("could not get IMDS identity metadata: %w", err)
```
ConnorJC3: NP: remove "identity" (i.e. should be "could not get IMDS metadata")

charts/aws-ebs-csi-driver/values.yaml
AndrewSirenko: I was 50/50 on this. I was originally thinking it's easier for customers to understand the default is "imds,kubernetes" when looking at chart and driver pod args.

But I think you're right, especially to avoid weirdness if/when we ever change this default. 

Will change next rev.

pkg/driver/options.go
```diff
@@ -104,6 +110,7 @@ func (o *Options) AddFlags(f *flag.FlagSet) {
...
 	f.StringVar(&o.MetricsKeyFile, "metrics-key-file", "", "The path to a key to use for serving the metrics server over HTTPS. If this is non-empty, --http-endpoint and --metrics-cert-file MUST also be non-empty.")
 	f.BoolVar(&o.EnableOtelTracing, "enable-otel-tracing", false, "To enable opentelemetry tracing for the driver. The tracing is disabled by default. Configure the exporter endpoint with OTEL_EXPORTER_OTLP_ENDPOINT and other env variables, see https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration.")
+	f.StringVar(&o.MetadataSources, "metadata-sources", "imds,kubernetes", "Dictates which sources are used to retrieve instance metadata. The driver will attempt to rely on each source in order until one succeeds. Valid options include 'imds' and 'kubernetes'.")
```
torredil: Consider using a [StringSliceVar](https://github.com/spf13/pflag/blob/5ca813443bd2a4d9f46a253ea0407d23b3790713/string_slice.go#L120) instead. This way, users can do either of these:

```
-metadata-sources=imds,kubernetes
-metadata-sources=imds --metadata-sources=kubernetes
```

with the second method you don't need to worry about quoting commas in bash or when using --set in Helm.

pkg/driver/options.go
```diff
@@ -137,11 +144,17 @@ func (o *Options) Validate() error {
...
+			return errors.New("--metrics-cert-file MUST be specified when using the metrics server with HTTPS")
 		case o.MetricsKeyFile == "":
-			return errors.New("--metrics-key-file MUST be specififed when using the metrics server with HTTPS")
+			return errors.New("--metrics-key-file MUST be specified when using the metrics server with HTTPS")
+		}
+	}
+
+	for _, source := range strings.Split(o.MetadataSources, ",") {
+		if source != metadata.SourceIMDS && source != metadata.SourceK8s {
+			return metadata.InvalidSourceErr(strings.Split(o.MetadataSources, ","), source)
```
torredil: If we use a StringSliceVar, it would be ideal to normalize and validate once here:

```
for i, s := range o.MetadataSources {
    s = strings.ToLower(strings.TrimSpace(s))
    switch s {
    case metadata.SourceIMDS, metadata.SourceK8s:
        o.MetadataSources[i] = s
    default:
        return metadata.InvalidSourceErr(o.MetadataSources, s)
    }
}

```

passing the slice downstream makes it so that `strings.Split` isn't needed in `main.go` again.

torredil (commented):
Largely lgtm, left some suggestions.

pkg/driver/options.go
AndrewSirenko: So I was also thinking about StringSliceVar originally, but ended up going with the StringVar because the helm chart and CLI UX felt clearer:

```
  metadataSources:
    - "imds"
    - "kubernetes"
```

Versus `metadataSources:'imds,kubernetes'`

Discussed before I put the PR out with @ConnorJC3 and we originally decided to go with regular StringVar instead.  

Also given that customers will be mostly overwriting the default with a single value (either imds or kubernetes), the quoting isn't too much of a papercut IMO.

But I was pretty 45/55 on this decision, so alright with switching if there is consensus from team.

torredil: I agree that

```
metadataSources:
    - "imds"
    - "kubernetes"
```

is a better UX.

`StringSliceVar` is more idiomatic, flexible, and cleaner. @ConnorJC3 Is there a strong reason why `StringVar` is preferable instead?

AndrewSirenko: Eddie mentioned offline stringSliceVar accepts both

```
// For example:
//   --ss="v1,v2" --ss="v3"
// will result in
//   []string{"v1", "v2", "v3"}
```

Will switch back to StringSliceVar. Thank you!

AndrewSirenko:
/hold

Final manual testing.

AndrewSirenko:
/unhold

Manually retested

AndrewSirenko:
/approve

If documentation needs edits we can do that in follow up pr

PR #2519, opened 2025-06-16
Use correct name for nvme_collector_duration_seconds

yuvipanda:
This is the actual name of the metric, as the namespace  `aws_ebs_csi_` is hardcoded in the code. Matches the other examples here too.

#### What type of PR is this?

/kind documentation

The example names for the node metrics were missing the prefix, making it slightly difficult
to validate that metrics collection is enabled.

I verified that these metrics *are* available in my installation.

```release-note
NONE
```

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -65,20 +65,20 @@ The EBS CSI Driver will emit [container storage Interface managed devices metric
 
 The metrics will appear in the following format: 
 ```sh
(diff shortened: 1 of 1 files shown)


torredil (approved):
Thank you for the contribution @yuvipanda : )

/ok-to-test
/lgtm

yuvipanda:
ty for the merge @torredil, and ty to everyone working on this software! Having this be in prometheus (rather than in cloudwatch) helps us connect with other metrics we care about much more easily - so thank you to whoever added these metrics in the first place.

PR #2534, opened 2025-06-20
Fast-fail e2e tests and allow 2 retries when test flakes

AndrewSirenko:
#### What type of PR is this?

/kind cleanup

ginkgo --flake-attempt=3 will allow a test to be retried twice when failing
ginkgo --fail-fast interrupts all test processes when a final failing test occurs 

By failing fast we increase frugality. 

Open to just allowing flake-attempt for windows tests or upstream kubernetes tests. 

```release-note
NONE
```

### DIFF
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -145,6 +145,7 @@ else
       --focus-regex="${GINKGO_FOCUS}" \
       --test-package-version=$(curl -L https://dl.k8s.io/release/stable-${packageVersion}.txt) \
       --parallel=${GINKGO_PARALLEL} \
+      --ginkgo-args="--flake-attempts=3 --fail-fast" \
       --test-args="-storage.testdriver=${PWD}/manifests.yaml -kubeconfig=${KUBECONFIG} -node-os-distro=${NODE_OS_DISTRO}"
     TEST_PASSED=$?
     set -e


hack/e2e/run.sh
```diff
@@ -145,6 +145,8 @@ else
       --focus-regex="${GINKGO_FOCUS}" \
       --test-package-version=$(curl -L https://dl.k8s.io/release/stable-${packageVersion}.txt) \
       --parallel=${GINKGO_PARALLEL} \
+      --flake-attempts=1 \
```
AndrewSirenko: Open to setting this with a GINKGO_FLAKE_ATTEMPT env var instead 

Also open to just using this on windows tests (but i've seen some other flaky tests on upstream ones as well)

AndrewSirenko:
CI Failing despite tests passing. Probably need to change our hack scripts a bit

AndrewSirenko:
/hold

Need to check few more things

ElijahQuinones (approved):
/lgtm

hack/e2e/run.sh
```diff
@@ -145,6 +145,7 @@ else
       --focus-regex="${GINKGO_FOCUS}" \
       --test-package-version=$(curl -L https://dl.k8s.io/release/stable-${packageVersion}.txt) \
       --parallel=${GINKGO_PARALLEL} \
+      --ginkgo-args="--flake-attempts=3 --fail-fast" \
```
ConnorJC3, line 148: Do we want to fast fail CI? This seems like a dangerous idea - we essentially won't know all of the tests that fail for a given run.

Second thing: Probably should add at least the `--flake-attempts` down around line 160 too

AndrewSirenko: > we essentially won't know all of the tests that fail for a given run.

Fair. Will remove. It is convenient to see which tests failed in CI instead of re-evaluating locally.

AndrewSirenko:
/close

Manu advocated for tasks to look into upstream e2e tests flaking instead.

PR #2539, opened 2025-06-25
Remove completed TODO comment

mdzraf:
#### What type of PR is this?

/kind cleanup

This PR removes the TODO comment to check if the device is attached before releasing as that is already being done in this function, here we defer the release: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/af089444c5b7e17907cef26415febc511b7699aa/pkg/cloud/cloud.go#L970 

Right before the function returns we check that the volume is in an detached state and if it isn't we log it:  https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/af089444c5b7e17907cef26415febc511b7699aa/pkg/cloud/cloud.go#L994-L1001

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -962,7 +962,6 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
 		return err
 	}
 
-	// TODO: check if attached
 	device, err := c.dm.GetDevice(instance, volumeID)
 	if err != nil {
 		return err


ElijahQuinones (commented):
/lgtm 

We also check if it is attached already on 971

```
if !device.IsAlreadyAssigned {
		klog.InfoS("DetachDisk: called on non-attached volume", "volumeID", volumeID)
	}
```

torredil (approved):
/lgtm
/approve

PR #2544, opened 2025-07-01
[nit] 

SamMayWork:
#### What type of PR is this?

/kind bug

It looks like the original intent was the check the errors coming out the invoked functions, but the code is actually checking the error from the previous call that got us into the else branch.

`make test` - visually the code looks to be wrong, and doing some quick grep through the issues and PRs on this repo, I couldn't find that this fix has been suggested before

```release-note
NONE
```


SamMayWork:
(fyi the company I work for is currently going through the organization CLA process, pushing to get the CLA signed but need to get some boxes checked)

mdzraf:
/lgtm

Thank you for the fix!

ConnorJC3 (approved):
/lgtm

As you know, this can't be merged until the CLA is signed because of Kubernetes/CNCF requirements, but other than that it looks ready to me

SamMayWork:
@ConnorJC3 (and everyone else) - Thanks for the reviews here, CLA is signed by the org, I think this is good to be merged?

torredil (approved):
Thank you for the contribution @SamMayWork 
/approve

PR #2545, opened 2025-07-01
Return Resource Exhausted When Exceeding Quotas

mdzraf:
#### What type of PR is this?

/kind feature

This PR ensures that the RESOURCE_EXHAUSTED gRPC error code is returned when the user exceeds their AWS quotas for Storage, IOPS, and/or number of snapshots. 

Ran `make update && make verify`, `make e2e/external`, `make test`, as well as a manual test results [here](https://gist.github.com/mdzraf/819e00c53b43fe09df7e72efc4fe6527)

```release-note
Return RESOURCE_EXHAUSTED gRPC error code when AWS quotas are exceeded
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -172,8 +172,8 @@ var (
 	// ErrInvalidRequest is returned if parameters were rejected by driver.
 	ErrInvalidRequest = errors.New("invalid request")
 
-	// ErrAttachmentLimitExceeded is returned if the attachment limit is exceeded.
-	ErrAttachmentLimitExceeded = errors.New("attachment limit exceeded")
+	// ErrLimitExceeded is returned if a user exceeds a quota.
+	ErrLimitExceeded = errors.New("limit exceeded")
 )
 
 // Set during build time via -ldflags.
@@ -664,6 +664,12 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 			c.latestClientTokens.Set(volumeName, &nextTokenNumber)
 			return nil, ErrIdempotentParameterMismatch
 		}
+		if isAWSErrorVolumeLimitExceeded(err) {
+			return nil, fmt.Errorf("%w: %w", ErrLimitExceeded, err)
+		}
+		if isAwsErrorMaxIOPSLimitExceeded(err) {
+			return nil, fmt.Errorf("%w: %w", ErrLimitExceeded, err)
+		}
 		return nil, fmt.Errorf("could not create volume in EC2: %w", err)
 	}
 
@@ -813,6 +819,12 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
 			// Wrap error to preserve original message from AWS as to why this was an invalid argument
 			return 0, fmt.Errorf("%w: %w", ErrInvalidArgument, err)
 		}
+		if isAWSErrorVolumeModificationSizeLimitExceeded(err) {
+			return 0, fmt.Errorf("%w: %w", ErrLimitExceeded, err)
+		}
+		if isAwsErrorMaxIOPSLimitExceeded(err) {
+			return 0, fmt.Errorf("%w: %w", ErrLimitExceeded, err)
+		}
 		return 0, err
 	}
 	// If the volume modification isn't immediately completed, wait for it to finish
@@ -934,7 +946,7 @@ func (c *cloud) AttachDisk(ctx context.Context, volumeID, nodeID string) (string
 				likelyBadDeviceNames.Store(device.Path, struct{}{})
 			}
 			if isAWSErrorAttachmentLimitExceeded(attachErr) {
-				return "", fmt.Errorf("%w: %w", ErrAttachmentLimitExceeded, attachErr)
+				return "", fmt.Errorf("%w: %w", ErrLimitExceeded, attachErr)
 			}
 			return "", fmt.Errorf("could not attach volume %q to node %q: %w", volumeID, nodeID, attachErr)
 		}
@@ -1283,6 +1295,9 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
 		o.Retryer = c.rm.createSnapshotRetryer
 	})
 	if err != nil {
+		if isAwsErrorSnapshotLimitExceeded(err) {
+			return nil, fmt.Errorf("%w: %w", ErrLimitExceeded, err)
+		}
 		return nil, fmt.Errorf("error creating snapshot of volume %s: %w", volumeID, err)
 	}
 	if res == nil {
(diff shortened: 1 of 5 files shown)


pkg/cloud/cloud_test.go
```diff
@@ -1366,6 +1366,30 @@ func TestCreateDisk(t *testing.T) {
...
+			volumeName: "vol-test-name",
+			diskOptions: &DiskOptions{
+				CapacityBytes: util.GiBToBytes(1),
+				Tags:          map[string]string{VolumeNameTagKey: "vol-test", AwsEbsDriverTagKey: "true"},
+			},
+			expDisk:              nil,
+			expCreateVolumeInput: &ec2.CreateVolumeInput{},
+			expCreateVolumeErr:   errors.New("MaxIOPSLimitExceeded"),
+			expErr:               fmt.Errorf("could not create volume in EC2: %w", errors.New("MaxIOPSLimitExceeded")),
+		},
```
mdzraf, line 1392: Put these here just in case but since there is already a test to check that a specific error is returned properly, not sure if it's worth keeping here, thoughts?

pkg/cloud/cloud.go
```diff
@@ -174,6 +174,18 @@ var (
...
+	ErrVolumeModificationSizeLimitExceeded = errors.New("modification size limit exceeded")
+
+	// ErrVolumeLimit is returned if the volume storage limit is exceeded.
+	ErrVolumeLimitExceeded = errors.New("volume limit exceeded")
+
+	// ErrMaxIOPSLimitExceeded is returned if the limit of IOPS usage is exceeded.
+	ErrMaxIOPSLimitExceeded = errors.New("max IOPS limit exceeded")
+
+	// ErrSnapshotLimitExceeded is returned if the limit of snapshots that can be created is exceeded.
+	ErrSnapshotLimitExceeded = errors.New("snapshot limit exceeded")
```
ConnorJC3: Given we're wrapping the EC2 error anyways, so the user can directly see the text from EC2, it would probably make more sense to have a generic `ErrLimitExceeded` rather than one for each type of EC2 limit exceeded errors

pkg/driver/controller_modify_volume.go
```diff
@@ -149,12 +149,16 @@ func executeModifyVolumeRequest(c cloud.Cloud) func(string, modifyVolumeRequest)
...
 			if err != nil {
-				if errors.Is(err, cloud.ErrInvalidArgument) {
+				switch {
+				case errors.Is(err, cloud.ErrInvalidArgument):
 					return 0, status.Errorf(codes.InvalidArgument, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
-				} else if errors.Is(err, cloud.ErrNotFound) {
+				case errors.Is(err, cloud.ErrNotFound):
 					return 0, status.Errorf(codes.NotFound, "Could not modify volume (not found) %q: %v", volumeID, err)
+				case errors.Is(err, cloud.ErrMaxIOPSLimitExceeded), errors.Is(err, cloud.ErrVolumeLimitExceeded), errors.Is(err, cloud.ErrVolumeModificationSizeLimitExceeded):
+					return 0, status.Errorf(codes.ResourceExhausted, "Could not modify volume (resource exhausted) %q: %v", volumeID, err)
```
ConnorJC3, line 158: Is `ResourceExhausted` actually a legal return value for modify? It's not in https://github.com/container-storage-interface/spec/blob/master/spec.md#controllermodifyvolume-errors

pkg/cloud/cloud.go
mdzraf: That sounds like a better way to do it, fixed in next rev.

pkg/driver/controller_modify_volume.go
mdzraf: You're right, I see it is not there, in that case there are two options:

1.) Leave as is even though it's not a valid return value for modify.

2.) Change it to return `Internal` and have a message `Could not modify volume (resource exhausted)` plus the wrapped EC2 error.

We must still fail the call somehow. 

Reviewers: Any thoughts on this ?

torredil: Its odd that the Modify Volume section in the CSI Spec does not include similar wording to the below:

> If the plugin is unable to complete the CreateSnapshot call successfully, it MUST return a non-ok gRPC code in the gRPC status. If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code. The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.

I understand this to mean that the plugin **can** return any non-ok gRPC, but that it **must** return the ones that are explicitly defined when those conditions are met. In other words, I think returning `ResourceExhausted` here is fine.

mdzraf: Thanks for that information Eddie, knowing that, I also agree that keeping it as is, is fine.

torredil (approved):
/lgtm

pkg/cloud/cloud.go
```diff
@@ -172,8 +172,7 @@ var (
 	// ErrInvalidRequest is returned if parameters were rejected by driver.
 	ErrInvalidRequest = errors.New("invalid request")
 
-	// ErrAttachmentLimitExceeded is returned if the attachment limit is exceeded.
-	ErrAttachmentLimitExceeded = errors.New("attachment limit exceeded")
+	ErrLimitExceeded = errors.New("limit exceeded")
```
ElijahQuinones, line 176: You removed the old error with the comment above that says when the error occurs (as all other errors here have) but you did not add a new comment for when this error occurs.

pkg/cloud/cloud.go
```diff
@@ -1673,6 +1687,30 @@ func isAWSErrorAttachmentLimitExceeded(err error) bool {
...
+}
+
+// isAwsErrorMaxIOPSLimitExceeded checks if the error is a MaxIOPSLimitExceeded error.
+// This error is reported when the limit on the IOPS usage for a region is exceeded.
+func isAwsErrorMaxIOPSLimitExceeded(err error) bool {
+	return isAWSError(err, "MaxIOPSLimitExceeded")
+}
+
+// isAwsErrorSnapshotLimitExceeded checks if the error is a SnapshotLimitExceeded error.
+// This error is reported when the limit on the number of snapshots that can be created is exceeded.
```
ElijahQuinones, line 1710: Very very tiny nit: Here for some of the errors we say "for a region" and others we do not. To me this indicates that the limits for say snapshots (where the "for a region") is absent seems to indicate that these limits are per account instead of per region.

mdzraf: Discussed Offline: Decided to keep as is because it is matching EC2 docs

pkg/cloud/cloud.go
mdzraf: Nice catch, fixed ✅

Rylon:
Quick question about this feature, as I understand it, we need the feature gate `MutableCSINodeAllocatableCount` which is Alpha in v1.33, but EKS does not support Alpha feature flags. If we install this CSI driver update in the meantime, will we at least benefit from this error being raised when the EBS volume attachment limit is reached, so the scheduler will try a different node?

ConnorJC3:
@Rylon this code is not about attachment limits, it is about create-time limits, so the mutable CSINode feature is not relevant.

This PR will cause the scheduler to re-select a different node if the create fails due to an EBS limit. However, most EBS limits are regional, not zonal, so re-selecting a new node is unlikely to make a meaningful difference unless you are hitting a zonal EBS limit.

Rylon:
@ConnorJC3 thanks for that, in our case it's when the attachment limit is reached on a specific node, so not a zonal/regional quota limit.

PR #2550, opened 2025-07-03
Replace provenance script with `BUILDX_NO_DEFAULT_ATTESTATIONS` env var

torredil:
#### What type of PR is this?
/kind bug

Our CI is currently failing because the buildx provenance script we maintain to disable attestation is failing (the client message the scrip checks for has been updated).

Instead of using a script to detect and disable buildx provenance, lets use the officially supported `BUILDX_NO_DEFAULT_ATTESTATIONS` env var.

This environment variable was introduced in Docker Buildx 0.10.4 specifically for this purpose and is the recommended way to disable default provenance attestations, see:

- https://docs.docker.com/build/release-notes/#0104
- https://github.com/docker/buildx/pull/1645

```release-note
None
```


ElijahQuinones (approved):
/lgtm
/approve

PR #2553, opened 2025-07-07
Adding Bottlerocket CI Test

mdzraf:
#### What type of PR is this?

/kind feature

This PR adds the necessary changes to the repo in order to be able to run Bottlerocket e2e tests

Locally by running `make test-e2e-external-bottlerocket-eks` 

```release-note
NONE
```


Makefile
```diff
@@ -55,6 +55,7 @@ CLUSTER_NAME?=ebs-csi-e2e.k8s.local
 CLUSTER_TYPE?=kops
 
 GINKGO_WINDOWS_SKIP?="\[Disruptive\]|\[Serial\]|\[Flaky\]|\[LinuxOnly\]|\[Feature:VolumeSnapshotDataSource\]|\(xfs\)|\(ext4\)|\(block volmode\)"
+GINKGO_BOTTLEROCKET_SKIP?="\[Disruptive\]|\[Serial\]|\[Flaky\]|should not mount / map unused volumes in a pod \[LinuxOnly\]"
```
mdzraf, line 58: Tests use nsenter which is being blocked by selinux used in bottlerocket.

mdzraf:
Looks like a flake /retest

hack/e2e/eksctl/eksctl.sh
```diff
@@ -43,6 +43,7 @@ function eksctl_create_cluster() {
     K8S_VERSION="${K8S_VERSION}" \
     ZONES="${ZONES}" \
     INSTANCE_TYPE="${INSTANCE_TYPE}" \
+    BOTTLEROCKET="${BOTTLEROCKET:-false}" \
     WINDOWS="${WINDOWS}" \
```
torredil: Suggest setting the default value in `hack/e2e/config.sh` for consistency:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ba7a117a614d1c61a2a49ab72cae3e1626f30dcb/hack/e2e/config.sh#L32

torredil (commented):
We should also update `docs/makefile.md` to reference the new env var, see example:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ba7a117a614d1c61a2a49ab72cae3e1626f30dcb/docs/makefile.md?plain=1#L59

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ba7a117a614d1c61a2a49ab72cae3e1626f30dcb/docs/makefile.md?plain=1#L92

hack/e2e/eksctl/cluster.yaml
```diff
@@ -30,7 +30,11 @@ iam:
         ebsCSIController: true
 managedNodeGroups:
   - name: ng-linux
+{{- if eq .Env.BOTTLEROCKET "true" }}
+    amiFamily: Bottlerocket 
+{{- else}}
```
ConnorJC3: Would it maybe be better to have something like an `AMI_FAMILY` environment variable to be flexible?

hack/prow-e2e.sh
```diff
@@ -64,6 +64,11 @@ test-e2e-external-eks-windows-fips)
   export CLUSTER_TYPE="eksctl"
   export WINDOWS="true"
   ;;
+test-e2e-external-eks-bottlerocket)
+  TEST="external-eks-bottlerocket"
+  export CLUSTER_TYPE="eksctl"
+  export BOTTLEROCKET="true"
+  ;;
```
ConnorJC3: np: move this after the a1 job so it's in the same relative place as the makefile and doesn't break up the windows tests in the middle

hack/e2e/eksctl/eksctl.sh
mdzraf: Done ✅  (now named AMI_FAMILY), set default to `AmazonLinux2023`

hack/e2e/eksctl/cluster.yaml
mdzraf: Makes sense, done ✅

hack/prow-e2e.sh
mdzraf: ✅

AndrewSirenko (approved):

mdzraf:
Got one `/lgtm` and one `Approve` 

/approve

PR #2564, opened 2025-07-14
Removing In-Driver Validation Already Done by EC2

mdzraf:
#### What type of PR is this?

/kind feature

This PR removes some of the pre-validation of parameters done in the driver's CreateVolume and ModifyVolume paths that is already done by EC2 the time of the call. It also deprecates the blockExpress parameter. 

Ran `make test`, CI e2e tests ~~(pending)~~, also manually tested.

```release-note
Deprecated `blockExpress` Parameter. This parameter no longer has any effect on the IOPS cap for io2 volumes. All io2 volumes will now have a cap of 256,000 IOPS.
```

# For Reviewers

This change will change the result of creating a resource with a storage class prior to this change when the IOPS were capped either at 64k or 256k depending on the `blockExpress` parameter, for example:

Using this PVC: 
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 65Gi
```

And this storage class: 
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
parameters: 
  iopsPerGB: "1000"
  type: io2
  ```
Today, this would result in a volume that is 64,000 IOPS because of the IOPS cap being set to 64,000 since the blockExpress param was not set to true. 

But after the change, the same configuration, would result in a 65,000 IOPS volume because of not having the non-BX and BX option for IOPS cap. 

This is something we already knew about at design but just wanted to point it out here <ins>to get some ideas on whether we should label this as a "breaking" change or not.<ins/>

### CODE REVIEW

File: pkg/driver/controller.go [hunk 1]
```diff
@@ -175,8 +174,8 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		case PVNameKey:
 			volumeTags[PVNameTag] = value
 			tProps.PVName = value
-		case BlockExpressKey:
-			blockExpress = isTrue(value)
+		case DeprecatedBlockExpressKey:
+			klog.InfoS("blockexpress key is deprecated, all io2 volumes have 256,000 IOPS cap and can be attached to all Nitro Based instances")
```
mdzraf: Possibly any better wording here?

AndrewSirenko: General comment: I wouldn't ever use the default [klog.InfoS verbosity](https://kubernetes.io/docs/concepts/cluster-administration/system-logs/#log-verbosity-level). 

Our codebase generally follows the following convention:
1: Critical issue
2: Something happened that everyone should know about
4: Detailed info about things happening
7: Noisy logs that in some world could be useful for debugging 

---

Specific comment: I think customers will complain that this makes their logs super noisy if they have many storage classes in production that refer to this parameter. Is there a world we only log first time we see this parameter?  @ConnorJC3 thoughts (I know ideally customer just updates all their SCs, but that might be scary for customers using that SC in production StatefulSets)?

mdzraf:
# For Reviewers

This change will change the result of creating a resource with a storage class prior to this change when the IOPS were capped either at 64k or 256k depending on the `blockExpress` parameter, for example:

Using this PVC: 
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 65Gi
```

And this storage class: 
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
parameters: 
  iopsPerGB: "1000"
  type: io2
  ```
Today, this would result in a volume that is 64,000 IOPS because of the IOPS cap being set to 64,000 since the blockExpress param was not set to true. 

But after the change, the same configuration, would result in a 65,000 IOPS volume because of not having the non-BX and BX option for IOPS cap. 

This is something we already knew about at design but just wanted to point it out here <ins>to get some ideas on whether we should label this as a "breaking" change or not.<ins/>

File: pkg/driver/controller.go (hunk 1 above)
Comment (mdzraf) [reply pkg/driver/controller.go]: Discussed offline: Decided to leave at current driver default verbosity which is 2.

Comment (ConnorJC3) [reply pkg/driver/controller.go]: Suggest wording:
```
"blockExpress key is deprecated and has no effect, all io2 volumes are now Block Express and share the same IOPS cap"
```

(I'm intentionally not referencing the number 256k or the Nitro system as they could change)

ConnorJC3: Discussed offline, recommending we follow precedent here of 2 or lower

ConnorJC3 (commented):
mostly lgtm two small things

pkg/driver/controller.go
mdzraf: Done ✅

AndrewSirenko:
/hold

Until release done

torredil (approved):
/lgtm
/approve

mdzraf:
/retest

Same flake...

PR #2565, opened 2025-07-15
Update dependencies

AndrewSirenko:

Upgrade deps

```release-note
NONE
```


pkg/cloud/volume_limits.go
```diff
@@ -32,7 +32,7 @@ const (
 func init() {
 	// This list of Nitro instance types have a dedicated Amazon EBS volume limit of up to 128 attachments, depending on instance size.
 	// The limit is not shared with other device attachments: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#nitro-system-limits
-	instanceFamilies := []string{"m8g", "m8gd", "m7i", "m7i-flex", "m7a", "c8g", "c8gd", "c7i", "c7i-flex", "c7a", "r7a", "r7i", "r7iz", "r8g", "r8gd", "x8g", "u7i", "u7inh", "g6", "g6e", "gr6", "i7i", "i7ie", "i8g", "p5", "p5e", "p5en", "u7i-6tb", "u7i-8tb", "u7i-12tb", "u7in-16tb", "u7in-24tb", "u7in-32tb", "u7inh-32tb", "f2", "trn2", "trn2u"}
+	instanceFamilies := []string{"m8g", "m8gd", "m7i", "m7i-flex", "m7a", "c8g", "c8gd", "c8gn", "c7i", "c7i-flex", "c7a", "r7a", "r7i", "r7iz", "r8g", "r8gd", "x8g", "u7i", "u7inh", "g6", "g6e", "gr6", "gb200", "i7i", "i7ie", "i8g", "p5", "p5e", "p5en", "p6-b200", "p6e-b200", "u7i-6tb", "u7i-8tb", "u7i-12tb", "u7in-16tb", "u7in-24tb", "u7in-32tb", "u7inh-32tb", "f2", "trn2", "trn2u", "x8g"}
```
AndrewSirenko: Please someone check my work on these!

ElijahQuinones: Typo `P6e-B200` ->  `P6e-GB200`

pkg/cloud/volume_limits.go
```diff
@@ -32,7 +32,7 @@ const (
 func init() {
```
ElijahQuinones, line 32: May need to run `generate-instance-store-table.sh` i see about 51 missing instances from the instance store  table

AndrewSirenko: Good catch!

Sidenote, any reason this can't be a makefile target and we substitute the code in that file? I'll add to backlog if not!

ElijahQuinones: I think that would be a really good idea !

Just have to make sure the account it is run from has all of the opt in regions

PR #2568, opened 2025-07-15
Add blockAttachUntilInitialized SC parameter

AndrewSirenko:
#### What type of PR is this?

/kind feature

Allows users to delay ControllerPublishVolume success (and therefore start of workload) until a volume restored from snapshot is fully initialized. 

This is achieved by polling `EC2 DescribeVolumeStatus` after volume attached. 

If feature is popular, In future we may add:
- ability to override blocking via VAC parameter
- Parallel polling for volume attachment and for volume initialized

- Our new E2E test (initialization rate + blockAttachUntilInitialized)
- Manually testing blockAttachUntilInitialized on volume restored from snapshot WITHOUT initializationRate
- Manually testing blockAttachUntilInitialized on non-restored volume

```release-note
Added StorageClass parameter 'blockAttachUntilInitialized' for users who want to delay ControllerPublishVolume success (and therefore start of workload) until a volume restored from snapshot is fully initialized 
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -88,7 +88,8 @@ var (
 )
 
 const (
-	cacheForgetDelay = 1 * time.Hour
+	cacheForgetDelay        = 1 * time.Hour
+	volInitCacheForgetDelay = 6 * time.Hour
 )
 
 // AWS provisioning limits.
@@ -104,6 +105,12 @@ const (
 	MaxTagKeyLength = 128
 	// MaxTagValueLength represents the maximum value length for a tag.
 	MaxTagValueLength = 256
+)
+
+// VolumeStatusInitializingState is const reported by EC2 DescribeVolumeStatus which AWS SDK does not have type for.
+const (
+	VolumeStatusInitializingState = "initializing"
+	VolumeStatusInitializedState  = "completed"
 )
 
 // Defaults.
@@ -135,7 +142,13 @@ const (
 	snapshotTagBatcher
 
 	batchDescribeTimeout = 30 * time.Second
-	batchMaxDelay        = 500 * time.Millisecond // Minimizes RPC latency and EC2 API calls. Tuned via scalability tests.
+
+	// Minimizes RPC latency and EC2 API calls. Tuned via scalability tests.
+	batchMaxDelay = 500 * time.Millisecond
+
+	// Tuned for EC2 DescribeVolumeStatus -- as of July 2025 it takes up to 5 min for initialization info to be updated.
+	slowVolumeStatusBatchMaxDelay = 2 * time.Minute
+	fastVolumeStatusBatchMaxDelay = 500 * time.Millisecond
 )
 
 var (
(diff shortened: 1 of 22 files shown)


pkg/cloud/cloud.go
```diff
@@ -1015,6 +1037,144 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
...
+}
+
+func isVolumeStatusInitializing(vsi types.VolumeStatusItem) bool {
+	for _, detail := range vsi.VolumeStatus.Details {
+		if detail.Name == types.VolumeStatusNameInitializationState && detail.Status != nil && *detail.Status == VolumeStatusInitializingState {
+			return true
+		}
+	}
+	return false
+}
```
AndrewSirenko, line 1119: We can replace this with checking whether volume initialization progress is 100. Some manual testing since release shows it is stable. Thoughts? Or is this fine. CC @ElijahQuinones @ConnorJC3

pkg/cloud/cloud.go
```diff
@@ -1015,6 +1037,144 @@ func (c *cloud) DetachDisk(ctx context.Context, volumeID, nodeID string) error {
...
+		volumeStatusItem, err = c.describeVolumeStatus(volumeID, true /* callASAP */)
+	// Case 2: We already know volume is initialized. Don't call DVS.
+	case volInit.initialized:
+		return true, nil
+	// Case 3: We know volume is initializing, but there is no SLA. Call DVS eventually during next slow batch.
+	case volInit.estimatedInitializationTime.IsZero():
+		volumeStatusItem, err = c.describeVolumeStatus(volumeID, false /* callASAP */)
+	// Case 4: We have an estimated time for initialization. Wait to call DVS again until then unless RPC ctx is done.
+	case !volInit.initialized:
+		//
```
ElijahQuinones: nit: empty comment

pkg/cloud/cloud.go
ElijahQuinones: Is it ever possible for volume to be in volume initializing status before we hit 100 on initialization progress ? Is the volume not in Initializing status as soon as initialization progress is >0 ?

tests/e2e/requires_aws_api.go
```diff
@@ -246,6 +247,32 @@ var _ = Describe("[ebs-csi-e2e] [single-az] [requires-aws-api] Dynamic Provision
...
+
+				statuses := resultDVS.VolumeStatuses
+				if len(statuses) != 1 {
+					Fail(fmt.Sprintf("expected 1 volume status, got %d", len(statuses)))
+				}
+
+				// Purposefully don't rely on helper function from cloud pkg
+				volumeInitialized := false
+				for _, detail := range statuses[0].VolumeStatus.Details {
+					if detail.Name == types.VolumeStatusNameInitializationState && detail.Status != nil && *detail.Status == awscloud.VolumeStatusInitializedState {
```
ElijahQuinones, line 269: From e2e tests
```
./requires_aws_api.go:269:120: undefined: awscloud.VolumeStatusInitializedState
./requires_aws_api.go:274:82: undefined: awscloud.VolumeStatusInitializedState
```

AndrewSirenko: Yeah I'm figuring out why it's not working 🤔, hence why no manual tests yet

AndrewSirenko: Oh figured I deleted it

pkg/cloud/cloud.go
AndrewSirenko: If inititializationProgress != 100 THEN we return true

ElijahQuinones: I think how we have it now is a bit safer as we are checking for the status to equal `VolumeStatusInitializingState` which to me seems a bit safer from things like Nil edge cases.

AndrewSirenko:
/retest

arm64 flake

PR #2579, opened 2025-07-23
helm: Add ebs-csi-node readiness probe

AndrewSirenko:
#### What type of PR is this?

/kind feature

Today when deploying on Kubernetes ebs-csi-node ebs-plugin pods have liveness probe but not readiness probe. This means that container/pod will be marked ready immediately despite not having metadata source or serving CSI RPCs. This means `Ready` containers without metadata will eventually restart due to liveness probe, confusing users. 

Put 15s sleep in main.go before metadata:
- Before PR container marked ready immediately
- After PR container marked ready after 15s.

Also validated by turning IMDS off for an instance on cluster and setting metadataSources to imds. 

```release-note
Helm: Add ebs-csi-node ebs-plugin readiness probe so that pod is not marked ready until metadata source acquired and starts serving CSI RPCs
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/_node-windows.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_node-windows.tpl
@@ -148,6 +148,13 @@ spec:
             timeoutSeconds: 3
             periodSeconds: 10
             failureThreshold: 5
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
+            timeoutSeconds: 3
+            periodSeconds: 5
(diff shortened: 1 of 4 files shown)


ConnorJC3:
/lgtm

Why would this be breaking though?

torredil (approved):
/lgtm

AndrewSirenko:
> Why would this be breaking though?

Suddenly node pods going to be marked not ready for X seconds. Not sure if that will mess up some folks' metrics.

mdzraf (approved):
/lgtm

PR #2580, opened 2025-07-23
Add explicit check that device looks like actual device before callin…

ConnorJC3:
…g /dev/

#### What type of PR is this?

/kind cleanup

Adds an explicit check before calling `lsblk` that the device name looks like an actual device name. This is intended to check that the device name doesn't look like `--malicious-option=do_evil`, as the device name can be user-controlled in some circumstances.

`make test` + E2E tests

```release-note
NONE
```


ConnorJC3:
/retest

Flaky test

ConnorJC3:
/retest

World's most flaky e2e test

pkg/mounter/mount_linux.go
```diff
@@ -160,25 +160,31 @@ func execRunner(name string, arg ...string) ([]byte, error) {
 
 // verifyVolumeSerialMatch checks the volume serial of the device against the expected volume.
 func verifyVolumeSerialMatch(canonicalDevicePath string, strippedVolumeName string, execRunner func(string, ...string) ([]byte, error)) error {
+	// As a security precaution, check the device name looks like a real device name before passing anything to exec
+	cleanDevice := filepath.Clean(canonicalDevicePath)
+	if !regexp.MustCompile(`^/dev/[A-Za-z0-9]+$`).MatchString(cleanDevice) {
+		return fmt.Errorf("refusing to mount %s (raw: %s) because it does not appear to be a valid device", cleanDevice, canonicalDevicePath)
+	}
```
torredil, line 167: Can we split this out into a helper so that we can unit test it?

pkg/mounter/mount_linux_test.go
mdzraf, line 1: Should we add a unit test here to check that an invalid name does not pass?

pkg/mounter/mount_linux.go
ConnorJC3: So turns out we already had unit tests for this function entirely and accidentally deleted them - so I restored them and added 2 new tests for the new functionality

pkg/mounter/mount_linux_test.go
ConnorJC3: See my above comment, same thing

pkg/mounter/mount_linux.go
torredil: Thank you for adding more code coverage 👍

PR #2588, opened 2025-07-28
Increase robustness of taint removal

### LINKED ISSUE #2575: 1.45.0 sometimes fails to remove taint
/kind bug

**What happened?**
We have upgraded some clusters to EKS 1.33 and ebs-csi v1.45.0-eksbuild.2 from EKS 1.32 and ebs-csi v1.42.0-eksbuild.1. We have seen a small number of nodes get stuck with the ebs-csi taint not getting cleared. This leaves the node in a state where workload pods will not get scheduled and Karpenter will not remove the node as it has not finished initializing yet.

Taint to be removed.

Unfortunately this is a rare race condition that we can't reproduce reliably. 

Looking through the audit logs it looks like the ebs-csi pod is not refreshing the node's current taints when the update request fails. It keeps retrying with the same outdated state.

Sample from CloudWatch logs:
Taints are:
```
responseObject.spec.taints.0 {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
responseObject.spec.taints.1 {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
responseObject.spec.taints.2 {"key":"node.kubernetes.io/not-ready","effect":"NoExecute","timeAdded":"2025-07-22T12:17:19Z"}
``` 

node-controller removes a taint leaving just the csi controllers:
```
@timestamp                   1753186659667
requestObject.spec.taints.0  {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.spec.taints.1  {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestReceivedTimestamp     2025-07-22T12:17:39.545267Z
responseObject.spec.taints.0 {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
responseObject.spec.taints.1 {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
verb                         patch
```

ebs-csi fails to update:
```
@timestamp               1753186660233
requestObject.0.op       test
requestObject.0.path     /spec/taints
requestObject.value.0    {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.value.1    {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.value.2    {"key":"node.kubernetes.io/not-ready","effect":"NoExecute","timeAdded":"2025-07-22T12:17:19Z"}
requestObject.1.op       replace
requestObject.1.path     /spec/taints
requestObject.value.0    {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.value.1    {"key":"node.kubernetes.io/not-ready","effect":"NoExecute","timeAdded":"2025-07-22T12:17:19Z"}
requestReceivedTimestamp 2025-07-22T12:17:40.096491Z
responseObject.code      422
responseObject.kind      Status
responseObject.message   the server rejected our request due to an error in our request
responseObject.reason    Invalid
responseObject.status    Failure
responseStatus.code      422
responseStatus.message   the server rejected our request due to an error in our request
responseStatus.reason    Invalid
responseStatus.status    FailurerequestObject.value.0  {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
verb                     patch
```

Keeps trying with same input and result:
```
@timestamp 1753186660483
requestReceivedTimestamp 2025-07-22T12:17:40.194728Z

@timestamp 1753186663240
requestReceivedTimestamp 2025-07-22T12:17:43.202173Z

@timestamp 1753186663240
requestReceivedTimestamp 2025-07-22T12:17:43.104818Z

@timestamp 1753186667752
requestReceivedTimestamp 2025-07-22T12:17:47.711254Z

@timestamp 1753186667752
requestReceivedTimestamp 2025-07-22T12:17:47.614086Z

@timestamp 1753186674520
requestReceivedTimestamp 2025-07-22T12:17:54.470557Z

@timestamp 1753186674520
requestReceivedTimestamp 2025-07-22T12:17:54.379558Z

@timestamp 1753186674520
requestReceivedTimestamp 2025-07-22T12:17:54.372631Z

@timestamp 1753186676525
requestReceivedTimestamp 2025-07-22T12:17:56.387603Z
```

efs-csi removes its taint:
```
@timestamp                   1753186678278
requestObject.0.op           test
requestObject.0.path         /spec/taints
requestObject.0.value.0      {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.0.value.1      {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.1.op           replace
requestObject.1.path         /spec/taints
requestObject.1.value.0      {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestReceivedTimestamp     2025-07-22T12:17:58.113754Z
responseObject.spec.taints.0 {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
verb                         patch
```

ebs-csi continues trying to replace old taints:
```
@timestamp               1753186679531
requestObject.0.op       test
requestObject.0.path     /spec/taints
requestObject.value.0    {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.value.1    {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.value.2    {"key":"node.kubernetes.io/not-ready","effect":"NoExecute","timeAdded":"2025-07-22T12:17:19Z"}
requestObject.1.op       replace
requestObject.1.path     /spec/taints
requestObject.value.0    {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
requestObject.value.1    {"key":"node.kubernetes.io/not-ready","effect":"NoExecute","timeAdded":"2025-07-22T12:17:19Z"}
requestReceivedTimestamp 2025-07-22T12:17:59.395297Z
responseObject.code      422
responseObject.kind      Status
responseObject.message   the server rejected our request due to an error in our request
responseObject.reason    Invalid
responseObject.status    Failure
responseStatus.code      422
responseStatus.message   the server rejected our request due to an error in our request
responseStatus.reason    Invalid
responseStatus.status    Failure
```

The condition does not resolve itself. We've seen nodes that are 4d+ old.

**Environment**
- Kubernetes version (use `kubectl version`): v1.33.1-eks-595af52
- Driver version: v1.45.0-eksbuild.2

ConnorJC3:
#### What type of PR is this?

/kind bug

Attempts to fix #2575 by:
- Abort retries when the error the API server returned will never be fixed by a retry (not found, invalid object, etc)
- Resync every 5 seconds to hopefully re-drive any weird informer failures

```release-note
Increase robustness of taint removal via resync
```

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
@@ -35,6 +36,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -72,7 +74,7 @@ var (
 
 const (
 	// taintWatcherDuration is the maximum duration for the not-ready taint watcher to run.
-	taintWatcherDuration = 1 * time.Minute
+	taintWatcherDuration = 10 * time.Minute
 )
 
 // NodeService represents the node service of CSI driver.
@@ -89,7 +91,7 @@ func NewNodeService(o *Options, md metadata.MetadataService, m mounter.Mounter, 
 	if k != nil {
 		// Watch for the agent‑not‑ready taint for up to one minute and remove it
 		// as soon as allocatable is available.
-		startNotReadyTaintWatcher(k, taintWatcherDuration)
+		go startNotReadyTaintWatcher(k, taintWatcherDuration)
 	}
 
 	return &NodeService{
@@ -862,30 +864,44 @@ func startNotReadyTaintWatcher(clientset kubernetes.Interface, maxWatchDuration 
 
 	factory := informers.NewSharedInformerFactoryWithOptions(
 		clientset,
-		0, // No resync
+		5*time.Second, // Resync every 5 seconds in case of networking or other rare issue
 		informers.WithTweakListOptions(func(lo *metav1.ListOptions) {
 			lo.FieldSelector = "metadata.name=" + nodeName
 		}),
 	)
 	informer := factory.Core().V1().Nodes().Informer()
 
+	var mutex sync.Mutex
+	ctx, cancel := context.WithTimeout(context.Background(), maxWatchDuration)
+	defer cancel()
 	attemptTaintRemoval := func(n *corev1.Node) {
 		if !hasNotReadyTaint(n) {
+			// Node has no taint, do nothing
 			return
 		}
+
+		if !mutex.TryLock() {
+			// Another removal thread is already running, do nothing
+			return
+		}
+		defer mutex.Unlock()
 
 		backoff := wait.Backoff{
 			Duration: 2 * time.Second,
 			Factor:   1.5,
 			Steps:    5,
 		}
-
-		ctx, cancel := context.WithTimeout(context.Background(), maxWatchDuration)
-		defer cancel()
-
 		err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
 			if err := removeNotReadyTaint(ctx, clientset, n); err != nil {
 				klog.ErrorS(err, "Failed to remove agent-not-ready taint, retrying", "node", n.Name)
+				if apierrors.IsBadRequest(err) || apierrors.IsInvalid(err) || apierrors.IsNotFound(err) {
+					// Our node is probably stale, get a new copy
+					var err error
+					n, err = clientset.CoreV1().Nodes().Get(ctx, n.Name, metav1.GetOptions{})
+					if err != nil {
+						klog.ErrorS(err, "Failed to update potentially stale node", "node", n.Name)
+					}
+				}
 				return false, nil // Continue retrying
 			}
 			klog.V(2).InfoS("Successfully removed agent-not-ready taint", "node", n.Name)
@@ -913,26 +929,28 @@ func startNotReadyTaintWatcher(clientset kubernetes.Interface, maxWatchDuration 
 		return
 	}
 
-	stopCh := make(chan struct{})
+	factory.Start(ctx.Done())
+	if ok := cache.WaitForCacheSync(ctx.Done(), informer.HasSynced); !ok {
+		klog.ErrorS(nil, "Taint-watcher: cache sync failed")
+	}
 
-	go func() {
-		factory.Start(stopCh)
-
-		if ok := cache.WaitForCacheSync(stopCh, informer.HasSynced); !ok {
-			klog.ErrorS(nil, "Taint-watcher: cache sync failed")
+	// Immediate scan in case the taint is already present and no event fires.
+	if obj, exists, err := informer.GetStore().GetByKey(nodeName); err == nil && exists {
+		if n, ok := obj.(*corev1.Node); ok {
+			attemptTaintRemoval(n)
 		}
+	}
 
-		// Immediate scan in case the taint is already present and no event fires.
-		if obj, exists, err := informer.GetStore().GetByKey(nodeName); err == nil && exists {
-			if n, ok := obj.(*corev1.Node); ok {
-				attemptTaintRemoval(n)
-			}
-		}
+	<-time.After(maxWatchDuration)
+	klog.V(8).InfoS("Taint-watcher timeout reached; stopping")
 
-		<-time.After(maxWatchDuration)
-		klog.V(8).InfoS("Taint-watcher timeout reached; stopping")
-		close(stopCh)
-	}()
+	// Try to remove the taint one last time in case we got extremely unlucky with the informer
+	lastChanceNode, err := clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		klog.ErrorS(err, "Failed to get node for last chance", "node", nodeName)
+	} else {
+		attemptTaintRemoval(lastChanceNode)
+	}
 }
 
 func hasNotReadyTaint(n *corev1.Node) bool {


pkg/driver/node.go
```diff
@@ -862,7 +863,7 @@ func startNotReadyTaintWatcher(clientset kubernetes.Interface, maxWatchDuration
 
 	factory := informers.NewSharedInformerFactoryWithOptions(
 		clientset,
-		0, // No resync
+		5, // Resync every 5 seconds in case of networking or other rare issue
```
torredil: 5 second re-sync seems a bit aggressive, I was thinking along the lines of 30s would be appropriate - deferring to your final rec.

pkg/driver/node.go
```diff
@@ -885,6 +886,12 @@ func startNotReadyTaintWatcher(clientset kubernetes.Interface, maxWatchDuration
 
 		err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
 			if err := removeNotReadyTaint(ctx, clientset, n); err != nil {
+				if apierrors.IsBadRequest(err) || apierrors.IsInvalid(err) || apierrors.IsNotFound(err) {
+					// Retrying the same request will never fix these errors
+					// Abort by returning error and wait for next resync/informer update
+					klog.ErrorS(err, "Failed to remove agent-not-ready taint, retry not useful", "node", n.Name)
+					return false, err
+				}
```
torredil: I agree that we should stop retrying with the old node object. However, thoughts on getting a fresh node object once we reach this state? I was doing some testing on my side and managed to solve the problem described in 2575 doing similar to below:

```
if statusErr, ok := err.(*errors.StatusError); ok && statusErr.Status().Code == 422 {
   // get fresh node object
   freshNode, err := clientset.CoreV1().Nodes().Get(context.TODO(), staleNode.Name, metav1.GetOptions{})
   if err != nil {
	return fmt.Errorf("failed to get fresh node object: %v", err)
   }
  // attempt to remove taint using new node object
```

Since our last offline conversation about 2575, I learned that a re-sync (plays back all the events held in the informer cache) is different from a re-list (hits the API server to get refreshed data). So I'm concerned that relying on re-sync is not sufficient to solve 2575 if the informer never receives the update.

ConnorJC3: So the reason I didn't get a fresh node object here is that we will then end up with multiple threads racing themselves to get and update the Node (this will be even worse with the re-sync, but even without the re-sync there were 6 such threads on a single node in #2575). This could easily triple or worse the API server load in cases where many nodes (1k+) are spun up at once.

It also doesn't actually fix any underlying issue except by luck. If there's scenarios where the informer get stuck on stale data we need to figure that out and fix it, because there are going to be nodes where our `UpdateFunc` is never called _at all_.

pkg/driver/node.go
ConnorJC3: My theory on 5 seconds is that we immediately exit in the function if there is no taint, so calling it more often gives us more chances to re-attempt a failure and is practically free. 30 seconds would only give us 1 or 2 (it's race-y) retries in the event of a weird failure.

pkg/driver/node.go
torredil: I understand the need to be conservative but before we implemented the informer the driver periodically reached out to the API server to get a fresh copy of the node. I think its reasonable to get a fresh copy of the object when patching fails due to a 422, as we know we are working with stale data atp. We cannot afford to not remove this taint given the impact of the inability to do so.

 > It also doesn't actually fix any underlying issue except by luck
 
 Not sure that I agree with this. As I understand it, its a pretty reliable solution. If the taint is present we should stop attempting to patch the node object with stale data, and at least try with an updated copy. (and yes, I understand there is a race condition where we retrieve an updated copy and it becomes stale before the patch goes through - in this case, we should retry, and continue working towards the desired state).
 
 > If there's scenarios where the informer get stuck on stale data we need to figure that out and fix it
 
 As I see it, unless we root cause why the informer doesn't receive an update before next release we should strongly consider either:
 
 1. Reverting the informer approach.
 2. Continue using the informer approach, but attempt to patch by fetching a fresh node object when we hit 422 (which seems to be an extremely rare scenario to begin with).
 
 2 seems much more desirable, given we are making ~1 call to the API server in rare circumstances, rather than many calls, always.

ConnorJC3: > I think its reasonable to get a fresh copy of the object when patching fails due to a 422, as we know we are working with stale data atp.

This is the point of the informer - if we think informers are so broken there are some scenarios under which they get stuck on stale data indefinitely, then we need to revert their use entirely as we fundamentally rely on that assumption and we cannot ever fix #2575 if that is not true.

pkg/driver/node.go
AndrewSirenko: I agree with the 5 seconds number in order to have more tries before livenessprobe shuts us down. 

It's a moderate level of aggressiveness between 1 and 30 seconds.

torredil: Okay, sounds good.

pkg/driver/node.go
AndrewSirenko: > It also doesn't actually fix any underlying issue except by luck

I agree with Eddie that if we're getting stale data from one approach (without knowing the reason), it would be helpful to fallback to an alternative approach that relies on a completely different mechanism.

And to ensure we don't spam API server, how about we only attempt to make a non-informer GET call one time per node driver restart? (with a flag)

ConnorJC3: @torredil @AndrewSirenko If we genuinely believe that informers are just broken, and can randomly fail to call the `UpdateFunc`, then what is our solution to when our UpdateFunc is called zero times?

AndrewSirenko: > If we genuinely believe that informers are just broken, and can randomly fail to call the UpdateFunc, then what is our solution to when our UpdateFunc is called zero times?

If we can't confirm our taint was removed within 10s, we make *one* manual get call and post right after that?

AndrewSirenko: Team agreed offline on:

Non-negotiable: Taint MUST eventually be removed. Better to restart-loop csi-node-pod forever than have taint up. 

Short term improvements:
1 This PR
2 Extend informer timeout to 10 min to prevent watch time bug
3 ONE API Call on 422 per driver restart. (Eddie's initial suggestion)
4 Mutex to prevent thread races.
5 One final API Call no matter what after 10s-60s. 

Longterm fixes:
a. Hemant's scheduler KEP that makes the startup taints unneccessary. 
b. Putting taint removal on ebs-csi-controller with Sylvia's new node-watcher.

torredil: We discussed this offline and agreed to:

1. Add a mutex to make `attemptTaintRemoval` atomic.
2. Increase the lifespan of the informer.
3. Fetch a new copy of the node object when patching fails due to 422.
4. Fetch a new copy of the node and attempt to remove the taint if is still present before the informer is terminated.

ElijahQuinones (approved):
/lgtm

Issue #1551, opened 2023-03-28
CSI controller Health Check doesn't properly validate controller health.


stobias123:
/kind bug

**What happened?**
Volumes could not attach. I had a combination of these two errors. After restarting the CSI driver deployment, the volumes attached properly.

```
E0328 19:16:27.445502       1 driver.go:120] "GRPC error" err="rpc error: code = Internal desc = Could not attach volume \"vol-02a81309e3377eebb\" to node \"i-0c9f3cd3ae6b2d20d\": attachment of disk \"vol-02a81309e3377eebb\" failed, expected device to be attached but was attaching"
I0328 19:16:58.979598       1 cloud.go:611] "Ignoring error from describe volume, will retry" volumeID="vol-02a81309e3377eebb" err=<
        RequestCanceled: request context canceled
        caused by: context deadline exceeded
```

**What you expected to happen?**

I'd expect this time out failure to surface in the healthcheck and cause the pod to be restarted. If the healthcheck fails, and restarts the pod, things would have auto healed without user input. 

**How to reproduce it (as minimally and precisely as possible)?**

**Anything else we need to know?**:
I had a support case open for this. 12358592881. 

**Environment**
- Kubernetes version (use `kubectl version`): `v1.23.16-eks-48e63af`
- Driver version: `v1.17.0-eksbuild.1`

stobias123:
Bummer! Any reason we can't keep this open? I think my concern is still valid here.

AndrewSirenko:
/remove-lifecycle rotten

@stobias123 yes we should be keeping this open. I'll bring this issue back up at today's standup. Thank you for your patience.

mdzraf:
/remove-lifecyle stale

PR is under review to fix

PR #2590, opened 2025-07-30
Add Readiness/Liveness EC2 API call probe to Controller Service

AndrewSirenko:
#### What type of PR is this?

/kind feature

Lets customers realize that their EBS CSI Driver ebs-csi-controller pods aren't actually ready due to some Auth/networking error. 

Should really help customer troubleshooting for new customers (and ensuring networking and auth issues aren't root cause of user problems)  

Should work for other COs because this is done through CSI Probe RPC

Deploy driver without right IAM role, see not ready 

Add IAM role, see ready

Remove role again, see driver eventually become not ready

Add role back and driver becomes ready again

```release-note
Breaking: Add Readiness probe via EC2 API call to Controller Service. Clusters with missing IAM roles or networking issues may see ebs-csi-controller pod restarts. 
```


pkg/driver/identity.go
```diff
@@ -58,6 +61,16 @@ func (d *Driver) GetPluginCapabilities(ctx context.Context, req *csi.GetPluginCa
 }
 
 func (d *Driver) Probe(ctx context.Context, req *csi.ProbeRequest) (*csi.ProbeResponse, error) {
-	klog.V(6).InfoS("Probe: called", "args", req)
+	// Controller Service will make dry-run EC2 call to ensure proper auth/networking
+	// TODO Q: Should we NOT do this in ALL mode?
```
AndrewSirenko: Thoughts?

pkg/cloud/cloud.go
```diff
@@ -1604,6 +1616,28 @@ func (c *cloud) EnableFastSnapshotRestores(ctx context.Context, availabilityZone
 	return response, nil
 }
 
+// DryRun will make a dry-run EC2 API call. Nil return value means we successfully received EC2 DryRunOperation error code.
+func (c *cloud) DryRun(ctx context.Context) error {
+	if c.attemptDryRun.Load() {
+		_, apiErr := c.ec2.DescribeAvailabilityZones(ctx,
```
AndrewSirenko, line 1796: Chose DAZ because default instance role has describeVolumes

Though should we be worried that DAZ gets added to default instance role?! Thoughts?

pkg/driver/identity.go
```diff
@@ -58,6 +61,16 @@ func (d *Driver) GetPluginCapabilities(ctx context.Context, req *csi.GetPluginCa
...
+	if d.controller != nil && d.controller.cloud != nil {
+		err := d.controller.cloud.DryRun(ctx)
+		if err != nil {
+			errStr := fmt.Sprintf("EBS CSI Controller failed readiness check, ensure your EBS CSI Controller has proper IAM credentials and network connection, %v", err)
+			klog.V(2).InfoS("Probe: failed", "err", errStr)
```
AndrewSirenko: For your review convenience, error message will look smth like: 

```
I0730 18:01:51.482755       1 identity.go:70] "Probe: failed" err="EBS CSI Controller failed readiness check, ensure your EBS CSI Cont
roller has proper IAM credentials and network connection, dry-run EC2 API call failed, EC2 API err: operation error EC2: DescribeAvail
abilityZones, https response error StatusCode: 403, RequestID: 98e619ff-f3a0-491a-9b28-98800ffe2c99, api error UnauthorizedOperation:
You are not authorized to perform this operation. User: arn:aws:sts::208536326202:assumed-role/eksctl-usual-podidentityrole-kube-syste
m-ebs--Role1-MQCYoktbbw8T/eks-usual-ebs-csi-co-047f0dc7-530f-458d-b076-7ba8d78ea01f is not authorized to perform: ec2:DescribeAvailabi
lityZones because no identity-based policy allows the ec2:DescribeAvailabilityZones action"
E0730 18:01:51.482770       1 driver.go:108] "GRPC error" err="rpc error: code = FailedPrecondition desc = EBS CSI Controller failed r
eadiness check, ensure your EBS CSI Controller has proper IAM credentials and network connection, dry-run EC2 API call failed, EC2 API
 err: operation error EC2: DescribeAvailabilityZones, https response error StatusCode: 403, RequestID: 98e619ff-f3a0-491a-9b28-98800ff
e2c99, api error UnauthorizedOperation: You are not authorized to perform this operation. User: arn:aws:sts::208536326202:assumed-role
/eksctl-usual-podidentityrole-kube-system-ebs--Role1-MQCYoktbbw8T/eks-usual-ebs-csi-co-047f0dc7-530f-458d-b076-7ba8d78ea01f is not aut
horized to perform: ec2:DescribeAvailabilityZones because no identity-based policy allows the ec2:DescribeAvailabilityZones action"
```

torredil: Since the error gets bubbled up to `Run` and logged there at the `klog.ErrorS` verbosity, I think we should remove this line to avoid duplicating information in the logs.:

```
klog.V(2).InfoS("Probe: failed", "err", errStr)
```

pkg/cloud/cloud.go
```diff
@@ -1604,6 +1616,28 @@ func (c *cloud) EnableFastSnapshotRestores(ctx context.Context, availabilityZone
...
+				o.Retryer = aws.NopRetryer{} // Don't retry so we can catch network failures. CO should retry liveness check multiple times.
+				o.APIOptions = nil           // Don't add our logging/metrics middleware because we expect errors.
+			})
+		if apiErr != nil {
+			var awsErr smithy.APIError
+			if errors.As(apiErr, &awsErr) && awsErr.ErrorCode() == "DryRunOperation" {
+				c.attemptDryRun.Store(false)
+				return nil
+			}
+			return fmt.Errorf("dry-run EC2 API call failed, EC2 API err: %w", apiErr)
```
ConnorJC3: Suggest changing string to `dry-run EC2 API call failed: %w`, as the error may not be from the EC2 API (could be a networking/DNS issue for example, or from STS). This also cuts a few characters off an otherwise long error message.

pkg/cloud/cloud.go
ConnorJC3: Unlikely and we can always change the API call in the future. Suggest you add a comment with this reasoning though so we remember in the future.

pkg/cloud/cloud_test.go
```diff
@@ -3626,6 +3623,63 @@ func TestIsVolumeInitialized(t *testing.T) {
...
+			c := &cloud{
+				region: "test-region",
+				ec2:    mockEC2,
+			}
+			c.attemptDryRun.Store(tc.attemptDryRunBefore)
+
+			if tc.attemptDryRunBefore {
+				mockEC2.EXPECT().DescribeAvailabilityZones(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, &smithy.GenericAPIError{
+					Code: tc.errCode,
+				}).Times(1)
```
torredil: Can we add a unit test where DAZ is called more than once?

pkg/cloud/cloud.go
```diff
@@ -400,7 +400,19 @@ func newEC2Cloud(region string, awsSdkDebugLog bool, userAgentExtra string, batc
...
+		dryRunTicker:          time.NewTicker(dryRunInterval),
 	}
+
+	// Ensure an EC2 Dry-run API call is made on startup and every dryRunInterval
+	c.attemptDryRun.Store(true)
+	go func() {
+		for {
+			<-c.dryRunTicker.C
+			c.attemptDryRun.Store(true)
+		}
```
ConnorJC3, line 422: Given that we don't need to access the ticker anywhere else, I think we could simplify this and remove the ticker variable entirely:
```
go func() {
    for range time.Tick(dryRunInterval) {
            c.attemptDryRun.Store(true)
    }
}()
```

pkg/driver/identity.go
ConnorJC3: +1 - duplicating the same error in the log is confusing, and having an Info log for an Error event is also confusing (also see my other comment on line 69 this ties into)

pkg/driver/identity.go
```diff
@@ -58,6 +61,16 @@ func (d *Driver) GetPluginCapabilities(ctx context.Context, req *csi.GetPluginCa
 }
 
 func (d *Driver) Probe(ctx context.Context, req *csi.ProbeRequest) (*csi.ProbeResponse, error) {
-	klog.V(6).InfoS("Probe: called", "args", req)
+	// Controller Service will make dry-run EC2 call to ensure proper auth/networking
+	// TODO Q: Should we NOT do this in ALL mode?
+	if d.controller != nil && d.controller.cloud != nil {
+		err := d.controller.cloud.DryRun(ctx)
+		if err != nil {
+			errStr := fmt.Sprintf("EBS CSI Controller failed readiness check, ensure your EBS CSI Controller has proper IAM credentials and network connection, %v", err)
```
ConnorJC3: Suggest the following changes:
- Don't reference the controller - somebody could be running in combined mode, and that terminology isn't necessarily universal anyways on other COs (Anybody seeing this error knows where it came from because they're looking at logs from a specific component anyways)
- Don't use the term "rediness check", as we don't actually know why the probe was called (could be a _liveness_ check or running on another CO entirely)
- Along with the below suggestion by Eddie to not use `InfoS`, move this into the return so the error object can be properly wrapped rather than passed as a string

Therefore, suggest instead replacing lines 69-71 with:
```
return &csi.ProbeResponse{}, status.Errorf(codes.FailedPrecondition, "Failed health check (verify network connection and IAM credentials): %w", errStr)
```

pkg/driver/identity.go
ConnorJC3: Why would we not run in `all` mode? If the driver is using the AWS API, this check should be run.

pkg/cloud/cloud_test.go
AndrewSirenko: Sure, but that would be a test that crosses from identity.go to cloud.go (calling probe twice!) 

That fine? 

It'll be our first identity.go test lol

pkg/driver/identity.go
AndrewSirenko: Can't use %w in status.Errorf so will use %v, but otherwise agree!

ConnorJC3:
Nitpick: Can we add "Fixes #1551" to the PR description?

Also this isn't a breaking change, that should be removed from the changelog entry.

AndrewSirenko:
> Also this isn't a breaking change, that should be removed from the changelog entry.

Can we switch to warning during changelog? I fear some customers with misconfigured drivers will be very concerned for a second after upgrading 😓

torredil (approved):

ElijahQuinones:
/hold 

holding for release

ConnorJC3 (commented):
Mostly lgtm, one question: Have we tested what this looks like for a timeout scenario? Do we need to raise this value to avoid the dry run cutting out?

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L187

AndrewSirenko:
> Have we tested what this looks like for a timeout scenario? Do we need to raise this value to avoid the dry run cutting out?

Have tested for timeout scenario (purposefull broke my policy AND also tested breaking networking)

However good callout, I'll raise timeout to 10s for live and readiness, and then raise liveness failure threshold to 10 (so pod becomes not ready first, THEN starts restarting)

AndrewSirenko:
/hold

For final manual test

AndrewSirenko:
/unhold

Did final test of:
- Deploy CSI Driver (it works)
- Break IAM policy
- Wait 3 hours --> Driver enters unready -> Driver enters crashbackoffLoo
- Fix IAM policy
- Driver upon next container restart (4 min?) works!

ConnorJC3 (changes requested):
Tested manually, doesn't correctly log the problem when it is due to a timeout, instead logs "context cancelled"

e.g. testing with `AWS_EC2_ENDPOINT` set to `https://www.google.com:81` to simulate a timeout:
```
E0826 20:44:21.041364       1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response error StatusCode: 0, RequestID: , canceled, context canceled"
E0826 20:44:29.349716       1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response error StatusCode: 0, RequestID: , canceled, context canceled"
E0826 20:44:31.041825       1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response error StatusCode: 0, RequestID: , canceled, context canceled"
```

AndrewSirenko:
So by setting probe-timeout as connor recommended now ebs-plugin logs flip between correct and misleading logs:

```
E0827 22:00:27.189582       1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify
 network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response
error StatusCode: 0, RequestID: , request send failed, Post \"https://www.google.com:81/\": dial tcp 142.250.73.132:81: i/o timeout"
E0827 22:00:27.337601       1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify
 network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response
error StatusCode: 0, RequestID: , request send failed, Post \"https://www.google.com:81/\": dial tcp 142.250.73.132:81: i/o timeout"
E0827 22:00:30.346150       1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify
 network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response
error StatusCode: 0, RequestID: , canceled, context canceled"
E0827 22:00:35.003252       1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify
 network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response
error StatusCode: 0, RequestID: , canceled, context canceled"
```

---

Looks like the detailed FailedPrecondition errors are from when csi sidecars probe Controller Service, but context canceled ones are from LivenessProbe logs. 

---

Tested with:

```
controller:
  env:
    - name: AWS_EC2_ENDPOINT
      value: https://www.google.com:81
```

AndrewSirenko:
Latest commit leads to proper timeouts and all expected GRPC errors.

```
1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify
 network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response
error StatusCode: 0, RequestID: , request send failed, Post \"https://www.google.com:81/\": dial tcp 142.250.73.132:81: i/o timeout"
1 driver.go:112] "GRPC error" err="rpc error: code = FailedPrecondition desc = Failed health check (verify
 network connection and IAM credentials): dry-run EC2 API call failed: operation error EC2: DescribeAvailabilityZones, https response
```

ConnorJC3 (approved):
/lgtm

PR #2591, opened 2025-07-31
Add EC2 metadata-labeler metadata source

sylviah23:
#### What type of PR is this?
/kind feature

This PR gets the Volume and ENI count of each node via the EC2 API. It then adds this information to the metadata labels of each node in the cluster, so Kubernetes can get this information without IMDS access.

Unit testing and E2E testing and manually testing with EKS cluster by adding a new node.

```release-note
Add ALPHA metadata-labeler sidecar and metadata source
```


Makefile
```diff
@@ -239,6 +239,12 @@ sub-push-a1compat:
 .PHONY: all-push
 all-push: sub-push sub-push-fips sub-push-a1compat
 
+# QUESTION: should my tests be run in the current CI jobs or a separate job
+
+.PHONY: ec2Labels-test
+ec2Labels-test:
+	cd tests/e2e && ginkgo -v -focus="disruptive"
```
AndrewSirenko: Low priority, but we should follow what we do for other `e2e/test-name` targets (see lines 128-198) instead of CD-ing into tests/e2e and running ginkgo. 

(Also nit, but we use dash-case instead of camelCase for makefile targets)

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -90,6 +90,14 @@ sidecars:
         type: RuntimeDefault
       readOnlyRootFilesystem: true
       allowPrivilegeEscalation: false
+  patcher:
+    # Additional parameters provided by livenessprobe.
```
AndrewSirenko: Provided by livenessprobe?

cmd/main.go
```diff
@@ -81,8 +86,71 @@ func main() {
 		klog.FlushAndExit(klog.ExitFlushTimeout, 0)
 	case string(driver.ControllerMode), string(driver.NodeMode), string(driver.AllMode):
 		options.Mode = driver.Mode(cmd)
+
+	// TODO question: I think it makes sense to check if cmd == "patcher" here because this is where the other
+	// cmd checks are. However, this would result in a lot of code duplication as the patcher container needs
```
AndrewSirenko: Yeah I think we can:
- break the pre-stop-hook case out into an if statement
- Move most of the shared boilerplate below pre-stop-hook case but above switch statement

Does team have any other thoughts?

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		lockName      = "my-lock"
+		lockNamespace = "kube-system"
+	)
+
+	l, err := rl.NewFromKubeconfig(
+		rl.LeasesResourceLock,
+		lockNamespace,
+		lockName,
+		rl.ResourceLockConfig{
+			Identity: instanceID, //TODO: question: what's a better identity to use
```
AndrewSirenko: Perhaps podname via [Downward API | Kubernetes](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/)? Thoughts @ConnorJC3 ?

pkg/cloud/metadata/k8s.go
```diff
@@ -107,6 +110,39 @@ func KubernetesAPIInstanceInfo(clientset kubernetes.Interface) (*Metadata, error
...
+	volumes := 0
+
+	if ec2Labels {
+		backoff := wait.Backoff{
+			Duration: 1 * time.Second,
+			Factor:   1.5,
+			Steps:    6,
+		}
+
+		//TODO: need to shorten to 5 seconds later
```
AndrewSirenko: Looks like you already shortened this

pkg/cloud/metadata/k8s.go
```diff
@@ -146,9 +182,52 @@ func KubernetesAPIInstanceInfo(clientset kubernetes.Interface) (*Metadata, error
...
+		if err != nil {
+			klog.ErrorS(err, "failed to convert number of ENIs label to int")
+			return 1, err
+		}
+	} else {
+		klog.V(2).InfoS("label not found on node", "node", node.Name, "label", ENIsLabel)
+		return 1, fmt.Errorf("%s label not found on node", ENIsLabel)
+	}
+	return enis, nil
+}
```
AndrewSirenko, line 266: I'm seeing some duplicate logic here (getting label value as int, returning errors, etc.). Perhaps something like the following would be a bit cleaner? 

```
func getEC2ENIsVolumes(node *corev1.Node) (int, int, error) {
	eni, err := getLabelAsInt(node, ENIsLabel, 1)
	if err != nil {
		return 0, 0, err
	}
	vol, err := getLabelAsInt(node, VolumesLabel, 0)
	if err != nil {
		return 0, 0, err
	}
	return eni, vol, nil
}

func getLabelAsInt(node *corev1.Node, label string, defaultValue int) (int, error) {
	val, ok := node.GetLabels()[label]
	if !ok {
		klog.V(2).InfoS("label not found on node", "node", node.Name, "label", label)
		return defaultValue, fmt.Errorf("%s label not found on node", label)
	}

	result, err := strconv.Atoi(val)
	if err != nil {
		klog.ErrorS(err, "failed to convert label to int", "label", label)
		return defaultValue, err
	}
	return result, nil
}
```

pkg/cloud/metadata/k8s.go
```diff
@@ -107,6 +110,39 @@ func KubernetesAPIInstanceInfo(clientset kubernetes.Interface) (*Metadata, error
 		return nil, fmt.Errorf("error getting Node %v: %w", nodeName, err)
 	}
 
+	enis := 1
```
AndrewSirenko: There used to be a comment explaining `// All nodes have at least 1 attached ENI, so we'll use that`. Can we add it back here? 

And maybe add a comment that we already factor in root volume elsewhere

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+
+const (
+	// VolumesLabel is the label name for the number of volumes on a node.
+	VolumesLabel = "ebs.csi.aws.com/non-csi-ebs-volumes-count"
+
+	// ENIsLabel is the label name for the number of ENIs on a node.
+	ENIsLabel = "ebs.csi.aws.com/enis-count"
+
+	// RenewDeadline is lease duration of the resource lock for leader election to update Nodes with additional ebs-csi-driver metadata labels.
+	RenewDeadline = 10
```
AndrewSirenko: Nit: Any reason this is exported? And I think you forgot to use it lower in file.

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+			if *vol.Ebs.VolumeId == pv {
+				nonCSIVolumes -= 1
+			}
+		}
+	}
+	return nonCSIVolumes
+}
+
+func isEbsCsiVolume(pv *v1.PersistentVolume) bool {
+	if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == "ebs.csi.aws.com" {
```
AndrewSirenko: Medium priority: There are also volumes that have migrated to ebs csi that we have to worry about.

I believe migrated PVs have an extra label: pv.kubernetes.io/migrated-to : ebs.csi.aws.com

BOTH regular ebs csi volumes AND migrated volumes should be counted as isEbsCsiVolume.

We'll also need a unit test for this.

pkg/cloud/metadata/metadata.go
```diff
@@ -71,9 +73,24 @@ func NewMetadataService(cfg MetadataServiceConfig, region string) (MetadataServi
...
 				klog.ErrorS(err, "Retrieving IMDS metadata failed")
 			}
+		case SourceEC2LabelsK8s:
+			klog.V(2).InfoS("Attempting to retrieve instance metadata from ec2Labels Kubernetes API")
+			metadata, err := retrieveK8sMetadata(cfg.K8sAPIClient, true)
+			if err == nil {
+				klog.V(2).InfoS("Retrieved metadata from ec2Labels Kubernetes")
+				return metadata.overrideRegion(region), nil
+			}
+			klog.ErrorS(err, "Retrieving ec2Labels Kubernetes metadata failed, fallback to Kubernetes metadata")
```
AndrewSirenko: We don't know if we're falling back to Kubernetes metadata because customer can override order. Would change to "Retrieving ec2Labels Kubernetes metadata failed"

pkg/cloud/metadata/metadata.go
```diff
@@ -71,9 +73,24 @@ func NewMetadataService(cfg MetadataServiceConfig, region string) (MetadataServi
...
+			klog.V(2).InfoS("Attempting to retrieve instance metadata from Kubernetes API")
+			metadata, err = retrieveK8sMetadata(cfg.K8sAPIClient, false)
+			if err == nil {
+				klog.V(2).InfoS("Retrieved metadata from Kubernetes")
+				return metadata.overrideRegion(region), nil
+			}
+			klog.ErrorS(err, "Retrieving Kubernetes metadata failed")
```
AndrewSirenko: We don't have to explicitly fallback to K8sMetadata from inside this case, because the larger for loop will automatically fallback to SourceK8s case. 

Can we delete these lines of code?

charts/aws-ebs-csi-driver/templates/clusterrole-patcher.yaml
```diff
@@ -0,0 +1,11 @@
...
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: label-patcher
+rules:
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["patch"]
+{{- end -}}
```
ConnorJC3: np: missing newline at the end of this file

This also affects `clusterrolebinding-patcher.yaml`

charts/aws-ebs-csi-driver/templates/clusterrole-patcher.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if not .Values.nodeComponentOnly -}}
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: label-patcher
```
ConnorJC3: Please include the Chart labels, example here: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/aa63b9cc120a5f6a12c570d7d995c82d07e30cff/charts/aws-ebs-csi-driver/templates/clusterrole-resizer.yaml#L7-L8

charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -422,6 +422,38 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
         {{- end }}
+        - name: patcher
+          image: {{ include "aws-ebs-csi-driver.fullImagePath" $ }} #TODO: question: should we have this be separate image path with just this extra container
```
ConnorJC3: > should we have this be separate image path with just this extra container

Did we not decide on in-container in the design? Or am I misremembering?

charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -422,6 +422,38 @@ spec:
...
+            - patcher
+            - --v={{ .Values.sidecars.patcher.logLevel }}
+            {{- range .Values.sidecars.patcher.additionalArgs }}
+            - {{ . }}
+            {{- end }}
+          env:
+            - name: CSI_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
```
ConnorJC3, line 506: Is this necessary in the patcher sidecar? I don't think it should be

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -90,6 +90,14 @@ sidecars:
...
+    logLevel: 4
```
ConnorJC3: Probably should be `2` which is our default everywhere else

docs/install.md
```diff
@@ -22,6 +22,10 @@ If the driver is able to access IMDS, it will utilize that as a preferred source
 
 In order for the driver to access IMDS, it either must be run in host networking mode, or with a [hop limit of at least 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-existing-instances.html#modify-PUT-response-hop-limit).
 
+#### EC2 Labels Kubernetes Metadata
+
+TODO: placeholder
```
ConnorJC3: np: don't forget to update this

Makefile
ConnorJC3: Will this even work as is? It won't deploy the driver right?

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	var resp *ec2types.Instance
+	var err error
+	var respList []*ec2types.Instance
+
+	if len(nodeIds) > 1 {
+		respList, err = cloud.GetInstances(ctx, nodeIds)
+	} else if len(nodeIds) == 1 {
+		resp, err = cloud.GetInstance(ctx, nodeIds[0])
+		respList = []*ec2types.Instance{resp}
+	}
```
ConnorJC3: This feels overly complicated, why not just always call `GetInstances`? Ultimately they both end up making an EC2 `DescribeInstances` call anyways

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+
+	if err != nil {
+		klog.ErrorS(err, "failed to describe instances")
```
ConnorJC3: np: Log messages should begin with a capital letter. Not going to comment it a dozen times but this file has several other `klog` entries with the same issue.

(Note however that messages in error objects should start with a lowercase letter, only log messages should be capitalized)

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+// parseNodes gets the instance name from parsing the provider ID.
+func parseNode(providerID string) string {
+	if providerID != "" {
+		parts := strings.Split(providerID, "/")
+		if len(parts) > 0 {
+			return parts[len(parts)-1]
+		}
+	}
+	return ""
+}
```
ConnorJC3: Maybe consider regex-ing out the intance ID like we do in the metadata code? https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/aa63b9cc120a5f6a12c570d7d995c82d07e30cff/pkg/cloud/metadata/k8s.go#L115-L118

pkg/cloud/metadata/labels.go
ConnorJC3: Couldn't we just check `pv.Spec.AwsElasticBlockStore != nil` for migrated volumes?

docs/options.md
```diff
@@ -20,4 +20,4 @@ There are a couple of driver options that can be passed as arguments when starti
...
-| metadata-sources                      | imds         | imds,kubernetes                                  | Dictates which sources are used to retrieve instance metadata. The driver will attempt to rely on each source in order until one succeeds. Valid options include 'imds' and 'kubernetes'.                                                                                                                                                                                                                                                    |
+| metadata-sources                      | imds         | imds,kubernetes,ec2labelskubernetes                                  | Dictates which sources are used to retrieve instance metadata. The driver will attempt to rely on each source in order until one succeeds. Valid options include 'imds', 'ec2labelskubernetes', and 'kubernetes'.                                                                                                                                                                                                                                                    |
```
ConnorJC3: I'm not sure if this PR is the best place to have it, but we need to have a discussion about the name for this source. I think "ec2labelskubernetes" is hard for customers to understand (what is an "EC2 label", I have the full context and I'm not even sure).

I suggest something like `kubernetes-enhanced` or `kubernetes-with-patcher` or something like that.

cmd/main.go
ConnorJC3: We don't actually need the instance ID for the patcher right, just the region?

It's a little bit janky but what if we do something like

```
var (
	metadataRequiredModes = map[string]struct{}{
		driver.ControllerMode: {},
		driver.NodeMode: {},
		driver.AllMode:  {},
		"patcher": {},
	}
)
```

Then, move the metadata above the switch statement like:
```
if(_, ok := metadataRequiredModes[cmd]) {
	// Put metadata stuff here
}

switch cmd {
// existing logic goes here
```

There's also precedence for doing things like this: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/aa63b9cc120a5f6a12c570d7d995c82d07e30cff/pkg/driver/node.go#L57-L64

pkg/cloud/cloud.go
```diff
@@ -1664,6 +1665,31 @@ func (c *cloud) getInstance(ctx context.Context, nodeID string) (*types.Instance
...
+// GetInstances returns the instance info associated with each node ID in `nodeIDs`.
+func (c *cloud) GetInstances(ctx context.Context, nodeIDs []string) ([]*types.Instance, error) {
+	request := &ec2.DescribeInstancesInput{
+		InstanceIds: nodeIDs,
+	}
+
+	instances := []*types.Instance{}
+	// A batcher isn't used here because GetInstances is only called once in awhile to update
+	// metadata labels of each node.
+	response, err := c.ec2.DescribeInstances(ctx, request)
```
ConnorJC3: What if there are 10k, 20k, or more instances in the cluster? We need to deal with pagination here I think, and maybe cap the number of IDs in the call too.

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	for _, instance := range respList {
+		numAttachedENIs := 1
+		if instance.NetworkInterfaces != nil {
+			numAttachedENIs = len(instance.NetworkInterfaces)
+		}
+		numBlockDeviceMappings := 0
+		if instance.BlockDeviceMappings != nil {
+			// -1 for root volume because we eventually add this back in when calculating allocatable count in getVolumesLimit()
+			nonCSIManagedVolumes := getNonCSIManagedVolumes(pvInformer, instance.BlockDeviceMappings)
+			numBlockDeviceMappings = nonCSIManagedVolumes - 1
```
ConnorJC3: creating a variable just to subtract 1 from it is unnecessary, collapse to
```
numBlockDeviceMappings = getNonCSIManagedVolumes(pvInformer, instance.BlockDeviceMappings) - 1
```

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		instanceID := *instance.InstanceId
+		ENIsVolumesMap[instanceID] = enisVolumes{ENIs: numAttachedENIs, Volumes: numBlockDeviceMappings}
```
ConnorJC3: same thing - no need to create a variable for a one-off use on the next line, collapse to
```
ENIsVolumesMap[*instance.InstanceId] = ...
```

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		resp, err = cloud.GetInstance(ctx, nodeIds[0])
+		respList = []*ec2types.Instance{resp}
+	}
+
+	if err != nil {
+		klog.ErrorS(err, "failed to describe instances")
+		return nil, err
+	}
+
+	ENIsVolumesMap := make(map[string]enisVolumes)
```
ConnorJC3: Non-exported go variables should start with a lowercase letter
Also, as a minor optimization, suggest `make(map[string]enisVolumes, len(respList))` to avoid unnecessary memory allocations

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, &v1.Node{})
+		if err != nil {
+			klog.ErrorS(err, "failed to create two way merge", "node", node.Name)
+			return err
+		}
+		if _, err := clientset.CoreV1().Nodes().Patch(ctx, node.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{}); err != nil {
+			klog.ErrorS(err, "Failed to patch node", "node", node.Name)
+			return err
+		}
+		klog.V(4).InfoS("patched labels", "node", node.Name, "num volumes label", VolumesLabel, "num ENIs label", ENIsLabel)
```
ConnorJC3: Bump this up to something like log level 6 - it is going to absolutely flood the logs

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+			return err
+		}
+		patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, &v1.Node{})
+		if err != nil {
+			klog.ErrorS(err, "failed to create two way merge", "node", node.Name)
+			return err
+		}
+		if _, err := clientset.CoreV1().Nodes().Patch(ctx, node.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{}); err != nil {
+			klog.ErrorS(err, "Failed to patch node", "node", node.Name)
+			return err
```
ConnorJC3: Returning here feels bad - it means we basically bail the entire operation if one patch fails (and therefore will probably spend another DIs call). What about just doing a `continue`, or maybe storing the number of failed patches and only bailing if it hits something like 3-5?

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		instanceID := *instance.InstanceId
+		ENIsVolumesMap[instanceID] = enisVolumes{ENIs: numAttachedENIs, Volumes: numBlockDeviceMappings}
+	}
+
+	return ENIsVolumesMap, nil
+}
+
+// patchNodes patches the labels of each node to have the number of ENIs and non-CSI managed volumes attached to each node.
+func patchNodes(ctx context.Context, nodes *v1.NodeList, enisVolumeMap map[string]enisVolumes, clientset kubernetes.Interface) error {
+	for _, node := range nodes.Items {
```
ConnorJC3: Have we tested how fast this is on large clusters? Do we need a worker group here or some similar construct?

pkg/cloud/metadata/k8s.go
```diff
@@ -22,28 +22,31 @@ import (
...
 )
 
-type KubernetesAPIClient func() (kubernetes.Interface, error)
+type KubernetesAPIClient func() (kubernetes.Interface, *rest.Config, error)
 
 func DefaultKubernetesAPIClient(kubeconfig string) KubernetesAPIClient {
-	return func() (clientset kubernetes.Interface, err error) {
-		var config *rest.Config
+	return func() (clientset kubernetes.Interface, cfg *rest.Config, err error) {
+		var config *rest.Config // needed for leader election
```
ConnorJC3: Can we clarify what exactly is needed for leader election?

1. All we do here is construct a default `rest.Config` object, so couldn't we just do that wherever it's needed for the leader election?
2. As far as I can tell, we don't use this in the leader election code at all - is it even needed?

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+
+type enisVolumes struct {
+	ENIs    int
+	Volumes int
+}
+
+// Uses leader election so that only one controller pod calls containuousUpdateLabels().
+func ContinuousUpdateLabelsLeaderElection(clientset kubernetes.Interface, k8sConfig *rest.Config, instanceID string, cloud cloud.Cloud, updateTime int) {
+	var (
+		lockName      = "my-lock"
```
ConnorJC3: This should be something like `node-patcher-ebs.csi.aws.com` which would be similar to what the official sidecars do

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		lockNamespace = "kube-system"
```
ConnorJC3: We probably need a CLI parameter for this because users can theoretically install the driver in whatever namespace they want (I believe some installers like gardener at least used to use a custom namespace)

pkg/cloud/metadata/labels.go
ConnorJC3: We should probably consider using the CSI Lib Utils wrapper library like everyone else: https://github.com/kubernetes-csi/csi-lib-utils/blob/master/leaderelection/leader_election.go

Even if not, suggest `os.Hostname()` as the identity like they do

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		RenewDeadline: time.Second * 10,
+		RetryPeriod:   time.Second * 2,
+		Name:          lockName,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				klog.V(2).InfoS("This container became the leader for updating Nodes with additional ebs-csi-driver metadata labels")
+				continuousUpdateLabels(ctx, clientset, cloud, updateTime)
+			},
+			OnStoppedLeading: func() {
+				klog.V(2).InfoS("This container is no longer the leader for updating Nodes with additional ebs-csi-driver metadata labels")
```
ConnorJC3: We need to actually do something here to stop updating (Another good reason to use the lib utils library)

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	nodeInformer := metadataInformer(ctx, k8sClient, cloud, pvInformer)
+	nodeInformerStopCh := make(chan struct{})
+	nodeInformer.Start(nodeInformerStopCh)
+	nodeInformer.WaitForCacheSync(nodeInformerStopCh)
+
+	ticker := time.NewTicker(time.Duration(updateTime) * time.Minute)
+
+	defer ticker.Stop()
+	updateLabels(ctx, k8sClient, cloud, pvInformer)
+	for range ticker.C {
```
ConnorJC3: Can be simplified to eliminate ticker variable entirely:
```
for range time.Tick(/* time to tick here */) {
```

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	el.Run(context.Background())
+}
+
+// continuousUpdateLabels is a go routine that updates the metadata labels of each node once every
+// `updateTime` minutes and uses an informer to update the labels of new nodes that join the cluster.
+// A PV informer is also used to keep track of CSI managed volumes when updating labels to avoid
+// double counting.
+func continuousUpdateLabels(ctx context.Context, k8sClient kubernetes.Interface, cloud cloud.Cloud, updateTime int) {
```
ConnorJC3: Don't take the time as an `int`, take it as a `time.Time` object

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	pvInformerFactory, pvInformer := pvInformer(k8sClient)
```
ConnorJC3: Suggest refactoring `pvInformer` to take factory as an argument rather than return it, as it's weird to return the factory in one informer function but not the other

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	pvInformerStopCh := make(chan struct{})
+	pvInformerFactory.Start(pvInformerStopCh)
+	pvInformerFactory.WaitForCacheSync(pvInformerStopCh)
+
+	nodeInformer := metadataInformer(ctx, k8sClient, cloud, pvInformer)
+	nodeInformerStopCh := make(chan struct{})
+	nodeInformer.Start(nodeInformerStopCh)
+	nodeInformer.WaitForCacheSync(nodeInformerStopCh)
```
ConnorJC3: Don't start each informer separately, call `Start` on the factory. Also, pass in `ctx.Done()` as the channel, don't make a channel that will never be used.

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	return volumeHandles
+}
+
+func getNonCSIManagedVolumes(pvInformer cache.SharedIndexInformer, volumes []ec2types.InstanceBlockDeviceMapping) int {
+	nonCSIVolumes := len(volumes)
+	pvs := getPvVolumeIDs(pvInformer)
+	for _, vol := range volumes {
+		for _, pv := range pvs {
+			if *vol.Ebs.VolumeId == pv {
```
ConnorJC3: Can `vol.Ebs` be `nil` (e.g. instance store volume)? Do we need a check for that?

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+		return true
+	}
+
+	return false
+}
+
+// pvInformer creates an informer that watches for CSI managed EBS volumes.
+func pvInformer(clientset kubernetes.Interface) (informers.SharedInformerFactory, cache.SharedIndexInformer) {
+	factory := informers.NewSharedInformerFactoryWithOptions(clientset, 0)
+	pvInformer := factory.Core().V1().PersistentVolumes().Informer()
```
ConnorJC3: Unfortunately it looks like we can't use it yet, but we should filter this after https://github.com/kubernetes/kubernetes/issues/115803 is implemented

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	_, err := pvInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(newObj interface{}) {
+			if pv, ok := newObj.(*v1.PersistentVolume); ok {
+				if !isEbsCsiVolume(pv) {
+					if err := pvInformer.GetStore().Delete(newObj); err != nil {
```
ConnorJC3: I have no hard evidence, but I am very worried this is unsafe. Are we sure that `Delete` is thread safe and that it's ok to mess with the informer internals in its own handlers? Is it even worth doing this to save a few kB?

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+						klog.ErrorS(err, "unable to delete pv from pv informer cache")
```
ConnorJC3: If we end up keeping this, remove the error log or set it to super high verbosity, as errors here are almost certainly irrelevant to the operation of the patcher as a whole

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+				if !isEbsCsiVolume(pv) {
+					if err := pvInformer.GetStore().Delete(obj); err != nil {
+						klog.ErrorS(err, "unable to delete pv from pv informer cache")
+					}
+				}
+			}
+		},
+	})
+	if err != nil {
+		klog.ErrorS(err, "unable to add event handler for pv informer")
```
ConnorJC3: This function should return the error here so it can be handled up the stack and not ignored

pkg/cloud/metadata/metadata.go
ConnorJC3: +1 as designed we should not hardcode a fallback, also we should return the error

cmd/main.go
```diff
@@ -40,6 +40,11 @@ var (
 	featureGate = featuregate.NewFeatureGate()
 )
 
+const (
+	// LabelRefreshTime is the time in minutes that it takes for node labels to update volume and ENI count.
+	LabelRefreshTime = 60
```
ConnorJC3: Probably shouldn't be exported

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,335 @@
...
+	ticker := time.NewTicker(time.Duration(updateTime) * time.Minute)
+
+	defer ticker.Stop()
+	updateLabels(ctx, k8sClient, cloud, pvInformer)
+	for range ticker.C {
+		updateLabels(ctx, k8sClient, cloud, pvInformer)
+	}
+}
+
+func getPvVolumeIDs(pvInformer cache.SharedIndexInformer) []string {
```
ConnorJC3: I'm not an expert on informers, but I think we could use an index to prevent the below quadratic algorithm in `getNonCSIManagedVolumes`.

See `AddIndexers`: https://pkg.go.dev/k8s.io/client-go@v0.33.3/tools/cache#SharedIndexInformer

pkg/cloud/metadata/metadata_test.go
```diff
@@ -115,29 +137,97 @@ func TestNewMetadataService(t *testing.T) {
...
-							corev1.LabelTopologyZone:       "us-west-2a",
-						},
-					},
-					Spec: corev1.NodeSpec{
-						ProviderID: "aws:///us-west-2a/i-1234567890abcdef0",
-					},
-				}
-				return fake.NewSimpleClientset(node), nil
+				node := tc.node
+				return fake.NewSimpleClientset(node), nil, nil
```
ConnorJC3: Collapse to
```
return fake.NewSimpleClientset(tc.node), nil, nil
```

pkg/cloud/metadata/labels_test.go
```diff
@@ -0,0 +1,405 @@
...
+				},
+				Spec: corev1.NodeSpec{
+					ProviderID: "example/i-001",
+				},
+			},
+			expectedMetadata: map[string]enisVolumes{
+				"i-001": {ENIs: 2, Volumes: 2},
+			},
+			expErr: nil,
+		},
```
ConnorJC3, line 62: Only a success test case?

pkg/cloud/metadata/labels_test.go
```diff
@@ -0,0 +1,405 @@
...
+				if err != nil {
+					return false, nil, err
+				}
+				close(pvWatcherStarted)
+				return true, watch, nil
+			})
+			pvInformerFactory, pvInformer := pvInformer(pvClientset)
+			pvInformerFactory.Start(ctx.Done())
+			cache.WaitForCacheSync(ctx.Done())
+			<-pvWatcherStarted
```
ConnorJC3: Lot of duplication on setting up watchers for tests, can we make a function for that?

pkg/cloud/metadata/labels_test.go
```diff
@@ -0,0 +1,405 @@
...
+			nodeInformer.Start(ctx.Done())
+			cache.WaitForCacheSync(ctx.Done())
+			<-nodeWatcherStarted
+
+			_, err := nodeClientset.CoreV1().Nodes().Create(t.Context(), tc.newNode, metav1.CreateOptions{})
+			if err != nil {
+				t.Fatalf("error injecting node add: %v", err)
+			}
+
+			time.Sleep(5e8)
```
ConnorJC3: Use something like `0.5 * time.Second` or `500 * time.Millisecond`, I don't know what `5e8` nanoseconds is off the top of my head

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
...
+			By("Deleting the EBS CSI node pod to trigger label update")
+			deletePod(labeledMetadata[firstChangedNonCSIVolumeInstance].NodeID, cs)
+			checkCSINodesUpdated(cs, labeledMetadata, expectedMetadata)
+
+			// For the instance that the new volume is attached to, the volume labels and allocatable count should not change
+			By("Creating and attaching a new CSI managed volume")
+			createStorageClass(cs)
+			cleanUp = append(cleanUp, func() { cleanUpStorageClass(cs, "ebs-sc") })
+			pvc := createPVC(cs)
+			cleanUp = append(cleanUp, func() { cleanUpPVC(cs, "default", "ebs-claim") })
```
ConnorJC3: Don't hardcode the `default` namespace - use the namespace the framework provides for you (i.e. `f.Namespace` object)

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
...
+	var (
+		ec2Client        *ec2.Client
+		cs               clientset.Interface
+		expectedMetadata map[string]*metadata
+		labeledMetadata  map[string]*metadata
+		cleanUp          []func()
+	)
+
+	BeforeEach(func() {
+		cfg, err := config.LoadDefaultConfig(context.TODO())
```
ConnorJC3: Use `context.Background()` instead (ditto everywhere else `TODO` is used in this file, unless it's a situation where `Background` is unsuitable and you're intentionally leaving it as TODO)

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
...
+			cleanUp = append(cleanUp, func() { cleanUpPod(cs, "default", "app") })
+			changedCSIVolumeInstance := createCSIManagedVolume(cs, pvc, pod)
+
+			// Because the previous step should not change volume labels/allocatable count, we add a non CSI managed volume to know that the
+			// volume labels/allocatable count updated accordingly
+			By("Creating a new non CSI managed volume")
+			changedNonCSIVolumeInstance, createdNonCSIVolumeID := createNonCSIManagedVolume(ec2Client, expectedMetadata, changedCSIVolumeInstance)
+			cleanUp = append(cleanUp, func() { cleanUpVolume(createdNonCSIVolumeID, changedNonCSIVolumeInstance, ec2Client, expectedMetadata) })
+			By("Attaching the new non CSI managed volume")
+			attachVolume(ec2Client, createdNonCSIVolumeID, changedNonCSIVolumeInstance, "/dev/sdh", expectedMetadata)
```
ConnorJC3: Suggest `/dev/sdz` to minimize the chance this conflicts with an existing volume (from other tests or something like that).

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
...
+
+const (
+	// VolumesLabel is the label name for the number of volumes on a node
+	VolumesLabel = "ebs.csi.aws.com/non-csi-ebs-volumes-count"
+
+	// ENIsLabel is the label name for the number of ENIs on a node
+	ENIsLabel = "ebs.csi.aws.com/enis-count"
+)
+
+var _ = Describe("[disruptive] EBS CSI Driver Node Labeling", func() {
```
ConnorJC3: np: Capitalize `Disruptive`

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
...
+}
+
+func cleanUpStorageClass(cs kubernetes.Interface, name string) {
+	By("Deleting StorageClass")
+	err := cs.StorageV1().StorageClasses().Delete(context.TODO(), name, metav1.DeleteOptions{})
+	Expect(err).NotTo(HaveOccurred())
+
+	Eventually(func() bool {
+		_, err := cs.StorageV1().StorageClasses().Get(context.TODO(), name, metav1.GetOptions{})
+		return err != nil
```
ConnorJC3: Can we use `errors.IsNotFound(err)` from https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors instead to avoid false positives? (Ditto for the other versions of this function)

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
...
+	Volumes          int
+	InstanceType     string
+	AllocatableCount int32
+	NodeID           string
+	AvailabilityZone string
+}
+
+const (
+	// VolumesLabel is the label name for the number of volumes on a node
+	VolumesLabel = "ebs.csi.aws.com/non-csi-ebs-volumes-count"
```
ConnorJC3: Get this from the `cloud` package (I see it's exported there) rather than defining it twice (ditto with the other label)

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+	return ""
+}
+
+func createStorageClass(cs kubernetes.Interface) *v1.StorageClass {
+	storageClass := &storagev1.StorageClass{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "storage.k8s.io/v1",
+			Kind:       "StorageClass",
+		},
```
ConnorJC3: Remove `TypeMeta` from your create calls, it isn't needed (and actually will break the code if for example there's ever a V2 StorageClass)

pkg/cloud/cloud.go
```diff
@@ -1664,6 +1665,31 @@ func (c *cloud) getInstance(ctx context.Context, nodeID string) (*types.Instance
 	}
 }
 
+// GetInstances returns the instance info associated with each node ID in `nodeIDs`.
+func (c *cloud) GetInstances(ctx context.Context, nodeIDs []string) ([]*types.Instance, error) {
```
ConnorJC3: Maybe this should be called something like `GetInstancesUnbatched` or `GetInstancesForPatching` so that a hapless developer doesn't in the future use this function thinking it's a generic function like `GetInstance` is

tests/e2e/metadata-labeler.go
```diff
@@ -0,0 +1,561 @@
...
+	}, "2m", "2s").Should(BeTrue(), "CSI node allocatable count were not updated with correct count")
+}
+
+func createNonCSIManagedVolume(ec2svc *ec2.Client, metadata map[string]*metadata, changedCSIVolumeInstance string) (string, string) {
+	var instanceID string
+	if changedCSIVolumeInstance != "" {
+		instanceID = changedCSIVolumeInstance
+	} else {
+		for k := range metadata {
+			instanceID = k
```
ConnorJC3, line 324: np: put a `break` after this line
also maybe a comment indicating you're choosing a random instance assuming that's the intent

tests/e2e/ec2-metadata.go
```diff
@@ -0,0 +1,561 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
ConnorJC3: np: 2023 -> 2025

charts/aws-ebs-csi-driver/templates/clusterrole-patcher.yaml
```diff
@@ -0,0 +1,11 @@
...
+rules:
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["patch"]
```
ConnorJC3: Don't we also need `list`/`watch`/maybe `get` on PVs for the informer?

mdzraf:
/test pull-aws-ebs-csi-driver-image-test

Fixed issue with ECR repo being full.

ConnorJC3 (commented):
First review round: I have *not* manually tested yet, intend to do that in my next review. I also don't think I was as in depth with looking at some parts (in particular the tests) as I should have been, so I might have missed some bugs lurking there.

The comments about the leader election are probably the most critical comments in this review - as written, this code will keep running after the leader election is switched away, causing two or more instances of the driver to simultaneously run the patcher and race each other.

pkg/cloud/metadata/labels.go
AndrewSirenko: Oh right now that we're not in same container as controller service batching isn't an optimization

docs/options.md
AndrewSirenko: IMO ec2 somewhere in name is helpful to help realize this comes directly from EC2 api. IMO Kubernetes enhanced would not help new customer realize what's going on at all. 

But yeah agreed post-scrum is a better forum, but also name is something that can be changed near end.

ConnorJC3: Does it come any more directly from the EC2 API than the standard Kubernetes metadata?

With this solution it's basically
Driver -> Kubernetes API -> Patcher -> EC2

With the existing metadata it's
Driver -> Kubernetes API -> AWS CCM -> EC2

pkg/cloud/metadata/interface.go
```diff
@@ -30,7 +31,7 @@ type MetadataService interface {
 	GetNumAttachedENIs() int
 	GetNumBlockDeviceMappings() int
 	GetOutpostArn() arn.ARN
-	UpdateMetadata() error
+	UpdateMetadata(k8sClient kubernetes.Interface) error
```
torredil: At this time, no changes to the MetadataService interface are needed (k8sClient isn't actually used in UpdateMetadata).

docs/options.md
```diff
@@ -20,4 +20,4 @@ There are a couple of driver options that can be passed as arguments when starti
...
 | legacy-xfs                            | true                    | false                                            | Warning: This option will be removed in a future release. It is a temporary workaround for users unable to immediately migrate off of older kernel versions. Formats XFS volumes with `bigtime=0,inobtcount=0,reflink=0`, so that they can be mounted onto nodes with linux kernel ≤ v5.4. Volumes formatted with this option may experience issues after 2038, and will be unable to use some XFS features (for example, reflinks).         |
-| metadata-sources                      | imds         | imds,kubernetes                                  | Dictates which sources are used to retrieve instance metadata. The driver will attempt to rely on each source in order until one succeeds. Valid options include 'imds' and 'kubernetes'.                                                                                                                                                                                                                                                    |
```
torredil, line 23: Please document that this new metadata source is being introduced as an alpha feature.

pkg/driver/node.go
```diff
@@ -79,10 +79,11 @@ const (
...
 type NodeService struct {
-	metadata metadata.MetadataService
-	mounter  mounter.Mounter
-	inFlight *internal.InFlight
-	options  *Options
+	metadata  metadata.MetadataService
+	mounter   mounter.Mounter
+	inFlight  *internal.InFlight
+	options   *Options
+	k8sClient kubernetes.Interface
```
torredil: We shouldn't update this struct to include `k8sClient` (it is not used by `UpdateMetadata` anyway).

`k8sClient` should be a property of `Metadata`. So, instead of injecting it into the Node service (to pass it in to `UpdateMetadata`), if `UpdateMetadata` needs access to the kubernetes client, it should access it similar to how it [already accesses the IMDS client](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/aa63b9cc120a5f6a12c570d7d995c82d07e30cff/pkg/cloud/metadata/metadata.go#L94).

torredil (commented):
Thank you for the contribution and adding comprehensive code coverage on the first revision @sylviah23 . Overall, this is looking solid 🙂.

The leader election code needs some work, as Connor suggested https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2591#discussion_r2247746564, leveraging the CSI lib-utils wrapper will be a nice improvement. That code is battle tested and already addresses some of the core challenges we need to solve implementing this enhancement.

Makefile
```diff
@@ -239,6 +239,17 @@ sub-push-a1compat:
 .PHONY: all-push
 all-push: sub-push sub-push-fips sub-push-a1compat
 
+# QUESTION: should my tests be run in the current CI jobs or a separate job
```
AndrewSirenko: Team can resolve this after internship maybe. But I'm leaning towards separate job just in case folks add future disruptive tests. Lets delete comment but add this to your list of "Post internship action items" along with the Eddie KEP enhancement to your project.

charts/aws-ebs-csi-driver/templates/controller.yaml
AndrewSirenko: I think what Sylvia means here is whether we want to have a SEPARATE values.yaml variable as image path for patcher container in case folks want to test separate images for driver and patcher. 

Sylvia is this what you meant? If so I think for now it's clearer to just share aws-ebs-csi-driver.fullImagePath, it's a 2 way door and we won't touch this feature too too much. 

Otherwise connor is right we decided in-container in design so we can delete todo.

charts/aws-ebs-csi-driver/templates/controller.yaml
sylviah23: I think that in case IMDS is disabled we would need the controller to get metadata through Kubernetes metadata, which I think needs the CSI_NODE_NAME, as my patcher code requires the region to work.

pkg/cloud/metadata/labels_test.go
sylviah23: As of the newest commit this test function is just testing to make sure that the node informer patches on node labels for newly added nodes. I can't think of another test to add, but open to suggestions!

Makefile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 The Kubernetes Authors.
+# Copyright 2025 The Kubernetes Authors.
```
AndrewSirenko, line 1: PR-wide comment: Observed the following race:

- Install driver, labels applied to node
- Attach additional ENI
- Reinstall/upgrade driver
- Node pods get old labels BEFORE controller calculates and posts new labels

This leads to incorrect CSINode allocatable

On bright side:
  - Shouldn't happen too often in practice because it means ENI was attached after hourly poll but before driver upgrade.
  - Helped out by eddie's KEP, where after x minutes Kubelet will get new correct allocatables count.

AndrewSirenko:
Hmm doesn't look like flake, opened [Patcher test d4c5](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2611) to see if your latest commit introduced a regression for default helm params! As well as a fake PR against master to see if CI is bugging.

charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -428,6 +428,38 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
         {{- end }}
+        - name: patcher
```
AndrewSirenko: Because we are deciding to call this an alpha feature, lets hide some of these manifests behind a feature gate set default to false in values.yaml. That way if there is some scalability issue due to the new container, customers have a way to turn it off. 

Maybe something like `controller.deployPatcherContainer`

You would wrap the relevant helm templates with a helm `{{ if ... }`  statements similar to what's used in relevant code.

You'd also wrap the clusterrole-patcher and clusterrolebinding-patcher with this helm if statement as well. (similar to how `{{- if not .Values.nodeComponentOnly -}}` wraps the yaml there. 

---

We would then need to document that this helm value must also be true to use the feature.

AndrewSirenko: The documentation about what this new metadata source does, and what is required to enable the feature should be included in the associated install.md section IMO.

AndrewSirenko:
Ran pre_allocated and karpenter scale tests with 2500 PVCs (therefore 25-40 nodes) and setting metadataSources=[ec2labelskubernetes]. Everything worked and proper allocatable counts were there!

Before moving to GA we should try suddenly installing driver on a XXL node cluster to see if anything chokes.

cmd/main.go
```diff
@@ -82,28 +180,17 @@ func main() {
 		}
 		klog.FlushAndExit(klog.ExitFlushTimeout, 0)
 	case string(driver.ControllerMode), string(driver.NodeMode), string(driver.AllMode):
-		options.Mode = driver.Mode(cmd)
```
ConnorJC3, line 83: Accidentally removed line? (How did our tests not catch this???)

hack/e2e/run.sh
```diff
@@ -154,19 +152,36 @@ else
...
+        --skip="${GINKGO_SKIP}" \
+        --junit-report="${REPORT_DIR}/junit.xml" \
+        "${TEST_PATH}" \
+        -- \
+        -kubeconfig="${KUBECONFIG}" \
+        -gce-zone="${FIRST_ZONE}"
+      TEST_PASSED=$?
+      set -e
+      set +x
+    fi
```
ConnorJC3: This code is very duplicated, we could probably combine it?

tests/sanity/fake_sanity_cloud.go
```diff
@@ -138,6 +138,14 @@ func (d *fakeCloud) GetSnapshotByName(ctx context.Context, name string) (*cloud.
 	return nil, cloud.ErrNotFound
 }
 
+func (d *fakeCloud) GetInstance(ctx context.Context, nodeID string) (*types.Instance, error) {
+	return &types.Instance{}, nil
+}
```
ConnorJC3: This function can be dropped I think

pkg/cloud/cloud.go
```diff
@@ -935,6 +942,7 @@ func (c *cloud) AttachDisk(ctx context.Context, volumeID, nodeID string) (string
 	}
 
 	instance, err := c.getInstance(ctx, nodeID)
+
```
ConnorJC3: np: remove random newline

pkg/cloud/cloud.go
```diff
@@ -134,6 +134,13 @@ const (
 	fastVolumeStatusBatchMaxDelay = 500 * time.Millisecond
 )
 
+const (
+	// maxInstancesDescribed is the maximum number of instances described in each EC2 Describe Instances call
+	// TODO Question: what should a good number for this constant be? https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-pagination.html just says
```
ConnorJC3: 1000 is fine in my opinion, no reason to use less, and we'll get paginated with more

pkg/cloud/cloud.go
```diff
@@ -1835,6 +1843,62 @@ func (c *cloud) getInstance(ctx context.Context, nodeID string) (*types.Instance
 	}
 }
 
+// GetInstancesPatching returns the instance info associated with each node ID in `nodeIDs` and uses pagination
+// to get instances for large clusters. The instances are also described in batches of size up to `maxInstancesDescribed`.
+func (c *cloud) GetInstancesPatching(ctx context.Context, nodeIDs []string) ([]*types.Instance, error) {
```
ConnorJC3, line 2070: Is there ever a "one bad instance spoils the bunch" type scenario here? what if an instance is one of those weird ones like EKS's self-managed feature?

pkg/cloud/metadata/k8s.go
```diff
@@ -114,17 +116,70 @@ func KubernetesAPIInstanceInfo(clientset kubernetes.Interface) (*Metadata, error
 		return nil, fmt.Errorf("error getting Node %v: %w", nodeName, err)
 	}
 
-	providerID := node.Spec.ProviderID
-	if providerID == "" {
-		return nil, errors.New("node providerID empty, cannot parse")
+	numAttachedENIs := 1        // Default: All nodes have at least 1 attached ENI
+	numBlockDeviceMappings := 0 // Default: 0
+	sageMakerLabels := true
```
ConnorJC3: I think the code would be easier to reason about and more correct if this defaulted to `false`

cmd/main.go
```diff
@@ -68,6 +83,89 @@ func main() {
...
+		cloud, err = cloudPkg.NewCloud(region, accountID, options.AwsSdkDebugLog, options.UserAgentExtra, options.Batching, options.DeprecatedMetrics)
+		if err != nil {
+			klog.ErrorS(err, "failed to create cloud service")
+			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+		}
+
+		k8sClient, err = cfg.K8sAPIClient()
+		if err != nil {
+			klog.V(2).InfoS("Failed to setup k8s client", "err", err)
+		}
```
ConnorJC3: Is it safe to have these empty in non-metadata cases? I don't think we did that before

docs/install.md
```diff
@@ -22,6 +22,10 @@ If the driver is able to access IMDS, it will utilize that as a preferred source
 
 In order for the driver to access IMDS, it either must be run in host networking mode, or with a [hop limit of at least 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-existing-instances.html#modify-PUT-response-hop-limit).
 
+#### EC2 Labels Kubernetes Metadata
+
+For this metadata source to work, the driver pods must have access to the Kubernetes API server. The driver controller pods must also have Kubernetes RBAC permission to patch nodes. This metadata service gets the instance ID, type, region, and AZ in the same way as Kubernetes Metadata. The volume and ENI count are additionally obtained from the EC2 API.
```
ConnorJC3: todo: add info here about enabling the feature once we add helm flag

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,300 @@
...
+
+// Uses leader election so that only one controller pod calls containuousUpdateLabels().
+func ContinuousUpdateLabelsLeaderElection(clientset kubernetes.Interface, cloud cloud.Cloud, updateTime time.Duration) error {
+	var (
+		lockName = "node-patcher-ebs.csi.aws.com"
+	)
+	le := leaderelection.NewLeaderElection(clientset, lockName, func(ctx context.Context) {
+		err := continuousUpdateLabels(ctx, clientset, cloud, updateTime)
+		if err != nil {
+			klog.ErrorS(err, "failed to patch node labels with volume/ENI count")
```
ConnorJC3: np: Capitalize log messages

pkg/cloud/metadata/labels.go
```diff
@@ -0,0 +1,300 @@
...
+
+type enisVolumes struct {
+	ENIs    int
+	Volumes int
+}
+
+// Uses leader election so that only one controller pod calls containuousUpdateLabels().
+func ContinuousUpdateLabelsLeaderElection(clientset kubernetes.Interface, cloud cloud.Cloud, updateTime time.Duration) error {
+	var (
+		lockName = "node-patcher-ebs.csi.aws.com"
```
ConnorJC3: Can we base this, as well as `VolumesLabel` and `ENIsLabel` off the driver name constant (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/driver.go#L49)

I know the existing code already makes this mistake in a few places, but it'd be great if we don't add even more broken hardcoded driver name spots

cmd/main.go
sylviah23: I had to move this line up because it's needed for parsing the extra command line arguments, which is needed to get the metadata

pkg/cloud/metadata/labels.go
sylviah23: I thought we wanted error messages to be lowercase

AndrewSirenko:
> I still think we need to have a naming discussion, especially re the metadata source name and the patcher sidecar name

Final names: `metadata-labeler` for both metadata source and new container.

sylviah23:
@AndrewSirenko LabelRefreshTime in cmd/main.go should be changed to 60 * time.Minute

pkg/cloud/metadata/labels.go
AndrewSirenko:
```
❯ make update
go fix ./...
package github.com/kubernetes-sigs/aws-ebs-csi-driver/cmd
	imports github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/metadata from main.go
	imports github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver from labels.go
	imports github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/metadata from driver.go: import cycle not allowed
make: *** [update/gofix] Error 1
```

@ConnorJC3 looks like we get an import cycle if we do this :/ . Should we make a separate `constants` package?

pkg/cloud/metadata/labels.go
AndrewSirenko: @sylviah23  error message != log message

klog.ErrorS emits a log (that mentions the err)

pkg/cloud/metadata/labels.go
AndrewSirenko: Done, will rebase once #2628 merged

pkg/cloud/cloud.go
AndrewSirenko: Real concern. Decided in Wednesday standup that this will be required for graduating feature to GA. 

For now we will put disclaimer that alpha feature might not work on heterogenous clusters. 

Putting on backlog ability to get out of bad instance spoils bunch feature for this and for batcher. (can also reuse mechanism for "one fake volume spoils the bunch" too)

Because EC2 reports Which instance-id/volume-id is invalid, we can have a cache for known bad ids. While not ideal because not persisted accross restarts, eventually driver will get out of bad loop.

Makefile
AndrewSirenko, line 1: Top level: Hmm seeing many  `reflector.go:450] "Forcing resync" reflector="/gomodcache/k8s.io/client-go@v0.33.3/tools/cache/reflector.go:285"` in ebs-csi-node logs if MutableCSINode turned on. Will investigate  if something is up with informers before unholding.

docs/install.md
```diff
@@ -22,6 +22,12 @@ If the driver is able to access IMDS, it will utilize that as a preferred source
 
 In order for the driver to access IMDS, it either must be run in host networking mode, or with a [hop limit of at least 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-existing-instances.html#modify-PUT-response-hop-limit).
 
+#### Metadata Labeler
```
AndrewSirenko: Need to mention what 'alpha' means. Expand on what "volume and eni count are obtained from EC2 Api" means. Have fuller explanation on how to enable that includes `metadataSources`

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -251,6 +259,7 @@ controller:
   logLevel: 2
   userAgentExtra: "helm"
   nodeSelector: {}
+  deployMetadataLabelerContainer: false
```
ConnorJC3: Can we put this parameter as something like `enabled` in the `sidecars.metadataLabeler` section? (Similar to what we do for snapshotter but don't call it `forceEnable`)

docs/install.md
```diff
@@ -22,6 +22,12 @@ If the driver is able to access IMDS, it will utilize that as a preferred source
...
+
+For this metadata source to work, the driver pods must have access to the Kubernetes API server. The driver controller pods must also have Kubernetes RBAC permission to patch nodes. This metadata service gets the instance ID, type, region, and AZ in the same way as Kubernetes Metadata. The volume and ENI count are additionally obtained from the EC2 API. 
+
+This is an alpha feature and is disabled by default. To enable this feature set `controller.deployMetadataContainer` to `true`.
```
ConnorJC3: Suggest also adding a thing about including it in `node.metadataSources`

cmd/main.go
```diff
@@ -40,6 +41,11 @@ var (
 	featureGate = featuregate.NewFeatureGate()
 )
 
+const (
+	// LabelRefreshTime is the time in minutes that it takes for node labels to update volume and ENI count.
+	LabelRefreshTime = 60 * time.Minute
```
AndrewSirenko: I think we might have to make this a driver parameter.

Otherwise we can't test this in e2e tests without restarting ebs-csi-controller pod (which is fine workaround)

Downside is that customers can override it and spam ec2 backend. Though we can avoid making it an explicit helm parameter and rely on controller extra args.

AndrewSirenko: Team Decided to just restart pod.

AndrewSirenko: Done in latest rev

AndrewSirenko:
/unhold

I will make any other last small changes from personal computer and personal AWS account

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -90,6 +90,15 @@ sidecars:
         type: RuntimeDefault
       readOnlyRootFilesystem: true
       allowPrivilegeEscalation: false
+  metadataLabeler:
+    enabled: true
+    logLevel: 2
+    # Additional parameters provided by metadataLabeler.
+    additionalArgs: [ ]
```
ConnorJC3: TODO: Remove space here

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -417,7 +426,8 @@ node:
   # Enable the linux daemonset creation
   enableLinux: true
   enableWindows: true
-  metadataSources:
+  # Comma separated list of metadata sources that override the default used by the EBS CSI Driver. Valid sources include 'imds', 'kubernetes', and 'metadatalabeler'
+  metadataSources: "metadatalabeler"
```
ConnorJC3: TODO: Remove this default value

cmd/main.go
```diff
@@ -54,6 +55,15 @@ func main() {
 		options  = driver.Options{}
 	)
 
+	var (
+		metadataRequiredModes = map[string]struct{}{
+			string(driver.ControllerMode): {},
+			string(driver.NodeMode):       {},
+			string(driver.AllMode):        {},
+			"metadataLabeler":             {},
```
ConnorJC3: TODO: Make this a constant

AndrewSirenko: FYI Had to just make it a string not a typed Mode because of some go case switch linter. Didn't want to refactor other code because a separate PR should probably do the same thing for pre-stop-hook mode.

ConnorJC3:
Test run from latest revision:
```
$ make e2e/disruptive
TEST_PATH=./tests/e2e/... \
GINKGO_FOCUS="\[ebs-csi-e2e\] \[Disruptive\]" \
GINKGO_SKIP="\[Flaky\]" \
GINKGO_PARALLEL=1 \
EBS_INSTALL_SNAPSHOT=false \
HELM_EXTRA_FLAGS="--set=sidecars.metadataLabeler.enabled=true,node.metadataSources='metadata-labeler'" \
./hack/e2e/run.sh
...
Will run 1 of 78 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 1 of 78 Specs in 109.456 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 77 Skipped
```

ElijahQuinones:
Needs make update

cmd/main.go
```diff
@@ -66,6 +76,65 @@ func main() {
...
+	}
+
+	if _, ok := metadataRequiredModes[cmd]; ok {
+		region := os.Getenv("AWS_REGION")
+		var metadataErr error
+
+		if region != "" {
+			klog.InfoS("Region provided via AWS_REGION environment variable", "region", region)
+			if options.Mode != driver.ControllerMode {
+				klog.InfoS("Node service requires metadata even if AWS_REGION provided, initializing metadata")
```
mdzraf, line 112: np: 
```suggestion
				klog.InfoS("Node service requires metadata even if AWS_REGION is provided, initializing metadata")
```

cmd/main.go
```diff
@@ -66,6 +76,65 @@ func main() {
...
+			}
+		} else {
+			klog.InfoS("Initializing metadata")
+			md, metadataErr = metadata.NewMetadataService(cfg, region)
+		}
+
+		if metadataErr != nil {
+			klog.ErrorS(metadataErr, "Failed to initialize metadata when it is required")
+			if options.Mode == driver.ControllerMode {
+				klog.InfoS("The region can be manually supplied via the AWS_REGION environment variable")
```
mdzraf, line 123: Is there a reason why we print this, is the region missing the only way we would get to this point?

ConnorJC3: This is how the existing code is, this line was just moved, can I suggest we don't debate it here

ConnorJC3:
/retest
/unhold

Disruptive tests: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/2591/pull-aws-ebs-csi-driver-e2e-disruptive/1981810623859658752

PR #2613, opened 2025-08-11
Add ability to use az-id for volume provisioning

ElijahQuinones:
#### What type of PR is this?

/kind feature

This PR adds the ability to create a volume using AZ-ID as the volumes topology
Manually tested by deploying modified driver to cluster and provisioning a volume using az-id topology key successfully 

```release-note
Add ability to provision volumes using AZ-ID 
```


tests/e2e/go.mod
```diff
@@ -160,6 +160,10 @@ require (
 replace k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20250628140032-d90c4fd18f59
 
 replace (
+	github.com/aws/aws-sdk-go-v2 v1.36.3 => /home/elijahlq/AZ-ID_SDK/awsSdkGoV2
+	github.com/aws/aws-sdk-go-v2/config v1.29.14 => /home/elijahlq/AZ-ID_SDK/awsSdkGoV2/config
+	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.16.30 => /home/elijahlq/AZ-ID_SDK/awsSdkGoV2/feature/ec2/imds
+	github.com/aws/aws-sdk-go-v2/service/ec2 v1.218.0 => /home/elijahlq/AZ-ID_SDK/awsSdkGoV2/service/ec2
```
torredil: The replace directive specifies `/home/elijahlq/` for these dependencies, that needs to be fixed.

ElijahQuinones: Ahhh thanks Eddie will remove on now

torredil:
/hold 

for 2PR, feel free to unhold.

ElijahQuinones:
@AndrewSirenko 
>Do we need a simple e2e test for this? Or are we sufficiently positive that protecting regressions via unit tests is enough.

Do you believe there is anything that the unit tests would not catch that an e2e test would for this change ? The current set of unit tests check to make sure that we are passing the correct information to AWS and returning the correct errors in cases where we don't (ie AZ and AZ-ID). To me at least the only thing a e2e test adds is making sure the API behavior is as documented and as we expect it which is not something we typically test for as the API is already throughly tested and has documentation with it that contracts how it should behave. 

That being said is there any reason you think a e2e test would be beneficial for us to add ? If so I would have no problem adding one but I currently don't see the value in it.

ConnorJC3:
/lgtm
/approve
/remove-hold

(Applying both because @torredil already lgtm'd and the only diff is rebase)

ConnorJC3:
/retest
flake

PR #2635, opened 2025-08-20
Adding IOPS Increase On Resize If iopspergb & allowautoiopsincreaseonresize are set

mdzraf:
#### What type of PR is this?

/kind feature

This PR adds a way for allowing IOPS to be increased if a volume is resized and `IopsPerGB` is set. This way the ratio will remain as expected after a resize, in order to use this feature `allowautoiopsincreaseonresize` must be set to true as well. 

This PR also adds `iopspergb` as a VAC parameter so that it can be changed. 

## **As of this first revision, no unit tests have been added or existing ones have been modified, manual testing listed below. Unit tests will be a part of revision 2 after initial rev is reviewed**

Manual Tests Performed:
- Resizing and ensuring IOPS are adjusted to match set ratio ✅ 
- Resizing + increasing `iopspergb` ratio and ensuring IOPS are adjusted correctly to new size and ratio ✅ 
- Resizing with `iopspergb` set but `allowautoiopsincreaseonresize` not set/set to false, ensure that resize happens but no IOPS are adjusted ✅ 
- `iopspergb` is increased and IOPS are adjusted accordingly depending on volume size ✅ 
- `allowautoiopsincreaseonresize` is set to true after initial volume creation, and then resizes. IOPS are adjusted accordingly ✅ 

```release-note
Added AllowAutoIopsIncreaseOnResize SC and VAC parameter To Allow For Maintaining desired IOPS:GB Ratio When A Volume is Resized and Has IopsPerGB Set. Added ability to modify IopsPerGB parameter through VAC. 
```


mdzraf:
/hold 

Unit tests need to be added after initial rev is reviewed.

docs/parameters.md
```diff
@@ -11,6 +11,7 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
...
 | "allowAutoIOPSPerGBIncrease" | true, false                                     | false   | When `"true"`, the CSI driver increases IOPS for a volume when `iopsPerGB * <volume size>` is too low to fit into IOPS range supported by AWS. This allows dynamic provisioning to always succeed, even when user specifies too small PVC capacity or `iopsPerGB` value. On the other hand, it may introduce additional costs, as such volumes have higher IOPS than requested in `iopsPerGB`. |
+| "allowAutoIOPSIncreaseOnResize" | true, false                                  | false   | When `"true"`, the CSI Driver adjusts IOPS for a volume during resizing if `iopsPerGB` is set. This ensures that the volume maintains the desired IOPS/GiB ratio. If `iopsPerGB * <volume size>` exceed volume limits, the IOPS will be capped at the maximum allowed for that volume type and the call will succeed.
```
AndrewSirenko: Can we also make this a VAC parameter so that customers can enable this for volumes that have been around forever?

mdzraf: It is set as a VAC parameter. Did I miss something?

AndrewSirenko: Forgot we don't have separate docs

pkg/driver/validation.go
```diff
@@ -52,6 +52,12 @@ func validateExtraTags(tags map[string]string, warnOnly bool) error {
 		if k == cloud.SnapshotNameTagKey {
 			return fmt.Errorf("tag key '%s' is reserved", cloud.SnapshotNameTagKey)
 		}
+		if k == IopsPerGBKey {
+			return fmt.Errorf("tag key '%s' is reserved", IopsPerGBKey)
+		}
+		if k == AllowAutoIopsIncreaseOnResizeKey {
+			return fmt.Errorf("tag key '%s' is reserved", AllowAutoIopsIncreaseOnResizeKey)
+		}
```
ConnorJC3, line 58: This function is getting full of `if` statements, should we use a map instead?

pkg/cloud/cloud.go
```diff
@@ -821,27 +811,58 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
 	if newSizeBytes != 0 {
 		req.Size = aws.Int32(newSizeGiB)
 	}
-	if options.IOPS != 0 {
-		req.Iops = aws.Int32(options.IOPS)
-	}
 	if options.VolumeType != "" {
 		req.VolumeType = types.VolumeType(options.VolumeType)
+	} else {
+		req.VolumeType = volume.VolumeType
```
ConnorJC3: We shouldn't be setting things on the request object based on the existing object, this introduces a race condition, use a separate variable

pkg/cloud/cloud.go
```diff
@@ -821,27 +811,58 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
 	}
-	needsModification, volumeSize, err := c.validateModifyVolume(ctx, volumeID, newSizeGiB, options)
+	request := &ec2.DescribeVolumesInput{
+		VolumeIds: []string{volumeID},
+	}
+	volume, err := c.getVolume(ctx, request)
+	if err != nil {
+		return 0, err
+	}
+	needsModification, volumeSize, err := c.validateModifyVolume(ctx, volumeID, newSizeGiB, options, *volume)
```
ConnorJC3: Don't we need to do the math before `validateModifyVolume` for the duplicate requests detection to work?

pkg/cloud/cloud.go
```diff
@@ -821,27 +811,58 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+		return 0, err
+	}
+
+	if iopsPerGBIsSet && allowautoincreaseIsSet && options.IOPS == 0 {
+		sizeToUse := newSizeGiB
+		if req.Size != nil {
+			sizeToUse = *req.Size
+		}
+		req.Iops = aws.Int32(capIOPS(string(req.VolumeType), sizeToUse, (sizeToUse * iopsPerGbVal), minIops, maxIops, maxIopsPerGb, false))
+	}
```
ConnorJC3, line 995: This code is not DRY - we shouldn't have the same if statement and `capIOPS` call twice

pkg/cloud/cloud.go
```diff
@@ -821,27 +811,58 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+	if err != nil {
+		return 0, err
+	}
+
+	if options.IOPS != 0 {
+		sizeToUse := newSizeGiB
+		if req.Size != nil {
+			sizeToUse = *req.Size
+		}
+		req.Iops = aws.Int32(capIOPS(string(req.VolumeType), sizeToUse, options.IOPS, minIops, maxIops, maxIopsPerGb, false)) //TODO: See if we want to set allowIncrease to always true instead of false on modify?
```
ConnorJC3: Do we need a tag for allowIncrease? I think we do to make the behavior consistent, or we could treat it as always true

pkg/cloud/cloud.go
```diff
@@ -2421,3 +2433,74 @@ func capIOPS(volumeType string, requestedCapacityGiB int32, requestedIops int32,
...
+		throughputOut = throughput
+	default:
+		return 0, 0, 0, 0, fmt.Errorf("invalid AWS VolumeType %q", volumeType)
+	}
+	return maxIops, minIops, maxIopsPerGb, throughputOut, nil
+}
+
+// Checks if a volume's IOPS can be increased on expansion to adhere to IopsPerGB ratio.
+func (c *cloud) CheckIfIopsIncreaseOnExpansion(ctx context.Context, volumeID string) (iopsPerGBIsSet bool, allowautoincreaseIsSet bool, iopsPerGbVal int32, err error) {
+	req := &ec2.DescribeTagsInput{
```
ConnorJC3: Why are we calling describe tags? Do we not already call DV?

pkg/driver/controller_modify_volume.go
```diff
@@ -224,6 +236,33 @@ func parseModifyVolumeParameters(params map[string]string) (*modifyVolumeRequest
...
+		}
+		if req.newSize != 0 {
+			req.modifyDiskOptions.IOPS = int32(iopsPerGb) * util.BytesToGiB(req.newSize)
+		} else {
+			req.modifyDiskOptions.IOPS = int32(iopsPerGb) * disk.CapacityGiB
+			req.newSize = util.GiBToBytes(disk.CapacityGiB) // Make sure to always pass in a size to capIOPS
+		}
+	}
+	return req.modifyDiskOptions.IOPS, nil
+}
```
ConnorJC3: Why are we not just passing iopsPerGb to cloud? This causes us to double describe the volume

pkg/cloud/cloud.go
```diff
@@ -2421,3 +2433,74 @@ func capIOPS(volumeType string, requestedCapacityGiB int32, requestedIops int32,
...
+
+	existingTags, err := c.ec2.DescribeTags(ctx, req)
+	if err != nil {
+		klog.ErrorS(err, "failed to describe tags", "volumeID", volumeID)
+		return false, false, 0, err
+	}
+
+	if existingTags.Tags != nil {
+		for _, tag := range existingTags.Tags {
+			klog.InfoS(*tag.Key, ":", tag.Value)
```
ConnorJC3: Debug print left in?

pkg/cloud/cloud.go
```diff
@@ -116,6 +116,10 @@ const (
 	KubernetesTagKeyPrefix = "kubernetes.io"
 	// AwsEbsDriverTagKey is the tag to identify if a volume/snapshot is managed by ebs csi driver.
 	AwsEbsDriverTagKey = util.DriverName + "/cluster"
+	// AllowIopsIncreaseOnResize is the tag key for allowing IOPS increase on resizing if IopsPerGB is set to ensure desired ratio is maintained.
+	AllowAutoIopsIncreaseOnResizeKey = "allowautoiopsincreaseonresize"
```
ConnorJC3: This would be more readable camel cased, ditto the below key

ConnorJC3: Also should be prefixed with `util.DriverName`

pkg/cloud/cloud.go
```diff
@@ -2421,3 +2433,74 @@ func capIOPS(volumeType string, requestedCapacityGiB int32, requestedIops int32,
 	}
 	return iops
 }
+
+// Setting volume IOPS Supported AWS limits depending on volume type.
+func getVolumeLimits(volumeType string, throughput int32) (maxIops int32, minIops int32, maxIopsPerGb int32, throughputOut int32, err error) {
```
ConnorJC3: Why is throughput an input to this function? The limits don't depend on the throughput do they?

ConnorJC3: Also we should really consider returning a struct here, maybe taking that struct as an input to `capIOPS` as well

ConnorJC3 (changes requested):

pkg/driver/validation.go
mdzraf: When you say use a map what do you mean? The tags are already in a map, do you mean like use a switch statement instead or?

ConnorJC3: A map of the disallowed names, something along the lines of `ValidFSTypes` in `node.go`

pkg/cloud/cloud.go
mdzraf: Also considered that but wanted to get teams input on it since I am not sure if there is an obvious option here.

pkg/driver/validation.go
mdzraf: Ah I see, that makes sense, will have for next rev. Though we will still have to have a separate check for the prefix case.

pkg/cloud/cloud.go
mdzraf: Good catch, fixed next rev ✅

pkg/cloud/cloud.go
mdzraf: Are you suggesting to instead directly pass volume.VolumeType to the getVolumeLimits call in the else instead of assigning the value to req.Volumetype and then passing in req.Volumetype to the getVolumeLimits call?

pkg/cloud/cloud.go
mdzraf: Good catch, extracted common logic and fixed in next rev ✅

pkg/cloud/cloud.go
mdzraf: > Why is throughput an input to this function? The limits don't depend on the throughput do they?

It is only needed because prior to putting this logic into a function this switch statement in CreateDisk was where throughput is assigned if it is a GP3 volume. We could remove that from the function and just pass diskOptions.Throughput in CreateDisk directly, for ResizeOrModifyDisk we don't need it all as it is already passed in through options.Throughput, so I think it would actually be best to remove it as well. Will do next rev ✅

mdzraf: > Also we should really consider returning a struct here, maybe taking that struct as an input to capIOPS as well

Done ✅

pkg/cloud/cloud.go

pkg/cloud/cloud.go
mdzraf: Fixed Next Rev ✅

pkg/driver/controller_modify_volume.go
mdzraf: Solved in next rev ✅

pkg/cloud/cloud.go
```diff
@@ -571,36 +581,22 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
-		maxIops = io2MaxTotalIOPS
-		minIops = io2MinTotalIOPS
-		maxIopsPerGb = io2MaxIOPSPerGB
-	case VolumeTypeGP3:
-		maxIops = gp3MaxTotalIOPS
-		minIops = gp3MinTotalIOPS
-		maxIopsPerGb = gp3MaxIOPSPerGB
-		throughput = diskOptions.Throughput
-	default:
+	IopLimits, err := getVolumeLimits(createType)
```
ConnorJC3: Why capitalized? Also, weird name, suggest `iopsLimit`

pkg/cloud/cloud.go
```diff
@@ -2396,28 +2408,102 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	case VolumeTypeIO2:
+		iopLimits.maxIops = io2MaxTotalIOPS
+		iopLimits.minIops = io2MinTotalIOPS
+		iopLimits.maxIopsPerGb = io2MaxIOPSPerGB
+	case VolumeTypeGP3:
+		iopLimits.maxIops = gp3MaxTotalIOPS
+		iopLimits.minIops = gp3MinTotalIOPS
+		iopLimits.maxIopsPerGb = gp3MaxIOPSPerGB
+	default:
+		return iopLimits, fmt.Errorf("invalid AWS VolumeType %q", volumeType)
```
ConnorJC3: This effectively validates the volume type locally, so I think instead of returning an error we should just do nothing here and forward directly to the EC2 API.

pkg/cloud/cloud.go
```diff
@@ -2396,28 +2408,102 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
-		klog.V(5).InfoS("[Debug] Capped IOPS for volume", "volumeType", volumeType, "requestedCapacityGiB", requestedCapacityGiB, "maxIOPSPerGB", maxIOPSPerGB, "limit", iops)
+		klog.V(5).InfoS("[Debug] Capped IOPS for volume", "volumeType", volumeType, "requestedCapacityGiB", requestedCapacityGiB, "maxIOPSPerGB", iopLimits.maxIopsPerGb, "limit", iops)
 	}
 	return iops
 }
+
+// Setting volume IOPS Supported AWS limits depending on volume type.
+func getVolumeLimits(volumeType string) (iopLimits IopLimits, err error) {
+	switch strings.ToLower(volumeType) {
+	case VolumeTypeGP2, VolumeTypeSC1, VolumeTypeST1, VolumeTypeStandard:
```
ConnorJC3, line 2688: This will cause the cap to go to `0` I think? We should probably set `maxIops` and `maxIopsPerGb` to max int or something like that instead. Either that or continue to return a `0` here but check for a zero in `capIOPS`

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
 	}
+
+	var sizeToUse int32
+	if req.Size != nil {
+		sizeToUse = *req.Size
+	} else {
+		sizeToUse = *volume.Size
+	}
+
+	IopLimits, err := getVolumeLimits(string(volTypeToUse))
```
ConnorJC3: Same as previous comment - weird and capitalized variable name

pkg/cloud/cloud.go
```diff
@@ -2396,28 +2408,102 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+		iopLimits.minIops = gp3MinTotalIOPS
+		iopLimits.maxIopsPerGb = gp3MaxIOPSPerGB
+	default:
+		return iopLimits, fmt.Errorf("invalid AWS VolumeType %q", volumeType)
+	}
+	return iopLimits, nil
+}
+
+// Checks if a volume's IOPS can be increased on expansion to adhere to IopsPerGB ratio.
+func (c *cloud) CheckIfIopsIncreaseOnExpansion(ctx context.Context, volumeID string, existingTags []types.Tag) (iopsPerGBIsSet bool, allowautoincreaseIsSet bool, iopsPerGbVal int32, err error) {
```
ConnorJC3: We don't need to call this outside of `cloud` right? So name should start lowercase to not export

pkg/cloud/cloud.go
```diff
@@ -2396,28 +2408,102 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+		case AllowAutoIopsIncreaseOnResizeKey:
+			isSet, err := strconv.ParseBool(*tag.Value)
+			if err != nil {
+				return false, false, 0, fmt.Errorf("%w: %w", ErrInvalidArgument, err)
+			}
+			if isSet {
+				allowautoincreaseIsSet = true
+			} else {
+				allowautoincreaseIsSet = false
+			}
```
ConnorJC3:
```suggestion
			allowautoincreaseIsSet, err := strconv.ParseBool(*tag.Value)
			if err != nil {
				return false, false, 0, fmt.Errorf("%w: %w", ErrInvalidArgument, err)
			}
```

pkg/cloud/cloud.go
```diff
@@ -2396,28 +2408,102 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	}
+	return iopLimits, nil
+}
+
+// Checks if a volume's IOPS can be increased on expansion to adhere to IopsPerGB ratio.
+func (c *cloud) CheckIfIopsIncreaseOnExpansion(ctx context.Context, volumeID string, existingTags []types.Tag) (iopsPerGBIsSet bool, allowautoincreaseIsSet bool, iopsPerGbVal int32, err error) {
+	for _, tag := range existingTags {
+		switch *tag.Key {
+		case IopsPerGBKey:
+			iopsPerGBIsSet = true
```
ConnorJC3: Rather than having a separate "is set" variable, it probably makes more sense to either use `0` as unset or use a pointer

pkg/driver/controller.go
```diff
@@ -142,6 +142,9 @@ func (d *ControllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 				return nil, status.Errorf(codes.InvalidArgument, "Could not parse invalid iopsPerGB: %v", parseIopsPerGBKeyErr)
 			}
 			iopsPerGB = int32(parseIopsPerGBKey)
+			volumeTags[cloud.IopsPerGBKey] = strconv.Itoa(int(iopsPerGB))
+		case AllowAutoIopsIncreaseOnResizeKey:
+			volumeTags[cloud.AllowAutoIopsIncreaseOnResizeKey] = value
```
ConnorJC3: This should probably be wrapped with an `isTrue` if statement, otherwise a user could set the parameter to `foobar`

pkg/driver/constants.go
```diff
@@ -47,6 +47,9 @@ const (
 	// AllowAutoIOPSPerGBIncreaseKey represents key for allowing automatic increase of IOPS.
 	AllowAutoIOPSPerGBIncreaseKey = "allowautoiopspergbincrease"
 
+	// AllowIopsIncreaseOnResize represents key for allowing IOPS increase on resizing if IopsPerGB is set to ensure desired ratio is maintained.
+	AllowAutoIopsIncreaseOnResizeKey = "allowautoiopsincreaseonresize"
```
ConnorJC3: I think we should consider calling this "on modify" instead of "on resize", as it can't actually happen on resize alone from my understanding (as you can't shrink a volume)

docs/parameters.md
ConnorJC3: > If `iopsPerGB * <volume size>` exceed volume limits, the IOPS will be capped at the maximum allowed for that volume type and the call will succeed.

This should be put somewhere else because it doesn't have anything to do with this parameter, and putting it here implies the parameter enables this functionality (when in reality it is always enabled).

pkg/cloud/cloud.go
```diff
@@ -116,6 +116,10 @@ const (
 	KubernetesTagKeyPrefix = "kubernetes.io"
 	// AwsEbsDriverTagKey is the tag to identify if a volume/snapshot is managed by ebs csi driver.
 	AwsEbsDriverTagKey = util.DriverName + "/cluster"
+	// AllowIopsIncreaseOnResize is the tag key for allowing IOPS increase on resizing if IopsPerGB is set to ensure desired ratio is maintained.
+	AllowAutoIopsIncreaseOnResizeKey = util.DriverName + "/AllowAutoIopsIncreaseOnResizeKey"
```
ConnorJC3:
```suggestion
	AllowAutoIopsIncreaseOnResizeKey = util.DriverName + "/AllowAutoIopsIncreaseOnResize"
```

pkg/cloud/cloud.go
```diff
@@ -116,6 +116,10 @@ const (
...
+	// IopsPerGBKey represents the tag key for IOPS per GB.
+	IopsPerGBKey = util.DriverName + "/IopsPerGb"
```
ConnorJC3: Do we want to make this `IOPS` instead of `Iops`? I think we usually prefer the former for user/customer facing stuff in the driver. Ditto for the above variable.

pkg/cloud/cloud.go
```diff
@@ -227,9 +231,18 @@ type DiskOptions struct {
...
-	Throughput int32
+	VolumeType                string
+	IOPS                      int32
+	Throughput                int32
+	IOPSPerGB                 int32
+	AllowIopsIncreaseOnResize bool
+}
+
+// IopLimits represents the IOPS limits set by EBS of a volume dependent on the volume type.
+type IopLimits struct {
```
ConnorJC3: Should be `iopsLimits` (non-exported, using correct term)

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+	request := &ec2.DescribeVolumesInput{
+		VolumeIds: []string{volumeID},
+	}
+	volume, err := c.getVolume(ctx, request)
+	if err != nil {
+		return 0, err
+	}
 
-	if err != nil || !needsModification {
+	startModification, volumeSize, err := c.validateVolumeState(ctx, volumeID, newSizeGiB, *volume.Size, options)
```
ConnorJC3: np: Suggest retaining the previous `needsModification` naming, in my opinion it's a more clear variable name

pkg/cloud/cloud.go
```diff
@@ -2321,62 +2368,27 @@ func needsVolumeModification(volume types.Volume, newSizeGiB int32, options *Mod
...
 		if err != nil {
 			return true, oldSizeGiB, err
 		}
 		returnGiB, returnErr := c.checkDesiredState(ctx, volumeID, newSizeGiB, options)
 		return false, returnGiB, returnErr
 	}
 
-	if state == string(types.VolumeModificationStateOptimizing) {
-		return true, 0, fmt.Errorf("volume %q in OPTIMIZING state, cannot currently modify", volumeID)
-	}
```
ConnorJC3, line 2486: Having this check after the `needsVolumeModification` check is intentional - we can't move it earlier or it causes a bug.

See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1833#issuecomment-1802224607 for context

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+		iopsValue = sizeToUse * options.IOPSPerGB
+	} else if iopsPerGBIsSet && allowautoincreaseIsSet {
+		iopsValue = sizeToUse * iopsPerGbVal
+	}
+	if iopsValue != 0 {
+		// Even if volume type does not support IOPS, still pass value and let EC2 handle error.
+		if IopLimits.maxIops == 0 {
+			req.Iops = aws.Int32(iopsValue)
+		} else {
+			req.Iops = aws.Int32(capIOPS(string(volTypeToUse), sizeToUse, iopsValue, IopLimits, false))
```
ConnorJC3: Maybe best for a standup discussion but I think it probably makes more sense for the last parameter to be `true` rather than `false`?

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+		}
+		options.IOPS = *req.Iops
```
ConnorJC3, line 989: Would it make the code saner if we passed `req` into `validateModifyVolume` instead of `options`?

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+	IopLimits, err := getVolumeLimits(string(volTypeToUse))
+	if err != nil {
+		return 0, err
+	}
+	iopsPerGBIsSet, allowautoincreaseIsSet, iopsPerGbVal, err := c.CheckIfIopsIncreaseOnExpansion(ctx, volumeID, volume.Tags)
+	if err != nil {
+		return 0, err
+	}
+	var iopsValue int32
+	//nolint:gocritic // gocritic suggests rewriting all of the below conditionals a switch statement,
```
ConnorJC3: Suggest removing this and doing as the linter requests as it's more go-idiomatic

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+	// but there are different variables being checked and it is best readable as is.
+	if options.IOPS != 0 {
+		iopsValue = options.IOPS
+	} else if options.IOPSPerGB != 0 && ((options.AllowIopsIncreaseOnResize) || (allowautoincreaseIsSet)) {
```
ConnorJC3: Many extra parenthesis on this line

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+		iopsValue = sizeToUse * options.IOPSPerGB
+	} else if iopsPerGBIsSet && allowautoincreaseIsSet {
```
ConnorJC3: Missing check against `options.AllowIopsIncreaseOnResize` (should be like how it is in the previous `if`)

pkg/cloud/cloud.go
```diff
@@ -821,27 +817,78 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
+
+	IopLimits, err := getVolumeLimits(string(volTypeToUse))
+	if err != nil {
+		return 0, err
+	}
+	iopsPerGBIsSet, allowautoincreaseIsSet, iopsPerGbVal, err := c.CheckIfIopsIncreaseOnExpansion(ctx, volumeID, volume.Tags)
+	if err != nil {
+		return 0, err
+	}
+	var iopsValue int32
```
ConnorJC3: Suggest a name something like `newIops` or `iopsForModify` to be clearer

pkg/cloud/cloud.go
mdzraf: Yes I agree, will leave thread open to get some more opinions from reviewers
cc: @torredil @ElijahQuinones

docs/parameters.md
mdzraf: Good catch, moved to the `iopsPerGb` section.

pkg/cloud/cloud.go
mdzraf: Added check for `iopLimits.maxIops == 0` in capIOPS, I think that is the best approach instead of setting it to max int

pkg/cloud/cloud.go
mdzraf: That would also mean we would have to modify `validateVolumeState` which is called much earlier on before we construct the request input with the options, I would prefer to leave as is unless you feel there is a very strong reason to make the switch

ElijahQuinones:
/hold 

holding for release

mdzraf:
/unhold

Release is complete and rebase is done, ready for review.

pkg/cloud/cloud.go
ConnorJC3: Why would we have to modify `validateVolumeState`?

docs/parameters.md
```diff
@@ -9,8 +9,9 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
...
-| "iopsPerGB"                  |                                                 |         | I/O operations per second per GiB. Can be specified for IO1, IO2, and GP3 volumes.                                                                                                                                                                                                                                                                                                            |
+| "iopsPerGB"                  |                                                 |         | I/O operations per second per GiB. Can be specified for IO1, IO2, and GP3. If `iopsPerGB * <volume size>` exceed volume limits, the IOPS will be capped at the maximum allowed for that volume type and the call will succeed. volumes.                                                                                                                                                                                                                                                                                                            |
```
ConnorJC3: Several typos in this line

pkg/cloud/cloud.go
```diff
@@ -905,27 +911,89 @@ func (c *cloud) ResizeOrModifyDisk(ctx context.Context, volumeID string, newSize
...
 	}
+
+	var sizeToUse int32
+	if req.Size != nil {
+		sizeToUse = *req.Size
+	} else {
+		sizeToUse = *volume.Size
+	}
+
+	allowautoincreaseIsSet, iopsPerGbVal, err := c.checkIfIopsIncreaseOnExpansion(volume.Tags)
```
ConnorJC3: Camel case `allowautoincreaseIsSet`

pkg/cloud/cloud.go
mdzraf: Because it also calls `checkDesiredState` which would also have to be changed to take in `req` instead of `options` if we did this.

ConnorJC3: Ehhhhh

I don't like it, but it's not worth refactoring for this PR, resolving for now

ConnorJC3 (approved):
/lgtm

PR #2636, opened 2025-08-21
Not Returning Infeasible If Tags Were Modified

mdzraf:
#### What type of PR is this?

/kind bug

This PR makes it so that when there is a modifyVolume request and the request to modify tags which happens before the `ResizeOrModifyDisk`, succeeds, and then there is a failure when `ResizeOrModifyDisk` is called due to an InvalidParameter, the driver does not return codes.InvalidArgument which is considered an infeasible error by the resizer. Because there was _some_ modification done to the volume, an infeasible error is not to be returned as this would signal that no changes were done at all.

Running `make test` for unit tests.

```release-note
NONE
```

### DIFF
--- a/pkg/driver/controller_modify_volume.go
+++ b/pkg/driver/controller_modify_volume.go
@@ -151,6 +151,11 @@ func executeModifyVolumeRequest(c cloud.Cloud) func(string, modifyVolumeRequest)
 			if err != nil {
 				switch {
 				case errors.Is(err, cloud.ErrInvalidArgument):
+					// Returning Internal error instead of InvaliArgument because at this point any tag modifications have succeeded.
+					// It would not be correct to return an error that is considered infeasible by the resizer if the volume was already modified in any way.
+					if len(req.modifyTagsOptions.TagsToAdd) > 0 || len(req.modifyTagsOptions.TagsToDelete) > 0 {
+						return 0, status.Errorf(codes.Internal, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
+					}
 					return 0, status.Errorf(codes.InvalidArgument, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
 				case errors.Is(err, cloud.ErrNotFound):
 					return 0, status.Errorf(codes.NotFound, "Could not modify volume (not found) %q: %v", volumeID, err)


AndrewSirenko (commented):
I thought we wanted to undo the tag modification then return InvalidArgument in this case? 

By returning internal error here customer will not be able to rollback vacname AND we will spam EC2 because there is no delayed retry interval.

ConnorJC3:
Is undoing the tags reasonably possible? Don't we get completely wrecked by restarts/crashes? I thought we were going to add a MV dry run, and bail infreasibly on that.

mdzraf:
Offline Discussion: We will merge this PR and work on designing a long-term fix to ensure that the volume is not changed at all when we return an infeasible error.

pkg/driver/controller_modify_volume.go
```diff
@@ -151,7 +151,7 @@ func executeModifyVolumeRequest(c cloud.Cloud) func(string, modifyVolumeRequest)
 			if err != nil {
 				switch {
 				case errors.Is(err, cloud.ErrInvalidArgument):
-					return 0, status.Errorf(codes.InvalidArgument, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
+					return 0, status.Errorf(codes.Internal, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
```
torredil: Suggest adding a comment here in the source code so that a future maintainer understands why the driver isn't returning `codes.InvalidArgument` here when cloud throws `ErrInvalidArgument`.

mdzraf: Good idea, done ✅

torredil (approved):

AndrewSirenko:
Going to have one more standup discussion on next steps because:
- Changing invalidArg -> internal leads to spamming EC2 API Calls on wrong parameters
- ~[KEP-3751: add error handling](https://github.com/kubernetes/enhancements/pull/5482) may constrain whether customers can roll forward or roll back internal vs InvalidArg errors~ Nvm final wording is relaxed about this. 

Instead a dry-run modifyVolume before createTags would help customers catch invalid VACs before modifying volume at all.

AndrewSirenko:
/hold 

Until standup discussion

pkg/driver/controller_modify_volume.go
```diff
@@ -151,7 +151,9 @@ func executeModifyVolumeRequest(c cloud.Cloud) func(string, modifyVolumeRequest)
 			if err != nil {
 				switch {
 				case errors.Is(err, cloud.ErrInvalidArgument):
-					return 0, status.Errorf(codes.InvalidArgument, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
+					// Returning Internal error instead of InvaliArgument because at this point any tag modifications have succeeded.
+					// It would not be correct to return an error that is considered infeasible by the resizer if the volume was already modified in any way.
+					return 0, status.Errorf(codes.Internal, "Could not modify volume (invalid argument) %q: %v", volumeID, err)
```
ConnorJC3: Offline discussion notwithstanding, I think we should only return `Internal` here if a tag modification was actually made, so we should probably pass that into this function and branch the return code based on that.

AndrewSirenko (commented):
/unhold

pkg/driver/controller_modify_volume.go
mdzraf: Makes sense, done ✅

pkg/driver/controller_modify_volume.go
```diff
@@ -151,6 +151,11 @@ func executeModifyVolumeRequest(c cloud.Cloud) func(string, modifyVolumeRequest)
 			if err != nil {
 				switch {
 				case errors.Is(err, cloud.ErrInvalidArgument):
+					// Returning Internal error instead of InvaliArgument because at this point any tag modifications have succeeded.
+					// It would not be correct to return an error that is considered infeasible by the resizer if the volume was already modified in any way.
+					if len(req.modifyTagsOptions.TagsToAdd) > 0 && len(req.modifyTagsOptions.TagsToDelete) > 0 {
```
ConnorJC3: This should be `||` instead of `&&`

ConnorJC3:
/retest
flake

ConnorJC3 (approved):
/lgtm

PR #2641, opened 2025-08-27
Makefile: Add validation to prevent Windows clusters with kOps

torredil:
#### What type of PR is this?

/kind cleanup

Currently, running `make cluster/create` with `WINDOWS=true` and the default `CLUSTER_TYPE=kops` creates a cluster with no Windows nodes, which is confusing for developers expecting Windows support.

```
make cluster/create WINDOWS=true                                                                                                                                                                             
Error: Windows clusters are not supported with kops. Please set CLUSTER_TYPE=eksctl when WINDOWS=true
make: *** [cluster/create] Error 1
```

```release-note
NONE
```


ConnorJC3 (commented):
/lgtm

PR #2646, opened 2025-09-02
Add e2e/disruptive placeholder

AndrewSirenko:
#### What type of PR is this?

/kind cleanup

For #2591 

For now leaving out the additional helm parameters from PR because those aren't implemented yet.

Locally

```release-note
NONE
```


ElijahQuinones:
Re running unit test job seems to have gotten hung

PR #2558, opened 2025-07-08
Align node and controller metrics to consistent experience

ConnorJC3:
#### What type of PR is this?

/kind feature

#### What is this PR about? / Why do we need it?

The goal of this PR is to make the metrics experience between the node and controller consistent, via these changes:
- Enable sidecar metrics in `controller.enableMetrics`
- Add appropriate `prometheus.io/scrape` and `prometheus.io/port` to the controller metrics service (Included a parameter to disable this if needed, as otherwise it could result in double scrapes like @sidewinder12s mentions below)
- Add a `ServiceMonitor` object for the node metrics
- Set default additional labels for both `ServiceMonitor` objects to an empty object (instead of setting a `release` label which is likely to be confusing to Helm users)
- Update `metrics.md` to remove confusing or incorrect information, and greatly improve clarity

Fixes #2556 

#### How was this change tested?

Check new serives and service monitors:
```
$ helm template aws-ebs-csi-driver ./charts/aws-ebs-csi-driver --set controller.enableMetrics=true --set controller.serviceMonitor.forceEnable=true --set node.enableMetrics=true --set node.serviceMonitor.forceEnable=true -s templates/metrics.yaml 
---
# Source: aws-ebs-csi-driver/templates/metrics.yaml
apiVersion: v1
kind: Service
metadata:
  name: ebs-csi-controller
  namespace: default
  labels:
    app: ebs-csi-controller
  annotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "3301"
spec:
  selector:
    app: ebs-csi-controller
  ports:
    - name: metrics
      port: 3301
      targetPort: 3301
  type: ClusterIP
... VERY long, cut for brevity, run locally to check full output ...
```

Check `enablePrometheusAnnotations` parameter:
```
$ helm template aws-ebs-csi-driver ./charts/aws-ebs-csi-driver --set node.enableMetrics=true --set node.enablePrometheusAnnotations=false -s templates/metrics.yaml 
---
# Source: aws-ebs-csi-driver/templates/metrics.yaml
apiVersion: v1
kind: Service
metadata:
  name: ebs-csi-node
  namespace: default
  labels:
    app: ebs-csi-node
spec:
  selector:
    app: ebs-csi-node
  ports:
    - name: metrics
      port: 3302
      targetPort: 3302
  type: ClusterIP
```

#### Does this PR introduce a user-facing change?

```release-note
[...]

PR #2651, opened 2025-09-05
Fix service monitor for ebs-csi-node and ebs-csi-controller

timonegk:
#### What type of PR is this?

/kind bug

Currently, the service monitor introduced in #2558 does not work correctly, because `targetPort` is used, but the ports are not specified on the Pod resources.
A ServiceMonitor has two ways to specify the port of the metrics endpoint. `port` is used for the port name of the Service, `targetPort` is used for the port number on the Pod. A ports used as `targetPort` also has to be specified in the port list of the Pod resource. Since this is not the case for ebs-csi-node and ebs-csi-controller, `port` has to be used.

If you prefer, I can also change the PR to keep using `targetPort` and add the correct port specifications to the Pods.

The change was tested by manually editing the ServiceMonitor to use `port`.

```release-note
Fix ServiceMonitor port discovery for ebs-csi-node and ebs-csi-controller
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/_metrics.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_metrics.tpl
@@ -43,7 +43,7 @@ spec:
     matchNames:
       - {{ .Release.Namespace }}
   endpoints:
-    - targetPort: {{ .Port }}
+    - port: metrics
       path: /metrics
       interval: {{ .ServiceMonitor.interval | default "15s"}}
       {{- if .ServiceMonitor.relabelings }}


mugdha-adhav:
We might not need this change if the issue is fixed in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2654

timonegk:
Yes, same issue, different fix. I don't mind which solution you prefer.

ConnorJC3:
#2654 also fixes some other related issues, so I think that's the preferred solution if it completely also solves the issue for this PR. Thank you very much for your contribution, but we'll probably close this out if that is merged.

For now, until we make a decision:
/hold

ConnorJC3:
/close

Superseded by #2654 (thanks for reporting the issue though, do let us know if that PR fixes this, based on my testing it should)

timonegk:
I also think it should fix the issue, I will test the next release as soon as it's available.

PR #2654, opened 2025-09-08
Add containerPort spec for metrics endpoint

### LINKED ISSUE #2653: Missing containerPort declarations for sidecar container metrics endpoints
/kind bug

**What happened?**
The AWS EBS CSI driver Helm chart configures sidecar containers with --http-endpoint arguments to expose metrics on specific ports, but the corresponding containerPort declarations are missing from the container specifications. This prevents proper service discovery by monitoring systems and can cause metrics collection failures.

Each sidecar container with a configured --http-endpoint should have a corresponding containerPort declaration in its container specification, similar to how the main ebs-plugin container is configured.

Test with monitoring system (SumoLogic example)

* Configure metrics collection for the EBS CSI controller (use default values provided in helm chart, and set `enableMetrics` to true)
* Observe that only the main container metrics (port 3301) are collected and forwarded to SumoLogic
* Sidecar metrics (ports 3302, 3303, 3304 and 3305) fail to be discovered/collected

When we try to check the availability of metrics by port-forwarding the service, it works. But the metrics are not scraped reliably by aggregators like SumoLogic.

For getting this work, we had to add custom patch for containerPorts on all sidecars.

**Environment**
- Kubernetes version (use `kubectl version`): v1.31
- Driver version: v1.48.0
- Helm Chart Version: v2.48.0
- Monitoring System: SumoLogic

torredil:
#### What type of PR is this?

/kind feature

#closes #2653

Verified that all sidecar containers now have matching containerPort declarations, conditionally added only when enableMetrics is true.

```release-note
Added containerPort declarations for containers in Helm chart to enable proper metrics discovery by monitoring systems
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/_node.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_node.tpl
@@ -152,6 +152,11 @@ spec:
             - name: healthz
               containerPort: 9808
               protocol: TCP
+            {{- if .Values.node.enableMetrics }}
+            - name: metrics
+              containerPort: 3302
+              protocol: TCP
+            {{- end }}
           livenessProbe:
             httpGet:
               path: /healthz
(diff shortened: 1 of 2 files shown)


torredil:
@ConnorJC3 Yep - its already done in `charts/aws-ebs-csi-driver/templates/_node.tpl`.

PR #2678, opened 2025-09-26
Update gp3MaxTotalIOPS to 80,000 from 16,000

mdzraf:
#### What type of PR is this?

/kind feature

This PR Updates the IOPS Limits for GP3 To The New Limits Set By EBS. 

Created GP3 Volume Using Driver in `us-west-2` with 20,000 IOPS (Higher than 16k Previous Limit)

```release-note
Increase gp3 volume maximum IOPS from 16,000 to 80,000 to match EC2 behavior
```


mdzraf:
#2682 has been merged, we will not update the hardcoded values.

PR #2680, opened 2025-09-30
Update docs to include updated Outpost information

ElijahQuinones:
#### What type of PR is this?

/kind documentation

This PR adds documentation for using the AWS EBS CSI Driver on outpost
Doc

```release-note
NONE
```

### DIFF
--- a/docs/outposts.md
+++ b/docs/outposts.md
@@ -1,0 +1,51 @@
+# Using EBS CSI Driver on AWS Outposts
+
+## Overview
+
+When using the EBS CSI Driver on AWS Outposts, there are several important considerations to be aware of.
+
+### IMDS Metadata Access
+The EBS CSI driver requires access to Instance Metadata Service (IMDS) to function properly on Outposts. Ensure that:
+- IMDS is not disabled on your EC2 instances
+- Ensure the hop limit is set to 2 
+
+### Volume Type 
+- You must explicitly specify a supported volume type (e.g., `gp2`). Check [AWS documentation](https://aws.amazon.com/outposts/rack/features/#topic-0) for the latest supported EBS volume types on your specific Outpost configuration.
+
+### Topology Considerations
+
+- You can use "WaitForFirstConsumer" which will ensure the volume will wait until a stateful pod is specified to 
+provision the volume to allow topology aware provisioning without specifying the Outpost ARN. 
+
+- Otherwise if you would like to specify the Outpost explicitly use `topology.ebs.csi.aws.com/outpost-id` as the topology key along with  `volumeBindingMode:` Immediate.
+
+### StorageClass Examples
+Example Outpost with `volumeBindingMode:` WaitForFirstConsumer
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: outpostExampleOne
+provisioner: ebs.csi.aws.com
+volumeBindingMode: WaitForFirstConsumer
+```
+
+Example with `volumeBindingMode:` Immediate
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: outpostExampleTwo
+provisioner: ebs.csi.aws.com
+parameters:
+  type: gp2  # Specify volume type supported on your outpost
+volumeBindingMode: Immediate
+allowedTopologies:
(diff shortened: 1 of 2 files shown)


docs/outposts.md
```diff
@@ -0,0 +1,53 @@
...
+
+### IMDS Metadata Access
+The EBS CSI driver requires access to Instance Metadata Service (IMDS) to function properly on Outposts. Ensure that:
+- IMDS is not disabled on your EC2 instances
+- If using IMDSv2 ensure the hop limit is set to 2 
+
+### First Generation Outpost Racks
+- **gp3 volumes are not available** on first generation Outpost racks
+- The driver defaults to `gp3` volume type, which will cause provisioning failures
+- You must explicitly specify a supported volume type (e.g., `gp2`). Check AWS documentation for the latest supported EBS volume types on your specific Outpost configuration.
```
torredil: I think this sentence is perfect; suggest removing the two points above so that it doesn't immediately become outdated when gp3 is supported in first gen outpost racks. Also might be a good idea to link to the relevant AWS docs here.

docs/outposts.md
```diff
@@ -0,0 +1,51 @@
...
+
+When using the EBS CSI Driver on AWS Outposts, there are several important considerations to be aware of.
+
+### IMDS Metadata Access
+The EBS CSI driver requires access to Instance Metadata Service (IMDS) to function properly on Outposts. Ensure that:
+- IMDS is not disabled on your EC2 instances
+- If using IMDSv2 ensure the hop limit is set to 2 
+
+### First Generation Outpost Racks
+- You must explicitly specify a supported volume type (e.g., `gp2`). Check AWS documentation for the latest supported EBS volume types on your specific Outpost configuration.
```
ConnorJC3: np: `Check AWS documentation` -> `Check the AWS documentation`

Also, maybe provide a link?

docs/parameters.md
```diff
@@ -24,7 +24,8 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
 | "volumeInitializationRate"   | integer                                           |         |  When creating a volume from a snapshot, this parameter can be used to request a provisioned initialization rate, in MiB/s.                             |
 
 ## Restrictions
-* `gp3` is currently not supported on outposts. Outpost customers need to use a different type for their volumes.
+
+* `gp3` is currently not supported on First Generation Outpost Racks. As the driver defaults to gp3 when no type is passed. Outpost customers using First Generation Outposts need to specify a different type for their volumes.
```
ConnorJC3: Suggest rewording this bullet point to

```
* The EBS CSI Driver defaults to `gp3` volumes when no volume type is specified. If the outpost does not support `gp3` volumes, specify a supported volume type via a `StorageClass`.
```

This avoids documenting how EBS works, and thus is evergreen

docs/outposts.md
```diff
@@ -0,0 +1,51 @@
...
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: nonGen1
+provisioner: ebs.csi.aws.com
+volumeBindingMode: WaitForFirstConsumer
+```
+
+Example of First Generation Outpost with `volumeBindingMode:` Immediate
```
ConnorJC3: Remove references in this section to "first generation" outposts. Just title this example something like "Example with `Immediate` binding mode and `gp2` volume type"

docs/outposts.md
```diff
@@ -0,0 +1,51 @@
+# Using EBS CSI Driver on AWS Outposts
+
+## Overview
+
+When using the EBS CSI Driver on AWS Outposts, there are several important considerations to be aware of.
+
+### IMDS Metadata Access
+The EBS CSI driver requires access to Instance Metadata Service (IMDS) to function properly on Outposts. Ensure that:
+- IMDS is not disabled on your EC2 instances
+- If using IMDSv2 ensure the hop limit is set to 2
```
ConnorJC3: Remove "If using IMDSv2" -> it's confusing and I don't believe the SDK even supports v1 anymore.

docs/outposts.md
ConnorJC3: Also also, see my other comment but basically don't say "first generation outposts" anywhere here.

ConnorJC3 (approved):
lgtm

PR #2682, opened 2025-10-01
Refactor To Get  Maximum IOPS for Capping Through Error Message

mdzraf:
#### What type of PR is this?

/kind feature

This PR adds functionality to retrieve and temporarily cache the maximum IOPS limit for volume types from the error message of a CreateVolume DryRun instead of using a hardcoded table. The driver will default to the hardcoded table if retrieving from the error fails.

Added Unit tests, ran existing ones, and tested manually.

For manual test I created a few PVCs, in the same AZ and added logging code that would show me what the max IOPS limit was set to before IOPS capping and whether it had to make the DryRun call again or if it was going to properly hit the cache if the entry already existed. 

```release-note
Retrieve maximum IOPS limit dynamically via `CreateVolume` dry-run to support increased IOPS limits for GP3 and other volume types
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -22,7 +22,9 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -66,15 +68,15 @@ const (
 // AWS provisioning limits.
 // Source: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html
 const (
-	io1MinTotalIOPS = 100
-	io1MaxTotalIOPS = 64000
-	io1MaxIOPSPerGB = 50
-	io2MinTotalIOPS = 100
-	io2MaxTotalIOPS = 256000
-	io2MaxIOPSPerGB = 1000
-	gp3MaxTotalIOPS = 16000
-	gp3MinTotalIOPS = 3000
-	gp3MaxIOPSPerGB = 500
+	io1MinTotalIOPS    = 100
+	io1FallbackMaxIOPS = 64000
+	io1MaxIOPSPerGB    = 50
+	io2MinTotalIOPS    = 100
+	io2FallbackMaxIOPS = 256000
+	io2MaxIOPSPerGB    = 1000
+	gp3FallbackMaxIOPS = 16000
+	gp3MinTotalIOPS    = 3000
+	gp3MaxIOPSPerGB    = 500
 )
 
 var (
@@ -90,8 +92,9 @@ var (
 )
 
 const (
-	cacheForgetDelay        = 1 * time.Hour
-	volInitCacheForgetDelay = 6 * time.Hour
+	cacheForgetDelay          = 1 * time.Hour
+	volInitCacheForgetDelay   = 6 * time.Hour
+	iopsLimitCacheForgetDelay = 12 * time.Hour
 
 	dryRunInterval = 3 * time.Hour
 
@@ -186,6 +189,17 @@ const (
 	ValidationException = "ValidationException"
 )
 
+// Regex Patterns.
+var (
+	// For getting IOPS limit from gp3/io1 error.
+	// Error example it is used for: "An error occurred (InvalidParameterValue) when calling the CreateVolume operation: Volume iops of 200000 is too high; maximum is 80000".
+	nonIo2ErrRegex = regexp.MustCompile(`(?i)volume iops.*is too high.*maximum is (\d+)`)
+
+	// For getting IOPS limit from io2 error.
+	// Error example it is used for: "An error occurred (InvalidParameterCombination) when calling the CreateVolume operation: io2 volumes configured with greater than 64 TiB or 256K IOPS or 1000:1 IOPS:GB ratio are not supported".
+	io2ErrRegex = regexp.MustCompile(`(?i)(\d+)K IOPS`)
+)
+
 var invalidParameterErrorCodes = map[string]struct{}{
 	"InvalidParameter":            {},
 	"InvalidParameterCombination": {},
@@ -234,6 +248,20 @@ type ModifyDiskOptions struct {
 	VolumeType string
 	IOPS       int32
 	Throughput int32
+}
+
+// iopsLimits represents the IOPS limits set by EBS of a volume dependent on the volume type.
+type iopsLimits struct {
+	maxIops      int32
+	minIops      int32
+	maxIopsPerGb int32
+}
+
+// getVolumeLimitsParams represents the AZ parameters that getVolumeLimits will use to make the DryRun CreateVolume call.
+type getVolumeLimitsParams struct {
+	availabilityZone   string
+	availabilityZoneId string
+	outpostArn         string
 }
 
 // ModifyTagsOptions represents parameter to modify the tags of an existing EBS volume.
@@ -339,6 +367,7 @@ type cloud struct {
 	likelyBadDeviceNames  expiringcache.ExpiringCache[string, sync.Map]
 	latestClientTokens    expiringcache.ExpiringCache[string, int]
 	volumeInitializations expiringcache.ExpiringCache[string, volumeInitialization]
+	latestIOPSLimits      expiringcache.ExpiringCache[string, iopsLimits]
 	accountID             string
 	accountIDOnce         sync.Once
 	attemptDryRun         atomic.Bool
@@ -412,6 +441,7 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
 		likelyBadDeviceNames:  expiringcache.New[string, sync.Map](cacheForgetDelay),
 		latestClientTokens:    expiringcache.New[string, int](cacheForgetDelay),
 		volumeInitializations: expiringcache.New[string, volumeInitialization](volInitCacheForgetDelay),
+		latestIOPSLimits:      expiringcache.New[string, iopsLimits](iopsLimitCacheForgetDelay),
 	}
 
 	// Ensure an EC2 Dry-run API call is made on startup and every dryRunInterval
@@ -572,9 +602,6 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		iops          int32
 		throughput    int32
 		err           error
-		maxIops       int32
-		minIops       int32
-		maxIopsPerGb  int32
 		requestedIops int32
 	)
 
@@ -590,36 +617,8 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		createType = VolumeTypeGP3
 	}
 
-	switch createType {
-	case VolumeTypeGP2, VolumeTypeSC1, VolumeTypeST1, VolumeTypeStandard:
-	case VolumeTypeIO1:
-		maxIops = io1MaxTotalIOPS
-		minIops = io1MinTotalIOPS
-		maxIopsPerGb = io1MaxIOPSPerGB
-	case VolumeTypeIO2:
-		maxIops = io2MaxTotalIOPS
-		minIops = io2MinTotalIOPS
-		maxIopsPerGb = io2MaxIOPSPerGB
-	case VolumeTypeGP3:
-		maxIops = gp3MaxTotalIOPS
-		minIops = gp3MinTotalIOPS
-		maxIopsPerGb = gp3MaxIOPSPerGB
-		throughput = diskOptions.Throughput
-	default:
-		return nil, fmt.Errorf("invalid AWS VolumeType %q", diskOptions.VolumeType)
-	}
-
 	if diskOptions.MultiAttachEnabled && createType != VolumeTypeIO2 {
 		return nil, errors.New("CreateDisk: multi-attach is only supported for io2 volumes")
-	}
-
-	if maxIops > 0 {
-		if diskOptions.IOPS > 0 {
-			requestedIops = diskOptions.IOPS
-		} else if diskOptions.IOPSPerGB > 0 {
-			requestedIops = diskOptions.IOPSPerGB * capacityGiB
-		}
-		iops = capIOPS(createType, capacityGiB, requestedIops, minIops, maxIops, maxIopsPerGb, diskOptions.AllowIOPSPerGBIncrease)
 	}
 
 	tags := make([]types.Tag, 0, len(diskOptions.Tags))
@@ -676,6 +675,26 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 	// EBS doesn't handle empty outpost arn, so we have to include it only when it's non-empty
 	if len(diskOptions.OutpostArn) > 0 {
 		requestInput.OutpostArn = aws.String(diskOptions.OutpostArn)
+	}
+
+	azParams := getVolumeLimitsParams{
+		availabilityZone:   zone,
+		availabilityZoneId: zoneID,
+		outpostArn:         diskOptions.OutpostArn,
+	}
+
+	iopsLimit, err := c.getVolumeLimits(ctx, createType, azParams)
+	if err != nil {
+		return nil, fmt.Errorf("invalid AWS VolumeType %q", diskOptions.VolumeType)
+	}
+
+	if iopsLimit.maxIops > 0 {
+		if diskOptions.IOPS > 0 {
+			requestedIops = diskOptions.IOPS
+		} else if diskOptions.IOPSPerGB > 0 {
+			requestedIops = diskOptions.IOPSPerGB * capacityGiB
+		}
+		iops = capIOPS(createType, capacityGiB, requestedIops, iopsLimit, diskOptions.AllowIOPSPerGBIncrease)
 	}
 
 	if len(diskOptions.KmsKeyID) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/cloud/cloud.go
```diff
@@ -90,8 +92,9 @@ var (
 )
 
 const (
-	cacheForgetDelay        = 1 * time.Hour
-	volInitCacheForgetDelay = 6 * time.Hour
+	cacheForgetDelay          = 1 * time.Hour
+	volInitCacheForgetDelay   = 6 * time.Hour
+	iopsLimitCacheForgetDelay = 24 * time.Hour
```
torredil: 2 questions here: 
- What is the motivation for using an expiring cache?
- Is there a reason why the cache is invalidated every 24 hours? if there is, please add a comment explaining why this is the recommended value.

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+
+	requestInputCopy := *requestInput // We do not want to modify the real input.
+	requestInputCopy.DryRun = aws.Bool(true)
+	requestInputCopy.Iops = aws.Int32(math.MaxInt32) // Ensure we fail with too high IOPS.
+	volType := strings.ToLower(string(requestInputCopy.VolumeType))
+
+	_, err = c.ec2.CreateVolume(ctx, &requestInputCopy)
+	if err != nil {
+		maxIops, err := extractMaxIOPSFromError(err.Error(), volType)
+		// Default To Hardcoded Limits if we can't get the max IOPS from the error message.
```
torredil, line 2529: Suggest adding a log statement here when executing this code path for debugging; we always expect to be able to retrieve the limit from the error.

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+
+	if volumeType == VolumeTypeIO2 {
+		re2 := regexp.MustCompile(`(?i)(\d+)K IOPS`)
+		if matches := re2.FindStringSubmatch(errorMsg); len(matches) > 1 {
+			if val, err := strconv.ParseInt(matches[1], 10, 32); err == nil {
+				result := val * 1000
+				// No real overflow concern here but adding for safety.
+				if result > math.MaxInt32 || result < math.MinInt32 {
+					return 0, fmt.Errorf("maximum IOPS value exceeds maximum value of int32: %d", val)
+				}
```
torredil, line 2582: result < math.MinInt32 is impossible since val is positive and multiplied by 1000, so that can be removed.

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	}
+
+	cacheKey := fmt.Sprintf("%s|%s|%s|%s", volumeType, az, azId, outpost)
+	if c.latestIOPSLimits != nil {
+		if value, ok := c.latestIOPSLimits.Get(cacheKey); ok {
+			return *value, nil
+		}
+	}
+
+	requestInputCopy := *requestInput // We do not want to modify the real input.
```
ConnorJC3: We shouldn't use the customer's input, as other random validations could break the call.

Instead, construct a new input with:
- AZ or AZID
- Outpost (if set)
- Volume Type
- DryRun: true
- Size (Required by EC2)
- Dummy EBS CSI Tag (required by IAM)

pkg/cloud/cloud.go
mdzraf: Agreed but gosec requires a more explicit check it seems because: 

With 
```
if result > math.MaxInt32 {
					return 0, fmt.Errorf("maximum IOPS value exceeds maximum value of int32: %d", val)
				}
				return int32(result), nil
```

On `make verify` I get:
```
pkg/cloud/cloud.go:2548:17: G115: integer overflow conversion int64 -> int32 (gosec)
                                return int32(result), nil
                                            ^
1 issues:
* gosec: 1
```

pkg/cloud/cloud.go
mdzraf: ✅

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+		re1 := regexp.MustCompile(`(?i)volume iops.*?is too high.*?maximum is (\d+)`)
+		if matches := re1.FindStringSubmatch(errorMsg); len(matches) > 1 {
+			if val, err := strconv.ParseInt(matches[1], 10, 32); err == nil {
+				return int32(val), nil
+			}
+		}
+	}
+
+	if volumeType == VolumeTypeIO2 {
+		re2 := regexp.MustCompile(`(?i)(\d+)K IOPS`)
```
ConnorJC3: Make this a top level constant so it isn't re-processed for each call (same with the other regex)

pkg/cloud/cloud.go
mdzraf: >What is the motivation for using an expiring cache?

The reason for this is so that if any IOPS limit changes are deployed to a region, we want to make sure that the driver is not using the old cached values. 

Note: We are aware that the cache will clear `iopsLimitCacheForgetDelay` after last use, so if customers are constantly creating volumes and the cache does not refresh they will have to restart driver in order to get new  limits.

> Is there a reason why the cache is invalidated every 24 hours? if there is, please add a comment explaining why this is the recommended value.

Offline discussion on this: Team decided that we will bring down to 12 hours.

pkg/cloud/cloud.go
```diff
@@ -236,6 +239,13 @@ type ModifyDiskOptions struct {
 	Throughput int32
 }
 
+// IopLimits represents the IOPS limits set by EBS of a volume dependent on the volume type.
+type IopLimits struct {
```
ConnorJC3: change this to `iopsLimits`:
- It shouldn't be exported
- It's typod - `iops` not `iop` (also needs to be fixed in a few variable names)

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	if iops > maxIopsByCapacity && maxIopsByCapacity >= iopLimits.minIops {
 		iops = maxIopsByCapacity
-		klog.V(5).InfoS("[Debug] Capped IOPS for volume", "volumeType", volumeType, "requestedCapacityGiB", requestedCapacityGiB, "maxIOPSPerGB", maxIOPSPerGB, "limit", iops)
+		klog.V(5).InfoS("[Debug] Capped IOPS for volume", "volumeType", volumeType, "requestedCapacityGiB", requestedCapacityGiB, "maxIOPSPerGB", iopLimits.maxIopsPerGb, "limit", iops)
 	}
 	return iops
 }
+
+// Gets IOPS limits for a specific volume type in a specific Zone and caches it. If the limits are cached, simply return limits.
+func (c *cloud) getVolumeLimits(ctx context.Context, volumeType string, requestInput *ec2.CreateVolumeInput) (iopLimits IopLimits, err error) {
```
ConnorJC3: Suggest passing in az/azid/output as their own variables (or creating a struct to pass them in), instead of passing in a `CreateVolumeInput` as that will need to be refactored when we add modify support.

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+		iopLimits.minIops = io2MinTotalIOPS
+		iopLimits.maxIopsPerGb = io2MaxIOPSPerGB
+	case VolumeTypeGP3:
+		iopLimits.minIops = gp3MinTotalIOPS
+		iopLimits.maxIopsPerGb = gp3MaxIOPSPerGB
+	default:
+		return iopLimits, fmt.Errorf("invalid AWS VolumeType %q", volumeType)
+	}
+
+	c.latestIOPSLimits.Set(cacheKey, &iopLimits)
```
ConnorJC3: We shouldn't cache the limits if we got hardcoded values. Otherwise our dry run could get a network timeout or a `RequestLimitExceeded` or whatever and then we cache the fallback forever.

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	volType := strings.ToLower(string(requestInputCopy.VolumeType))
+
+	_, err = c.ec2.CreateVolume(ctx, &requestInputCopy)
+	if err != nil {
+		maxIops, err := extractMaxIOPSFromError(err.Error(), volType)
+		// Default To Hardcoded Limits if we can't get the max IOPS from the error message.
+		if err != nil {
+			switch volType {
+			case VolumeTypeIO1:
+				iopLimits.maxIops = io1MaxTotalIOPS
```
ConnorJC3: Suggest renaming these variables to something like `io1FallbackMaxIOPS` or something like that to make it clear they are fallback values.

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	c.latestIOPSLimits.Set(cacheKey, &iopLimits)
+
+	return iopLimits, nil
+}
+
+// Get what the maxIops is from DryRun error message.
+func extractMaxIOPSFromError(errorMsg string, volumeType string) (int32, error) {
+	// io1 and gp3 have the same error message but io2 has different one.
+	if volumeType == VolumeTypeIO1 || volumeType == VolumeTypeGP3 {
+		re1 := regexp.MustCompile(`(?i)volume iops.*?is too high.*?maximum is (\d+)`)
```
ConnorJC3: Get rid of the two question marks in `.*?`, they're superfluous

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+
+	c.latestIOPSLimits.Set(cacheKey, &iopLimits)
+
+	return iopLimits, nil
+}
+
+// Get what the maxIops is from DryRun error message.
+func extractMaxIOPSFromError(errorMsg string, volumeType string) (int32, error) {
+	// io1 and gp3 have the same error message but io2 has different one.
+	if volumeType == VolumeTypeIO1 || volumeType == VolumeTypeGP3 {
```
ConnorJC3: Can we just make this if the type _isn't_ io2, assuming any future type will likely use this message?

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+		return iopLimits, fmt.Errorf("invalid AWS VolumeType %q", volumeType)
+	}
+
+	c.latestIOPSLimits.Set(cacheKey, &iopLimits)
+
+	return iopLimits, nil
+}
+
+// Get what the maxIops is from DryRun error message.
+func extractMaxIOPSFromError(errorMsg string, volumeType string) (int32, error) {
```
ConnorJC3, line 2573: Can we add a comment of example error messages so future maintainers

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2433,121 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	requestInputCopy := *requestInput // We do not want to modify the real input.
+	requestInputCopy.DryRun = aws.Bool(true)
+	requestInputCopy.Iops = aws.Int32(math.MaxInt32) // Ensure we fail with too high IOPS.
+	volType := strings.ToLower(string(requestInputCopy.VolumeType))
+
+	_, err = c.ec2.CreateVolume(ctx, &requestInputCopy)
+	if err != nil {
+		maxIops, err := extractMaxIOPSFromError(err.Error(), volType)
+		// Default To Hardcoded Limits if we can't get the max IOPS from the error message.
+		if err != nil {
```
ConnorJC3, line 2530: Can we add a debug log in this case?

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2457,141 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	cacheKey := fmt.Sprintf("%s|%s|%s|%s", volumeType, az, azId, outpostArn)
+	if c.latestIOPSLimits != nil {
+		if value, ok := c.latestIOPSLimits.Get(cacheKey); ok {
+			return *value, nil
+		}
+	}
+
+	dryRunRequestInput := &ec2.CreateVolumeInput{
+		VolumeType: types.VolumeType(volumeType),
+		Size:       aws.Int32(1),
```
torredil: Min valid volume size for io1/io2 is 4 GiB - should increase this accordingly to ensure DryRun call returns the error we are striving to induce.

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2457,141 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+	volType := strings.ToLower(string(dryRunRequestInput.VolumeType))
+	_, err = c.ec2.CreateVolume(ctx, dryRunRequestInput, func(o *ec2.Options) {
+		o.APIOptions = nil // Don't add our logging/metrics middleware because we expect errors.
+	})
+	usedFallBack := false
+	if err != nil {
+		maxIops, err := extractMaxIOPSFromError(err.Error(), volType)
+		// Default To Hardcoded Limits if we can't get the max IOPS from the error message.
+		if err != nil {
+			klog.V(5).Infof("[Debug] error getting IOPS limit, defaulting to hardcoded values for volume type %s, %s", volumeType, err.Error())
```
ConnorJC3: Use InfoS not Infof

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2457,141 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+
+	volType := strings.ToLower(string(dryRunRequestInput.VolumeType))
+	_, err = c.ec2.CreateVolume(ctx, dryRunRequestInput, func(o *ec2.Options) {
+		o.APIOptions = nil // Don't add our logging/metrics middleware because we expect errors.
+	})
+	usedFallBack := false
+	if err != nil {
+		maxIops, err := extractMaxIOPSFromError(err.Error(), volType)
+		// Default To Hardcoded Limits if we can't get the max IOPS from the error message.
+		if err != nil {
```
ConnorJC3, line 2530: IMPORTANT: if `err == nil` this code won't set an iops max limit and thus will scale the iops down to 0, we should handle that case

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2457,141 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+		Iops:       aws.Int32(math.MaxInt32),
+		DryRun:     aws.Bool(true),
+		// Required by IAM.
+		TagSpecifications: []types.TagSpecification{
+			{
+				ResourceType: types.ResourceTypeVolume,
+				Tags: []types.Tag{
+					{
+						Key:   aws.String(VolumeNameTagKey),
+						Value: aws.String("DryRunValue"),
```
ConnorJC3: np: Can we make this something like `IopsLimitDryRun` or similar so it's obvious in CloudTrail what the purpose is

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2457,141 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+			return *value, nil
+		}
+	}
+
+	dryRunRequestInput := &ec2.CreateVolumeInput{
+		VolumeType: types.VolumeType(volumeType),
+		Size:       aws.Int32(1),
+		Iops:       aws.Int32(math.MaxInt32),
+		DryRun:     aws.Bool(true),
+		// Required by IAM.
```
ConnorJC3: Make this comment more clear - e.g. "Required by default EBS CSI Driver IAM policy"

pkg/cloud/cloud.go
```diff
@@ -236,6 +250,20 @@ type ModifyDiskOptions struct {
...
 
+// iopsLimits represents the IOPS limits set by EBS of a volume dependent on the volume type.
+type iopsLimits struct {
+	maxIops      int32
+	minIops      int32
+	maxIopsPerGb int32
+}
+
+// paramsForGetVolumeLimits represents the AZ parameters that getVolumeLimits will use to make the DryRun CreateVolume call.
+type paramsForGetVolumeLimits struct {
```
ConnorJC3: `paramsForGetVolumeLimits` -> `getVolumeLimitsParams`

pkg/cloud/cloud.go
```diff
@@ -236,6 +250,20 @@ type ModifyDiskOptions struct {
...
+	AvailabilityZone   string
+	AvailabilityZoneId string
+	OutpostArn         string
```
ConnorJC3: Why are these exported?

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2457,141 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+		klog.V(5).InfoS("[Debug] Capped IOPS for volume", "volumeType", volumeType, "requestedCapacityGiB", requestedCapacityGiB, "maxIOPSPerGB", iopsLimits.maxIopsPerGb, "limit", iops)
 	}
 	return iops
 }
+
+// Gets IOPS limits for a specific volume type in a specific Zone and caches it. If the limits are cached, simply return limits.
+func (c *cloud) getVolumeLimits(ctx context.Context, volumeType string, azParams paramsForGetVolumeLimits) (iopsLimits iopsLimits, err error) {
+	az := azParams.AvailabilityZone
+	azId := azParams.AvailabilityZoneId
+	outpostArn := azParams.OutpostArn
```
ConnorJC3: np: Suggest removing these variables, they don't add much and make the code slightly harder to read due to indirection

pkg/cloud/cloud.go
```diff
@@ -2438,28 +2457,141 @@ func getVolumeAttachmentsList(volume types.Volume) []string {
...
+
+	cacheKey := fmt.Sprintf("%s|%s|%s|%s", volumeType, az, azId, outpostArn)
+	if c.latestIOPSLimits != nil {
```
ConnorJC3: Why the `nil` check? Do we expect it to ever be `nil`?

pkg/cloud/cloud.go
mdzraf: Good Catch, set it to use fallback values if we unexpectedly succeed, and added unit test case for it.

mdzraf:
/approve

Got the required lgtm's + approvals to merge.

PR #2683, opened 2025-10-02
Updating IOPS Limit For GP3

mdzraf:
#### What type of PR is this?

/kind cleanup

This PR Updates the documented limits for GP3 To The[ New Limits Set By EBS](https://aws.amazon.com/about-aws/whats-new/2025/09/amazon-ebs-size-provisioned-performance-gp3-volumes/). 

Created GP3 Volume Using Driver in `us-west-2` with 80,000 IOPS

```release-note
NONE
```


docs/parameters.md
```diff
@@ -24,15 +24,15 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
 | "volumeInitializationRate"   | integer                                           |         |  When creating a volume from a snapshot, this parameter can be used to request a provisioned initialization rate, in MiB/s.                             |
 
 ## Restrictions
-* `gp3` is currently not supported on outposts. Outpost customers need to use a different type for their volumes.
+* `gp3` is currently not supported on First-generation outposts. First-generation outposts customers need to use a different type for their volumes.
```
torredil: ginormous np: 

Remove that second sentence `First-generation outposts customers need to use a different type for their volumes.` In the future, it'll be one less thing to have to update and it doesn't particularly convey any useful information (if gp3 is not supported in a specific zone, its already implied you'll need to use a different volume type).

/lgtm

ConnorJC3: Can I suggest we just revert this change entirely and let #2680 deal with this?

docs/parameters.md
```diff
@@ -24,15 +24,15 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
...
 * If the requested IOPS (either directly from `iops` or from `iopsPerGB` multiplied by the volume's capacity) produces a value above the maximum IOPS allowed for the [volume type](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html), the IOPS will be capped at the maximum value allowed. If the value is lower than the minimal supported IOPS value per volume, either an error is returned (the default behavior), or the value is increased to fit into the supported range when `allowautoiopspergbincrease` is `"true"`.
 * You may specify either the "iops" or "iopsPerGb" parameters, not both. Specifying both parameters will result in an invalid StorageClass.
```
ConnorJC3, line 29: Suggest we add a bullet point explaining the auto-detection behavior.

Something like:
```
* When using `iopsPerGb`, the maximum supported IOPS will be automatically detected via a dry-run `CreateVolume` API call.
```

docs/parameters.md
torredil: Yeah that sounds good to me.

docs/parameters.md
```diff
@@ -27,12 +27,13 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
...
+* When using `iopsPerGb`, the maximum supported IOPS will be automatically detected via a dry-run `CreateVolume` API call.
 
 | Volume Type                | Min total IOPS | Max total IOPS | Max IOPS per GB |
 |----------------------------|----------------|---------------|-----------------|
 | io1                        | 100            | 64000         | 50              |
 | io2                        | 100            | 256000        | 1000            |
-| gp3                        | 3000           | 16000         | 500             |
+| gp3                        | 3000           | 80000         | 500             |
```
ElijahQuinones: I know you are editing what is already here but this will be inaccurate for some customers and will become outdated at some point can I suggest we simply link to the official AWS docs https://docs.aws.amazon.com/ebs/latest/userguide/ebs-volume-types.html

mdzraf: Do we drop the whole table?

ElijahQuinones: I would suggest it as we shouldn't be documenting things that are not driver specific and have AWS official docs that will be regularly updated but please take this as a non blocking comment

mdzraf: I think that's a good idea, I don't see why we couldn't do it, will get some more input from team to decide.

PR #2690, opened 2025-10-06
Disable nrext64 if LegacyXFSProgs is set

mdzraf:
#### What type of PR is this?

/kind feature

The EKS-D base image used by the EBS CSI Driver upgraded xfsprogs from 5.18.0 to 6.12.0. This appears to have caused XFS volumes to be formatted with the nrext64 feature, which is not supported by the Kernel used in Amazon Linux 2. This PR  disables the nrext64 if `legacyXFS` is set to true. 

- `make test`
- Manually by:
Putting up an EKS cluster that had the driver installed with these changes and the flag set, then provisioning a volume that has `csi.storage.k8s.io/fstype: xfs` and making sure it successfully creates and attaches. 

```release-note
Setting legacyXFS to true now disables nrext64. 
```

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -267,7 +267,7 @@ func (d *NodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		formatOptions = append(formatOptions, "-C", ext4ClusterSize)
 	}
 	if fsType == FSTypeXfs && d.options.LegacyXFSProgs {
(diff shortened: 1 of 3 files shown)


mdzraf:
Will test manually with AL2023 nodes as well.

mdzraf:
Manual tests on AL2023 Nodes have passed ✅

torredil (approved):
/lgtm

mdzraf:
Got required lgtm's and conducted proper manual tests

/approve

PR #2707, opened 2025-10-10
Update Dependabot to run at 15:30 (3:30PM) on Fridays

ConnorJC3:
#### What type of PR is this?

/kind cleanup

See title - more aligns Dependabot with when we do dependency bumps normally

N/A

```release-note
NONE
```


ConnorJC3:
/override pull-aws-ebs-csi-driver-verify

Dependabot-only change

PR #2715, opened 2025-10-13
Bump dependencies and sidecars

ElijahQuinones:
#### What type of PR is this?

/kind cleanup

Bump dependencies and sidecars. Dep bump resulted in new linters being enabled as such I made changes to comply with the linters.

make update && make test && make verify 
CI

```release-note
Bump dependencies and sidecars
```


.golangci.yml
```diff
@@ -81,6 +81,9 @@ linters:
       - linters:
           - gosec
         path: tests/sanity
+      - linters:
+          - godoclint
+        path: pkg/mounter/mount.go
```
ConnorJC3: Is this just because it's too hard to fix? If yes, can we add a TODO comment so we remember to fix?

go.mod
```diff
@@ -12,21 +12,21 @@ require (
 	github.com/container-storage-interface/spec v1.11.0
 	github.com/golang/mock v1.6.0
 	github.com/google/go-cmp v0.7.0
-	github.com/kubernetes-csi/csi-proxy/client v1.2.1
-	github.com/kubernetes-csi/csi-proxy/v2 v2.0.0-alpha.1
+	github.com/kubernetes-csi/csi-proxy/client v1.3.0
+	github.com/kubernetes-csi/csi-proxy/v2 v2.0.0-alpha.2
```
ConnorJC3: This is going to break the Windows with a panic, you need to revert it

.golangci.yml
ElijahQuinones: This is because mount.go and mock_mount.go both have a godoc package comment and mock_mount.go is auto generated so unless we want to remove the godoc comment from the real mount.go  we have to ignore it.

ConnorJC3: Can we just `nolint` the one bad line then?

ElijahQuinones: Sounds good will do

PR #2720, opened 2025-10-15
Dump pod logs to artifacts instead of stdout

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Instead of dumping pod logs to stdout and absolutely flooding the E2E test output, instead put them in the artifacts folder so we can access them if needed. Also turns down the verbosity of `ginkgo` as we get the full output via junit anyways.

CI

```release-note
NONE
```


ElijahQuinones:
/lgtm
/approve 
https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/2720/pull-aws-ebs-csi-driver-external-test/1978860579854487552/artifacts/

PR #2722, opened 2025-10-16
Minor increase to NVMe Metrics Performance

ConnorJC3:
#### What type of PR is this?

/kind cleanup

- Inline `nvmeReadLogPage` into `getNVMEMetrics` as it's only called in one place, and having a separate function forces us to copy the buffer while holding the fd open
- Only open the device as RDONLY, as it's not necessary to open RW for `ioctl` on linux (this is also a minor improvement to security stance)
- Allocate structs before opening the file, to minimize the amount of time we hold the file descriptor open and potentially blocking I/O

E2E tests

```release-note
Minor tweaks to NVMe metrics collector performance
```


mdzraf:
Lgtm except for the typo on the commit message

mdzraf (approved):
/lgtm
/approve

PR #2726, opened 2025-10-20
Support Node-Local Shared Cache Volume Access

### LINKED ISSUE #2552: [Feature Request] Support Node-Local Shared Cache Volume Access for Pods

In high-performance or data-intensive workloads on Amazon EKS, it’s common for multiple co-located pods on the same node to require access to a shared dataset—for example, files cached from Amazon S3 or data pre-populated on an EBS volume prior to node launch.

However, Kubernetes does not natively support exposing a single node-local volume—such as a pre-attached EBS volume—to multiple pods on the same node without relying on non-portable constructs like hostPath, using local PersistentVolumes with a single-node scheduling constraint, or building a custom controller to orchestrate volume sharing.

While the existing EBS CSI driver enables sharing a volume among multiple pods on a single same node, it is not designed to scale this pattern cluster-wide to ensure pods on each node can share their respective node-local volumes seamlessly.

I am seeking support for safely mounting a shared, node-local volume into multiple pods running on the same node. This capability could be enabled by enhancing the existing [EBS CSI Driver’s static provisioning](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/static-provisioning/manifests) functionality. Specifically, rather than requiring the EBS volume ID to be hardcoded in the PersistentVolume definition (e.g., `volumeHandle: vol-03c604538dd7d2f41`), the CSI driver could support a node-resolved identifier such as the device name (e.g., `deviceName: /dev/xvdbb`). This approach would allow the driver to identify the existing EBS volume already attached to the node where the pod is running.

With this enhancement, each pod would be able to share its respective node’s EBS volume independently of volumes on other nodes, enabling safe, scalable sharing of node-local volumes across multiple pods on the same node.

Existing static provisioning PV

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: test-pv
spec:
  # ... other spec fields ...
  csi:
    driver: ebs.csi.aws.com
    volumeHandle: vol-03c604538dd7d2f41
    # ... other CSI fields ...
```

Suggested static provisioning PV

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: test-pv
spec:
  # ... other spec fields ...
  csi:
    driver: ebs.csi.aws.com
    # Hardcoded value signalling the CSI driver to look up a locally attached volume on the node
    volumeHandle: local-ebs-volume
    volumeAttributes:
      # The device name or any other identifier for the CSI driver to use to find a locally attached volume on the node
      deviceName: /dev/xvdbb 
    # ... other CSI fields ...
```

This enhancement would:

* Allow a single PersistentVolume and PVC to be defined cluster-wide but scoped to a node.
* Enable multiple pods on the same node to mount the same volume concurrently.
* Each node has it’s own  EBS volume and all pods running on the same node sharing the node’s attached EBS volume.
* Leverage the existing EBS CSI plugin with minimal changes, focusing only on the static provisioning and node publishing path.

1. Dynamic Provisioning per Pod (Standard EBS CSI Driver Flow):  Each pod receives its own dynamically provisioned EBS volume, managed by the CSI driver. The problems are:
    1. Pods cannot natively share data across volumes.
    2. Not cost-effective due to one EBS volume per pod.
    3. Hits the per-instance EBS volume attachment limits in high-density pod scenarios.
    4. Slower startup times due to volume provisioning and attachment delays.

2. Custom Control Loop That Directly Creates Pods on Specific Nodes: This approach involves building a custom controller that bypasses Kubernetes’ native scheduler by directly creating pods with a specified nodeName, effectively controlling where each pod runs.  The problems are:
    1. Incompatible with workloads that rely on Kubernetes’ native scheduling to optimize placement across diverse node types.
    2. EKS environments often use a wide range of EC2 instance types with varying CPU, memory, and ephemeral storage configurations—bypassing the scheduler forfeits the ability to match pod resource requirements to the most suitable node.
    3. Adds operational complexity and reduces portability across environments.

torredil:
#### What type of PR is this?

/kind feature

Introduces a node-local static provisioning mode that enables a single cluster-wide PV/PVC to mount pre-attached, node-specific EBS volumes. When pods reference this PVC, each node independently mounts its own local EBS device and all pods on that node share the mount. This is useful for node-local caching scenarios where each node has a dedicated EBS volume already attached.

The feature is enabled via the `--enable-node-local-volumes` flag and uses a special `local-ebs://` volume handle prefix to identify node-local volumes. The controller resolves the actual volume ID from the device name on each node to support `ReadWriteMany` access mode.

closes #2552

```
make update && make verify && make test
```

```
kubectl get pvc                                                                                               
NAME                   STATUS   VOLUME                CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
node-local-cache-pvc   Bound    node-local-cache-pv   10Gi       RWX                           <unset>                 76s
```

```
kubectl get pv                                                                                               
NAME                  CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                          STORAGECLASS   VOLUMEATTRIBUTESCLASS   REASON   AGE
node-local-cache-pv   10Gi       RWX            Retain           Bound    default/node-local-cache-pvc                  <unset>                          106s
```

```
kubectl get pods -l app=cache-test -o wide                                                                    
NAME                          READY   STATUS    RESTARTS   AGE   IP               NODE                                           NOMINATED NODE   READINESS GATES
cache-test-6f7f5987bb-dztrr   1/1     Running   0          61s   192.168.19.49    ip-192-168-15-196.us-west-2.compute.internal   <none>           <none>
cache-test-6f7f5987bb-knx9q   1/1     Running   0          61s   192.168.76.200   ip-192-168-82-170.us-west-2.compute.internal   <none>           <none>
cache-test-6f7f5987bb-mvhjk   1/1     Running   0          62s   192.168.72.123   ip-192-168-82-170.us-west-2.compute.internal   <none>           <none>
cache-test-6f7f5987bb-wrllj   1/1     Running   0          61s   192.168.23.48    ip-192-168-15-196.us-west-2.compute.internal   <none>           <none>
```

```
kubectl logs ebs-csi-controller-59587c7bf7-7q4zc -n kube-system             
Defaulted container "ebs-plugin" out of: ebs-plugin, csi-provisioner, csi-attacher, csi-resizer, liveness-probe
I1019 22:57:51.693400       1 main.go:154] "Initializing metadata"
I1019 22:57:51.693748       1 metadata.go:66] "Attempting to retrieve instance metadata from IMDS"
I1019 22:57:51.698118       1 metadata.go:69] "Retrieved metadata from IMDS"
I1019 22:57:51.699515       1 driver.go:72] "Driver Information" Driver="ebs.csi.aws.com" Version="v1.51.0"
I1019 22:58:09.215535       1 controller.go:565] "ControllerPublishVolume: resolved node-local volume" volumeID="local-ebs://dev/xvdbz" realVolumeID="vol-0af977b743f8e9277" nodeID="i-050f3d45febd84cc5" deviceName="/dev/xvdbz"
I1019 22:58:09.215535       1 controller.go:565] "ControllerPublishVolume: resolved node-local volume" volumeID="local-ebs://dev/xvdbz" realVolumeID="vol-022eab8074d55582e" nodeID="i-0378636d42f6de9ce" deviceName="/dev/xvdbz"
```

```release-note
Add node-local volume support, enabling a single cluster-wide PV/PVC to mount node-specific EBS volumes with ReadWriteMany access.
```


pkg/cloud/cloud.go
```diff
@@ -1641,6 +1641,27 @@ func (c *cloud) GetDiskByID(ctx context.Context, volumeID string) (*Disk, error)
...
 }
 
+func (c *cloud) GetVolumeIDByNodeAndDevice(ctx context.Context, nodeID string, deviceName string) (string, error) {
+	instance, err := c.getInstance(ctx, nodeID)
+	if err != nil {
+		return "", fmt.Errorf("failed to get instance %s: %w", nodeID, err)
+	}
+
+	if instance.RootDeviceName != nil && *instance.RootDeviceName == deviceName {
+		return "", fmt.Errorf("device %s is the root device: %w", deviceName, ErrNotFound)
```
ElijahQuinones: Can we use `INVALID_ARGUMENT` here instead of not found as the volume is there but we are refusing to use it as it is an invalid volume to have used.

pkg/cloud/cloud_test.go
```diff
@@ -5140,6 +5140,115 @@ func createDescribeVolumesOutput(volumeIDs []*string, nodeID, path, state string
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			mockEC2 := NewMockEC2API(mockCtrl)
+			c := newCloud(mockEC2)
+
+			ctx := t.Context()
+
+			if tc.instanceErr != nil {
+				mockEC2.EXPECT().DescribeInstances(gomock.Any(), gomock.Any()).Return(nil, tc.instanceErr)
```
ElijahQuinones: In light of us trying to remove all instances of  `gomock.Any()` from our unit tests could we remove the gomock.Any() from all the tests here/leave a comment as to why we are using gomock.Any() such to avoid having to refactor these as part of the larger task.

pkg/cloud/cloud.go
torredil: Great suggestion 👍 

Looking at Cloud return codes:

```
/ ErrInvalidArgument is returned if parameters were rejected by cloud provider.
ErrInvalidArgument = errors.New("invalid argument")

// ErrInvalidRequest is returned if parameters were rejected by driver.
ErrInvalidRequest = errors.New("invalid request")
```

Lets update this to return `ErrInvalidRequest`, as that is most appropriate since the request is valid from the cloud provider perspective but we explicitly reject this condition at the driver layer. On controller side, this should be treated as an internal error.

pkg/cloud/cloud_test.go
torredil: Done.

mdzraf (approved):
/lgtm

mdzraf:
/retest

TestPatchNewNodes is flaking

PR #2732, opened 2025-10-21
Fix IOPS handling

ConnorJC3:
#### What type of PR is this?

/kind bug

Fix several places IOPS is handled incorrectly:
- Fixes clone E2E test which tries to create a GP2 volume with IOPS (previously worked due to a bug)
- Fixes `extractMaxIOPSFromError` to return `0` for cases when the volume type is invalid
- Fixes `CreateDisk`/`ResizeOrModifyDisk` to not fail if the volume type is unknown
- Fix `capIOPS` to not apply limits when they are 0

CI

```release-note
Improve handling of volume types that do not have hardcoded IOPS limits and/or do not support IOPS
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -677,18 +677,15 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		outpostArn:         diskOptions.OutpostArn,
 	}
 
-	iopsLimit, err := c.getVolumeLimits(ctx, createType, azParams)
-	if err != nil {
-		return nil, fmt.Errorf("invalid AWS VolumeType %q", diskOptions.VolumeType)
-	}
+	iopsLimits := c.getVolumeLimits(ctx, createType, azParams)
 
 	if diskOptions.IOPS > 0 {
 		iops = diskOptions.IOPS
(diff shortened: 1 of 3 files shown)


pkg/cloud/cloud.go
```diff
@@ -2725,18 +2711,22 @@ func (c *cloud) getVolumeLimits(ctx context.Context, volumeType string, azParams
 		iopsLimits.minIops = gp3MinTotalIOPS
 		iopsLimits.maxIopsPerGb = gp3MaxIOPSPerGB
 	default:
-		return iopsLimits, fmt.Errorf("invalid AWS VolumeType %q", volumeType)
+		klog.V(5).InfoS("[Debug] Unknown volume type, not setting limits", "volumeType", volumeType)
```
torredil: np: Suggest "IOPS not supported for volume type, not setting limits".

/lgtm

torredil (approved):

pkg/cloud/cloud.go
ConnorJC3: The point of this fallback is that if we reach it, we don't really know if the volume supports IOPS or not

I'm going to switch it to `No known limits for volume type, not performing capping`, let me know if you have any suggestions for that

Issue #2728, opened 2025-10-21
regression in throughput handling


cburroughs:
/kind bug

**What happened?**

We have a storage class along the lines of
```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "false"
  name: gp3-boost
parameters:
  csi.storage.k8s.io/fstype: xfs
  iops: "5001"
  throughput: "601"
  type: gp3
provisioner: ebs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
```

After updating from `v1.39.0-eksbuild.1` to `v1.51.0-eksbuild.1` we observed that `throughput` was no longer being set and volumes were instead being created with the default value of `125`.  We verified that in the AWS web console, observations in the container, and from the cloudtrail api logs:

<img width="755" height="112" alt="Image" src="https://github.com/user-attachments/assets/dc931125-d37d-404f-8802-970a61e2c1a4" />

After rolling back to `v1.49.1-eksbuild.1` we again saw non-default throughput numbers (601 in the example above) reported in the web console and observed from within the container.  That is, on that version everything was working as expected.

**Environment**
- Kubernetes version (use `kubectl version`):
```
 kubectl --context eqci02build version
Client Version: v1.33.4
Kustomize Version: v5.6.0
Server Version: v1.33.5-eks-113cf36
```
- Driver version: `    v1.51.0-eksbuild.1`

ConnorJC3:
Hi @cburroughs thanks for reporting this. We've merged a fix and are working on an urgent release, in the meantime if you use `v1.49.1` or older it won't contain this bug.

PR #2737, opened 2025-10-24
Log volume attributes after successful creation

torredil:
#### What type of PR is this?

/kind cleanup

This PR adds logging of volume attributes after successful volume creation to help cluster admins observe what EC2/EBS provisioned at the Kubernetes layer. This idea surfaced from https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2728. 

```
make update && make verify && make update
```

CI

```release-note
Added logging of volume attributes at controller.logLevel 7+ after successful volume creation for improved observability.
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -772,9 +772,13 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
 		return nil, errors.New("disk size was not returned by CreateVolume")
 	}
 
-	if err := c.waitForVolume(ctx, volumeID); err != nil {
+	volume, err := c.waitForVolume(ctx, volumeID)
+	if err != nil {
 		return nil, fmt.Errorf("timed out waiting for volume to create: %w", err)
 	}
+
+	klog.V(7).InfoS("CreateDisk: volume created successfully", "volumeName", volumeName, "volume", volume)
+
 	return &Disk{CapacityGiB: size, VolumeID: volumeID, AvailabilityZone: zone, SnapshotID: diskOptions.SnapshotID, SourceVolumeID: diskOptions.SourceVolumeID, OutpostArn: outpostArn}, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/cloud.go
```diff
@@ -772,9 +772,20 @@ func (c *cloud) CreateDisk(ctx context.Context, volumeName string, diskOptions *
...
 	}
+
+	klog.V(4).InfoS("CreateDisk: volume created successfully",
+		"volumeID", volumeID,
+		"volumeName", volumeName,
+		"volumeType", volume.VolumeType,
+		"size", aws.ToInt32(volume.Size),
+		"iops", aws.ToInt32(volume.Iops),
+		"throughput", aws.ToInt32(volume.Throughput),
+		"multiAttachEnabled", volume.MultiAttachEnabled)
```
ConnorJC3: I really like this iteration of the concept, rather than trying to log "mismatches" which we can't be sure of.

Two things:
- What does the output look like if the volume doesn't support `iops` or any other parameter? Does it log `0`, `nil`, etc?
- I was looking through [the CV API docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html#API_CreateVolume_ResponseElements) and there's actually a bunch of extra stuff we don't log here (e.g. AZ, initialization rate, KMS key, AZ ID). Would it ever make sense to just log the entire struct, like `"volume", volume`? (Downside of this approach: it'll bloat the log size a bit, but I don't think that's particularly bad given this is a higher-loglevel 'debug' style log)

ElijahQuinones (approved):
/lgtm

pkg/cloud/cloud.go
torredil: > What does the output look like if the volume doesn't support iops or any other parameter?

Here is an example statement for io2, which doesn't support throughput:

```
I1024 17:19:45.823719       1 cloud.go:780] "CreateDisk: volume created successfully" volumeID="vol-0847ebb9f5098d1f7" volumeName="pvc-572b8d4e-6a75-4f9f-b764-2308b6a07b98" volumeType="io2" size=4 iops=100 throughput=0 multiAttachEnabled=false
```

> Would it ever make sense to just log the entire struct, like "volume", volume? 

It does significantly bloat the logging (primarily due to tags, for example):

```
   "Volumes": [
        {
            "Iops": 100,
            "Tags": [
                {
                    "Key": "kubernetes.io/created-for/pvc/namespace",
                    "Value": "default"
                },
                {
                    "Key": "CSIVolumeName",
                    "Value": "pvc-572b8d4e-6a75-4f9f-b764-2308b6a07b98"
                },
                {
                    "Key": "kubernetes.io/created-for/pv/name",
                    "Value": "pvc-572b8d4e-6a75-4f9f-b764-2308b6a07b98"
                },
                {
                    "Key": "ebs.csi.aws.com/cluster",
                    "Value": "true"
                },
                {
                    "Key": "kubernetes.io/created-for/pvc/name",
                    "Value": "ebs-claim"
                }
            ],
```

That may not necessarily be completely undesirable, but I don't think log level 4 would be appropriate for it. We could either a) log the entire struct here and increase the v-level or b) have a separate log statement which logs the entire struct at higher v-level (8 or 10). Do you have a preference?

ConnorJC3: I think personally I'd lean towards only having 1 log message (i.e. changing this one and upping the verbosity) but I don't have a particularly strong opinion so I'd be ok with either.

I wouldn't go higher than level 7 though, as that's the level our Helm debug logs parameter uses: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/24495d387fcdbc75fa4864d8a3ef32a9103d1f34/charts/aws-ebs-csi-driver/templates/controller.yaml#L106-L108

torredil: Done.

ConnorJC3 (approved):
/lgtm
/approve

PR #2739, opened 2025-10-28
Bumping Dependencies Ahead of v1.52.0 release

mdzraf:
#### What type of PR is this?

Bumping Dependencies Ahead of v1.52.0 release

```
make update
make test
make verify
make cluster/image
```

```release-note
NONE
```


pkg/cloud/volume_limits_table.go
```diff
@@ -501,7 +497,7 @@ var volumeLimits = map[string]volumeLimit{
 	"m8a.large":           {32, "dedicated"},
 	"m8a.medium":          {32, "dedicated"},
 	"m8a.metal-24xl":      {64, "dedicated"},
-	"m8a.metal-48xl":      {79, "dedicated"},
```
torredil, line 504: Have we validated the correctness of this diff?

ConnorJC3: The API actually does report shared here, appears to be a regression:

```
$ aws ec2 describe-instance-types --instance-type m8a.metal-48xl | grep AttachmentLimitType
                "AttachmentLimitType": "shared"
```

But that's wrong so we need to add it to our list of hardcoded fixes in `volume_limits.go`

pkg/cloud/volume_limits_table.go
```diff
@@ -23,10 +23,6 @@ var nonNitroInstanceTypes = map[string]struct{}{
 	"f1.16xlarge":  {},
 	"f1.2xlarge":   {},
 	"f1.4xlarge":   {},
-	"g3.16xlarge":  {},
-	"g3.4xlarge":   {},
-	"g3.8xlarge":   {},
-	"g3s.xlarge":   {},
```
ConnorJC3, line 29: Did we run the script in all regions, including opt-in regions? These shouldn't be getting dropped.

hack/e2e/config.sh
```diff
@@ -50,7 +50,7 @@ LINUX_AMI=${LINUX_AMI:-}
 WINDOWS_AMI=${WINDOWS_AMI:-}
 
 EBS_INSTALL_SNAPSHOT=${EBS_INSTALL_SNAPSHOT:-"true"}
-EBS_INSTALL_SNAPSHOT_VERSION=${EBS_INSTALL_SNAPSHOT_VERSION:-"v8.3.0"}
+EBS_INSTALL_SNAPSHOT_VERSION=${EBS_INSTALL_SNAPSHOT_VERSION:-"8.4.0"}
```
ConnorJC3: I think we're missing a `v` here?

hack/tools/install.sh
```diff
@@ -40,7 +40,7 @@ HELM_VERSION="v3.19.0"
 KOPS_VERSION="v1.34.0-alpha.1"
```
torredil: Can bump up to non-alpha tag now: `v1.34.0`.

pkg/cloud/volume_limits_table.go
mdzraf: Ill manually re-add, this instance type is now listed as a [previous generation instance](https://docs.aws.amazon.com/ec2/latest/instancetypes/pg.html#:~:text=c4.4xlarge%20%7C%20c4.8xlarge-,G3,-g3.4xlarge%20%7C%20g3.8xlarge), maybe the script didn't catch it because of that.

hack/tools/install.sh
mdzraf: ✅

hack/e2e/config.sh
mdzraf: ✅

pkg/cloud/volume_limits_table.go
mdzraf: ✅

ConnorJC3: We need to fix this in `volume_limits.go`, not manually edit the table file. The table file should exactly as the script outputs or we end up with an infinite chain of manual fixes everytime somebody generates the file.

mdzraf: It's already there as of two weeks ago, https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b95d8b1b73cb6c8e6c9ba4566a2fc226c53fd9c3/pkg/cloud/volume_limits.go#L31

ConnorJC3: Ah, cool thanks, re-resolving then

ConnorJC3 (approved):
/lgtm
(other than maybe unit test failures?)

PR #2747, opened 2025-10-31
Add disableMutation parameter to v2.52.0 Helm Changelog

mdzraf:
#### What type of PR is this?

/kind documentation

Manually adding Helm Changelog entry that was missed by release-notes script.

```release-note
NONE
```

### DIFF
--- a/charts/aws-ebs-csi-driver/CHANGELOG.md
+++ b/charts/aws-ebs-csi-driver/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Feature
 
 - Bump driver version to `v1.52.0`.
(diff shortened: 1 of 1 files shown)


charts/aws-ebs-csi-driver/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Feature
 
 - Bump driver version to `v1.52.0`.
+- Add Helm parameter to disable mutation for node SA [#2723](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2723), [@ConnorJC3](https://github.com/ConnorJC3)
```
ConnorJC3: Can I suggest we use the release note from the PR?
```
Add Helm parameter `node.serviceAccount.disableMutation` to disable mutating RBAC permissions to the `ebs-csi-node` service account. When enabled, driver features such as taint removal may not function.
```

As written, it's confusing and arguably wrong

PR #2749, opened 2025-10-31
Fix off-by-one error in volume limits for default nitro instances

### LINKED ISSUE #2748: Volume Attachment Capacity Issue: regression in v1.49.0
/kind bug

**What happened?**

After upgrading the AWS EBS CSI Driver from `v1.47.0` to `v1.50.2` in our clusters, we observed a [Volume Attachment Capacity Issue (FAQ)](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#volume-attachment-capacity-issues) resulting in pod creation failures. Specifically, stateful pods were scheduled to nodes that did not have a free slot for EBS volume attachment, causing the pods to remain in a pending state until a volume was detached from the node.

Upon reviewing recent source code changes, we identified a potential **regression** introduced in version `v1.49.0` with the PR:

- #2660

cc @ConnorJC3 to raise awareness of this issue.

**What did you expect to happen?**

The driver should correctly set the volume attachment capacity limit for all VM instance types, in alignment with AWS documentation and actual instance capabilities.

We have not attempted to reproduce the issue in a minimal environment. However, the problem was consistently observed in our production setup, and our source code review provides sufficient evidence to report this issue.

**Anything else we need to know?**

- PR #2660 introduced a utility to generate a volume limits table: see [hack/generate-volume-limits-table/main.go](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/74ac0e5afdb8d05a7424d3485d8fa156b843da82/hack/generate-volume-limits-table/main.go).
	- The `volumeLimits` map contains instance types with non-standard volume limits. Non-nitro instances, and nitro instances with the default limit (**shared 27**) are excluded.
	- Instance types with the default limit of **27** slots are therefore not explicitly listed in the `volumeLimits` map (see [main.go#L57-L63](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/74ac0e5afdb8d05a7424d3485d8fa156b843da82/hack/generate-volume-limits-table/main.go#L57-L63), [main.go#L182](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/74ac0e5afdb8d05a7424d3485d8fa156b843da82/hack/generate-volume-limits-table/main.go#L182), and [volume_limits_table.go#L93-L95](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/74ac0e5afdb8d05a7424d3485d8fa156b843da82/pkg/cloud/volume_limits_table.go#L93-L95)).
	- The value **27** is consistent with AWS documentation ([gp_storage-ebs](https://docs.aws.amazon.com/ec2/latest/instancetypes/gp.html#gp_storage-ebs)), which lists "Up to 27 (Shared limit)" for most instance types.

However, in [pkg/cloud/volume_limits.go#L54](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/74ac0e5afdb8d05a7424d3485d8fa156b843da82/pkg/cloud/volume_limits.go#L54), the `GetVolumeLimits(instanceType string)` function, after checking the `volumeLimits` map, falls back to returning **28** as the default limit.

This fallback is inconsistent with both the generated volume limits table and the AWS documentation. The default should be **27**, not **28**.

As a result, the AWS EBS CSI Driver currently assumes a limit that is one higher than the actual value for all instance types with the default limit of 27 slots. This leads to the volume attachment capacity misalignment described above.

The fix appears straightforward: change the default in [pkg/cloud/volume_limits.go#L54](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/74ac0e5afdb8d05a7424d3485d8fa156b843da82/pkg/cloud/volume_limits.go#L54) from `28` to `27`. We have not submitted a PR immediately, as testing a local build of the driver was not feasible within our current time frame.

**Additional notes:**

We believe the source of confusion is as follows:

The general description of EBS volume attachment limits in the [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html#shared-limit) states:

> A volume attachment limit is **shared** between Amazon EBS volumes, network interfaces, and NVMe instance store volumes.
> You can attach any number of Amazon EBS volumes up to that limit, **less the number of attached network interfaces** and NVMe instance store volumes.
>
> Most Nitro instances support a maximum of **28** attachments.
> Keep in mind that every instance **must have at least one network interface**.
> Example: With an m5.xlarge instance with only the primary network interface, you can attach 27 EBS volumes. **28** volumes **- 1 network interface** = **27**

Thus, while the general limit is **28**, every instance has a mandatory primary network interface that counts towards this limit, resulting in an effective EBS volume attachment limit of **27**.

On the [instance type limits page](https://docs.aws.amazon.com/ec2/latest/instancetypes/gp.html#gp_storage-ebs), the default limit is consistently listed as **27**, with no mention of the general shared limit of 28. It appears AWS has already subtracted 1 for the mandatory network interface in these tables and in the metadata service, which is reflected in the generated volume limits table in the CSI driver codebase.

This implicit deduction is consistent with the implementation in [pkg/driver/node.go#L793-L798](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/74ac0e5afdb8d05a7424d3485d8fa156b843da82/pkg/driver/node.go#L793-L798), where the available attachments are calculated by subtracting one less than the number of network interfaces, since the primary interface is already accounted for in the default limit of 27.

cc @dimityrmirchev @dguendisch @kon-angelo @hebelsan

**Environment**
- Kubernetes version (use `kubectl version`): `1.33.4`
- Driver version: `v1.49.0` (the regression was likely introduced in `v1.49.0` and affects all subsequent versions)

ConnorJC3:
#### What type of PR is this?

/kind bug

Fixes #2748 

Manually + added a new unit test to guard this case

```release-note
Fix off-by-one error in volume limits for default nitro instances which would get PVCs stuck attempting to attach to a full instance.
```

### DIFF
--- a/pkg/cloud/volume_limits.go
+++ b/pkg/cloud/volume_limits.go
@@ -50,8 +50,8 @@ func GetVolumeLimits(instanceType string) (int, string) {
 		return limit.maxAttachments, limit.attachmentType
 	}
 
-	// Default to shared limit of 28
-	return 28, util.AttachmentShared
+	// Default to shared limit of 27
+	return 27, util.AttachmentShared
 }
 
 // KnownInstanceTypes returns all known instance types from the limits table.
--- a/pkg/driver/node_test.go
+++ b/pkg/driver/node_test.go
@@ -1074,6 +1074,21 @@ func TestGetVolumesLimit(t *testing.T) {
 			},
 		},
 		{
+			name: "m5.large_volume_attach_limit",
+			options: &Options{
+				VolumeAttachLimit:         -1,
+				ReservedVolumeAttachments: -1,
+			},
+			expectedVal: 27,
+			metadataMock: func(ctrl *gomock.Controller) *metadata.MockMetadataService {
+				m := metadata.NewMockMetadataService(ctrl)
+				m.EXPECT().GetNumBlockDeviceMappings().Return(0)
+				m.EXPECT().GetInstanceType().Return("m5.large")
+				m.EXPECT().GetNumAttachedENIs().Return(0)
+				return m
+			},
+		},
+		{
 			name: "ReservedVolumeAttachments_specified",
 			options: &Options{
 				VolumeAttachLimit:         -1,
--- a/tests/e2e/metadata-labeler.go
+++ b/tests/e2e/metadata-labeler.go
@@ -171,7 +171,7 @@ var _ = framework.Describe("[ebs-csi-e2e] [Disruptive] Metadata Labeler Sidecar"
 
 // getAllocatableCount returns the limit of volumes that the node supports.
 func getAllocatableCount(volumes, enis int) int32 {
-	return int32(28 - volumes - enis)
+	return int32(27 - volumes - enis)
 }
 
 // getVolENIs gets the expected metadata of each instance from the ec2 API


torredil (approved):
/lgtm
/approve

ConnorJC3:
/cherrypick release-1.49 release-1.50 release-1.51 release-1.52

ConnorJC3:
/override pull-aws-ebs-csi-driver-test-e2e-external-eks-windows

Speed up cherrypicks - the test passed and we're just waiting on the cluster to delete

PR #2780, opened 2025-11-07
Added Verification Of Expected Parameters for e2e tests

mdzraf:
#### What type of PR is this?

/kind cleanup

This PR has two purposes, the first is to verify that the volumes modified in the modify volume e2e tests are checked to ensure that the desired attributes have been actually modified on the volume. The second is adding some simple create volume tests that ensure that volumes are created with the inputted user values. 

```
make update && make verify && make test

Manually (cd into tests/e2e):
> ginkgo run -focus="Volume Properties Verification" -procs=4 -timeout=10m -v

Ran 4 of 82 Specs in 37.025 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 78 Skipped

> ginkgo run -focus="will modify associated PV and EBS Volume via external-resizer" -procs=9 -timeout=10m -v

Ran 9 of 82 Specs in 93.208 seconds
SUCCESS! -- 9 Passed | 0 Failed | 0 Pending | 73 Skipped

```

```release-note
NONE
```

### DIFF
--- a/tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
+++ b/tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
@@ -1,0 +1,46 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the 'License');
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testsuites
+
+import (
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/e2e/driver"
+	. "github.com/onsi/ginkgo/v2"
+	v1 "k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// DynamicallyProvisionedVolumePropertiesTest will create a pod along with a volume.
+// It will then wait until the volume is created and verify input parameters were
+// properly applied to the volume.
+type DynamicallyProvisionedVolumePropertiesTest struct {
+	CreateVolumeParameters map[string]string
+	ClaimSize              string
+}
+
+func (t *DynamicallyProvisionedVolumePropertiesTest) Run(c clientset.Interface, ns *v1.Namespace, ebsDriver driver.PVTestDriver) {
+	volumeDetails := CreateVolumeDetails(t.CreateVolumeParameters, t.ClaimSize)
+	testVolume, _ := volumeDetails.SetupDynamicPersistentVolumeClaim(c, ns, ebsDriver)
+	defer testVolume.Cleanup()
+
+	pod := createPodWithVolume(c, ns, "", testVolume, volumeDetails)
+	defer pod.Cleanup()
+	pod.WaitForSuccess()
+
+	By("verifying volume properties")
+	volumeID := testVolume.persistentVolume.Spec.CSI.VolumeHandle
+
+	expected := BuildExpectedParameters(t.CreateVolumeParameters, t.ClaimSize)
+	VerifyVolumeProperties(volumeID, expected)
+}
--- a/tests/e2e/testsuites/e2e_utils.go
+++ b/tests/e2e/testsuites/e2e_utils.go
@@ -17,8 +17,12 @@ package testsuites
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -44,6 +48,7 @@ const (
 	VolumeType = "type"
 	TagSpec    = "tagSpecification"
 	TagDel     = "tagDeletion"
+	Encrypted  = "encrypted"
 )
 
 var DefaultGeneratedVolumeMount = VolumeMountDetails{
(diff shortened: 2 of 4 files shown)


tests/e2e/testsuites/e2e_utils.go
```diff
@@ -181,3 +183,62 @@ func PrefixAnnotations(prefix string, parameters map[string]string) map[string]s
 	}
 	return result
 }
+
+type ExpectedParameters struct {
+	Size       *int32
+	IOPS       *int32
+	Throughput *int32
+	VolumeType *string
+}
```
mdzraf, line 196: Anything else we should check? I had tags here but removed it, open to suggestions.

ElijahQuinones: We should add the encrypted bool

mdzraf: > We should add the encrypted bool

✅  Done, added a test case as well.

tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/e2e/driver"
+	. "github.com/onsi/ginkgo/v2"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// DynamicallyProvisionedVolumePropertiesTest will create a pod along with a volume.
+// It will then wait until the volume is created in order to verify that the volume
+// was created with parametes inputted at by the user.
```
ElijahQuinones: nit: typo `parametes` ->parameters

tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/tests/e2e/driver"
+	. "github.com/onsi/ginkgo/v2"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// DynamicallyProvisionedVolumePropertiesTest will create a pod along with a volume.
+// It will then wait until the volume is created in order to verify that the volume
```
ElijahQuinones: suggest rewording 

`It will then wait until the volume is created in order to verify that the volume was created with parametes inputted at by the user.`

to 

`It will then wait until the volume is created and verify input parameters where properly applied to the volume`

tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
```diff
@@ -0,0 +1,81 @@
...
+	pod := createPodWithVolume(c, ns, "", testVolume, volumeDetails)
+	pod.WaitForSuccess()
+
+	By("verifying volume properties before cleanup")
+	volumeID := testVolume.persistentVolume.Spec.CSI.VolumeHandle
+
+	verification := ExpectedParameters{}
+
+	quantity, _ := resource.ParseQuantity(t.ClaimSize)
+	sizeGiB := int32(quantity.Value() / (1024 * 1024 * 1024))
```
ElijahQuinones: What is this line doing ?

ElijahQuinones: Can we replace this to just use `BytesToGiB` from https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/util/util.go#L71 ? the util file in testsuites already uses the opposite especially since we need to do this in multiple places https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/tests/e2e/testsuites/e2e_utils.go#L73

tests/e2e/testsuites/modify_volume_tester.go
```diff
@@ -112,6 +113,38 @@ func (modifyVolumeTest *ModifyVolumeTest) Run(c clientset.Interface, ns *v1.Name
...
+	verifyVolumeBeforeCleanup(testVolume, modifyVolumeTest.ModifyVolumeParameters, modifyVolumeTest.ShouldResizeVolume, updatedPvcSize)
+}
+
+func verifyVolumeBeforeCleanup(testVolume *TestPersistentVolumeClaim, modifyParams map[string]string, resized bool, updatedSize resource.Quantity) {
+	volumeID := testVolume.persistentVolume.Spec.CSI.VolumeHandle
+
+	verification := ExpectedParameters{}
+
+	if resized {
+		expectedSizeGiB := int32(updatedSize.Value() / (1024 * 1024 * 1024))
```
ElijahQuinones: Same comment as other file can we replace this with using `BytesToGiB`

tests/e2e/volume_properties.go
```diff
@@ -0,0 +1,102 @@
+// Copyright 2025 The Kubernetes Authors.
```
ElijahQuinones, line 1: Why not just add this to `requires_aws_api.go` under dynamic provisioning  instead of making a whole new file

tests/e2e/volume_properties.go
```diff
@@ -0,0 +1,102 @@
...
+//
```
ElijahQuinones, line 2: Can we add a test that has all parameters set?

tests/e2e/volume_properties.go
mdzraf: I thought about this but I think because we in the future may add many more parameters that we want to verify, I think this should be a separate suite since it will eventually get quite large. Feel free to reopen thread if you feel strongly about having it in requires_aws_api.go and we can re-evaluate.

tests/e2e/volume_properties.go
mdzraf: ✅

tests/e2e/testsuites/modify_volume_tester.go
mdzraf: ✅

tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
mdzraf: ✅

tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
mdzraf: ✅

tests/e2e/testsuites/dynamically_provisioned_volume_properties_tester.go
mdzraf: ✅

tests/e2e/testsuites/e2e_utils.go
```diff
@@ -181,3 +184,75 @@ func PrefixAnnotations(prefix string, parameters map[string]string) map[string]s
...
+		iops = *volume.Iops
+	}
+	if volume.Throughput != nil {
+		throughput = *volume.Throughput
+	}
+	if volume.Encrypted != nil {
+		encrypted = *volume.Encrypted
+	}
+	framework.Logf("Volume %s verified: Size=%d GiB, IOPS=%d, Throughput=%d, Type=%s, Encrypted=%t",
+		volumeID, size, iops, throughput, volume.VolumeType, encrypted)
```
torredil: Suggest removing these ~20 lines of code, its not adding any value.

tests/e2e/testsuites/modify_volume_tester.go
```diff
@@ -112,6 +113,38 @@ func (modifyVolumeTest *ModifyVolumeTest) Run(c clientset.Interface, ns *v1.Name
...
+		verification.IOPS = &iops
+	}
+
+	if throughputStr, ok := modifyParams[Throughput]; ok {
+		expectedThroughput, _ := strconv.ParseInt(throughputStr, 10, 32)
+		throughput := int32(expectedThroughput)
+		verification.Throughput = &throughput
+	}
+
+	if volType, ok := modifyParams[VolumeType]; ok {
```
torredil: Instead of duplicating the parsing logic across the different test suites, can we have a single function in e2e_utils.go that does the needful? that way, we can centralize that logic and reuse it.

https://gist.github.com/torredil/a522d2dec3fc80072028446e49724114

tests/e2e/testsuites/e2e_utils.go
mdzraf: ✅

ConnorJC3:
Skimming through, lgtm other than @torredil's comments above. I'll do a (hopefully) final review after those are addressed.

tests/e2e/testsuites/modify_volume_tester.go
mdzraf: ✅

PR #2781, opened 2025-11-10
Fix Max IOPS Extraction From Error Bug

mdzraf:
#### What type of PR is this?

/kind bug

This PR fixes two bugs in how the driver extract IOPS limits during Create and Resize/Modify:

1.) If both the AZ and AZ ID are passed to the DryRun call, it will fail with InvalidParameterCombination and default to hardcoded limits. With this change we pass either the AZ or the AZ ID but not both.

2.) If the volume type is io2,  the error returned for having to many IOPS (that is parsed to get the maximum IOPS) can change depending on the specific AZ. This PR ensures that regardless of the volume type, we try to parse both known error messages before defaulting to hardcoded values.

```
make verify && make test

Manually: 
- Create io2 volume
- Modify IOPS on the volume and ensure that regardless of the AZ it is still able to get the maximum IOPS
```

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -2730,12 +2730,13 @@ func (c *cloud) getVolumeLimits(ctx context.Context, volumeType string, azParams
 			},
 		},
 	}
+	// We only want either the Zone or ZoneID otherwise the DryRun call will fail with InvalidParameterCombination due to having both.
 	if azParams.availabilityZone != "" {
 		dryRunRequestInput.AvailabilityZone = aws.String(azParams.availabilityZone)
-	}
-	if azParams.availabilityZoneId != "" {
+	} else if azParams.availabilityZoneId != "" {
 		dryRunRequestInput.AvailabilityZoneId = aws.String(azParams.availabilityZoneId)
 	}
+
 	if azParams.outpostArn != "" {
 		dryRunRequestInput.OutpostArn = aws.String(azParams.outpostArn)
 	}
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud.go
```diff
@@ -2796,23 +2797,19 @@ func extractMaxIOPSFromError(errorMsg string, volumeType string) (int32, error)
...
-		if matches := nonIo2ErrRegex.FindStringSubmatch(errorMsg); len(matches) > 1 {
-			if val, err := strconv.ParseInt(matches[1], 10, 32); err == nil {
-				return int32(val), nil
+	// io1 and gp3 have the same error message but io2 has different one depending on the availability zone.
+	if matches := nonIo2ErrRegex.FindStringSubmatch(errorMsg); len(matches) > 1 {
+		if val, err := strconv.ParseInt(matches[1], 10, 32); err == nil {
+			return int32(val), nil
+		}
+	} else if matches := io2ErrRegex.FindStringSubmatch(errorMsg); len(matches) > 1 {
+		if val, err := strconv.ParseInt(matches[1], 10, 32); err == nil {
```
torredil, line 2806: Suggest updating `TestExtractMaxIOPSFromError` in cloud_test.go to add a test for the fallback behavior.

mdzraf: Done ✅

pkg/cloud/cloud.go
```diff
@@ -2730,12 +2730,13 @@ func (c *cloud) getVolumeLimits(ctx context.Context, volumeType string, azParams
 			},
 		},
 	}
+	// We only want either the Zone or ZoneID otherwise the DryRun call will with InvalidParameterCombination due to having both.
```
ElijahQuinones: `We only want either the Zone or ZoneID otherwise the DryRun call will with InvalidParameterCombination due to having both.` ->`We only want either the Zone or ZoneID otherwise the DryRun call will fail with InvalidParameterCombination due to having both.`

Issue #2387, opened 2025-03-10
EBS CSI DRIVER Volume attachment limit check on AL2023 Nodes


PRYeswanthReddy:
Currently we are using AL2023 with EKS nodes of type R5.4xlarge, and the node capacity for volumes are 27 (excluding ENI). Eventhough the node reaches the maximum volume attachment the EBS CSI driver is trying to attach the volumes to the same node as the pods get scheduled to the same node.

Can we implement a logic at CSI driver level to check for node count and then if its reached maximum number then try to migrate the volume along with pod to a  different node? is it possible to implement?

EBS CSI VERSION : v1.36.0-eksbuild.1
EKS version: 1.30

ElijahQuinones:
Hi @PRYeswanthReddy,

This is unfortunately a common issue that we are actively working on. The `R5.4xlarge ` has a shared limit between it's volumes and any ENIs as you have said. We have logic in the driver today that sets the CSI NODE allocatables on node start up. however the CSI NODE allocatable property is currently immutable leading to the issue you described when we run into a state where the CSI Node allocatable property is out of sync with the current attachment limit of the node. This can happen for a number of reasons such as attaching a volume outside of the driver or extra ENIs being attached after node startup (among other things.) Please see [this section of our faq](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#volume-attachment-capacity-issues) for workarounds for this issue. Additionally our team is actively working on a KEP to make the CSI NODE allocatable property mutable to help alleviate this issue. [#4876](https://github.com/kubernetes/enhancements/issues/4876).

makarov-roman:
Hi @ElijahQuinones, did I understand correctly that it was implemented in 1.33? Should we expect that this behaviour will be improved in released EKS 1.33?

torredil:
Hi @makarov-roman, the [Mutable CSI Node Allocatable Count](https://kubernetes.io/blog/2025/05/02/kubernetes-1-33-mutable-csi-node-allocatable-count/) enhancement is introduced as an alpha feature in Kubernetes 1.33. However, EKS does not currently enable alpha features by default and doesn't allow feature gate configuration on master components, see https://github.com/aws/containers-roadmap/issues/512.

This feature will need to graduate to beta status in a future Kubernetes release before it becomes available in EKS. Once it reaches beta, it will be included in the corresponding EKS release (currently targeting 1.34).

[...]

PR #2784, opened 2025-11-11
fix(chart): only enable nodeAllocatableUpdatePeriodSeconds for EKS 1.34+

maxsxu:
#### What type of PR is this?

/kind bug

> The MutableCSINodeAllocatableCount feature gate is enabled by default in EKS 1.34

Per the AWS [documentation](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions-standard.html#kubernetes-1-34)

There're also some issues reported when deploying on EKS < 1.34:

- https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2387#issuecomment-3381664889

For EKS < 1.34

1. For a 1.33 EKS 
2. Upgrade aws-ebs-csi-driver helm chart to this PR version

For EKS >= 1.34

1. Create a new 1.34 EKS 
3. Deploy this aws-ebs-csi-driver helm chart

```release-note
N/A
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-ebs-csi-driver/templates/csidriver.yaml
@@ -8,7 +8,7 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: false
-  {{- if semverCompare ">=1.33.0-0" .Capabilities.KubeVersion.Version }}
+  {{- if semverCompare ">=1.34.0-0" .Capabilities.KubeVersion.Version }}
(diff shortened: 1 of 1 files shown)


ConnorJC3:
/close

As @torredil explains, this is intentional behavior and will be auto-dropped by the api server if the feature gate is disabled.

If you have a weird edge case scenario where you need to disable the feature, you can do so via the Helm parameter `nodeAllocatableUpdatePeriodSeconds` (set it to `0` or `null` to disable the feature).

PR #2795, opened 2025-11-20
Fix metadata labeler test flakes

torredil:
#### What type of PR is this?

/kind cleanup

Update metadata labeler unit tests to be deterministic by removing instances of time.Sleep() and other anti-patterns in the testing code that make it brittle / prone to flakes. See testgrid: https://testgrid.k8s.io/provider-aws-ebs-csi-driver#ci-unit-test.

```
go test -race -v ./pkg/cloud/metadata -run "^Test(GetMetadata|PatchSingleNode|PatchNodes|VolumeIDIndexFunc|GetNonCSIManagedVolumes)$" -count=100 -timeout=10m
```

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/metadata/labels_test.go
+++ b/pkg/cloud/metadata/labels_test.go
@@ -17,20 +17,20 @@ package metadata
 import (
 	"context"
 	"errors"
-	"fmt"
-	reflect "reflect"
+	"maps"
+	"slices"
 	"strconv"
-	"sync"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
-	gomock "github.com/golang/mock/gomock"
+	"github.com/golang/mock/gomock"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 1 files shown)


pkg/cloud/metadata/labels_test.go
```diff
@@ -17,483 +17,439 @@ package metadata
...
-	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
-	clienttesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
 )
 
 func init() {
-	// Ensure variables are initialized
-	// TODO: Figure out a cleaner way to do this in tests
```
ConnorJC3, line 42: Don't drop this comment (possible rebase mistake?).

pkg/cloud/metadata/labels_test.go
```diff
@@ -17,483 +17,439 @@ package metadata
...
 	initVariables()
 }
 
-func TestPatchNewNodes(t *testing.T) {
```
ConnorJC3, line 46: This PR removes `TestPatchNewNodes` but doesn't replace it or add any equivalent tests, which loses all of our test coverage for `updateMetadataEC2` and `patchNewNodes` (which were mostly covered before), is that an oversight?

pkg/cloud/metadata/labels_test.go
```diff
@@ -17,483 +17,439 @@ package metadata
...
+		t.Fatalf("Failed to add indexer: %v", err)
+	}
+	stopCh := make(chan struct{})
+	t.Cleanup(func() { close(stopCh) })
+	factory.Start(stopCh)
+	cache.WaitForCacheSync(stopCh, pvInformer.HasSynced)
+	return pvInformer
+}
+
+func equalEnisVolumesMap(a, b map[string]enisVolumes) bool {
```
ConnorJC3: This function can be replaced by golang's builtin `maps.Equal`: https://pkg.go.dev/maps#Equal

PR #2797, opened 2025-11-25
Adding SnapshotLock Capabilities

mdzraf:
#### What type of PR is this?

/kind feature

This PR adds the ability for users to [lock snapshots](https://docs.aws.amazon.com/ebs/latest/userguide/snapshot-lock-concepts.html) on creation by passing in desired [LockSnapshot parameters](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LockSnapshot.html) through the VolumeSnapshotClass. 

Ran various unit tests as well as e2e tests that: 
- Created locked snapshots in governance mode
- Create locked snapshots in compliance mode
- Create locked snapshots with FSR 

Also manually tested by running a modified version of our snapshot example.
- In the `examples/kubernetes/snapshot` Change `snapshotclass.yaml` to:
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: csi-aws-vsc
driver: ebs.csi.aws.com
deletionPolicy: Delete
parameters: 
  snapshotLockMode: "governance"
  snapshotLockDuration: "1"
```
- Run through example, and just make sure to unlock snapshot through AWS CLI before doing cleanup by running: 
```
aws ec2 unlock-snapshot --snapshot-id  snap-<snapshot-id>
```

```release-note
Added VolumeSnapshotClass parameters to lock EBS snapshots after creation. 
```

### DIFF
--- a/docs/snapshot.md
+++ b/docs/snapshot.md
@@ -2,7 +2,11 @@
 | Parameter                          | Description of value                                      |
 |------------------------------------|-----------------------------------------------------------|
 | fastSnapshotRestoreAvailabilityZones | Comma separated list of availability zones                |
-| outpostArn                         | Arn of the outpost you wish to have the snapshot saved to | 
+| outpostArn                         | Arn of the outpost you wish to have the snapshot saved to |
+| lockMode                   | Lock mode (governance/compliance)                         |
+| lockDuration               | Lock duration in days                                     |
+| lockExpirationDate         | Lock expiration date (RFC3339 format)                    |
+| lockCoolOffPeriod          | Cool-off period in hours (compliance mode only)          | 
 
 The AWS EBS CSI Driver supports [tagging](tagging.md) through `VolumeSnapshotClass.parameters` (in v1.6.0 and later). 
 ## Prerequisites
@@ -43,6 +47,41 @@ parameters:
 ## Failure Mode 
 
 The driver will attempt to check if the availability zones provided are supported for fast snapshot restore before attempting to create the snapshot. If the `EnableFastSnapshotRestores` API call fails, the driver will hard-fail the request and delete the snapshot. This is to ensure that the snapshot is not left in an inconsistent state.
+
+# Snapshot Lock
+
+The EBS CSI Driver supports [EBS Snapshot Lock](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-snapshot-lock.html) via `VolumeSnapshotClass.parameters`. Snapshot locking protects snapshots from accidental or malicious deletion. A locked snapshot can't be deleted.
+
+**Example - Lock in Governance Mode with Specified Duration**
+```yaml
+apiVersion: snapshot.storage.k8s.io/v1
+kind: VolumeSnapshotClass
+metadata:
+  name: csi-aws-vsc-locked
+driver: ebs.csi.aws.com
+deletionPolicy: Delete
+parameters:
+  lockMode: "governance"
+  lockDuration: "7"
+```
+
+**Example - Lock in Compliance Mode with Expiration Date and Cool Off Period**
+```yaml
+apiVersion: snapshot.storage.k8s.io/v1
+kind: VolumeSnapshotClass
+metadata:
+  name: csi-aws-vsc-compliance
+driver: ebs.csi.aws.com
+deletionPolicy: Delete
+parameters:
+  lockMode: "compliance"
+  lockExpirationDate: "2030-12-31T23:59:59Z"
+  lockCoolOffPeriod: "24"
+```
+
+## Failure Mode
+
+If the `LockSnapshot` API call fails, the driver will hard-fail the request and delete the snapshot. This ensures that the snapshot is not left in an unlocked state when locking was explicitly requested.
 
 
 # Amazon EBS Local Snapshots on Outposts
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -313,10 +313,19 @@ type ListSnapshotsResponse struct {
 	NextToken string
 }
 
-// SnapshotOptions represents parameters to create an EBS volume.
+// SnapshotOptions represents parameters to create an EBS snapshot.
 type SnapshotOptions struct {
 	Tags       map[string]string
 	OutpostArn string
+}
+
+// SnapshotLockOptions represents parameters to lock an EBS snapshot.
+type SnapshotLockOptions struct {
+	SnapshotId     *string
+	LockMode       types.LockMode
+	CoolOffPeriod  *int32
+	ExpirationDate *time.Time
+	LockDuration   *int32
 }
 
 // ec2ListSnapshotsResponse is a helper struct returned from the AWS API calling function to the main ListSnapshots function.
@@ -1959,6 +1968,21 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
 		CreationTime:   aws.ToTime(res.StartTime),
 		ReadyToUse:     res.State == types.SnapshotStateCompleted,
 	}, nil
+}
+
+func (c *cloud) LockSnapshot(ctx context.Context, lockOptions *SnapshotLockOptions) error {
+	lockSnapshotInput := ec2.LockSnapshotInput{
+		SnapshotId:     lockOptions.SnapshotId,
+		LockMode:       lockOptions.LockMode,
+		CoolOffPeriod:  lockOptions.CoolOffPeriod,
+		ExpirationDate: lockOptions.ExpirationDate,
+		LockDuration:   lockOptions.LockDuration,
+	}
+	_, err := c.ec2.LockSnapshot(ctx, &lockSnapshotInput)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 func (c *cloud) DeleteSnapshot(ctx context.Context, snapshotID string) (success bool, err error) {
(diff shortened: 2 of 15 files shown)


mdzraf:
/hold 

Waiting on managed policy change.

pkg/cloud/cloud.go
```diff
@@ -306,12 +306,18 @@ type ListSnapshotsResponse struct {
...
 type SnapshotOptions struct {
 	Tags       map[string]string
 	OutpostArn string
 }
 
+// SnapshotLockOptions represents the snapshot lock specific parameters for locking en EBS snapshot.
+type SnapshotLockOptions struct {
+	SnapshotLockEnabled bool
+	LockSnapshotInput   ec2.LockSnapshotInput
+}
```
torredil: I suggest removing the `SnapshotLockEnabled` field from this struct, its redundant and creates unnecessary complexity for the end user (ie confusing state where someone could set `snapshotLockMode: "governance"` but forget `snapshotLockEnabled: "true"`, and the lock silently won't happen).

If a user specifies `snapshotLockMode` or `snapshotLockDuration`, the intent is clearly to lock the snapshot, so we should infer enablement from this. Its also more consistent with how enabling FSR works - users just provide the zones, there isn't an additional knob to specify enablement status.

mdzraf: Spoke offline, decided that we will remove the `SnapshotLockEnabled` field and instead check if ANY snapshot lock related parameters are set by the end user even if they did not input required parameters (Ie they specify `SnapshotLockCoolOffPeriod` and `SnapshotLockDuration` but not the required `SnapshotLockMode`) we make the `LockSnapshot` call and let the API return the error.

torredil (approved):
/lgtm

docs/snapshot.md
```diff
@@ -44,6 +48,49 @@ parameters:
 
 The driver will attempt to check if the availability zones provided are supported for fast snapshot restore before attempting to create the snapshot. If the `EnableFastSnapshotRestores` API call fails, the driver will hard-fail the request and delete the snapshot. This is to ensure that the snapshot is not left in an inconsistent state.
 
+# Snapshot Lock
+
+The EBS CSI Driver provides support for [EBS Snapshot Lock](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-snapshot-lock.html) via `VolumeSnapshotClass.parameters`. Snapshot locking protects snapshots from accidental or malicious deletion. A locked snapshot can't be deleted.
+
+Snapshot locking supports two modes:
```
ElijahQuinones: I would remove this section as this is AWS specific documentation and is included in the link you have above. We should only document how the EBS feature and the EBS-CSI-Driver interact with one another not document the feature itself. You could however document how to set the mode in the driver here.

mdzraf: ✅ Done

examples/kubernetes/snapshot/manifests/classes/snapshotclass.yaml
```diff
@@ -17,4 +17,4 @@ kind: VolumeSnapshotClass
 metadata:
   name: csi-aws-vsc
 driver: ebs.csi.aws.com
-deletionPolicy: Delete
+deletionPolicy: Delete
```
ConnorJC3: Drop this diff

pkg/cloud/cloud.go
```diff
@@ -309,12 +309,17 @@ type ListSnapshotsResponse struct {
...
 
-// SnapshotOptions represents parameters to create an EBS volume.
+// SnapshotOptions represents parameters to create an EBS snapshot.
 type SnapshotOptions struct {
 	Tags       map[string]string
 	OutpostArn string
 }
 
+// SnapshotLockOptions represents the snapshot lock specific parameters for locking en EBS snapshot.
+type SnapshotLockOptions struct {
```
ConnorJC3: Why is there a separate struct for this? Why not add to `SnapshotOptions`?

pkg/cloud/cloud.go
```diff
@@ -309,12 +309,17 @@ type ListSnapshotsResponse struct {
...
+	LockSnapshotInput ec2.LockSnapshotInput
```
ConnorJC3: `ec2` objects shouldn't leak outside this package, make this `string`s or the other relevant fields instead

pkg/driver/constants.go
```diff
@@ -126,6 +126,18 @@ const (
 const (
 	// FastSnapshotRestoreAvailabilityZones represents key for fast snapshot restore availability zones.
 	FastSnapshotRestoreAvailabilityZones = "fastsnapshotrestoreavailabilityzones"
+
+	// SnapshotLockMode represents a key for indicating whether snapshots are locked in Governance or Compliance mode.
+	SnapshotLockMode = "snapshotlockmode"
```
ConnorJC3: We should remove `snapshot` from all these flags? e.g. `snapshotlockmode` -> `lockmode` and `snapshotlockduration` -> `lockduration`

These fields can only be present in a `VolumeSnaphotClass`, so they are automatically and always about snapshots.

pkg/driver/controller.go
```diff
@@ -874,6 +878,26 @@ func (d *ControllerService) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		case SnapshotLockMode:
+			vsLock.LockSnapshotInput.LockMode = types.LockMode(value)
+		case SnapshotLockDuration:
+			lockDuration, err := strconv.ParseInt(value, 10, 32)
+			if err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Could not parse SnapshotLockDuration: %q", value)
+			}
+			vsLock.LockSnapshotInput.LockDuration = aws.Int32(int32(lockDuration))
+		case SnapshotLockExpirationDate:
+			expirationDate, err := time.Parse(time.RFC3339, value)
```
ConnorJC3, line 899: Why `RFC3339`? Doesn't the AWS API use 8601? https://docs.aws.amazon.com/AWSEC2/latest/APIReference/CommonParameters.html#:~:text=For%20more%20information%2C%20see%20ISO%208601

pkg/cloud/cloud.go
```diff
@@ -1872,6 +1877,15 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
 	}, nil
 }
 
+func (c *cloud) LockSnapshot(ctx context.Context, lockSnapshotInput ec2.LockSnapshotInput) (*ec2.LockSnapshotOutput, error) {
+	klog.InfoS("Attempting to lock Snapshot", "request parameters: ", lockSnapshotInput)
```
ConnorJC3: Debug log left in?

pkg/cloud/cloud.go
```diff
@@ -1872,6 +1877,15 @@ func (c *cloud) CreateSnapshot(ctx context.Context, volumeID string, snapshotOpt
 	}, nil
 }
 
+func (c *cloud) LockSnapshot(ctx context.Context, lockSnapshotInput ec2.LockSnapshotInput) (*ec2.LockSnapshotOutput, error) {
```
ConnorJC3: Why does this function return the output? What is it used for? Shouldn't it just return an error?

tests/e2e/requires_aws_api.go
```diff
@@ -351,6 +352,355 @@ var _ = Describe("[ebs-csi-e2e] [single-az] [requires-aws-api] Dynamic Provision
...
+					Fail(fmt.Sprintf("expected lock mode governance, got %s", lockedSnapshot.LockState))
+				}
+
+				if lockedSnapshot.LockDuration == nil || *lockedSnapshot.LockDuration != 1 {
+					Fail(fmt.Sprintf("expected lock duration 1 day, got %v", lockedSnapshot.LockDuration))
+				}
+
+				_, err = ec2Client.UnlockSnapshot(context.Background(), &ec2.UnlockSnapshotInput{
+					SnapshotId: aws.String(snapshotId),
+				})
```
ConnorJC3: Shouldn't this be in some kind of cleanup function so it's always called? We don't want to leak locked snapshots on the CI account.

docs/snapshot.md
```diff
@@ -44,6 +48,41 @@ parameters:
 
 The driver will attempt to check if the availability zones provided are supported for fast snapshot restore before attempting to create the snapshot. If the `EnableFastSnapshotRestores` API call fails, the driver will hard-fail the request and delete the snapshot. This is to ensure that the snapshot is not left in an inconsistent state.
 
+# Snapshot Lock
+
+The EBS CSI Driver provides support for [EBS Snapshot Lock](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-snapshot-lock.html) via `VolumeSnapshotClass.parameters`. Snapshot locking protects snapshots from accidental or malicious deletion. A locked snapshot can't be deleted.
```
ConnorJC3: passive voice: `provides support for` -> `supports`

pkg/driver/controller.go
mdzraf: RFC3339 is based on ISO 8601, it parses the YYYY‑MM‑DDThh:mm:ssZ format that AWS expects

pkg/driver/constants.go
mdzraf: Done ✅

pkg/cloud/cloud.go

tests/e2e/requires_aws_api.go
mdzraf: Done ✅

mdzraf: Done ✅

pkg/cloud/cloud.go
mdzraf: Should only return error, fixed ✅

pkg/driver/controller.go
ConnorJC3: Ah, I forgot about this jank - they're the same thing basically. Golang dosen't have an explicit ISO8601 preset so marking as resolved.

mdzraf:
/unhold

Managed Policy has been updated to have LockSnapshot permissions.

PR #2800, opened 2025-12-01
Pre-create report dir in config.sh

ConnorJC3:
#### What type of PR is this?

/kind cleanup

We're seeing failures in https://github.com/aws/csi-components/pull/23 because `REPORT_DIR` doesn't exist when it tries to dump the pod logs. I'm not exactly sure why this doesn't happen on the "real" CI, I suspect it has something to do with whether `ARTIFACTS` is set or not.

Regardless, always pre-create this directory so we know it's available.

CI

```release-note
NONE
```


ConnorJC3:
/cherrypick release-1.53

Cherrypicking after this is merged so that the `csi-components` PR can take advantage of it.

ConnorJC3:
/retest

flake

PR #2801, opened 2025-12-01
Fix a1 e2e tests

ElijahQuinones:
#### What type of PR is this?

/kind cleanup

This PR fixes up the a1 tests such that they can be run in CI again
Ran make e2e/external-a1 on an a1 cluster

```
Ran 106 of 7450 Specs in 348.622 seconds
SUCCESS! -- 106 Passed | 0 Failed | 0 Pending | 7344 Skipped


Ginkgo ran 1 suite in 5m51.955206757s
Test Suite Passed
+ TEST_PASSED=0
+ set -e
+ set +x
```

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,8 +159,8 @@ e2e/external: bin/helm bin/kubetest2
 	COLLECT_METRICS="true" \
 	./hack/e2e/run.sh
 
-.PHONY: e2e/external-a1-eks
-e2e/external-a1-eks: bin/helm bin/kubetest2
+.PHONY: e2e/external-a1
+e2e/external-a1: bin/helm bin/kubetest2
 	HELM_EXTRA_FLAGS="--set=a1CompatibilityDaemonSet=true" \
 	./hack/e2e/run.sh
 
--- a/hack/prow-e2e.sh
+++ b/hack/prow-e2e.sh
@@ -49,13 +49,13 @@ test-e2e-external-eks)
   TEST="external"
   export CLUSTER_TYPE="eksctl"
   ;;
-test-e2e-external-a1-eks)
-  TEST="external-a1-eks"
-  export K8S_VERSION_EKSCTL="1.30"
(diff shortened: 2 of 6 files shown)


hack/prow-e2e.sh
```diff
@@ -49,13 +49,13 @@ test-e2e-external-eks)
...
   export CLUSTER_TYPE="eksctl"
   ;;
-test-e2e-external-a1-eks)
-  TEST="external-a1-eks"
-  export K8S_VERSION_EKSCTL="1.30"
-  export INSTANCE_TYPE="a1.large"
+test-e2e-external-a1)
+  TEST="external-a1"
+  export AWS_AVAILABILITY_ZONES="us-west-2b,us-west-2c"
+  export INSTANCE_TYPE="a1.xlarge"
```
torredil, line 55: Do we also need to update build-image.sh?

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/18aa7b714c3d1521a657e076344f95e22677704c/hack/e2e/build-image.sh#L73

ElijahQuinones: Good catch yes we do

Makefile
```diff
@@ -159,8 +159,8 @@ e2e/external: bin/helm bin/kubetest2
 	COLLECT_METRICS="true" \
 	./hack/e2e/run.sh
 
-.PHONY: e2e/external-a1-eks
-e2e/external-a1-eks: bin/helm bin/kubetest2
+.PHONY: e2e/external-a1
+e2e/external-a1: bin/helm bin/kubetest2
```
ConnorJC3, line 163: This should also depend on `bin/gomplate` (Actually, all the kops targets now should I think)

ConnorJC3: We talked about this in person, it's not used during testing only during cluster create, so this isn't necessary

ConnorJC3 (approved):
/lgtm

PR #2803, opened 2025-12-02
Adjust error codes

ElijahQuinones:
#### What type of PR is this?

/kind bug

This PR is to address what error codes we return on createVolume failures when we are unsure if a volume could have been created already.

New change not applied 

1. Created volume with describe volumes removed from managed policy
2. deleted pvc 
3. volume leaks 
4. fixed managed policy
5. external provisioner never retried as it did not receive aborted error code

After change with aborted error code is applied.

1. Created volume with describe volumes removed from managed policy
2. deleted pvc 
3. volume leaks 
4. fixed managed policy
5. external provisioner retries 
6. pv is created 
7. pv deleted 
8. leaked volume is deleted 

```release-note
Adjust errors codes on create path to fix rare volume leak edge case
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -168,6 +168,9 @@ var (
 	// name, but different size, is found.
 	ErrDiskExistsDiffSize = errors.New("there is already a disk with same name and different size")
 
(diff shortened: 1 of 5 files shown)


torredil (approved):
/lgtm

PR #2807, opened 2025-12-11
[DO NOT MERGE] Testing Minimal Policy

mdzraf:
#### What type of PR is this?

Testing CI with Minimal Policy & review new proposed minimal policy.

Manual Tests Completed: 
```
* Create Volume (With Tags)
* Modify Volume (Changed volume type, iops, and deleted tag)
* Attach Volume
* Detach Volume
* Delete Volume 
* Resize Volume
* Create Snapshot From Volume (snapshot with tags, and FSR enabled)
* Restore Volume From Snapshot
* Delete snapshot
* Copy Volume (with tags)
* Create Snapshot with Lock
* Cannot Create/consume PV from pre-existing volume
* Can create/consume PV from pre-existing volume that had tags manually added
```

### DIFF
--- a/docs/example-iam-policy.json
+++ b/docs/example-iam-policy.json
@@ -1,10 +1,10 @@
 {
-  "Version" : "2012-10-17",
-  "Statement" : [
+  "Version": "2012-10-17",
+  "Statement": [
     {
-      "Sid":"ReadOnlyDescribeOperations",
-      "Effect" : "Allow",
-      "Action" : [
+      "Sid": "ReadOnlyDescribeOperations",
+      "Effect": "Allow",
+      "Action": [
         "ec2:DescribeAvailabilityZones",
         "ec2:DescribeInstances",
         "ec2:DescribeSnapshots",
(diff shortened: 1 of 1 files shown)


docs/example-iam-policy.json
```diff
@@ -1,67 +1,89 @@
...
+  "Statement" : [
     {
-      "Effect": "Allow",
-      "Action": [
+      "Effect" : "Allow",
+      "Action" : [
         "ec2:DescribeAvailabilityZones",
         "ec2:DescribeInstances",
         "ec2:DescribeSnapshots",
-        "ec2:DescribeTags",
```
mdzraf, line 10: We do not actually call this anywhere in the driver besides in a testing function that does not actually need it: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/33be568c96afd3729698531f6a0d3fe9f5b58868/pkg/cloud/cloud_test.go#L3666

So I have removed it from the policy.

docs/example-iam-policy.json
```diff
@@ -1,67 +1,89 @@
...
         "ec2:DescribeVolumes",
         "ec2:DescribeVolumesModifications",
         "ec2:DescribeVolumeStatus"
```
mdzraf, line 13: These calls have `*` permissions  because they only have `ec2:Region` as an allowed condition key, therefore we cannot pass in any condition key that would clearly identify a resource managed by the driver (I.e. tags on resource) see: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html

hack/e2e/eksctl/cluster.yaml
mdzraf, line 1: Only made changes here to test CI, feel free to skip review.

hack/e2e/kops/patch-cluster.yaml
mdzraf, line 1: Only made changes here to test CI, feel free to skip review.

docs/example-iam-policy.json
```diff
@@ -1,67 +1,89 @@
...
     {
-      "Effect": "Allow",
-      "Action": [
+      "Effect" : "Allow",
+      "Action" : [
+        "ec2:AttachVolume",
+        "ec2:DetachVolume"
+      ],
+      "Resource" : "arn:aws:ec2:*:*:instance/*"
+    },
```
mdzraf: Necessary for the EBS CSI Driver to perform the privileged task of attaching/detaching volumes on a host on behalf of the user.

mdzraf:
/retest 

Trying to see if it was a flake, the pre-provisioned tests should all pass since we actually do add the required tags.

docs/example-iam-policy.json
```diff
@@ -1,67 +1,89 @@
 {
```
ElijahQuinones, line 1: As this policy is scoped down now to only use volumes with the appropriate tags I think it would now be appropriate to add the ability to modify tags such that customers do not have to go out of their way to change their policy to use this feature.

mdzraf: That's a great suggestion, will add to next rev, thanks!

docs/example-iam-policy.json
```diff
@@ -70,122 +92,86 @@
...
+        "ec2:CreateVolume",
+        "ec2:CopyVolumes"
       ],
-      "Resource": "arn:aws:ec2:*:*:volume/*",
-      "Condition": {
-        "StringLike": {
-          "ec2:ResourceTag/CSIVolumeName": "*"
+      "Resource" : "arn:aws:ec2:*:*:volume/*",
+      "Condition" : {
+        "ForAnyValue:StringLike" : {
```
torredil: Per [docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-single-vs-multi-valued-context-keys.html#reference_policies_condition-key-points), remove the use of `ForAnyValue` throughout this policy with single valued context keys.

> Do not use condition set operators ForAllValues or ForAnyValue with single-valued context keys. Using condition set operators with single-valued context keys can lead to overly permissive policies.

docs/example-iam-policy.json
```diff
@@ -70,122 +92,86 @@
...
-      "Resource": "arn:aws:ec2:*:*:snapshot/*",
-      "Condition": {
-        "StringLike": {
-          "aws:RequestTag/CSIVolumeSnapshotName": "*"
+      "Resource" : "arn:aws:ec2:*:*:volume/*",
+      "Condition" : {
+        "ForAnyValue:StringLike" : {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster" : "true",
+          "ec2:ResourceTag/CSIVolumeName" : "*",
+          "ec2:ResourceTag/kubernetes.io/created-for/pvc/name" : "*"
```
torredil: Is the intent here to require all three tags for this action? If so, are we ultra sure that volumes will always have all 3 tags? that would be a change from the existing behavior, the current policy allows for volume deletion as long as ANY of the tags are present, it doesn't require all 3.

I was reading the docs and my understanding is that [ForAnyValue](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-single-vs-multi-valued-context-keys.html#reference_policies_condition-multi-valued-context-keys) on single value keys is essentially a no-op and evaluates the same as without it. In other words, this:

```
"Condition" : {
     "ForAnyValue:StringLike" : {
          "ec2:ResourceTag/ebs.csi.aws.com/cluster" : "true",
          "ec2:ResourceTag/CSIVolumeName" : "*",
          "ec2:ResourceTag/kubernetes.io/created-for/pvc/name" : "*"
```

 behaves identically to this:


```
"Condition" : {
     "StringLike" : {
          "ec2:ResourceTag/ebs.csi.aws.com/cluster" : "true",
          "ec2:ResourceTag/CSIVolumeName" : "*",
          "ec2:ResourceTag/kubernetes.io/created-for/pvc/name" : "*"
```

This also explains why single az `should use a pre-provisioned volume and delete PV with reclaimPolicy "Delete" ` is failing. The pre-provisioned vol only has the following tags: 

```
tags='{CSIVolumeName=pre-provisioned, ebs.csi.aws.com/cluster=true}'
```

its missing `ec2:ResourceTag/kubernetes.io/created-for/pvc/name`.

mdzraf: > Is the intent here to require all three tags for this action? 

No, I just noticed while I was debugging the test that `ForAnyValue:StringLike` does not actually evaluate it with OR logic for each statement. So what we actually want to do here is have separate statements for each to make sure that even if it has just one, we still allow deletion. This is also true for a few other statements in the policy so I will fix for next rev.

torredil: Okay cool, thanks!

docs/example-iam-policy.json
```diff
@@ -1,191 +1,232 @@
...
+        "ec2:DetachVolume"
       ],
-      "Resource": "arn:aws:ec2:*:*:volume/*",
-      "Condition": {
-        "StringLike": {
-          "aws:RequestTag/CSIVolumeName": "*"
+      "Resource" : "arn:aws:ec2:*:*:volume/*",
+      "Condition" : {
+        "StringLike" : {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster" : "true"
```
torredil: The driver should also be able to manage volumes previously created by the intree provisioner that only have the `kubernetes.io/created-for/pvc/name` tag. This is handled correctly in the `DeleteVolume` statements, but also needs to be accounted for in attach, detach, and modify.

As an example, consider the scenario where a user is modifying volumes that only have the `kubernetes.io/created-for/pvc/name` tag via the [volume-modifier-for-k8s](https://github.com/awslabs/volume-modifier-for-k8s) custom sidecar. That workflow works with the existing managed policy, and should also work with this new minimal one.

mdzraf: Nice catch, will fix in next rev ✅

mdzraf: Will also add for CopyVolumes

ElijahQuinones (approved):
/lgtm

PR #2817, opened 2025-12-16
Cache known bad ids to remove from batch calls

ElijahQuinones:
#### What type of PR is this?

/kind feature

This PR improves our batching process by adding a cache to put known bad IDs into for Volumes,snapshots and instances. When a request comes in with a known bad id we trim it from the batch request and try it separately from the batch. This ensures one bad ID does not constantly poison the whole batch.
### Volume Example

Statically provisioned a fake volume 
Modified the volume

Without Change

```
Error on first call in batcher

E0106 20:46:06.268871       1 handlers.go:86] "Error from AWS API" err="api error InvalidVolume.NotFound: The volume 'vol-05ae9e63bbfb1f23a' does not exist."                                                      │
│ E0106 20:46:06.269018       1 batcher.go:161] "execute: error executing batch" err="operation error EC2: DescribeVolumes, https response error StatusCode: 400, RequestID: 97bded0d-b90f-486b-8e9f-97d6fd70781d, a │
│ E0106 20:46:06.269108       1 driver.go:133] "GRPC error" err="rpc error: code = Internal desc = Could not modify volume \"vol-05ae9e63bbfb1f23a\": operation error EC2: DescribeVolumes, https response error Sta │
│ I0106 20:46:07.288117       1 controller.go:730] "ControllerModifyVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\" mutable_parameters:{key:\"csi.storage.k8s.io/pv/name\" value:\"test-pv\"} mutable_para │
│ I0106 20:46:09.288933       1 cloud.go:970] "Received Modify Disk request" volumeID="vol-05ae9e63bbfb1f23a" options={"VolumeType":"gp3","IOPS":4000,"Throughput":250,"IOPSPerGB":0,"AllowIopsIncreaseOnResize":fal │
│ E0106 20:46:09.877350       1 handlers.go:86] "Error from AWS API" err="api error InvalidVolume.NotFound: The volume 'vol-05ae9e63bbfb1f23a' does not exist."      

subsequent retries are all in the batcher posioning good requests in same batch
 
 │
│ E0106 20:46:09.877419       1 batcher.go:161] "execute: error executing batch" err="operation error EC2: DescribeVolumes, https response error StatusCode: 400, RequestID: f3b919ba-990d-490d-a570-a22ed91f075d, a │
│ E0106 20:46:09.877491       1 driver.go:133] "GRPC error" err="rpc error: code = Internal desc = Could not modify volume \"vol-05ae9e63bbfb1f23a\": operation error EC2: DescribeVolumes, https response error Sta │
│ I0106 20:46:11.895847       1 controller.go:730] "ControllerModifyVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\" mutable_parameters:{key:\"csi.storage.k8s.io/pv/name\" value:\"test-pv\"} mutable_para │
│ I0106 20:46:13.896998       1 cloud.go:970] "Received Modify Disk request" volumeID="vol-05ae9e63bbfb1f23a" options={"VolumeType":"gp3","IOPS":4000,"Throughput":250,"IOPSPerGB":0,"AllowIopsIncreaseOnResize":fal
```
With change

```
 │ I0108 19:57:34.169876       1 controller.go:686] "ControllerExpandVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\"  capacity_range:{required_bytes:21474836480}  volume_capability:{mount:{}  access_mode │
│ I0108 19:57:35.183239       1 controller.go:728] "ControllerModifyVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\"  mutable_parameters:{key:\"csi.storage.k8s.io/pv/name\"  value:\"ebs-pv\"}  mutable_pa │
│ I0108 19:57:36.171154       1 cloud.go:1020] "Received Resize and/or Modify Disk request" volumeID="vol-05ae9e63bbfb1f23a" newSizeBytes=21474836480 options={"VolumeType":"gp3","IOPS":4000,"Throughput":200,"IOPS │
│ E0108 19:57:36.798077       1 handlers.go:86] "Error from AWS API" err="api error InvalidVolume.NotFound: The volume 'vol-05ae9e63bbfb1f23a' does not exist."                                                      │
│ E0108 19:57:36.798191       1 batcher.go:161] "execute: error executing batch" err="operation error EC2: DescribeVolumes, https response error StatusCode: 400, RequestID: fd0d2b23-eb5f-47b4-89ce-6db798d5ae7e, a │
│ E0108 19:57:36.798315       1 driver.go:133] "GRPC error" err="rpc error: code = Internal desc = Could not modify volume \"vol-05ae9e63bbfb1f23a\": operation error EC2: DescribeVolumes, https response error Sta │
│ E0108 19:57:36.798376       1 driver.go:133] "GRPC error" err="rpc error: code = Internal desc = Could not resize volume \"vol-05ae9e63bbfb1f23a\": rpc error: code = Internal desc = Could not modify volume \"vo │
│ I0108 19:57:36.829929       1 controller.go:616] "ControllerGetCapabilities: called" args=""                                                                                                                       │
│ I0108 19:57:36.830555       1 controller.go:686] "ControllerExpandVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\"  capacity_range:{required_bytes:21474836480}  volume_capability:{mount:{}  access_mode │
│ I0108 19:57:38.826895       1 controller.go:728] "ControllerModifyVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\"  mutable_parameters:{key:\"csi.storage.k8s.io/pv/name\"  value:\"ebs-pv\"}  mutable_pa │
│ I0108 19:57:38.831021       1 cloud.go:1020] "Received Resize and/or Modify Disk request" volumeID="vol-05ae9e63bbfb1f23a" newSizeBytes=21474836480 options={"VolumeType":"gp3","IOPS":4000,"Throughput":200,"IOPS │
│ E0108 19:57:39.428054       1 handlers.go:86] "Error from AWS API" err="api error InvalidVolume.NotFound: The volume 'vol-05ae9e63bbfb1f23a' does not exist."                                                      │
│ E0108 19:57:39.428159       1 driver.go:133] "GRPC error" err="rpc error: code = NotFound desc = Could not modify volume (not found) \"vol-05ae9e63bbfb1f23a\": resource was not found"                            │
│ E0108 19:57:39.428270       1 driver.go:133] "GRPC error" err="rpc error: code = Internal desc = Could not resize volume \"vol-05ae9e63bbfb1f23a\": rpc error: code = NotFound desc = Could not modify volume (not
```

### Snapshots example

Create a Snapshot with a fake snapshot id

````
I0108 20:02:34.498164       1 controller.go:996] "ListSnapshots: called" args="snapshot_id:\"snap-0a1b033b1430b3b05\""                                                                                             │
│ E0108 20:02:35.061317       1 handlers.go:86] "Error from AWS API" err="api error InvalidSnapshot.NotFound: The snapshot 'snap-0a1b033b1430b3b05' does not exist."                                                 │
│ E0108 20:02:35.061504       1 batcher.go:161] "execute: error executing batch" err="operation error EC2: DescribeSnapshots, https response error StatusCode: 400, RequestID: 7f593571-5d9b-4d33-a0d6-911b3a0d47d6, │
│ E0108 20:02:35.061565       1 driver.go:133] "GRPC error" err="rpc error: code = Internal desc = Could not get snapshot ID \"snap-0a1b033b1430b3b05\": operation error EC2: DescribeSnapshots, https response erro │
│ I0108 20:02:35.072913       1 controller.go:616] "ControllerGetCapabilities: called" args=""                                                                                                                       │
│ I0108 20:02:35.073430       1 controller.go:996] "ListSnapshots: called" args="snapshot_id:\"snap-0a1b033b1430b3b05\""                                                                                             │
│ E0108 20:02:35.678695       1 handlers.go:86] "Error from AWS API" err="api error InvalidSnapshot.NotFound: The snapshot 'snap-0a1b033b1430b3b05' does not exist."                                                 │
│ I0108 20:02:35.678822       1 controller.go:1004] "ListSnapshots: snapshot not found, returning with success"                                                                                                      │
│ I0108 20:02:36.071131       1 controller.go:616] "ControllerGetCapabilities: called" args=""                                                                                                                       │
│ I0108 20:02:36.071655       1 controller.go:996] "ListSnapshots: called" args="snapshot_id:\"snap-0a1b033b1430b3b05\""                                                                                             │
│ E0108 20:02:36.646632       1 handlers.go:86] "Error from AWS API" err="api error InvalidSnapshot.NotFound: The snapshot 'snap-0a1b033b1430b3b05' does not exist."                                                 │
│ I0108 20:02:36.646756       1 controller.go:1004] "ListSnapshots: snapshot not found, returning with success"                                                                                                      │
│ I0108 20:02:40.647625       1 controller.go:616] "ControllerGetCapabilities: called" args=""
````

### Instance Example 

Crate a fake CSI-node object a fake Node object and a fake volume-Attachemnt

#### example yaml 
```
fake-node.yaml:
yaml
apiVersion: v1
kind: Node
metadata:
  name: fake-node
  labels:
    kubernetes.io/arch: amd64
    kubernetes.io/os: linux
    node.kubernetes.io/instance-type: m5.large
    topology.kubernetes.io/zone: us-west-2a
    node.kubernetes.io/instance-id: i-0b00dbff4a16b95ad
spec:
  taints:
  - effect: NoSchedule
    key: fake-node
status:
  conditions:
  - type: Ready
    status: "True"
  addresses:
  - type: InternalIP
    address: 10.0.1.100
  - type: Hostname
    address: fake-node


csi-node.yaml:
yaml
apiVersion: storage.k8s.io/v1
kind: CSINode
metadata:
  name: fake-node
spec:
  drivers:
  - name: ebs.csi.aws.com
    nodeID: i-0b00dbff4a16b95ad
    topologyKeys:


volume-attachment.yaml:
yaml
apiVersion: storage.k8s.io/v1
kind: VolumeAttachment
metadata:
  name: fake-attachment
spec:
  attacher: ebs.csi.aws.com
  nodeName: fake-node
  source:
    persistentVolumeName: test-pv
```

```
│ I0108 20:05:59.252170       1 controller.go:472] "ControllerPublishVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\"  node_id:\"i-0b00dbff4a16b95ad\"  volume_capability:{mount:{}  access_mode:{mode:SING │
│ I0108 20:05:59.252252       1 controller.go:493] "ControllerPublishVolume: attaching" volumeID="vol-05ae9e63bbfb1f23a" nodeID="i-0b00dbff4a16b95ad"                                                                │
│ E0108 20:05:59.825454       1 handlers.go:86] "Error from AWS API" err="api error InvalidInstanceID.NotFound: The instance ID 'i-0b00dbff4a16b95ad' does not exist"                                                │
│ E0108 20:05:59.825655       1 batcher.go:161] "execute: error executing batch" err="error listing AWS instances: operation error EC2: DescribeInstances, https response error StatusCode: 400, RequestID: 7ac9c220 │
│ I0108 20:05:59.825744       1 inflight.go:74] "Node Service: volume operation finished" key="vol-05ae9e63bbfb1f23ai-0b00dbff4a16b95ad"                                                                             │
│ E0108 20:05:59.825758       1 driver.go:133] "GRPC error" err="rpc error: code = Internal desc = Could not attach volume \"vol-05ae9e63bbfb1f23a\" to node \"i-0b00dbff4a16b95ad\": error listing AWS instances: o │
│ I0108 20:05:59.832940       1 controller.go:472] "ControllerPublishVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\"  node_id:\"i-0b00dbff4a16b95ad\"  volume_capability:{mount:{}  access_mode:{mode:SING │
│ I0108 20:05:59.833040       1 controller.go:493] "ControllerPublishVolume: attaching" volumeID="vol-05ae9e63bbfb1f23a" nodeID="i-0b00dbff4a16b95ad"                                                                │
│ E0108 20:06:00.410111       1 handlers.go:86] "Error from AWS API" err="api error InvalidInstanceID.NotFound: The instance ID 'i-0b00dbff4a16b95ad' does not exist"                                                │
│ I0108 20:06:00.410260       1 inflight.go:74] "Node Service: volume operation finished" key="vol-05ae9e63bbfb1f23ai-0b00dbff4a16b95ad"                                                                             │
│ E0108 20:06:00.410308       1 driver.go:133] "GRPC error" err="rpc error: code = NotFound desc = Volume \"vol-05ae9e63bbfb1f23a\" not found"                                                                       │
│ I0108 20:07:03.833967       1 controller.go:472] "ControllerPublishVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\"  node_id:\"i-0b00dbff4a16b95ad\"  volume_capability:{mount:{}  access_mode:{mode:SING │
│ I0108 20:07:03.834063       1 controller.go:493] "ControllerPublishVolume: attaching" volumeID="vol-05ae9e63bbfb1f23a" nodeID="i-0b00dbff4a16b95ad"                                         │

Request is tried outside of batcher

│ E0106 21:37:21.656333       1 handlers.go:86] "Error from AWS API" err="api error InvalidInstanceID.NotFound: The instance ID 'i-0b00dbff4a16b95ad' does not exist"                                                │
│ I0106 21:43:20.880366       1 controller.go:472] "ControllerPublishVolume: called" args="volume_id:\"vol-05ae9e63bbfb1f23a\" node_id:\"i-0b00dbff4a16b95ad\" volume_capability:{mount:{fs_type:\"ext4\"} access_mo │

As we will timeout on future requests because the batch call does not error out we hit deadline and try again. 

│ E0106 21:43:20.880460       1 driver.go:133] "GRPC error" err="rpc error: code = Aborted desc = An operation with the given Volume vol-05ae9e63bbfb1f23a already exists"
```

```release-note
Cache known bad ids and remove from batch calls
```

working on unit tests and finishing up manual testing WIP

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -217,6 +217,10 @@ var (
 	// For getting IOPS limit from io2 error.
 	// Error example it is used for: "An error occurred (InvalidParameterCombination) when calling the CreateVolume operation: io2 volumes configured with greater than 64 TiB or 256K IOPS or 1000:1 IOPS:GB ratio are not supported".
 	io2ErrRegex = regexp.MustCompile(`(?i)(\d+)K IOPS`)
+
+	volumeIDRegex   = regexp.MustCompile(util.VolumeIDRegex)
+	snapshotIDRegex = regexp.MustCompile(util.SnapshotIDRegex)
+	instanceIDRegex = regexp.MustCompile(util.InstanceIDRegex)
 )
 
 var invalidParameterErrorCodes = map[string]struct{}{
@@ -460,7 +464,6 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
 		ec2Client = plugin.GetEC2Client(cfg, ec2Options)
 		smClient = plugin.GetSageMakerClient(cfg, smOptions)
 	}
-
 	// Default clients if plugin is not in use or does not implement client override.
 	if ec2Client == nil {
 		ec2Client = ec2.NewFromConfig(cfg, ec2Options)
@@ -474,7 +477,6 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
 		klog.V(4).InfoS("NewCloud: batching enabled")
 		bm = newBatcherManager(ec2Client)
 	}
-
 	c := &cloud{
 		awsConfig:             cfg,
 		region:                region,
@@ -506,21 +508,25 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
 // Each batcher's `entries` set to maximum results returned by relevant EC2 API call without pagination.
 // Each batcher's `delay` minimizes RPC latency and EC2 API calls. Tuned via scalability tests.
 func newBatcherManager(svc util.EC2API) *batcherManager {
+	likelyNotFoundInstanceIDs := expiringcache.New[string, struct{}](cacheForgetDelay)
+	likelyNotFoundVolumeIDs := expiringcache.New[string, struct{}](cacheForgetDelay)
+	likelyNotFoundSnapshotIDs := expiringcache.New[string, struct{}](cacheForgetDelay)
+
 	return &batcherManager{
 		volumeIDBatcher: batcher.New(500, batchMaxDelay, func(ids []string) (map[string]*types.Volume, error) {
-			return execBatchDescribeVolumes(svc, ids, volumeIDBatcher)
+			return execBatchDescribeVolumes(svc, ids, volumeIDBatcher, likelyNotFoundVolumeIDs)
 		}),
 		volumeTagBatcher: batcher.New(500, batchMaxDelay, func(names []string) (map[string]*types.Volume, error) {
-			return execBatchDescribeVolumes(svc, names, volumeTagBatcher)
+			return execBatchDescribeVolumes(svc, names, volumeTagBatcher, likelyNotFoundVolumeIDs)
 		}),
 		instanceIDBatcher: batcher.New(50, batchMaxDelay, func(ids []string) (map[string]*types.Instance, error) {
-			return execBatchDescribeInstances(svc, ids)
+			return execBatchDescribeInstances(svc, ids, likelyNotFoundInstanceIDs)
 		}),
 		snapshotIDBatcher: batcher.New(1000, batchMaxDelay, func(ids []string) (map[string]*types.Snapshot, error) {
-			return execBatchDescribeSnapshots(svc, ids, snapshotIDBatcher)
+			return execBatchDescribeSnapshots(svc, ids, snapshotIDBatcher, likelyNotFoundSnapshotIDs)
 		}),
 		snapshotTagBatcher: batcher.New(1000, batchMaxDelay, func(names []string) (map[string]*types.Snapshot, error) {
-			return execBatchDescribeSnapshots(svc, names, snapshotTagBatcher)
+			return execBatchDescribeSnapshots(svc, names, snapshotTagBatcher, likelyNotFoundSnapshotIDs)
 		}),
 		volumeModificationIDBatcher: batcher.New(500, batchMaxDelay, func(names []string) (map[string]*types.VolumeModification, error) {
 			return execBatchDescribeVolumesModifications(svc, names)
@@ -534,23 +540,40 @@ func newBatcherManager(svc util.EC2API) *batcherManager {
 	}
 }
 
+func removeLikelyBadIds(cache expiringcache.ExpiringCache[string, struct{}], input []string) (goodIds []string, likelyBadIds []string) {
+	// Iterate backwards to safely remove values without affecting indices of remaining items
+	for i := len(input) - 1; i >= 0; i-- {
+		id := input[i]
+		_, exists := cache.Get(id)
+		if exists {
+			likelyBadIds = append(likelyBadIds, id)
+		} else {
+			goodIds = append(goodIds, id)
+		}
+	}
+
+	return goodIds, likelyBadIds
+}
+
 // execBatchDescribeVolumes executes a batched DescribeVolumes API call depending on the type of batcher.
-func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBatcherType) (map[string]*types.Volume, error) {
+func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBatcherType, cache expiringcache.ExpiringCache[string, struct{}]) (map[string]*types.Volume, error) {
+	goodVolumes, badVolumes := removeLikelyBadIds(cache, input)
+
 	var request *ec2.DescribeVolumesInput
 
 	switch batcher {
 	case volumeIDBatcher:
-		klog.V(7).InfoS("execBatchDescribeVolumes", "volumeIds", input)
+		klog.V(7).InfoS("execBatchDescribeVolumes", "volumeIds", goodVolumes)
 		request = &ec2.DescribeVolumesInput{
-			VolumeIds: input,
+			VolumeIds: goodVolumes,
 		}
 
 	case volumeTagBatcher:
-		klog.V(7).InfoS("execBatchDescribeVolumes", "names", input)
+		klog.V(7).InfoS("execBatchDescribeVolumes", "names", goodVolumes)
 		filters := []types.Filter{
 			{
 				Name:   aws.String("tag:" + VolumeNameTagKey),
-				Values: input,
+				Values: goodVolumes,
 			},
 		}
 		request = &ec2.DescribeVolumesInput{
@@ -564,9 +587,31 @@ func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBat
 	ctx, cancel := context.WithTimeout(context.Background(), batchDescribeTimeout)
 	defer cancel()
 
-	resp, err := describeVolumes(ctx, svc, request)
-	if err != nil {
-		return nil, err
+	var resp []types.Volume
+	var err error
+
+	if len(goodVolumes) >= 1 {
+		resp, err = describeVolumes(ctx, svc, request)
+		if err != nil {
+			knownErrorVols := volumeIDRegex.FindAllString(err.Error(), -1)
+			for _, volID := range knownErrorVols {
+				cache.Set(volID, &struct{}{})
+			}
+			return nil, err
+		}
+	}
+
+	for _, vol := range badVolumes {
+		request := &ec2.DescribeVolumesInput{
+			VolumeIds: []string{vol},
+		}
+		retryResp, err := describeVolumes(ctx, svc, request)
+		if err == nil && len(retryResp) > 0 {
+			cache.Remove(*retryResp[0].VolumeId)
+			resp = append(resp, retryResp...)
+		} else {
+			klog.V(7).InfoS("execBatchDescribeVolumes: failure", "err", err)
+		}
 	}
 
 	result := make(map[string]*types.Volume)
@@ -1096,18 +1141,42 @@ func (c *cloud) DeleteDisk(ctx context.Context, volumeID string) (bool, error) {
 }
 
 // execBatchDescribeInstances executes a batched DescribeInstances API call.
-func execBatchDescribeInstances(svc util.EC2API, input []string) (map[string]*types.Instance, error) {
-	klog.V(7).InfoS("execBatchDescribeInstances", "instanceIds", input)
+func execBatchDescribeInstances(svc util.EC2API, input []string, cache expiringcache.ExpiringCache[string, struct{}]) (map[string]*types.Instance, error) {
+	goodInstances, badInstances := removeLikelyBadIds(cache, input)
+
+	klog.V(7).InfoS("execBatchDescribeInstances", "instanceIds", goodInstances)
 	request := &ec2.DescribeInstancesInput{
-		InstanceIds: input,
+		InstanceIds: goodInstances,
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), batchDescribeTimeout)
 	defer cancel()
 
-	resp, err := describeInstances(ctx, svc, request)
-	if err != nil {
-		return nil, err
+	var resp []types.Instance
+	var err error
+
+	if len(goodInstances) >= 1 {
+		resp, err = describeInstances(ctx, svc, request)
+		if err != nil {
+			knownErrorInstances := instanceIDRegex.FindAllString(err.Error(), -1)
+			for _, instanceID := range knownErrorInstances {
+				cache.Set(instanceID, &struct{}{})
+			}
+			return nil, err
+		}
+	}
+
+	for _, instance := range badInstances {
+		request := &ec2.DescribeInstancesInput{
+			InstanceIds: []string{instance},
+		}
+		retryResp, err := describeInstances(ctx, svc, request)
+		if err == nil && len(retryResp) > 0 {
+			cache.Remove(*retryResp[0].InstanceId)
+			resp = append(resp, retryResp...)
+		} else {
+			klog.V(7).InfoS("execBatchDescribeInstances: failure", "err", err)
+		}
 	}
 
 	result := make(map[string]*types.Instance)
@@ -1722,22 +1791,24 @@ func getInstanceIDFromAssociatedResource(arn string) (string, error) {
 }
 
 // execBatchDescribeSnapshots executes a batched DescribeSnapshots API call depending on the type of batcher.
-func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapshotBatcherType) (map[string]*types.Snapshot, error) {
+func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapshotBatcherType, cache expiringcache.ExpiringCache[string, struct{}]) (map[string]*types.Snapshot, error) {
+	goodSnapshots, badSnapshots := removeLikelyBadIds(cache, input)
+
 	var request *ec2.DescribeSnapshotsInput
 
 	switch batcher {
 	case snapshotIDBatcher:
-		klog.V(7).InfoS("execBatchDescribeSnapshots", "snapshotIds", input)
+		klog.V(7).InfoS("execBatchDescribeSnapshots", "snapshotIds", goodSnapshots)
 		request = &ec2.DescribeSnapshotsInput{
-			SnapshotIds: input,
+			SnapshotIds: goodSnapshots,
 		}
 
 	case snapshotTagBatcher:
-		klog.V(7).InfoS("execBatchDescribeSnapshots", "names", input)
+		klog.V(7).InfoS("execBatchDescribeSnapshots", "names", goodSnapshots)
 		filters := []types.Filter{
 			{
 				Name:   aws.String("tag:" + SnapshotNameTagKey),
-				Values: input,
+				Values: goodSnapshots,
 			},
 		}
 		request = &ec2.DescribeSnapshotsInput{
@@ -1751,9 +1822,31 @@ func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapsho
 	ctx, cancel := context.WithTimeout(context.Background(), batchDescribeTimeout)
 	defer cancel()
 
-	resp, err := describeSnapshots(ctx, svc, request)
-	if err != nil {
-		return nil, err
+	var resp []types.Snapshot
+	var err error
+
+	if len(goodSnapshots) >= 1 {
+		resp, err = describeSnapshots(ctx, svc, request)
+		if err != nil {
+			knownErrorSnapshots := snapshotIDRegex.FindAllString(err.Error(), -1)
+			for _, snapID := range knownErrorSnapshots {
+				cache.Set(snapID, &struct{}{})
+			}
+			return nil, err
+		}
+	}
+
+	for _, snap := range badSnapshots {
+		request := &ec2.DescribeSnapshotsInput{
+			SnapshotIds: []string{snap},
+		}
+		retryResp, err := describeSnapshots(ctx, svc, request)
+		if err == nil && len(retryResp) > 0 {
+			cache.Remove(*retryResp[0].SnapshotId)
+			resp = append(resp, retryResp...)
+		} else {
+			klog.V(7).InfoS("execBatchDescribeSnapshots: failure", "err", err)
+		}
 	}
 
 	result := make(map[string]*types.Snapshot)
@@ -2068,9 +2161,6 @@ func describeInstances(ctx context.Context, svc util.EC2API, request *ec2.Descri
 	for {
 		response, err := svc.DescribeInstances(ctx, request)
 		if err != nil {
-			if isAWSErrorInstanceNotFound(err) {
-				return nil, ErrNotFound
-			}
 			return nil, fmt.Errorf("error listing AWS instances: %w", err)
 		}
 
@@ -2095,6 +2185,9 @@ func (c *cloud) getInstance(ctx context.Context, nodeID string) (*types.Instance
 	if c.bm == nil {
 		instances, err := describeInstances(ctx, c.ec2, request)
 		if err != nil {
+			if isAWSErrorInstanceNotFound(err) {
+				return nil, ErrNotFound
+			}
 			return nil, err
 		}
 
--- a/pkg/expiringcache/expiring_cache.go
+++ b/pkg/expiringcache/expiring_cache.go
@@ -41,6 +41,8 @@ type ExpiringCache[KeyType comparable, ValueType any] interface {
 	// Set operates identically to setting a value in a map, adding an entry
 	// or overriding the existing value for a given key
 	Set(key KeyType, value *ValueType)
+	// Remove operates identically to removing a value in a map
+	Remove(key KeyType)
 }
 
 type timedValue[ValueType any] struct {
@@ -94,4 +96,11 @@ func (c *expiringCache[KeyType, ValueType]) Set(key KeyType, value *ValueType) {
 			value: value,
 		}
 	}
+}
+
+func (c *expiringCache[KeyType, ValueType]) Remove(key KeyType) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	// In the case we call Remove on a key that does not exist delete is a no op
+	delete(c.values, key)
 }
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -42,6 +42,10 @@ const (
 	// AttachmentShared volume attachment type constant.
 	AttachmentShared    = "shared"
 	AttachmentDedicated = "dedicated"
+
+	VolumeIDRegex   = "vol-[a-z0-9]+"
+	InstanceIDRegex = "i-[a-z0-9]+"
+	SnapshotIDRegex = "snap-[a-z0-9]+"
 )
 
 var (
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -5320,3 +5320,209 @@ func TestCheckIfIopsIncreaseOnExpansion(t *testing.T) {
 		})
 	}
 }
+
+func TestRemoveLikelyBadIds(t *testing.T) {
+	testCases := []struct {
+		name                 string
+		input                []string
+		cachedBadIds         []string
+		expectedResultInput  []string
+		expectedResultBadIds []string
+	}{
+		{
+			name:                "success: no cached bad IDs",
+			input:               []string{"vol-0f8a2c4e6b9d1e5787", "vol-0c7e1a5f8b2d4c6939", "vol-0d9f2b8e4a7c1f53301"},
+			cachedBadIds:        []string{},
+			expectedResultInput: []string{"vol-0d9f2b8e4a7c1f53301", "vol-0c7e1a5f8b2d4c6939", "vol-0f8a2c4e6b9d1e5787"},
+		},
+		{
+			name:                 "success: filter out cached bad IDs",
+			input:                []string{"vol-0f8a2c4e6b9d1e5787", "vol-0c7e1a5f8b2d4c6939", "vol-0d9f2b8e4a7c1f53301"},
+			cachedBadIds:         []string{"vol-0c7e1a5f8b2d4c6939"},
+			expectedResultInput:  []string{"vol-0d9f2b8e4a7c1f53301", "vol-0f8a2c4e6b9d1e5787"},
+			expectedResultBadIds: []string{"vol-0c7e1a5f8b2d4c6939"},
+		},
+		{
+			name:                 "success: filter out multiple cached bad IDs",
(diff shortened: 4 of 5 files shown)


pkg/cloud/cloud.go
```diff
@@ -488,8 +487,16 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
...
+		badInstanceIDs:        expiringcache.New[string, bool](cacheForgetDelay),
+		badVolumeIDs:          expiringcache.New[string, bool](cacheForgetDelay),
+		badSnapshotIDs:        expiringcache.New[string, bool](cacheForgetDelay),
 	}
 
+	if batchingEnabled {
+		klog.V(4).InfoS("NewCloud: batching enabled")
+		bm = newBatcherManager(ec2Client, c)
+	}
+	c.bm = bm
```
ConnorJC3: Why was this code moved?

pkg/cloud/cloud.go
```diff
@@ -488,8 +487,16 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
 		latestClientTokens:    expiringcache.New[string, int](cacheForgetDelay),
 		volumeInitializations: expiringcache.New[string, volumeInitialization](volInitCacheForgetDelay),
 		latestIOPSLimits:      expiringcache.New[string, iopsLimits](iopsLimitCacheForgetDelay),
+		badInstanceIDs:        expiringcache.New[string, bool](cacheForgetDelay),
+		badVolumeIDs:          expiringcache.New[string, bool](cacheForgetDelay),
+		badSnapshotIDs:        expiringcache.New[string, bool](cacheForgetDelay),
```
ConnorJC3: "Bad" is vague, maybe something like "unsafeVolumes" or "likelyNotFoundVolumes"?

pkg/cloud/cloud.go
```diff
@@ -535,7 +542,20 @@ func newBatcherManager(svc util.EC2API) *batcherManager {
 }
 
 // execBatchDescribeVolumes executes a batched DescribeVolumes API call depending on the type of batcher.
-func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBatcherType) (map[string]*types.Volume, error) {
+func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBatcherType, c *cloud) (map[string]*types.Volume, error) {
```
ConnorJC3: Don't pass in all of cloud, pass in the map you need, same as we do with the EC2API param
(obviously ditto all the other functions this was changed in)

pkg/cloud/cloud.go
```diff
@@ -566,9 +586,16 @@ func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBat
 
 	resp, err := describeVolumes(ctx, svc, request)
 	if err != nil {
+		volumeIDRegex := regexp.MustCompile("vol-[a-zA-Z0-9]+")
+		badVolumes := volumeIDRegex.FindAllString(err.Error(), -1)
```
ConnorJC3: Pick a different name for this variable, the shadowing makes it hard to read

pkg/cloud/cloud.go
```diff
@@ -566,9 +586,16 @@ func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBat
...
+		for _, volID := range badVolumes {
+			c.badVolumeIDs.Set(volID, aws.Bool(true))
+		}
 		return nil, err
 	}
 
+	retryBadVolumes(c, badVolumes, svc)
```
ConnorJC3: Given this is only used in one place, does a separate function really make sense?
Also, you need to not discard the response(s) if the call succeeds.

pkg/cloud/cloud.go
```diff
@@ -488,8 +487,16 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
 		latestClientTokens:    expiringcache.New[string, int](cacheForgetDelay),
 		volumeInitializations: expiringcache.New[string, volumeInitialization](volInitCacheForgetDelay),
 		latestIOPSLimits:      expiringcache.New[string, iopsLimits](iopsLimitCacheForgetDelay),
+		badInstanceIDs:        expiringcache.New[string, bool](cacheForgetDelay),
```
ConnorJC3: Why bool? Shouldn't it just be `struct{}`?

pkg/cloud/cloud.go
```diff
@@ -585,6 +612,21 @@ func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBat
...
+	ctx, cancel := context.WithTimeout(context.Background(), batchDescribeTimeout)
+	defer cancel()
+
+	for _, vol := range badVolumes {
+		request := &ec2.DescribeVolumesInput{
+			VolumeIds: []string{vol},
+		}
+		rsp, err := describeVolumes(ctx, svc, request)
+		if err == nil && len(rsp) > 0 {
+			c.badVolumeIDs.Set(*rsp[0].VolumeId, aws.Bool(false))
```
ConnorJC3: This should just be a remove, not set to false, to not waste space

pkg/cloud/cloud.go
```diff
@@ -535,7 +542,20 @@ func newBatcherManager(svc util.EC2API) *batcherManager {
 }
 
 // execBatchDescribeVolumes executes a batched DescribeVolumes API call depending on the type of batcher.
-func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBatcherType) (map[string]*types.Volume, error) {
+func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBatcherType, c *cloud) (map[string]*types.Volume, error) {
+	var goodVolumes []string
+	var badVolumes []string
+	for _, vol := range input {
+		isBadVolume, exists := c.badVolumeIDs.Get(vol)
```
ConnorJC3: If you switch to using `struct{}` and remove (see my other comments), this can become `_, isBadVolume := ...` and simplify the below if statement

pkg/cloud/cloud.go
```diff
@@ -1753,9 +1843,16 @@ func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapsho
 
 	resp, err := describeSnapshots(ctx, svc, request)
 	if err != nil {
+		snapshotIDRegex := regexp.MustCompile("snap-[a-zA-Z0-9]+")
```
ConnorJC3: Can we make these constants (maybe in our util package) so they can be reused elsewhere if needed? We may even already have volume/instance ID regexes elsewhere in the code.

pkg/cloud/cloud.go
```diff
@@ -1096,7 +1138,20 @@ func (c *cloud) DeleteDisk(ctx context.Context, volumeID string) (bool, error) {
...
-func execBatchDescribeInstances(svc util.EC2API, input []string) (map[string]*types.Instance, error) {
+func execBatchDescribeInstances(svc util.EC2API, input []string, c *cloud) (map[string]*types.Instance, error) {
+	var goodInstances []string
+	var badInstances []string
+	for _, vol := range input {
+		isBadInstance, exists := c.badInstanceIDs.Get(vol)
+		if exists && *isBadInstance {
+			badInstances = append(badInstances, vol)
+		} else {
+			goodInstances = append(goodInstances, vol)
```
ConnorJC3: Given we expect the vast majority of volumes to be good, maybe it makes more sense to remove bad volumes from `input` via https://pkg.go.dev/slices#Delete instead of always copying every good volume to a new slice?

pkg/cloud/cloud.go
ElijahQuinones: I can add it if necessary but our expiring cache does not have a remove or delete.

```
type ExpiringCache[KeyType comparable, ValueType any] interface {
	// Get operates identically to retrieving from a map, returning
	// the value and/or boolean indicating if the value existed in the map
	//
	// Multiple callers can receive the same value simultaneously from Get,
	// it is the caller's responsibility to ensure they are not modified
	Get(key KeyType) (value *ValueType, ok bool)
	// Set operates identically to setting a value in a map, adding an entry
	// or overriding the existing value for a given key
	Set(key KeyType, value *ValueType)
}
```

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/expiringcache/expiring_cache.go

pkg/cloud/cloud.go
ElijahQuinones: Using a Bool here as our expiring cache does not have a remove or delete function but as discussed below I am open to adding one.

pkg/cloud/cloud.go
ElijahQuinones: I was thinking of making a way that all the describe bash functions will call one retryBadDescribe function with an extra variable that tells you what describe call to do as the logic for all is super similar but still thinking on it now. In next rev will ether implement that idea or remove the function and have the logic here.

Will make sure the responses are not discarded as well.

pkg/cloud/cloud.go
ElijahQuinones: will do

pkg/cloud/cloud.go
ElijahQuinones: Ah yeah will do

pkg/cloud/cloud.go
ElijahQuinones: Sounds good

pkg/cloud/cloud.go
ElijahQuinones: Because batching manager now needs cloud to be initialized before it is created in the previous location cloud had not been initialized yet

pkg/cloud/cloud.go
ElijahQuinones: We do not currently have a remove method for our expiring cache I can add one if we want to but this is why I am using a bool here as well as this funky logic.

pkg/cloud/cloud.go
ElijahQuinones: sounds good

pkg/cloud/cloud.go
```diff
@@ -1722,7 +1781,17 @@ func getInstanceIDFromAssociatedResource(arn string) (string, error) {
...
+func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapshotBatcherType, cache *expiringcache.ExpiringCache[string, struct{}]) (map[string]*types.Snapshot, error) {
+	var badSnapshots []string
+	for i := len(input) - 1; i >= 0; i-- {
+		vol := input[i]
+		_, exists := (*cache).Get(vol)
+		if exists {
+			badSnapshots = append(badSnapshots, vol)
+			input = slices.Delete(input, i, i+1)
+		}
+	}
```
mdzraf: Would it be possible to extract this logic into a function that can be re-used since we also have it in  `execBatchDescribeInstances` and `execBatchDescribeVolumes` ?

pkg/cloud/cloud.go
```diff
@@ -1722,7 +1781,17 @@ func getInstanceIDFromAssociatedResource(arn string) (string, error) {
 }
 
 // execBatchDescribeSnapshots executes a batched DescribeSnapshots API call depending on the type of batcher.
-func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapshotBatcherType) (map[string]*types.Snapshot, error) {
+func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapshotBatcherType, cache *expiringcache.ExpiringCache[string, struct{}]) (map[string]*types.Snapshot, error) {
+	var badSnapshots []string
+	for i := len(input) - 1; i >= 0; i-- {
+		vol := input[i]
```
mdzraf: nit `snap := ...` ?

pkg/cloud/cloud.go
```diff
@@ -1096,7 +1129,17 @@ func (c *cloud) DeleteDisk(ctx context.Context, volumeID string) (bool, error) {
 }
 
 // execBatchDescribeInstances executes a batched DescribeInstances API call.
-func execBatchDescribeInstances(svc util.EC2API, input []string) (map[string]*types.Instance, error) {
+func execBatchDescribeInstances(svc util.EC2API, input []string, cache *expiringcache.ExpiringCache[string, struct{}]) (map[string]*types.Instance, error) {
+	var badInstances []string
+	for i := len(input) - 1; i >= 0; i-- {
+		vol := input[i]
```
mdzraf: nit: `instance := ...` ?

pkg/cloud/cloud.go
ConnorJC3: +1, and, it took me awhile to realize why we're decrementing backwards, can we leave a comment explaining that so I don't have to spend 10 minutes thinking about it next time I read this code?

Otherwise lgtm

pkg/expiringcache/expiring_cache.go
```diff
@@ -95,3 +97,14 @@ func (c *expiringCache[KeyType, ValueType]) Set(key KeyType, value *ValueType) {
 		}
 	}
 }
+
+func (c *expiringCache[KeyType, ValueType]) Remove(key KeyType) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	_, exists := c.Get(key)
```
torredil: `c.Get(key)` also tries to [acquire the same mutex](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/34809358a819d8aaf61d0172911dcf60c45d503d/pkg/expiringcache/expiring_cache.go#L67), causing a deadlock.

ElijahQuinones: Just seen this when I added the unit tests removed the c.Get(key) as it is unneeded as delete is a no op when the key does not exist

mdzraf:
/lgtm

Waiting on manual testing instructions for approval

pkg/cloud/cloud.go
```diff
@@ -2068,9 +2144,6 @@ func describeInstances(ctx context.Context, svc util.EC2API, request *ec2.Descri
 	for {
 		response, err := svc.DescribeInstances(ctx, request)
 		if err != nil {
-			if isAWSErrorInstanceNotFound(err) {
-				return nil, ErrNotFound
-			}
```
torredil, line 2069: Should this check also be removed from `getInstancesPatchingBatch` ? https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/53109e3522ac5a8883a32ad27dee38ffeeef08d4/pkg/cloud/cloud.go#L2145-L2147

pkg/cloud/cloud.go
```diff
@@ -534,8 +542,24 @@ func newBatcherManager(svc util.EC2API) *batcherManager {
 	}
 }
 
+func removeLikelyBadIds(cache *expiringcache.ExpiringCache[string, struct{}], input *[]string) (likelyBadIds []string) {
```
torredil: Pointer to interfaces are unconventional in Go, suggest updating `*expiringcache.ExpiringCache[string` -> `expiringcache.ExpiringCache[string` and instances of `(*cache).Get()` to just `cache.Get()`, its more idiomatic.

pkg/cloud/cloud.go
```diff
@@ -1753,9 +1813,25 @@ func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapsho
 
 	resp, err := describeSnapshots(ctx, svc, request)
 	if err != nil {
+		snapshotIDRegex := regexp.MustCompile(util.SnapshotIDRegex)
```
torredil: Rather than compiling here every time this code path is invoked, lets do it once at the package level. Similar to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/53109e3522ac5a8883a32ad27dee38ffeeef08d4/pkg/util/util.go#L48

pkg/util/util.go
```diff
@@ -42,6 +42,10 @@ const (
 	// AttachmentShared volume attachment type constant.
 	AttachmentShared    = "shared"
 	AttachmentDedicated = "dedicated"
+
+	VolumeIDRegex   = "vol-[a-zA-Z0-9]+"
+	InstanceIDRegex = "i-[a-zA-Z0-9]+"
+	SnapshotIDRegex = "snap-[a-zA-Z0-9]+"
```
torredil: np: AWS IDs use lowercase hex only, so we can tighten up the regex.

pkg/cloud/cloud.go
```diff
@@ -566,9 +590,25 @@ func execBatchDescribeVolumes(svc util.EC2API, input []string, batcher volumeBat
...
 
+	for _, vol := range badVolumes {
+		request := &ec2.DescribeVolumesInput{
+			VolumeIds: []string{vol},
+		}
+		retryResp, err := describeVolumes(ctx, svc, request)
+		if err == nil && len(retryResp) > 0 {
+			(*cache).Remove(*retryResp[0].VolumeId)
+			resp = append(resp, retryResp...)
+		}
```
torredil, line 614: It would be best to log the error (even at a high v-level), otherwise it just silently fails.

pkg/expiringcache/expiring_cache.go
```diff
@@ -95,3 +97,10 @@ func (c *expiringCache[KeyType, ValueType]) Set(key KeyType, value *ValueType) {
 		}
 	}
 }
+
+func (c *expiringCache[KeyType, ValueType]) Remove(key KeyType) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	// In the case we call Remove on a key that does not exist delete is a no op
+	delete(c.values, key)
```
torredil, line 105: Do we also need to stop the expiring cache timer here?

pkg/cloud/cloud.go
ElijahQuinones: I might be wrong here but if we do not use a pointer here I was under the impression that go was pass by value and thus the real cache would not be updated ? The thought process behind the pointers was ensuring we where modifying the real cache and not just a passed by value version of it.

pkg/expiringcache/expiring_cache.go
ElijahQuinones: I do not believe so the timer as I understand it is value specific and says don't remove this value until this time. As we are removing the value here there is no timer for said value to update.

pkg/cloud/cloud.go
ElijahQuinones: Sounds good to me.

pkg/cloud/cloud.go
```diff
@@ -505,22 +513,22 @@ func NewCloud(region string, awsSdkDebugLog bool, userAgentExtra string, batchin
 // newBatcherManager initializes a new instance of batcherManager.
 // Each batcher's `entries` set to maximum results returned by relevant EC2 API call without pagination.
 // Each batcher's `delay` minimizes RPC latency and EC2 API calls. Tuned via scalability tests.
-func newBatcherManager(svc util.EC2API) *batcherManager {
+func newBatcherManager(svc util.EC2API, c *cloud) *batcherManager {
```
ConnorJC3: This sort of cyclic dependency makes our code hard to read, can we just pass in `c.likelyNotFoundVolumeIDs` here so we can revert the changes in `NewCloud`?

pkg/expiringcache/expiring_cache.go
ConnorJC3: As of Go 1.23, running timers can be garbage collected, so it's safe to just delete here and let the GC cleanup the timer: https://go.dev/wiki/Go123Timer

pkg/cloud/cloud.go
ConnorJC3: In go, an interface is automatically and always a pointer. It is impossible to pass by value to an argument using an interface as its type.

We should also fix this in all the `exec` functions. Note that we do need to pass the input as a pointer to be able to modify it, because `[]string` is not an interface, so keep that.

pkg/cloud/cloud.go
ConnorJC3: No, it should both stay there **and it needs to be added to `getInstance` if we're removing it here**.

These cloud errors are how we determine which gRPC error code to send to the client.

ElijahQuinones: Yeah I will have to move where we convert it to a ErrNotFound such that we are not swallowing the error in the batcher but we are still able to return the correct gRPC error code.

pkg/cloud/cloud.go
ElijahQuinones: Sounds good thank you !

pkg/cloud/cloud.go
```diff
@@ -1096,7 +1136,9 @@ func (c *cloud) DeleteDisk(ctx context.Context, volumeID string) (bool, error) {
 }
 
 // execBatchDescribeInstances executes a batched DescribeInstances API call.
-func execBatchDescribeInstances(svc util.EC2API, input []string) (map[string]*types.Instance, error) {
+func execBatchDescribeInstances(svc util.EC2API, input []string, cache *expiringcache.ExpiringCache[string, struct{}]) (map[string]*types.Instance, error) {
+	badInstances := removeLikelyBadIds(cache, &input)
```
ElijahQuinones: Need to add len check so as to not return all Instances Volumes and snapshots in account when all requests in batch are in the cache.

pkg/cloud/cloud.go
```diff
@@ -534,8 +542,24 @@ func newBatcherManager(svc util.EC2API) *batcherManager {
...
 }
 
+func removeLikelyBadIds(cache *expiringcache.ExpiringCache[string, struct{}], input *[]string) (likelyBadIds []string) {
+	// Iterate backwards to safely remove values without affecting indices of remaining items
+	for i := len(*input) - 1; i >= 0; i-- {
+		id := (*input)[i]
+		_, exists := (*cache).Get(id)
+		if exists {
+			likelyBadIds = append(likelyBadIds, id)
+			*input = slices.Delete(*input, i, i+1)
```
ElijahQuinones: Copies of slices in go unfortunately contain a pointer to the original object as part of the copy .... as such we can not modify input we need to make a new array  to avoid deleting things from the actual batch.

ElijahQuinones:
> This revision looks great. The logic is sound, edge cases are handled correctly, and code coverage is comprehensive. If we could run this through a scale test before release that would be sweet.
> 
> /lgtm

Running one as we speak :)

Edit:

Scale test has been run and all looks normal

pkg/cloud/cloud.go
```diff
@@ -1751,9 +1822,31 @@ func execBatchDescribeSnapshots(svc util.EC2API, input []string, batcher snapsho
...
+			SnapshotIds: []string{snap},
+		}
+		retryResp, err := describeSnapshots(ctx, svc, request)
+		if err == nil && len(retryResp) > 0 {
+			cache.Remove(*retryResp[0].SnapshotId)
+			resp = append(resp, retryResp...)
+		} else {
+			klog.V(7).InfoS("execBatchDescribeSnapshots: failure", "err", err)
+		}
 	}
```
mdzraf, line 1850: I see this logic is repeated in `execBatchDescribeInstances` and `execBatchDescribeVolumes`, I wonder if we can somehow consolidate this into a re-usable function for all? Though not sure if it would be worth doing since it will require a complex restructure and likely negatively impact readability for only a few lines saved and some maintainability gains, so I can also see why we may just want to keep it as is.

Thoughts on this?

ElijahQuinones: I did think of this but it in my opinion it both makes the code harder to read and does not even save us much in the way of repeated code. We have to separate these out regardless as they use different structs for the input for the api call and the api call is different so what you would end up with is a function that does all three but has separation for each of the api calls regardless.

ie 
```
if is volume 
 loop through all volumes
  create describeVolume input 
  call describe volumes
else if is snapshot .....
```

Issue #2812, opened 2025-12-15
driver will never re-register if daemonset pods are duplicated


sjung-anthropic:
/kind bug

**What happened?**

When the apiserver is under high load, the daemonset controller can assign multiple pods to the same node. Each daemonset pod has its own node-driver-registrar container, which will register the ebs.csi.aws.com driver when it starts up, and unregister it when it shuts down. When the daemonset controller recovers, it will delete all the duplicate pods, which will unregister the driver, and only one pod will be left behind. This pod already registered the driver (during its own startup) and will never reregister it (because from its perspective, the job was already done). As a result, this node will now permanently have no ebs.csi.aws.com csi driver on it. This state will persist until an operator restarts the pod/container manually.

**What you expected to happen?**

The system should recover on its own after the daemonset controller recovers, ie the registrar container should register itself again.

**How to reproduce it (as minimally and precisely as possible)?**

Delete the daemonset resource from the cluster, then create two daemonset pods on the same node. (If the daemonset exists, then it will try to deduplicate the pods before they can start.) Once both daemonset pods have started, delete one of them manually (simulating the daemonset controller's behavior of recovering from duplication).

The deleted pod will unregister the driver. Then the remaining pod's ebs-plugin container will fail its liveness probe and restart. However, after that, it will not restart again; it will simply log messages complaining that the driver is not registered. Notably the driver socket will exist, so the liveness-probe container (which hosts /healthz for the ebs-plugin liveness probe) will connect successfully. Meanwhile the node-driver-registrar container will never restart, and will remain unregistered indefinitely.

**Anything else we need to know?**:

The current manifest code references the registrar container's liveness probe command (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/v1.54.0/deploy/kubernetes/base/node.yaml#L127-L132). However this command was deprecated by upstream and intentionally returns 0 (https://github.com/kubernetes-csi/node-driver-registrar/blob/v2.15.0/cmd/csi-node-driver-registrar/main.go#L148-L152), so the registrar container will never fail its liveness probe and thus will not be restarted automatically.

[...]

PR #2819, opened 2025-12-30
Update nodeDriverRegistrar livenessProbe

mdzraf:
#### What type of PR is this?

/kind bug

This PR updates the driver registrars liveness probe to use the supported /healthz endpoint so that it can properly monitor container health.

/closes https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2812

Manually tested by: 
1. Creating cluster and installing driver
2. Deleting `ebs-csi-node` daemonset
3. Deploying two test pods on the same node
4. Deleting one of the pods (simulates duplication recovery behavior)
5. Monitor the remaining pod to ensure that the `ebs-plugin` container restarts after failing liveness probe, followed by `node-driver-registrar` container properly failing a health check and restarting. 

```
   Warning  Unhealthy  20m (x3 over 23m)  kubelet  Liveness probe failed: HTTP probe failed with statuscode: 404
   Normal   Killing    20m                kubelet  Container node-driver-registrar failed liveness probe, will be restarted
```

```release-note
Replaced deprecated node-driver-registrar liveness probe command with /healthz endpoint.
```

### DIFF
--- a/charts/aws-ebs-csi-driver/values.schema.json
+++ b/charts/aws-ebs-csi-driver/values.schema.json
@@ -1141,35 +1141,7 @@
             },
             "livenessProbe": {
               "type": "object",
-              "additionalProperties": false,
-              "properties": {
-                "exec": {
-                  "type": "object",
-                  "additionalProperties": false,
-                  "properties": {
-                    "command": {
-                      "type": "array",
-                      "default": [
-                        "/csi-node-driver-registrar",
-                        "--kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)",
-                        "--mode=kubelet-registration-probe"
-                      ]
-                    }
-                  }
-                },
-                "initialDelaySeconds": {
-                  "type": "integer",
-                  "default": 30
-                },
-                "periodSeconds": {
-                  "type": "integer",
-                  "default": 90
-                },
-                "timeoutSeconds": {
-                  "type": "integer",
-                  "default": 15
-                }
-              }
+              "additionalProperties": true
             }
           }
         },
--- a/charts/aws-ebs-csi-driver/values.yaml
+++ b/charts/aws-ebs-csi-driver/values.yaml
@@ -154,11 +154,9 @@ sidecars:
       readOnlyRootFilesystem: true
       allowPrivilegeEscalation: false
     livenessProbe:
-      exec:
-        command:
-          - /csi-node-driver-registrar
-          - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
-          - --mode=kubelet-registration-probe
+      httpGet:
+        path: /healthz
+        port: healthz
       initialDelaySeconds: 30
       periodSeconds: 90
       timeoutSeconds: 15
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -118,18 +118,20 @@ spec:
           args:
             - --csi-address=$(ADDRESS)
             - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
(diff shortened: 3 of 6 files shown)


charts/aws-ebs-csi-driver/values.schema.json
```diff
@@ -1143,17 +1143,17 @@
               "type": "object",
               "additionalProperties": false,
               "properties": {
```
ConnorJC3: This is going to break existing valid customizations of the liveness probe. We should just drop this `properties` section entirely and flip `additionalProperties` to true for this param

charts/aws-ebs-csi-driver/values.yaml
```diff
@@ -148,17 +148,16 @@ sidecars:
       tag: "v2.15.0-eksbuild.3"
     logLevel: 2
     # Additional parameters provided by node-driver-registrar.
-    additionalArgs: []
+    additionalArgs: 
+      - --http-endpoint=0.0.0.0:9809
```
ConnorJC3: Left in from debugging? Drop this

ConnorJC3 (changes requested):

charts/aws-ebs-csi-driver/values.yaml
mdzraf: The health check endpoint is only exposed if the http endpoint is set, am I correct in that we need to enable a default for this so that it works out of the box? Maybe not do it  additionalArgs, but maybe leave it as a dedicated field? Something like:
```
  nodeDriverRegistrar:
    env: []
    image:
      pullPolicy: IfNotPresent
      repository: public.ecr.aws/csi-components/csi-node-driver-registrar
      tag: "v2.15.0-eksbuild.3"
    logLevel: 2
    # Default health check endpoint, can be modified through additional args
    httpEndpoint: "0.0.0.0:9809"
```
And then if a customer wants to customize it they can do so through the additionalArgs?

charts/aws-ebs-csi-driver/values.schema.json
mdzraf: Will do, going to keep the `initialDelaySeconds, periodSeconds, timeoutSeconds` defaults though since the k8s defaults are very low (0s, 10s, 1s)

mdzraf: Will do

charts/aws-ebs-csi-driver/values.yaml
ConnorJC3: Don't do anything with additional args, just put it in the default argument set.

deploy/kubernetes/base/node.yaml
```diff
@@ -119,17 +119,16 @@ spec:
...
               value: /var/lib/kubelet/plugins/ebs.csi.aws.com/csi.sock
           livenessProbe:
-            exec:
-              command:
-              - /csi-node-driver-registrar
-              - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
-              - --mode=kubelet-registration-probe
+            httpGet:
+              path: /healthz
+              port: 9809
```
torredil: The `spec.container.ports` field is informational but as a best practice, I suggest adding it to the manifest here, ie:

```
ports:
   - containerPort: 9809
      name: healthz
livenessProbe:
     httpGet:
       path: /healthz
       port: healthz
```

Similar to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/base/node.yaml#L82

mdzraf: Done ✅

ConnorJC3 (approved):
/lgtm

mdzraf:
/approve 

Got two necessary lgtms.

PR #2820, opened 2026-01-06
Add encryption support ext4 format option to StorageClass params

### LINKED ISSUE #2394: [Feature] Expose `-O encrypt` format option

In brief, we're in need to encrypt data at rest by using a per-tenant key, when storing customer data on EBS. We're currently thinking that the best way to fulfill this requirement is to use [fscrypt](https://github.com/google/fscrypt), which works at the Kernel level and just requires the `ext4` filesystem [to be formatted with](https://wiki.archlinux.org/title/Fscrypt#ext4) the `-O encrypt` option.

This option cannot be set at the `StorageClass` level with any option, as far as I can see.

/feature

FWIW, `mount-utils` [supports arbitrary options](https://github.com/kubernetes/mount-utils/blob/d561b6cfba27ec5c35c8b3255bc69d36f2dd7321/mount_linux.go#L554) to be passed to `mkfs.ext4`, so it would be just a matter of adding one more `ext4` specific option to [the storage class parameters](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/parameters.md).

This option could be plumbed through and passed to `mount-utils` similarly to other options ([example here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/668d94ebd180a0456cf9b1063297f28c78d10657/pkg/driver/node.go#L270)).

We've considered doing this on our own, basically through a `DaemonSet` that looks for EBS volumes mounted in the host that match certain characteristics, and call `tune2fs -O encrypt /dev/device` on them, but it sounds like a brittle solution.

Obviously this would be the first step, in order to get this to production, but ideally we hope that we can get rid of this workaround and rely on a first-class option in the `StorageClass`.

**Additional context**
More details about `fscrypt` [here](https://wiki.archlinux.org/title/Fscrypt) and [here]((https://github.com/google/fscrypt).

ConnorJC3:
#### What type of PR is this?

/kind feature

Fixes #2394, allowing encryption of data at rest via non-EBS encryption methods.

Added CI and unit tests.

```release-note
Add new StorageClass parameter `ext4EncryptionSupport` to enable encryption support in ext4 for non-EBS encryption at rest
```


pkg/driver/constants.go
```diff
@@ -111,6 +111,9 @@ const (
 	// Ext4ClusterSizeKey configures the cluster size when formatting an ext4 volume with the bigalloc option enabled.
 	Ext4ClusterSizeKey = "ext4clustersize"
 
+	// Ext4EncryptionSupportKey enables the encrypt option when formatting an ext4 volume.
+	Ext4EncryptionSupportKey = "ext4encryptionsupport"
```
torredil, line 115: Wondered if `fsEncryptionSupport` would be more appropriate to future proof this parameter.  

I think the current name is the right call, its clear and consistent with existing ext4BigAlloc/ext4ClusterSize params and doesn't mislead users into thinking it works on xfs/ntfs.

docs/parameters.md
```diff
@@ -22,6 +22,7 @@ The AWS EBS CSI Driver supports [tagging](tagging.md) through `StorageClass.para
...
 | "ext4ClusterSize"            |                                                 |         | The cluster size to use when formatting an `ext4` filesystem when the `bigalloc` feature is enabled. Note: The `ext4BigAlloc` parameter must be set to true. See our [FAQ](/docs/faq.md).                                                                                                                                                                                                     |
+| "ext4EncryptionSupport"      | true, false                                     | false   | Enables the `encrypt` feature when formatting an `ext4` filesystem. Only supported on linux nodes with fstype `ext4`.                                                                                                                                                                                                                                                                         |
```
torredil: For bigalloc we briefly mention checking kernel compatibility, similarly I think it would be a great idea to mention this functionality requires kernel 4.1+ and specific kernel configuration (`CONFIG_FS_ENCRYPTION` or `CONFIG_EXT4_ENCRYPTION`).

ConnorJC3: I'll add a mention of the config, but I don't think it's worth mentioning 4.1 because it's long EOL (https://kernel.org/) and thus just adds noise

pkg/driver/constants.go
ConnorJC3: Ooops didn't see this one before somehow: I agree with you the current name is better. It's possible we walk ourselves into an unfortunate future, but here's how I see it:

- Worst case scenario with `fsEncryptionSupport`: We end up having to create an `xfsEncryptionSupport` if it differs per filesystem (e.g. there's different modes or something more complicated than a true/false): extremely confusing for customers
- Worst case scenario with `ext4EncryptionSupport`: We end up having to create an `xfsEncryptionSupport`, which isn't ideal but isn't confusing

docs/parameters.md
ConnorJC3: Ok, doing some digging, it looks like `CONFIG_EXT4_ENCRYPTION` is only for outdated 4.XX kernels, and recent kernels use `CONFIG_FS_ENCRYPTION` so I mentioned that. Also added some extra wording to make what this parameter does more clear, and clarify it doesn't actually encrypt anything on its own, just enable the ext4 feature so pods can do the encryption.

PR #2821, opened 2026-01-07
Remove instances of gomock Any/AssignableToTypeOf/AnyTimes

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Greatly improves the robustness of unit tests by removing all instances of `gomock.Any()` and `.AnyTimes()` from the code.

A few edge cases:
- Some values (contexts and EC2 option functions), are nearly impossible to match on with `gomock.Eq()`. I created custom matchers for these cases. A new `testutil` package was created so these matchers can be reused in all tests.
- In some cases we are mocking a value passed in by non-driver code (primarily, the List and Watch calls from informers). To avoid testing Kubernetes code, we use `gomock.AssignableToTypeOf()` in those cases.
- In the request coalescing test, the input doesn't matter much and the parts that do are validated in the `DoAndReturn`, so I used `AssignableToTypeOf`
- A few instances of `AnyTimes` are left in that are exceptionally difficult to remove without major refactoring of the tests
- EC2 Inputs are difficult to match because of tag ordering, the EC2 Input matcher will be made to match more strictly at a later date.

`make test` and `make verify`

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -50,6 +50,16 @@ linters:
     - noinlineerr # We use inline errors
     - embeddedstructfieldcheck # We have a few instances of this
   settings:
+    forbidigo:
+      forbid:
+        # Forbid testing anti-patterns Any/AssignableToTypeOf
+        - pattern: ^gomock\.Any$
+          msg: gomock.Any() is forbidden. Use gomock.Eq() or testutil matchers instead.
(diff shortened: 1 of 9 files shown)


torredil (approved):
/approve

mdzraf (approved):
/lgtm

PR #2833, opened 2026-01-22
Detect, abort, and retry stuck attachments

ConnorJC3:
#### What type of PR is this?

/kind bug

Abort attachments that appear to be in a permanently broken state (have been attaching for at least 90 seconds) by calling `DetachVolume` and forcing a sidecar-level retry.

Unit tests added.

```release-note
Detect, abort, and force a retry of attachments stuck for longer than 90 seconds
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -100,6 +100,10 @@ const (
 	dryRunInterval = 3 * time.Hour
 
 	getCallerIdentityRetryDelay = 30 * time.Second
+
+	// stuckAttachingTimeout is the duration after which an attachment stuck in "attaching" state
+	// will be detached to allow a retry.
+	stuckAttachingTimeout = 90 * time.Second
 )
 
 // VolumeStatusInitializingState is const reported by EC2 DescribeVolumeStatus which AWS SDK does not have type for.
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud_test.go
```diff
@@ -4445,6 +4454,17 @@ func TestWaitForAttachmentState(t *testing.T) {
 				mockEC2.EXPECT().DescribeVolumes(testutil.AnyContext(), testutil.EC2Input(&ec2.DescribeVolumesInput{})).Return(&ec2.DescribeVolumesOutput{Volumes: []types.Volume{hyperpodAttachedVol}}, nil).MinTimes(1)
 			case "failure: multiple attachments with Multi-Attach disabled":
 				mockEC2.EXPECT().DescribeVolumes(testutil.AnyContext(), testutil.EC2Input(&ec2.DescribeVolumesInput{})).Return(&ec2.DescribeVolumesOutput{Volumes: []types.Volume{multipleAttachmentsVol}}, nil).MinTimes(1)
+			case "failure: stuck attaching triggers detach":
+				stuckAttachTime := time.Now().Add(-100 * time.Second)
```
ConnorJC3, line 4458: "Isn't there a `.Sub` method?"

There is but it doesn't work here, because it returns a duration. Go's official docs tell you to use `.Add` for this case.

pkg/cloud/cloud_test.go
```diff
@@ -4445,6 +4454,17 @@ func TestWaitForAttachmentState(t *testing.T) {
 				mockEC2.EXPECT().DescribeVolumes(testutil.AnyContext(), testutil.EC2Input(&ec2.DescribeVolumesInput{})).Return(&ec2.DescribeVolumesOutput{Volumes: []types.Volume{hyperpodAttachedVol}}, nil).MinTimes(1)
 			case "failure: multiple attachments with Multi-Attach disabled":
 				mockEC2.EXPECT().DescribeVolumes(testutil.AnyContext(), testutil.EC2Input(&ec2.DescribeVolumesInput{})).Return(&ec2.DescribeVolumesOutput{Volumes: []types.Volume{multipleAttachmentsVol}}, nil).MinTimes(1)
+			case "failure: stuck attaching triggers detach":
```
ConnorJC3, line 4457: Yeah I know it sucks to depend on test name here but this PR isn't the appropriate place to refactor this massively bad test.

pkg/cloud/cloud.go
```diff
@@ -100,6 +100,10 @@ const (
 	dryRunInterval = 3 * time.Hour
 
 	getCallerIdentityRetryDelay = 30 * time.Second
+
+	// stuckAttachingTimeout is the duration after which an attachment stuck in "attaching" state
+	// will be detached to allow a retry.
+	stuckAttachingTimeout = 90 * time.Second
```
ElijahQuinones, line 106: Largely looks go to me but is there any situation in which we expect an attachment to take longer than 90 seconds to attach where detaching and trying again will put us in a loop of almost being attached and then prematurely detaching before the attach can finish. I can't think of any but wanted to make sure before approving.

mdzraf (approved):
/lgtm

No manual tests but code lgmt

pkg/cloud/cloud.go
ConnorJC3: I have never seen an EBS attachment take longer than 10-15 seconds except when it gets stuck, so this is a very conservative estimate of when we can abort. To the best of my knowledge this would be excessively far outside the normal bounds of how long an attach operation should take.

PR #2827, opened 2026-01-13
Initialize name-dependent variables in metadata-labaler node path

ConnorJC3:
#### What type of PR is this?

/kind bug

#### What is this PR about? / Why do we need it?

Fixes a regression introduced in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2786 that breaks the node path for the alpha `metadata-labeler` feature. The globals being relied upon in this code path were not initialized because the node does not call `ContinuousUpdateLabelsLeaderElection`.

#### How was this change tested?

`make test-e2e-disruptive`

#### Does this PR introduce a user-facing change?

```release-note
Prevent startup crash when alpha `metadata-labeler` feature is enabled by initializing name-dependent variables in node path
```

PR #2843, opened 2026-01-24
feat: Expand plugin interface GetSegments()

engedaam:
#### What type of PR is this?

/kind feature 

This change extends the EBS CSI driver plugin interface to support custom node topology segments that can be propagated to Kubernetes resources.

- Added GetNodeSegments() method to the EbsCsiPlugin interface, allowing plugins to provide additional topology segments
- Introduced SetNodeSegments() and GetNodeSegments() utility functions to manage plugin-provided segments globally
- Updated controller and node services to merge plugin segments with existing topology information (zone, OS, outpost ARN)
- Modified all relevant tests to account for the new segment merging behavior using lo.Assign()

Dependencies:
- Added github.com/samber/lo for convenient map merging operations

- `make test`

```release-note
NONE
```

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -35,7 +35,6 @@ import (
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/internal"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util/template"
-	"github.com/samber/lo"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -1210,7 +1209,7 @@ func newCreateVolumeResponse(disk *cloud.Disk, ctx map[string]string) *csi.Creat
 	}
 
 	segments := map[string]string{WellKnownZoneTopologyKey: disk.AvailabilityZone}
-	lo.Assign(segments, util.GetNodeSegments())
+	maps.Copy(segments, util.GetNodeSegments())
 
 	arn, err := arn.Parse(disk.OutpostArn)
 
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"math/rand"
 	"reflect"
 	"strings"
@@ -36,7 +37,6 @@ import (
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/internal"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/testutil"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
-	"github.com/samber/lo"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/codes"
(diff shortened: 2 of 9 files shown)


pkg/driver/controller.go
```diff
@@ -1209,6 +1210,7 @@ func newCreateVolumeResponse(disk *cloud.Disk, ctx map[string]string) *csi.Creat
 	}
 
 	segments := map[string]string{WellKnownZoneTopologyKey: disk.AvailabilityZone}
+	lo.Assign(segments, util.GetNodeSegments())
```
torredil: lo.Assign does not modify the map in place, so the return value is being discarded here and plugin segments are never actually added.

pkg/driver/node.go
```diff
@@ -615,6 +616,7 @@ func (d *NodeService) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
 		WellKnownZoneTopologyKey: zone,
 		OSTopologyKey:            osType,
 	}
+	lo.Assign(segments, util.GetNodeSegments())
```
torredil: Same issue from controller.go, this new merged map is never assigned and is getting discarded.

pkg/driver/controller.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/driver/internal"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util/template"
+	"github.com/samber/lo"
```
torredil: Can we avoid adding a new dependency for this? The codebase already uses maps.Copy from the standard lib in several places, and it does exactly what we need here.

pkg/plugin/plugin_common.go
```diff
@@ -71,25 +71,31 @@ type EbsCsiPlugin interface {
 	// GetDriverName replaces the driver name in use (normally "ebs.csi.aws.com")
 	// This function can be called before Init and should not depend on it
 	GetDriverName() string
+	// GetSegments provides addational segments to be added as part of the driver and controllers
```
torredil: np: addational -> additional

pkg/driver/controller_test.go
```diff
@@ -663,7 +664,9 @@ func TestCreateVolume(t *testing.T) {
 					VolumeContext: map[string]string{},
 					AccessibleTopology: []*csi.Topology{
 						{
-							Segments: map[string]string{WellKnownZoneTopologyKey: expZone},
+							Segments: lo.Assign(map[string]string{
```
torredil: GetNodeSegments() returns an empty map here since nothing calls SetNodeSegments with actual data, so this test passes whether or not the merge works. Can we add a test that sets some segments and checks they show up in the response?

pkg/driver/controller.go
engedaam: Sure, I can switch to this function

pkg/driver/controller.go
engedaam: Dropped `lo.Assign()`, so this has been updated to a function that will update the map in place

pkg/driver/controller_test.go
engedaam: This is expected. Once a consumer has loaded the test with a plugin the segments will be updated to the correct value. This is how the remaining of the tests have been configured for the plugin

pkg/util/util.go
```diff
@@ -83,6 +85,21 @@ func GetDriverName() string {
 	return driverName
 }
 
+func SetNodeSegments(segments map[string]string) {
```
ConnorJC3, line 88: `SetDriverName` is an unfortunate anti-pattern we adopted to get ourselves out of a bad situation with how the driver name was handled in a pre-plugin world, but I'd like to avoid expanding on that anti-pattern as it:
- Makes tracing the code path used harder
- Greatly increases the chance of bugs slipping in, for example this one related to SetDriverName: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2827
- Increases the difficulty of testing changes

Because the same constraints that applied to the name don't apply to topology segments, can we just directly call the plugin in `node.go` and `controller.go`?

Something like:
```
plugin := plugin.GetPlugin()
if plugin != nil {
  maps.Copy(segments, plugin.GetTopologySegments())
}
```

pkg/plugin/plugin_common.go
```diff
@@ -71,6 +71,8 @@ type EbsCsiPlugin interface {
 	// GetDriverName replaces the driver name in use (normally "ebs.csi.aws.com")
 	// This function can be called before Init and should not depend on it
 	GetDriverName() string
+	// GetSegments provides additional segments to be added as part of the driver and controllers
+	GetNodeSegments() map[string]string
```
ConnorJC3, line 75: If we're also going to reuse this in the controller, can we call it something like `GetTopologySegments` or otherwise not reference `Node`?

ConnorJC3 (changes requested):

pkg/util/util.go
jmdeal: Sounds good, I've made this update in my PR.

pkg/plugin/plugin_common.go
jmdeal: I've cut a PR to continue this one, there I actually separated this out into two methods: `GetDiskTopologySegments` and `GetNodeTopologySegments`. The idea is that these two segments shouldn't be bound because they're properties of the instance and the volume respectively. I've also updated the methods to take in the metadata for the instance and disk respectively.

jmdeal:
I've discussed this with @engedaam offline, I'm going to take over this PR. Since I can't edit this one directly I just created a new PR with the requested changes: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2858

ConnorJC3:
/close

Closing this in favor of #2858

PR #2848, opened 2026-01-28
feat(chart/aws-ebs-csi-driver): enable choosing between Deployment/DaemonSet Kind for controller service

rcmorano:
#### What type of PR is this?

/kind feature

This feature allows you to change the kind of object used for the controller service. I'm using `DaemonSet` so volume provisioning works on my multi-regional deployment.  

I've uninstalled the original chart, reinstalled it using `controller.kind=DaemonSet` and successfully provisioned a pvc for a sts that I'm running in a different region than my eks control plane.

```release-note
`controller.kind` value is now configurable (only `Deployment` and `DaemonSet` are allowed).
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -1,6 +1,7 @@
 {{- if not .Values.nodeComponentOnly -}}
 # Controller Service
-kind: Deployment
+{{- $kind := .Values.controller.kind | default "Deployment" }}
+kind: {{ $kind }}
 apiVersion: apps/v1
 metadata:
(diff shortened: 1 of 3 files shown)


rcmorano:
Sorry it seems I was just lucky multiple times, this approach is not feasible as requests will randomly go through only one of the controllers.

One workaround could be to deploy multiple csi drivers with different names (ie `ebs-us-east-1.csi.aws.com`, `ebs-eu-west-1.csi.aws.com`) and use different storage classes with topology keys, but unfortunately, driver name is hardcoded and therefore not customizable via helm chart. 
I will look into some other ways to achieve multi-region and might open a PR to make it -optionally- customizable.

PR #2858, opened 2026-02-05
feat(plugin): configure node and volume topology segments

jmdeal:
#### What type of PR is this?

/kind feature

NOTE: This is a continuation of #2843

Adds the following methods to the `EbsCsiPlugin` interface:
- `GetDiskTopologySegments(*cloud.Disk)`
- `GetNodeTopologySegments(metadata.MetadataService)`

This enables plugins to inject topological constraints into the disks provisioned by the driver, and provide additional topological information for nodes running the driver. The original PR had these combined into a single function, `GetNodeSegments`, but I've separated them out since these topologies may not always be the same. Additionally, I've parameterized the methods to take in metadata for the disk and instance respectively which can be used by plugins to generate the topology segments.

`make test`

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -42,6 +42,7 @@ import (
 	dm "github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/devicemanager"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/expiringcache"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/metrics"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/plugin"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
@@ -415,13 +416,6 @@ type cloud struct {
 
 var _ Cloud = &cloud{}
 
-type AWSClientProvider interface {
-	// GetEC2Client replaces the AWS EC2 client the driver uses
-	GetEC2Client(cfg aws.Config, optFns ...func(*ec2.Options)) util.EC2API
-	// GetSageMakerClient replaces the AWS EC2 client the driver uses
-	GetSageMakerClient(cfg aws.Config, optFns ...func(*sagemaker.Options)) util.SageMakerAPI
-}
-
 // initVariables initializes variables that depend on driver name.
 // Separated into a separate function from NewCloud so it can be called in tests.
 func initVariables() {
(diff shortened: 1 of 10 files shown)


pkg/cloud/cloud.go
jmdeal, line 1: The motivation for these changes was to break the circular dependency between `cloud` and `plugin` that was created when making `cloud.Disk` a parameter for one of the plugin methods. I think this approach makes more sense than lifting out `cloud.Disk` into a third package and I think the justification to take a dependency on `cloud.Disk` in plugin is solid.

ElijahQuinones:
Hi @jmdeal 

You need to run `make verify` and address some of the  issues it pointed out such as 

`pkg/driver/controller_test.go:2203:1: File is not properly formatted (gci)`

and 

```
pkg/driver/node_test.go:2177:15: avoid direct access to proto field tc.expectedResp.AccessibleTopology.Segments, use tc.expectedResp.GetAccessibleTopology().GetSegments() instead (protogetter)
				maps.Copy(tc.expectedResp.AccessibleTopology.Segments, p.GetNodeTopologySegments(metadataService))
				          ^
pkg/util/util.go:59:2: var pluginSegments is unused (unused)
	pluginSegments = map[string]string{}
```

jmdeal:
I'll convert to draft for the time being, `make verify` isn't returning this error for me locally. I haven't been able to dive into it this morning yet. I'll convert out of draft once it's addressed

jmdeal:
~~@ElijahQuinones Do you have any opinion on skipping this check for this test though? I'm intentionally mutating that map so accessing it through a Getter doesn't make a ton of sense IMO. There are other workarounds I can take, but it seems like an appropriate place to configure it to skip that lint.~~

Nevermind, didn't realize the getters return pointer types. The indirection is a little strange to me, but it works 🤷

jmdeal:
Looks like this tests is failing here, but it is passing for me locally. I'll retest here, but is this test known to be flakey?

TestPatchNewNodes

/retest

jmdeal:
@ElijahQuinones looks like all tests are passing now, is there anything else outstanding?

ConnorJC3:
I'm a bit worried about using `cloud.Disk` and `metadata.MetadataService` as the inputs to the functions - the intent of the plugin system is to reduce the maintenance burden on plugin authors caused by breaking changes during EBS CSI updates, and we don't really consider those stable ABIs we'd be willing to commit to never changing. It also creates the ugly workaround we have to use in this PR to avoid an import loop.

Is there specific information from the Disk/Metadata you're looking for to determine the new topology that maybe would be better to pass in?

Also, FYI, we're currently working on a CI issue which means this PR will likely fail the single-az tests upon re-test by prow, that's "expected" and just ignore it for now.

jmdeal:
> Is there specific information from the Disk/Metadata you're looking for to determine the new topology that maybe would be better to pass in?

Fundamentally these topologies are properties of the disk and the instance respectively. A plugin should have some method to get the properties of those items to generate the topologies. Currently this isn't required for our use-case though, so I'd be fine with punting on this for now and revisiting when a need arises if we don't consider those APIs stable.

jmdeal:
I've updated the PR to remove the metadata source arguments, this should be good for review again.

ConnorJC3 (commented):
/lgtm
/retest

I (or somebody else) should do a final round of manual testing before it's approved but I agree this is the best approach right now and no issues noted in the code.

ConnorJC3:
/approve

Ran the tests locally and I haven't seen any other dissent on the PR in the last week.

PR #2867, opened 2026-02-17
Rename ports to avoid collisions

### LINKED ISSUE #2866: node-driver-registrar and ebs-plugin both register the same port name `healthz`
/kind bug

**What happened?**
Both containers within the same pod register the same port, which means that any check that references them by name will get the first one.
```Warning: spec.template.spec.containers[1].ports[0]: duplicate port name "healthz" with spec.template.spec.containers[0].ports[0], services and probes that select ports by name will use spec.template.spec.containers[0].ports[0]```

For each container to have their own healthz endpoint.

The default configuration is enough to reproduce the problem.

**Environment**
- Kubernetes version (use `kubectl version`): `v1.34.3`
- Driver version: `v1.55.0`

### PR DESCRIPTION
**Pull request by ElijahQuinones**
#### What type of PR is this?

/kind cleanup
Fixes #2866 and fixes this for the metrics ports as well 
CI

```release-note
Rename health and metrics ports 
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/controller.yaml
+++ b/charts/aws-ebs-csi-driver/templates/controller.yaml
@@ -278,7 +278,7 @@ spec:
               mountPath: /var/lib/csi/sockets/pluginproxy/
           {{- if .Values.controller.enableMetrics }}
           ports:
-            - name: metrics
+            - name: metrics-prov
               containerPort: 3302
               protocol: TCP
           {{- end }}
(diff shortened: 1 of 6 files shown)


charts/aws-ebs-csi-driver/templates/controller.yaml
```diff
@@ -278,7 +278,7 @@ spec:
               mountPath: /var/lib/csi/sockets/pluginproxy/
           {{- if .Values.controller.enableMetrics }}
           ports:
-            - name: metrics
+            - name: metrics-pro
```
ConnorJC3: Can we we `-prov` as "pro" implies something different? otherwise lgtm

PR #2868, opened 2026-02-17
Adding Helm Parameter CI Tests

mdzraf:
#### What type of PR is this?

/kind feature

This PR adds individual tests for the various helm parameters that can be configured when deploying the driver. 

Tested manually by running:

```
mkdir -p /tmp/e2e-results && export REPORT_DIR=/tmp/e2e-results  && make e2e/parameters-all
```

Also ran in CI under the existing single-az testing grid. It will have it's own test grid after this merges but put under single-az for reviewers to see results and JUnit outputs.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,22 @@ cluster/uninstall: bin/helm bin/aws
 ## E2E targets
 # Targets to run e2e tests
 
+.PHONY: test/helm-template
+test/helm-template: bin/helm
+	cd tests/helm-template && go test -v -count=1 ./...
+
+## e2e/parameters and e2e/parameters-all are Parameter-specific e2e tests
+# Usage: make e2e/parameters PARAM_SET=<name> or make e2e/parameters-all
+# See hack/e2e/param-sets.sh for available sets and their definitions.
+ 
+.PHONY: e2e/parameters
+e2e/parameters: bin/helm bin/ginkgo
+	./hack/e2e/param-sets.sh run $(PARAM_SET)
+
+.PHONY: e2e/parameters-all
+e2e/parameters-all: bin/helm bin/ginkgo test/helm-template
+	./hack/e2e/param-sets.sh run-all
+
 .PHONY: e2e/single-az
 e2e/single-az: bin/helm bin/ginkgo
 	AWS_AVAILABILITY_ZONES=us-west-2a \
--- a/hack/e2e/param-sets.sh
+++ b/hack/e2e/param-sets.sh
@@ -1,0 +1,148 @@
+#!/bin/bash
+
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Parameter set definitions for e2e parameter tests that require a live cluster.
+# Tests that only assert on rendered Kubernetes object specs run via
+# `go test ./tests/helm-template/...` and do not need a cluster.
+#
+# Each set only needs to set GINKGO_FOCUS. By default, Helm values come from
+# tests/helm-template/testdata/e2e-<set>.yaml (or a file pointed at by
+# VALUES_FILE), shared with the helm-template Go tests as a single source
+# of truth. A set may override HELM_EXTRA_FLAGS directly when the set only
+# needs one or two flags and a values file isn't worth it.
+#
+# Sets in PARAM_SETS_ALL:
+#   standard            - Volume tagging (EC2 API) and defaultFsType (mount check)
+#   miscellaneous       - Metadata labeler node labels and additional DaemonSet scheduling
+#   node-component-only - Deploys only node DaemonSet without controller
+#   fips                - Builds FIPS image then validates it is deployed
+#   legacy-compat       - legacyXFS behavior
+
+set -euo pipefail
+
+BASE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+VALUES_DIR="${BASE_DIR}/../../tests/helm-template/testdata"
+
+PARAM_SETS_ALL="standard miscellaneous node-component-only fips legacy-compat"
+
+param_set_standard() {
+  GINKGO_FOCUS="\[param:(extraCreateMetadata|k8sTagClusterId|extraVolumeTags|defaultFsType)\]"
+  # Opt out of install_driver()'s hardcoded --set controller.k8sTagClusterId=$CLUSTER_NAME
+  # so the value from e2e-standard.yaml takes effect (otherwise --set would beat --values).
+  HELM_OVERRIDE_K8S_TAG_CLUSTER_ID=true
+}
+
+param_set_miscellaneous() {
+  GINKGO_FOCUS="\[param:(metadataLabeler|additionalDaemonSets)\]"
+}
+
+param_set_node-component-only() {
+  GINKGO_FOCUS="\[param:nodeComponentOnly\]"
+  # Shares values with the helm-template TestNodeComponentOnly (identical content).
+  VALUES_FILE="${VALUES_DIR}/node-component-only.yaml"
+}
+
+param_set_legacy-compat() {
+  GINKGO_FOCUS="\[param:legacyXFS\]"
+  HELM_EXTRA_FLAGS="--set=node.legacyXFS=true"
+}
+
+param_set_fips() {
+  GINKGO_FOCUS="\[param:fips\]"
+  # Single flag; inlined instead of maintaining a one-line values file.
+  HELM_EXTRA_FLAGS="--set=fips=true"
+  FIPS_TEST=true
+}
+
+# Load a parameter set by name, exporting GINKGO_FOCUS and HELM_EXTRA_FLAGS.
+# By default HELM_EXTRA_FLAGS is derived from a values file convention
+# (e2e-<name>.yaml). A per-set function may override HELM_EXTRA_FLAGS or
+# VALUES_FILE directly when the values-file convention doesn't fit.
+load_param_set() {
+  local name="$1"
+  local func="param_set_${name}"
+  if ! declare -f "$func" >/dev/null 2>&1; then
+    echo "Unknown parameter set: ${name}" >&2
+    echo "Available sets: ${PARAM_SETS_ALL}" >&2
+    exit 1
+  fi
+  HELM_EXTRA_FLAGS=""
+  VALUES_FILE="${VALUES_DIR}/e2e-${name}.yaml"
+  # Clear state that a previous set's param_set_* function may have set,
+  # so sets are independent of run order.
+  unset HELM_OVERRIDE_K8S_TAG_CLUSTER_ID
+  unset FIPS_TEST
+  "$func"
+  if [[ -z "$HELM_EXTRA_FLAGS" ]]; then
+    HELM_EXTRA_FLAGS="--values=${VALUES_FILE}"
+  fi
+  export GINKGO_FOCUS HELM_EXTRA_FLAGS
+  export GINKGO_PARALLEL="${GINKGO_PARALLEL:-5}"
+  export AWS_AVAILABILITY_ZONES="${AWS_AVAILABILITY_ZONES:-us-west-2a}"
+  export TEST_PATH="${TEST_PATH:-./tests/e2e/...}"
+  export JUNIT_REPORT="${REPORT_DIR:-/logs/artifacts}/junit-params-${name}.xml"
+  if [[ -n "${EBS_INSTALL_SNAPSHOT+x}" ]]; then export EBS_INSTALL_SNAPSHOT; fi
+  if [[ -n "${FIPS_TEST+x}" ]]; then export FIPS_TEST; fi
+  if [[ -n "${HELM_OVERRIDE_K8S_TAG_CLUSTER_ID+x}" ]]; then export HELM_OVERRIDE_K8S_TAG_CLUSTER_ID; fi
+}
+
+# Run a single parameter set
+run_param_set() {
+  load_param_set "$1"
+  if [[ "${FIPS_TEST:-}" == "true" ]]; then
+    echo "### Building FIPS image for param set: $1"
+    FIPS_TEST=true make cluster/image || {
+      echo "FIPS image build failed!" >&2
+      return 1
+    }
+  fi
+  echo "### Running parameter set: $1"
+  ./hack/e2e/run.sh
+}
+
+# Run all standard parameter sets sequentially
+run_all_param_sets() {
+  echo "Running all parameter sets sequentially..."
+  local failed_sets=()
+  for set in $PARAM_SETS_ALL; do
+    run_param_set "$set" || failed_sets+=("$set")
+  done
+  if [[ ${#failed_sets[@]} -gt 0 ]]; then
+    echo "Failed parameter sets: ${failed_sets[*]}" >&2
+    return 1
+  fi
+  echo "All parameter sets completed successfully!"
+}
+
+# Allow direct invocation: ./hack/e2e/param-sets.sh run <name> or ./hack/e2e/param-sets.sh run-all
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+  case "${1:-}" in
+  run)
+    [[ -z "${2:-}" ]] && {
+      echo "Usage: $0 run <param-set-name>" >&2
+      exit 1
+    }
+    run_param_set "$2"
+    ;;
+  run-all)
+    run_all_param_sets
+    ;;
+  *)
+    echo "Usage: $0 {run <param-set-name>|run-all}" >&2
+    exit 1
+    ;;
+  esac
+fi
--- a/tests/e2e/parameter_tests.go
+++ b/tests/e2e/parameter_tests.go
@@ -1,0 +1,335 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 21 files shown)


hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,185 @@
...
+
+param_set_legacy-compat() {
+  GINKGO_FOCUS="\[param:(useOldCSIDriver|legacyXFS)\]"
+  HELM_EXTRA_FLAGS="--set=useOldCSIDriver=true,node.legacyXFS=true"
+}
+
+param_set_selinux() {
+  GINKGO_FOCUS="\[param:selinux\]"
+  HELM_EXTRA_FLAGS="--set=node.selinux=true"
+}
```
mdzraf, line 61: These are not in the PARAMETERS_ALL test for now, I will see if they need a separate cluster or not.

mdzraf:
/hold

Have to reverse single-az makefile target before merging

mdzraf:
Noticed some tests are being skipped, looking into it.

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,189 @@
...
+# limitations under the License.
+
+# Parameter set definitions for e2e parameter tests.
+# Each set defines GINKGO_FOCUS, HELM_EXTRA_FLAGS, and optionally other env vars.
+#
+# Sets in PARAM_SETS_ALL (no "special" cluster config needed):
+#   standard            - Behavioral params (tagging, metrics, logging, storage classes, etc.)
+#   other               - Volume modification, volume attach limit, and metadata labeler
+#   debug               - debugLogs=true overrides individual logLevel settings
+#   infra               - Infrastructure/deployment params (resources, security, strategy, etc.)
```
ElijahQuinones: Why is storage classes in standard and not in infra it is a cluster resource

ElijahQuinones: Also why are we dividing these into categories at all ? If we end up running them all anyways

Makefile
```diff
@@ -128,14 +128,23 @@ cluster/uninstall: bin/helm bin/aws
...
+# yet have their own testgrid but reviewers need to see how JUnit output is after tests run for this PR. 
 .PHONY: e2e/single-az
 e2e/single-az: bin/helm bin/ginkgo
-	AWS_AVAILABILITY_ZONES=us-west-2a \
-	TEST_PATH=./tests/e2e/... \
-	GINKGO_FOCUS="\[ebs-csi-e2e\] \[single-az\]" \
-	GINKGO_PARALLEL=5 \
-	HELM_EXTRA_FLAGS="--set=controller.volumeModificationFeature.enabled=true,sidecars.provisioner.additionalArgs[0]='--feature-gates=VolumeAttributesClass=true',sidecars.resizer.additionalArgs[0]='--feature-gates=VolumeAttributesClass=true',node.enableMetrics=true" \
-	./hack/e2e/run.sh
+	./hack/e2e/param-sets.sh run-all
```
torredil: Thanks for leaving a comment about it, I'm including it in the review so that we don't forget to revert this before the merge.

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,195 @@
...
+# Merge per-set JUnit XMLs into a single file with duplicate skipped tests removed.
+# Each Ginkgo run reports ALL specs (most as skipped), so the same skipped test appears
+# in every per-set file. This merges all results into one file, keeping non-skipped results
+# (passed/failed) over skipped duplicates, and emitting each skipped test only once.
+merge_junit_results() {
+  local report_dir="${REPORT_DIR:-/logs/artifacts}"
+  local output="${report_dir}/junit-params.xml"
+
+  python3 - "$report_dir" "$output" <<'PYEOF'
+import glob, sys, xml.etree.ElementTree as ET
```
torredil: I'd drop this merge script and let the JUnit files stand on their own.

Testgrid handles multiple files per job fine, and keeping them separate actually helps debugging since you can tell which param set a failure came from just by the filename. The skipped test noise is a not an issue,(the external test jobs already report 7000+ skips), but if we want to clean it up properly, the right move is putting these parameter tests in their own package so Ginkgo only discovers the specs we care about.

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,195 @@
...
+  # Remove per-set files only if merge succeeded, so CI only sees the merged result
+  if [[ $? -eq 0 && -f "$output" ]]; then
+    rm -f "${report_dir}"/junit-params-*.xml
+    echo "Merged JUnit results into ${output}"
+  else
+    echo "WARNING: JUnit merge failed, keeping per-set files" >&2
+  fi
+}
+
+# Run all standard parameter sets sequentially
```
torredil, line 116: Why are param sets running sequentially? can we parallelize?

Another thing, most of these tests don't need a live cluster. Things like "does the controller deployment have 3 replicas" or "does the node daemonset have this toleration" are assertions about rendered kubernetes objects, not runtime behavior. We could test the vast majority with helm template + assertions on the rendered YAML.

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,195 @@
...
...r.k8sTagClusterId=e2e-param-test,controller.extraVolumeTags.TestKey=TestValue,controller.enableMetrics=true,node.enableMetrics=true,controller.batching=true,controller.defaultFsType=xfs,controller.loggingFormat=json,node.loggingFormat=json,controller.logLevel=5,node.logLevel=5,sidecars.provisioner.logLevel=5,sidecars.attacher.logLevel=5,sidecars.snapshotter.logLevel=5,sidecars.resizer.logLevel=5,sidecars.nodeDriverRegistrar.logLevel=5,defaultStorageClass.enabled=true,storageClasses[0].name=test-sc,storageClasses[0].parameters.type=gp3,volumeSnapshotClasses[0].name=test-vsc,volumeSnapshotClasses[0].deletionPolicy=Delete,sidecars.snapshotter.forceEnable=true,controller.userAgentExtra=e2e-test,controller.enablePrometheusAnnotations=true,node.enablePrometheusAnnotations=true,node.kubeletPath=/var/lib/kubelet,node.tolerateAllTaints=true,controller.podDisruptionBudget.enabled=true,sidecars.provisioner.leaderElection.enabled=true,sidecars.attacher.leaderElection.enabled=true,sidecars.resizer.leaderElection.enabled=true,node.reservedVolumeAttachments=2,node.hostNetwork=true,node.serviceAccount.disableMutation=true,node.terminationGracePeriodSeconds=60,nodeAllocatableUpdatePeriodSeconds=30"
```
torredil: Say someone adds a new helm value `controller.foo=bar`,  here's what they have to do:

1. Write the go test with the right `[param:foo]` tag
2. Figure out which param set it belongs in (standard? infra? other? new one?)
3. Add foo to the `GINKGO_FOCUS` regex for that param set
4. Add `controller.foo=bar` to the `HELM_EXTRA_FLAGS` string for that param set
5. Hardcode "bar" in the go test assertion and hope it matches what they put in step 4
6. Hope they didn't typo anything in this massive 1000+ character strings

we need to simplify this, and have a single source of truth for what to deploy and what to assert. I suggest  defining the test values once, in one place, such as by using a values yaml file per param set and have the go tests read expected values from it (or a shared const).

mdzraf: Changes to value files per set to make this easier instead of having the extra flags string. For the ones that are not many parameters set, I left as the string since I didn't think it was worth having a whole file for.

hack/e2e/param-sets.sh
mdzraf: Split tests into ones that just assert rendering and ones that need live cluster. They are ran sequentially because that way we can use one cluster for all tests instead of spinning up separate clusters.

hack/e2e/param-sets.sh
mdzraf: Done ✅

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,148 @@
...
+}
+
+param_set_node-component-only() {
+  GINKGO_FOCUS="\[param:nodeComponentOnly\]"
+  # Shares values with the helm-template TestNodeComponentOnly (identical content).
+  VALUES_FILE="${VALUES_DIR}/node-component-only.yaml"
+}
+
+param_set_legacy-compat() {
+  GINKGO_FOCUS="\[param:(useOldCSIDriver|legacyXFS)\]"
```
ElijahQuinones: This test set is defined here but no corresponding tests are run. 

So when they are run they will "pass" but all we are doing is installing the driver with these parameters without checking anything.

An idea for testing this would be looking for the presence of `fsGroupPolicy` which should not exist when we set this parameter.

As for legacy xfs an idea would be provisioning a pod with an xfs volume then using 

`"xfs_info /mnt/test-1 | grep -q 'reflink=0'"` to ensure reflink is disabled

tests/helm-template/testdata/miscellaneous.yaml
```diff
@@ -0,0 +1,37 @@
+# Copyright 2026 The Kubernetes Authors.
```
ElijahQuinones, line 1: Would strongly suggest not naming this other.yaml

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,148 @@
...
+PARAM_SETS_ALL="standard other node-component-only fips"
+
+param_set_standard() {
+  GINKGO_FOCUS="\[param:(extraCreateMetadata|k8sTagClusterId|extraVolumeTags|defaultFsType)\]"
+  # Opt out of install_driver()'s hardcoded --set controller.k8sTagClusterId=$CLUSTER_NAME
+  # so the value from e2e-standard.yaml takes effect (otherwise --set would beat --values).
+  HELM_OVERRIDE_K8S_TAG_CLUSTER_ID=true
+}
+
+param_set_other() {
```
ElijahQuinones: Can we have a more descriptive name than other ?

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,148 @@
...
+}
+
+param_set_legacy-compat() {
+  GINKGO_FOCUS="\[param:(useOldCSIDriver|legacyXFS)\]"
+  # Only two flags; inlined instead of maintaining a near-empty values file.
+  HELM_EXTRA_FLAGS="--set=useOldCSIDriver=true,node.legacyXFS=true"
+}
+
+param_set_selinux() {
+  GINKGO_FOCUS="\[param:selinux\]"
```
ElijahQuinones: Same as comment on 62 

We do not actually test this at a minimum can we look to ensure setting the parameter changes the csi driver spec to have `seLinuxMount=true` 

A suggestion for testing would also be to check the node ebsPlugin's mount paths for `/sys/fs/selinux` and `/etc/selinux/config`

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,148 @@
...
+  local name="$1"
+  local func="param_set_${name}"
+  if ! declare -f "$func" >/dev/null 2>&1; then
+    echo "Unknown parameter set: ${name}" >&2
+    echo "Available sets: ${PARAM_SETS_ALL} legacy-compat selinux" >&2
+    exit 1
+  fi
+  HELM_EXTRA_FLAGS=""
+  VALUES_FILE="${VALUES_DIR}/e2e-${name}.yaml"
+  unset HELM_OVERRIDE_K8S_TAG_CLUSTER_ID
```
ElijahQuinones, line 86: Should we also add 

```
unset FIPS_TEST            
unset EBS_INSTALL_SNAPSHOT
```

such that we are always unsetting all  these env vars no matter what order we run these tests in

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,148 @@
...
+  GINKGO_FOCUS="\[param:selinux\]"
+  # Single flag; inlined instead of maintaining a one-line values file.
+  HELM_EXTRA_FLAGS="--set=node.selinux=true"
+}
+
+param_set_fips() {
+  GINKGO_FOCUS="\[param:fips\]"
+  # Single flag; inlined instead of maintaining a one-line values file.
+  HELM_EXTRA_FLAGS="--set=fips=true"
+  FIPS_TEST=true make cluster/image
```
ElijahQuinones: Feels super weird that all of these other functions only set env vars and this one builds an image along with setting the env vars. On top of that if make cluster/image fails we just continue. 

Can we add something like the following to run_param_set

```
if [[ "${FIPS__TEST}" == "true" ]]; then
          echo "### Building FIPS image for param set: $1"
           make cluster/image || { echo "FIPS image build failed!" >&2; return 1; }
      fi
```

tests/helm-template/infra_test.go
```diff
@@ -0,0 +1,514 @@
...
+	ebsPlugin := findContainer(t, cPS, "ebs-plugin")
+
+	ds := mustFind(t, resources, "DaemonSet", "ebs-csi-node")
+	dsSpec := nested(t, ds, "spec")
+	nPS := podSpec(t, ds)
+	nodePlugin := findContainer(t, nPS, "ebs-plugin")
+
+	t.Run("controllerReplicaCount", func(t *testing.T) {
+		replicas, ok := nestedFloat(dSpec, "replicas")
+		if !ok || replicas != 3 {
```
ElijahQuinones: Can we make sure these aren't hardcoded ? If I update infra.yaml this test will suddenly fail even if the test is properly adding my new requested replicas. 

Suggest pulling all the values we are checking here from infra.yaml as opposed to hard coding them

tests/helm-template/standard_test.go
```diff
@@ -0,0 +1,262 @@
...
+	t.Run("hostNetwork", func(t *testing.T) {
+		hn, ok := nPS["hostNetwork"].(bool)
+		if !ok || !hn {
+			t.Error("node pod should use hostNetwork")
+		}
+	})
+
+	t.Run("nodeTerminationGracePeriod", func(t *testing.T) {
+		tgp, ok := nestedFloat(nPS, "terminationGracePeriodSeconds")
+		if !ok || tgp != 60 {
```
ElijahQuinones: Same comment as on infra_test.go can we pull the value we are looking for here from standard.yaml such that we do not need to update the value in two places.

tests/e2e/parameter_tests.go
```diff
@@ -0,0 +1,297 @@
...
+		test := testsuites.DynamicallyProvisionedCmdVolumeTest{
+			CSIDriver: driver.InitEbsCSIDriver(),
+			Pods:      pods,
+		}
+		test.Run(cs, f.Namespace)
+	})
+
+	It("[param:nodeComponentOnly] should deploy only node DaemonSet without controller", func() {
+		_, err := cs.AppsV1().Deployments(ebsNamespace).Get(context.Background(), "ebs-csi-controller", metav1.GetOptions{})
+		Expect(err).To(HaveOccurred(), "Controller deployment should not exist when nodeComponentOnly=true")
```
ElijahQuinones: Can we check specially for a 404 here not any error. This test will also pass on `context.DeadlineExceeded` or any other possible error now.


Something like
 `Expect(apierrors.IsNotFound(err)).To(BeTrue(), "Controller deployment should not exist, but got error: %v", err)`

hack/e2e/param-sets.sh
mdzraf: For upcoming rev I added unset FIPS_TEST alongside the existing unset HELM_OVERRIDE_K8S_TAG_CLUSTER_ID, since `param_set_fips` is the only function that sets it and we do want it cleared between sets.

Held off on `unset EBS_INSTALL_SNAPSHOT` because no param_set_* function sets it, and clearing it inside `load_param_set` overrides whatever a caller may export if running locally. config.sh defaults it to "true". So unsetting it forces the snapshotter to reinstall every run even when the user explicitly set EBS_INSTALL_SNAPSHOT=false. That said, let me know if you believe we should re-evaluate this.

tests/e2e/parameter_tests.go
mdzraf: ✅ Done

tests/helm-template/standard_test.go
mdzraf: ✅ Done

tests/helm-template/infra_test.go
mdzraf: ✅  Done

hack/e2e/param-sets.sh
mdzraf: ✅ Done

hack/e2e/param-sets.sh
mdzraf: ✅ Done

hack/e2e/param-sets.sh
mdzraf: ✅ Changed to `miscellaneous`

tests/helm-template/testdata/miscellaneous.yaml
mdzraf: ✅ Changed to `miscellaneous`

hack/e2e/param-sets.sh
mdzraf: ✅ Done

Makefile
```diff
@@ -128,14 +128,27 @@ cluster/uninstall: bin/helm bin/aws
 ## E2E targets
 # Targets to run e2e tests
 
+.PHONY: test/helm-template
+test/helm-template: bin/helm
+	cd tests/helm-template && go test -v -count=1 ./...
+
+## e2e/parameters and e2e/parameters-all are Parameter-specific e2e tests
+# Usage: make e2e/parameters PARAM_SET=<name> or make e2e/paramaters-all
```
ElijahQuinones: nit: paramaters → parameters

hack/e2e/param-sets.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# Run all standard parameter sets sequentially
+run_all_param_sets() {
+  echo "Running all parameter sets sequentially..."
+  local failed=0
+  for set in $PARAM_SETS_ALL; do
+    run_param_set "$set" || failed=1
+  done
+  if [[ $failed -ne 0 ]]; then
+    echo "One or more parameter sets failed!"
```
ElijahQuinones: Suggestion: can we have an array that keeps track of test failures and is printed at the end hear so we don't need to fish through the log for what tests exactly failed?

Something like:

```
 local failed_sets=()
  for set in $PARAM_SETS_ALL; do
    run_param_set "$set" || failed_sets+=("$set")
  done
  if [[ ${#failed_sets[@]} -gt 0 ]]; then
    echo "Failed parameter sets: ${failed_sets[*]}" >&2
    return 1
  fi
```

ElijahQuinones:
Largely looks good to me just two more minor non blocking comments.
/lgtm

hack/e2e/param-sets.sh
mdzraf: ✅

Makefile
mdzraf: ✅

mdzraf:
/hold 

Once approved, I will remove single-az test hijack and then merge.

mdzraf:
/unhold
/approve

Merging as we discussed offline.

PR #2870, opened 2026-02-20
Release v1.56.0

ElijahQuinones:
Releases 1.56.0 of the ebs-csi-driver
CI

```release-note
None
```

### DIFF
--- a/charts/aws-ebs-csi-driver/CHANGELOG.md
+++ b/charts/aws-ebs-csi-driver/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Helm chart
 
+## 2.56.0
+
+- Bump driver version to `v1.56.0`.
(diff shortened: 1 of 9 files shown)


charts/aws-ebs-csi-driver/CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Helm chart
 
+## 2.56.0
+
+- Bump driver version to `v1.56.0`.
+- Bump sidecars to latest.
```
torredil: Suggest leaving a note here stating that the new sidecar builds address:
- a bug where retries could cause volumes to be provisioned in the wrong availability zone.
- a bug where the csi-resizer would crash with "concurrent map writes" when multiple resize operations failed simultaneously.

PR #2878, opened 2026-02-26
Add flake retries for Windows volume expansion tests

### LINKED ISSUE #2877: Windows Resize Failures Observed in CI
/kind failing-test

### Description

Windows volume expansion tests fail intermittently in CI with:

```
Get-PartitionSupportedSize : Failed
    + FullyQualifiedErrorId : StorageWMI 4,Get-PartitionSupportedSize
{"SizeMax":  null}
```

Once triggered for a volume, every kubelet retry fails identically until the defragsvc state changes.

### Root cause

csi-proxy calls `Get-PartitionSupportedSize` to determine the max partition size before resizing. This method has a hard dependency on the defragsvc service. When defragsvc is disabled or busy holding volume locks, the call fails with StorageWMI error code 4. There is no alternative Microsoft API to get the max supported partition size without this dependency.

I reproduced this on a standalone Windows Server 2025 EC2 instance by disabling defragsvc, which produces the exact same error, re-enabling fixes it immediately.

Looking more into this, I found that csi-proxy's `ResizeVolumeRequest` has a `SizeBytes` field. When SizeBytes > 0, csi-proxy skips `Get-PartitionSupportedSize` entirely and resizes directly. I implemented this (passing RequiredBytes from the kubelet through to SizeBytes) as a fix, and confirmed the plumbing works, but unfortunately RequiredBytes is the disk size, not the max partition size and the partition can't be as large as the disk due to GPT + MSR overhead (~17 MiB on small disks, ~129 MiB on large disks), so Resize-Partition rejects it with "Not enough available capacity." In other words we just trade one error for another.

Computing the max partition size from disk geometry (either in the driver or csi-proxy) isn't really viable either - it would only work for single partition disks and replaces a Windows API with unsafe custom logic. The driver also can't do this computation since it runs as an unprivileged container with no direct host access.

### Proposed fix

I think we should add retries for Windows volume expansion tests. The `defragsvc` contention is transient, and kubelet already retries `NodeExpandVolume` in production, so this aligns CI with prod behavior. I'm not a huge fan of retries as this can mask bugs, but for this specific issue I think its safe to do since the failure here is well understood and deterministic in its cause. We would enable retries for the small set of problematic tests on the Windows CI job (we still maintain no retries for resize tests on Linux).

### Affected versions

Both csi-proxy v1.2.1 (PowerShell-based) and v1.3.0 (CIM-based) have the same defragsvc dependency. The underlying WMI method is the same regardless of call path so the issue exists in both versions.

torredil:
#### What type of PR is this?

/kind cleanup

Windows volume expansion tests can fail intermittently due to defragsvc contention causing GetSupportedSize (StorageWMI error 4) failures.

This PR splits the Windows external test run into two passes:
- All non volume-expand tests run without retries
- Volume expansion tests run with --flake-attempts

closes #2877

- CI
- make verify && make test

```release-note
NONE
```

### DIFF
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -138,16 +138,62 @@ else
     set -x
     set +e
     # kubetest2 looks for deployers/testers in $PATH
-    PATH="${BIN}:${PATH}" "${BIN}/kubetest2" noop \
-      --run-id="e2e-kubernetes" \
-      --test=ginkgo \
-      -- \
-      --skip-regex="${GINKGO_SKIP}" \
-      --focus-regex="${GINKGO_FOCUS}" \
-      --test-package-version=$(curl -L https://dl.k8s.io/release/stable-${packageVersion}.txt) \
-      --parallel=${GINKGO_PARALLEL} \
-      --test-args="-storage.testdriver=${PWD}/manifests.yaml -kubeconfig=${KUBECONFIG} -node-os-distro=${NODE_OS_DISTRO}"
-    TEST_PASSED=$?
+
+    # Regex matching volume expansion tests susceptible to transient failures on Windows due to defragsvc contention.
+    WINDOWS_VOLUME_EXPAND_REGEX="volume-expand|expansion of pvcs created for ephemeral"
+
+    TEST_PACKAGE_VERSION=$(curl -L https://dl.k8s.io/release/stable-${packageVersion}.txt)
+
+    run_kubetest2() {
+      local run_id="$1"
+      local skip="$2"
+      local focus="$3"
+      local extra_ginkgo_args="${4:-}"
+
+      PATH="${BIN}:${PATH}" "${BIN}/kubetest2" noop \
+        --run-id="${run_id}" \
+        --test=ginkgo \
+        -- \
+        --skip-regex="${skip}" \
+        --focus-regex="${focus}" \
+        --test-package-version="${TEST_PACKAGE_VERSION}" \
+        --parallel=${GINKGO_PARALLEL} \
+        ${extra_ginkgo_args:+--ginkgo-args="${extra_ginkgo_args}"} \
+        --test-args="-storage.testdriver=${PWD}/manifests.yaml -kubeconfig=${KUBECONFIG} -node-os-distro=${NODE_OS_DISTRO}"
+    }
+
+    if [[ "${WINDOWS}" == true ]]; then
+      # Pass 1: Run all tests except volume-expand (no retries).
+      loudecho "Running non-volume-expand tests (no retries)"
+      run_kubetest2 "e2e-kubernetes" \
+        "${GINKGO_SKIP}|${WINDOWS_VOLUME_EXPAND_REGEX}" \
+        "${GINKGO_FOCUS}"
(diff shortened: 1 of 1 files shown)


hack/e2e/run.sh
```diff
@@ -138,16 +138,64 @@ else
...
+
+      # Pass 2: Run only volume-expand tests with flake retries to tolerate
+      # transient defragsvc contention on Windows (StorageWMI error 4).
+      loudecho "Running volume-expand tests (with flake retries)"
+      PATH="${BIN}:${PATH}" "${BIN}/kubetest2" noop \
+        --run-id="e2e-kubernetes-volume-expand" \
+        --test=ginkgo \
+        -- \
+        --skip-regex="${GINKGO_SKIP}" \
+        --focus-regex="${GINKGO_FOCUS}.*(${WINDOWS_VOLUME_EXPAND_REGEX})" \
```
ConnorJC3: Potentially dumb question: Why the `(` and `)` in this regex?

hack/e2e/run.sh
```diff
@@ -138,16 +138,64 @@ else
...
+      PATH="${BIN}:${PATH}" "${BIN}/kubetest2" noop \
+        --run-id="e2e-kubernetes" \
+        --test=ginkgo \
+        -- \
+        --skip-regex="${GINKGO_SKIP}" \
+        --focus-regex="${GINKGO_FOCUS}" \
+        --test-package-version=$(curl -L https://dl.k8s.io/release/stable-${packageVersion}.txt) \
+        --parallel=${GINKGO_PARALLEL} \
+        --test-args="-storage.testdriver=${PWD}/manifests.yaml -kubeconfig=${KUBECONFIG} -node-os-distro=${NODE_OS_DISTRO}"
+      TEST_PASSED=$?
```
ConnorJC3, line 195: Repeating this same block of code 3 times feels really non-DRY, should we make a function for it?

hack/e2e/run.sh
```diff
@@ -138,16 +138,64 @@ else
...
+        -- \
+        --skip-regex="${GINKGO_SKIP}" \
+        --focus-regex="${GINKGO_FOCUS}.*(${WINDOWS_VOLUME_EXPAND_REGEX})" \
+        --test-package-version=$(curl -L https://dl.k8s.io/release/stable-${packageVersion}.txt) \
+        --parallel=${GINKGO_PARALLEL} \
+        --ginkgo-args="--flake-attempts=2" \
+        --test-args="-storage.testdriver=${PWD}/manifests.yaml -kubeconfig=${KUBECONFIG} -node-os-distro=${NODE_OS_DISTRO}"
+      VOLUME_EXPAND_PASSED=$?
+
+      if [[ ${TEST_PASSED} -ne 0 || ${VOLUME_EXPAND_PASSED} -ne 0 ]]; then
```
ConnorJC3: np: I don't think this if case is needed, you could just have the inner if at line 182

hack/e2e/run.sh
torredil: Without the parentheses the second pattern loses the External.Storage prefix so it matches any driver's ephemeral expansion test.

PR #2886, opened 2026-03-18
Append -dev to non-release user agents

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Append `-dev` to UA of all non-release containers (CI, local dev environments, etc), add build release containers with `RELEASE_BUILD=true` to disable that.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,12 @@
 ## Variables/Functions
 
 VERSION?=v1.56.0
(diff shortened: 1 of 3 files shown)


ConnorJC3:
/hold

good to go but don't want to merge midway through release, will unhold after

PR #2891, opened 2026-03-20
Bump go and driver version

ElijahQuinones:
#### What type of PR is this?

/kind cleanup

Bumps go and trivy versions
CI

```release-note
NONE
```


ConnorJC3:
/override pull-aws-ebs-csi-driver-security

Failing because prow hasn't bumped their stdlib yet but we expect this to not be in the image (will be manually validating after it's built)

torredil:
make security failure has nothing to do with Prow and still needs to be addressed, `golang.org/x/net` should be bumped to `v0.51.0` and `go 1.26.0` -> `go 1.26.1` in go.mod.

PR #2899, opened 2026-03-24
Tag volumes and snapshots with cluster identity

torredil:
#### What type of PR is this?

/kind feature

When --k8s-tag-cluster-id is set, the driver now also writes an `ebs.csi.aws.com/cluster-name` tag on volumes and snapshots, enabling cluster-scoped IAM policies.

e2e test / unit test

```release-note
When --k8s-tag-cluster-id is set, tag volumes and snapshots with `ebs.csi.aws.com/cluster-name` to support cluster-scoped IAM policies.
```

### DIFF
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -181,6 +181,9 @@ const (
 	// the external provisioner sidecar is started with --extra-create-metadata=true and
 	// thus provides such metadata to the CSI driver.
 	PVNameTag = "kubernetes.io/created-for/pv/name"
+
+	// ClusterNameTagKey is the resource tag key for cluster-scoped IAM policies.
+	ClusterNameTagKey = "ebs.csi.aws.com/cluster-name"
 )
 
 // constants for default command line flag values.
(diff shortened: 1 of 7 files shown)


pkg/driver/constants.go
```diff
@@ -181,6 +181,9 @@ const (
 	// the external provisioner sidecar is started with --extra-create-metadata=true and
 	// thus provides such metadata to the CSI driver.
 	PVNameTag = "kubernetes.io/created-for/pv/name"
+
+	// EKSClusterNameTagKey is the tag key used for cluster-scoped IAM policies.
+	EKSClusterNameTagKey = "eks-cluster-name"
```
ConnorJC3: Why `eks`? This isn't an eks-specific feature and I feel like a much better name is something like `ebs.csi.aws.com/cluster-name` or something along that line.

torredil: Pod Identity injects a fixed set of session tags, [eks-cluster-name](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-abac.html) is one of them. Using any other key means Pod Identity users lose the zero-config path and have to manually tag their IAM role, same as IRSA users. So we have to trade here, either generic naming vs automatic Pod Identity support.

mdzraf: Is doing something in the policy like:  
```
"Condition": {
    "StringEquals": {
        "ec2:ExistingObjectTag/cluster-name": "${aws:PrincipalTag/eks-cluster-name}"
    }
}
```
not allowed? If it isn't then I think we should keep it as `eks-cluster-name`

ConnorJC3: ^^^ seconded, I don't understand why the tag would have to be the same name

torredil: Updated to `ebs.csi.aws.com/cluster-name` in latest rev

pkg/driver/constants.go
```diff
@@ -181,6 +181,9 @@ const (
 	// the external provisioner sidecar is started with --extra-create-metadata=true and
 	// thus provides such metadata to the CSI driver.
 	PVNameTag = "kubernetes.io/created-for/pv/name"
+
+	// ClusterNameTagKey is the resource tag key for cluster-scoped IAM policies.
+	ClusterNameTagKey = "ebs.csi.aws.com/cluster-name"
```
ConnorJC3, line 186: We should also add this to the list of tags we don't let users mess with: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/39037f20efe888b06c90550349163a4945e935c1/pkg/driver/validation.go#L28

Otherwise lgtm

mdzraf (approved):
/lgtm
/approve

PR #2901, opened 2026-03-26
[DO NOT MERGE] Review for New Managed Policies

mdzraf:
Review for New Managed Policies

Set up: 
- EKS 1.35 cluster
- snapshot-controller installed via Addon
- EBS CSI Driver Custom image with tagging changes deployed by using make cluster/image and make cluster/install. 
- Auth: Verified that only intended IAM policy is being used by Pod Identity Association.

----- Test #1 ------

- Ran `make e2e/external` everything passed:
Ran 99 of 7671 Specs in 355.634 seconds
SUCCESS! -- 99 Passed | 0 Failed | 0 Pending | 7572 Skipped

Ginkgo ran 1 suite in 5m59.149713404s
Test Suite Passed

--- Test # 2 ----

---- Manual Tests Through Driver Happy Path Test Cases (Resource name tag with matching value of role cluster name tag value) ----

- Create volume ✅
- Attach volume ✅
- Copy Volume ✅
- Modify Volume ✅
- Detach volume ✅
- Delete volume ✅
- Create snapshot ✅
- Lock Snapshot ✅
- Create Volume from snapshot ✅
- Delete snapshot ✅

--- Test # 3 ----

---- Manual Negative Testing Through (Resource tag value does not match role cluster name tag value) ---

- Cannot attach Volume w/o Cluster Name Tag Value that Matches ✅
- Cannot detach Volume w/o Cluster Name Tag Value that Matches ✅
- Cannot create snapshot from volume w/o Cluster Name Tag Value that Matches ✅
- Cannot cannot create volume from snapshot w/o Cluster Name Tag Value that Matches ✅
- Cannot delete volume w/o Cluster Name Tag Value that Matches ✅
- Cannot modify volume w/o Cluster Name Tag Value that Matches ✅
- Cannot Copy a volume w/o Cluster Name Tag Value that Matches ✅

### DIFF
--- a/docs/AmazonEBSCSIDriverEKSClusterScopedPolicy.json
+++ b/docs/AmazonEBSCSIDriverEKSClusterScopedPolicy.json
@@ -1,0 +1,188 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "ReadOnlyDescribeOperations",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DescribeAvailabilityZones",
+        "ec2:DescribeInstances",
+        "ec2:DescribeInstanceTypes",
+        "ec2:DescribeSnapshots",
+        "ec2:DescribeVolumes",
+        "ec2:DescribeVolumesModifications",
+        "ec2:DescribeVolumeStatus"
+      ],
+      "Resource": "*"
+    },
+    {
+      "Sid": "CreateAndCopyVolumesWithClusterTag",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume",
+        "ec2:CopyVolumes"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:RequestTag/ebs.csi.aws.com/cluster-name": "${aws:PrincipalTag/eks-cluster-name}"
+        }
+      }
+    },
+    {
+      "Sid": "CopyClusterVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CopyVolumes"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/vol-*",
+      "Condition": {
+        "StringEquals": {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster-name": "${aws:PrincipalTag/eks-cluster-name}"
+        }
+      }
+    },
+    {
+      "Sid": "CreateSnapshotsWithClusterTag",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateSnapshot"
+      ],
+      "Resource": "arn:aws:ec2:*:*:snapshot/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:RequestTag/ebs.csi.aws.com/cluster-name": "${aws:PrincipalTag/eks-cluster-name}"
+        }
+      }
+    },
+    {
+      "Sid": "CreateSnapshotsFromClusterVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateSnapshot"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster-name": "${aws:PrincipalTag/eks-cluster-name}"
+        }
+      }
+    },
+    {
+      "Sid": "ManageClusterVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:ModifyVolume",
+        "ec2:AttachVolume",
+        "ec2:DetachVolume",
+        "ec2:DeleteVolume"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster-name": "${aws:PrincipalTag/eks-cluster-name}"
+        }
+      }
+    },
+    {
+      "Sid": "CreateVolumesFromAndEnableFSROnClusterSnapshots",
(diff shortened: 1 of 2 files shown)


docs/AmazonEBSCSIDriverEKSClusterScopedPolicy.json
```diff
@@ -0,0 +1,242 @@
...
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster-name": "${aws:PrincipalTag/eks-cluster-name}"
+        }
+      }
+    },
+    {
+      "Sid": "AttachDetachVolumesToClusterInstance",
```
torredil, line 102: We should add a second attach/detach statement that allows attaching to instances tagged with something the customer controls since non-EKS instances won't have this tag.

docs/cluster-scoped-policy.json
```diff
@@ -0,0 +1,242 @@
...
+    {
+      "Sid": "CreateVolumesWithClusterTag",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringLike": {
+          "aws:RequestTag/ebs.csi.aws.com/cluster": "true"
```
torredil: The policy does not need to enforce this since the cluster name condition is already strictly stronger. Same comment for CopyVolumesWithClusterTag and CreateSnapshotsWithClusterTag.

torredil (commented):
2 comments, otherwise lgtm.

1 np: suggest renaming the files:
- `cluster-scoped-policy.json` -> `AmazonEBSCSIDriverEKSClusterScopedPolicy.json`
- `v2-policy.json` -> `AmazonEBSCSIDriverPolicyV2.json`

docs/AmazonEBSCSIDriverEKSClusterScopedPolicy.json
mdzraf: Done, added AttachDetachVolumesToManuallyTaggedInstance ✅

docs/cluster-scoped-policy.json
mdzraf: Done, thank you! ✅

torredil (approved):

docs/AmazonEBSCSIDriverEKSClusterScopedPolicy.json
```diff
@@ -0,0 +1,247 @@
+{
```
ConnorJC3, line 1: High level comment: There's a lot of duplication in this policy and the other. For example, in this policy, couldn't `CreateVolumesWithClusterTag` and `CopyVolumesWithClusterTag` be merged into one? This is especially pertinent given we were facing problems with policy length iirc.

docs/AmazonEBSCSIDriverPolicyV2.json
```diff
@@ -0,0 +1,281 @@
...
+    },
+    {
+      "Sid": "CreateVolumesWithManagedTag",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringLike": {
```
ConnorJC3: Shouldn't this be `StringEquals`? Ditto basically everywhere else this specific condition exists.

ElijahQuinones (approved):
There are still some minor formatting issues between these polices and `AmazonEBSCSIDriverPolicy` but the content of the polices looks good to me

docs/AmazonEBSCSIDriverEKSClusterScopedPolicy.json
```diff
@@ -0,0 +1,247 @@
...
+    {
+      "Sid": "ReadOnlyDescribeOperations",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DescribeAvailabilityZones",
+        "ec2:DescribeInstances",
+        "ec2:DescribeSnapshots",
+        "ec2:DescribeVolumes",
+        "ec2:DescribeVolumesModifications",
+        "ec2:DescribeVolumeStatus"
```
ConnorJC3, line 14: Should we just add DescribeInstanceTypes here now?

ElijahQuinones: Saves us another approval round but we would also have to add it to the `AmazonEBSCSIDriverPolicy`

mdzraf: Besides what Elijah mentioned, I don't see why not, will the feature needing this be in on time for next release though?

ConnorJC3: Does it matter? Frankly I don't think we can reliably predict that at this point but would it change our decision?

mdzraf: I think it would because historically we merge on the day that the permission is added. Not sure if we would be allowed to add in this permissions that is not necessary for the driver to function at that specific point in time. That said, I don't have a strong opinion on this so we can just add it now and see if it gets approved. It'll save us from going through two approval rounds further down the line like Eli mentioned.

docs/AmazonEBSCSIDriverPolicyV2.json
mdzraf: ✅ Done in new rev

docs/AmazonEBSCSIDriverEKSClusterScopedPolicy.json
mdzraf: ✅ Done in new rev

ConnorJC3 (commented):
lgtm (intentionally not using command)

torredil (approved):

mdzraf:
New managed policies have been published.

PR #2907, opened 2026-03-31
Append driver mode to user agent string for metadata labeler container

torredil:
#### What type of PR is this?

/kind cleanup

The metadata labeler sidecar makes EC2 API calls via its own cloud client, but its user agent string is identical to the controller's. This appends the driver mode (metadataLabeler) to the user agent so EC2 API calls from the sidecar are
distinguishable.

This PR also refactors `--user-agent-extra` and `--aws-sdk-debug-log` CLI options into a shared flag block for all modes that create a cloud client, and types `MetadataLabelerMode` for consistency with the other mode constants.

Manually and unit testing:

```
make cluster/create

make cluster/image

HELM_EXTRA_FLAGS="--set=sidecars.metadataLabeler.enabled=true,node.metadataSources='metadata-labeler',sidecars.metadataLabeler.additionalArgs[0]='--aws-sdk-debug-log=true'" \
make cluster/install

kubectl logs ebs-csi-controller-b89cc4f4d-9z26r -n kube-system -c metadata-labeler | grep "User-Agent"
User-Agent: aws-sdk-go-v2/1.41.4 ua/2.1 os/linux lang/go#1.26.1 md/GOOS#linux md/GOARCH#amd64 exec-env/aws-ebs-csi-driver-v1.57.1-helm-metadataLabeler api/ec2#1.296.0 m/z
```

```release-note
Append driver mode to user agent string for metadata labeler container.
```

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -45,7 +45,7 @@ const (
 	AllMode Mode = "all"
 
 	// MetadataLabelerMode is the mode that starts the metadata labeler.
-	MetadataLabelerMode = "metadataLabeler"
+	MetadataLabelerMode Mode = "metadataLabeler"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/driver/driver.go
```diff
@@ -45,7 +45,7 @@ const (
 	AllMode Mode = "all"
 
 	// MetadataLabelerMode is the mode that starts the metadata labeler.
-	MetadataLabelerMode = "metadataLabeler"
+	MetadataLabelerMode Mode = "metadataLabeler"
```
mdzraf, line 48: typo?

mdzraf: Nvm, I see this is how they are all defined

torredil: Intentional change so that we don't have to cast everywhere else

mdzraf:
Got two lgtms

/approve

PR #2914, opened 2026-04-08
Set user-agent-extra to 'dev' when installed via Makefile

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Set user-agent-extra to `"dev"` when installed via the Makefile

N/A

```release-note
NONE
```


torredil (approved):
/lgtm
/approve

PR #2915, opened 2026-04-08
Pass volume limit from DescribeInstanceTypes via metadata-labeler

ConnorJC3:
#### What type of PR is this?

/kind feature

Instead of using our hardcoded table, in the `metadata-labeler` path we now call DIT at runtime and pass the result via node label. This enables users to immediately get updates from EC2 (e.g. new instance releases) without having to wait on a driver release.

Manually (Create a cluster with `metadata-labeler` enabled, observe labels are added) and with new unit tests and modification to the E2E test for metadata-labeler.

```release-note
Pass volume limits from DescribeInstanceTypes when using metadata-labeler
```

### DIFF
--- a/pkg/cloud/metadata/labels.go
+++ b/pkg/cloud/metadata/labels.go
@@ -26,6 +26,7 @@ import (
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/kubernetes-csi/csi-lib-utils/leaderelection"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud/limits"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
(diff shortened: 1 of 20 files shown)


pkg/cloud/metadata/labels.go
```diff
@@ -236,7 +247,7 @@ func updateMetadataEC2(ctx context.Context, kubeclient kubernetes.Interface, clo
 }
 
 // getMetadata calls the EC2 API to get the number of ENIs and non-CSI managed volumes attached to each node.
-func getMetadata(ctx context.Context, cloud cloud.Cloud, nodes *v1.NodeList, pvInformer cache.SharedIndexInformer) (map[string]enisVolumes, error) {
+func getMetadata(ctx context.Context, c cloud.Cloud, nodes *v1.NodeList, pvInformer cache.SharedIndexInformer) (map[string]enisVolumes, error) {
```
ConnorJC3, line 250: We needed to rename because the `cloud` variable was shadowing the `cloud` package, thus preventing access to `cloud.InstanceTypeInfo`

PR #2916, opened 2026-04-08
Call DescribeInstanceTypes at runtime to get multi-card info

ConnorJC3:

Call DIT at runtime to get the latest EBS multi-card info for an instance rather than using a built in hardcoded table.

Unit tests + E2E tests (modified to mock new API calls)

```release-note
Call DescribeInstanceTypes at runtime for EBS multi-card instances
```

### DIFF
--- a/docs/example-iam-policy.json
+++ b/docs/example-iam-policy.json
@@ -6,6 +6,7 @@
       "Action": [
         "ec2:DescribeAvailabilityZones",
         "ec2:DescribeInstances",
+        "ec2:DescribeInstanceTypes",
         "ec2:DescribeSnapshots",
         "ec2:DescribeTags",
         "ec2:DescribeVolumes",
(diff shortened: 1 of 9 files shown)


hack/generate-volume-limits-table/main.go
```diff
@@ -69,13 +68,6 @@ type volumeLimit struct {
 	maxAttachments int
 	attachmentType string
 }
-
-// ebsCardCounts contains instance types with MaximumEbsCards > 1
-var ebsCardCounts = map[string]int{
-{{- range $instanceType, $cards := .EbsCardCounts}}
-	"{{$instanceType}}": {{$cards}},
-{{- end}}
-}
```
torredil, line 78: I think we should consider phasing the table out. In the upcoming release, we can add the DIT call but keep the static table as a fallback,  if the API call fails we fall back to the table instead of defaulting to 1 and then in a subsequent release remove the static table. 

That way users on multi card instances get correct behavior regardless of whether their IAM is updated yet, and the runtime API call is purely an improvement.

torredil (commented):
Largely lgtm, one suggestion.

hack/generate-volume-limits-table/main.go
ConnorJC3: Fully agree, fixed in rev2 :+1:

torredil (approved):
/lgtm

PR #2919, opened 2026-04-16
Validate reserved tag keys on tag deletion path during volume modification

torredil:
/kind bug

Reserved tag keys are validated when adding tags but not when deleting them during volume modification. This adds the same validation to the deletion path.

Unit tests and e2e validation on a kOps cluster.

```release-note
Prevent deletion of reserved tag keys during volume modification.
```


ConnorJC3 (approved):
/lgtm

ConnorJC3:
/lgtm

if no negative comments in the next few hours I intend to merge

mdzraf (approved):
/lgtm 
/approve

PR #2921, opened 2026-04-17
Updating Docs For Policy Changes

mdzraf:
#### What type of PR is this?

/kind documentation

Updating documentations to point to `AmazonEBSCSIDriverPolicyV2` in place of `AmazonEBSCSIDriverPolicy`. 

#### Test

```
> make cluster/image && make cluster/install

[...cut for brevity] 
DESCRIPTION: Install complete
NOTES:

    kubectl get pod -n kube-system -l "app.kubernetes.io/name=aws-ebs-csi-driver,app.kubernetes.io/instance=aws-ebs-csi-driver"

Two new AWS-managed IAM policies are available: AmazonEBSCSIDriverPolicyV2 scopes the EBS CSI driver to volumes and snapshots explicitly tagged for use by the driver. AmazonEBSCSIDriverEKSClusterScopedPolicy scopes the driver to volumes and snapshots belonging to a specific EKS cluster. If interested in migrating, see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2918.
+ set +x
###
## Sucessfully installed driver via helm
#
```

```release-note
NONE
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/NOTES.txt
+++ b/charts/aws-ebs-csi-driver/templates/NOTES.txt
@@ -2,6 +2,4 @@ To verify that aws-ebs-csi-driver has started, run:
 
     kubectl get pod -n {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "aws-ebs-csi-driver.name" . }},app.kubernetes.io/instance={{ .Release.Name }}"
 
-The "a1CompatibilityDaemonSet" parameter has been removed. For more information see the EBS CSI Helm Chart changelog:
-https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/CHANGELOG.md#2550
-
+Two new AWS-managed IAM policies are available: AmazonEBSCSIDriverPolicyV2 scopes the EBS CSI driver to volumes and snapshots explicitly tagged for use by the driver. AmazonEBSCSIDriverEKSClusterScopedPolicy scopes the driver to volumes and snapshots belonging to a specific EKS cluster. If interested in migrating, see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2918.
--- a/docs/example-iam-policy.json
+++ b/docs/example-iam-policy.json
@@ -168,7 +168,8 @@
     {
       "Effect": "Allow",
       "Action": [
-        "ec2:DeleteSnapshot"
+        "ec2:DeleteSnapshot",
+        "ec2:LockSnapshot"
       ],
       "Resource": "arn:aws:ec2:*:*:snapshot/*",
       "Condition": {
@@ -180,7 +181,8 @@
     {
       "Effect": "Allow",
       "Action": [
-        "ec2:DeleteSnapshot"
+        "ec2:DeleteSnapshot",
+        "ec2:LockSnapshot"
       ],
       "Resource": "arn:aws:ec2:*:*:snapshot/*",
       "Condition": {
--- a/docs/install.md
+++ b/docs/install.md
@@ -54,9 +54,17 @@ To enable this metadata source:
 > [!NOTE]  
 > The example policy and documentation below use the [`aws` partition in ARNs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html). When installing the EBS CSI Driver on other partitions, replace instances of `arn:aws:` with the local partition, such as `arn:aws-us-gov:` for AWS GovCloud.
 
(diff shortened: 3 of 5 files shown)


docs/example-iam-policy.json
torredil, line 1: Now that we have several managed policies, I think we should consider either updating the docs in this repo to have `docs/AmazonEBSCSIDriverPolicy.json`, `docs/AmazonEBSCSIDriverPolicyV2.json`, etc or remove it altogether and provide a brief description + link directly to the source of truth for each, ie: https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEBSCSIDriverPolicy.html, somewhere like install.md

mdzraf: Sure, I think the best approach would be to remove it and just link directly + give brief description in install.md. Will do in next rev.

ConnorJC3: I have some vague recollection of a project (terraform?) hardcoding this specific filename of `example-iam-policy.json`. So my suggestion for the short term would be:

1. Keep this file as the V1 policy so we don't randomly break those people
2. Create two new files with the example policies or link to them (whichever we think is better)
3. (Eventually, not in this PR) Decide what we're going to do with `example-iam-policy.json` and announce it
4. Wait a few months and do whatever we decided in step 3

charts/aws-ebs-csi-driver/templates/NOTES.txt
```diff
@@ -2,6 +2,6 @@ To verify that aws-ebs-csi-driver has started, run:
 
     kubectl get pod -n {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "aws-ebs-csi-driver.name" . }},app.kubernetes.io/instance={{ .Release.Name }}"
 
-The "a1CompatibilityDaemonSet" parameter has been removed. For more information see the EBS CSI Helm Chart changelog:
-https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/CHANGELOG.md#2550
-
+Two new AWS-managed IAM policies are available: AmazonEBSCSIDriverPolicyV2 scopes the EBS CSI driver to volumes and 
+snapshots explicitly tagged for use by the driver. AmazonEBSCSIDriverEKSClusterScopedPolicy scopes the driver to volumes and 
+snapshots belonging to a specific EKS cluster. If interested in migrating, see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2918.
```
ConnorJC3: This wording looks good but we should put it on a single line so it wraps to the user's terminal.

docs/example-iam-policy.json
mdzraf: Since we have customers that may be hardcoding the file name (I didn't know this) are we good for this PR to add two new policy files in /docs, reference the V2 file in install.md, and also include direct links + the brief description of all policies  in install.md as well?

ConnorJC3: > add two new policy files in /docs

If we want, I don't have a strong opinion.

> reference the V2 file in install.md

Yes, I would keep those changes as-is.

> and also include direct links + the brief description of all policies in install.md as well?

Personally I think this is better suited for an `iam.md` or something like that rather than further polluting `install.md`. The majority of self-provisioning customers are not likely to benefit from the non-V2 policies and I'd rather keep `install.md` shorter rather than longer.

My suggestion for this PR would be to just skip this for now.

docs/install.md
```diff
@@ -54,9 +54,17 @@ To enable this metadata source:
...
 
-The driver requires IAM permissions to talk to Amazon EBS to manage the volume on user's behalf. [The example policy here](./example-iam-policy.json) defines these permissions. AWS maintains a [managed policy version of the example policy](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEBSCSIDriverPolicy.html), available at ARN `arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy`.
+The driver requires IAM permissions to talk to Amazon EBS to manage the volume on user's behalf. [The example policy here](./AmazonEBSCSIDriverPolicyV2.json) defines these permissions. AWS maintains a [managed policy version of the example policy](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEBSCSIDriverPolicyV2.html), available at ARN `arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicyV2`.
 
-The baseline example policy excludes permissions for some rarer and potentially dangerous usecases. For these usecases, additional statements are necessary:
+The baseline example policy scopes permissions to tagged resources. Review the following if this affects your workflow:
```
ConnorJC3: "tagged resources" makes no sense to the user. Say something like "volumes and snapshots created by the EBS CSI Driver".

torredil (approved):

PR #2922, opened 2026-04-17
Fix batcher silently dropping results on unbuffered channels

torredil:
#### What type of PR is this?

/kind bug

We've observed unit tests intermittently timing out after 10 minutes in CI, rarely, but for a long time. See most recent example here https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/2899/pull-aws-ebs-csi-driver-unit/2036476445512962048

Looking into this, I've concluded its a race condition in the batcher's [execute](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/987e845fd9856f77b8e69da88ab6fbccbb263b3a/pkg/batcher/batcher.go#L152) method. The current implementation delivers results using a non-blocking send with a default case that just drops the result if the caller hasn't reached `<-ch` yet:

```
klog.V(7).InfoS("execute: sending batch results", "batch", batch)
for _, task := range batch {
	r := resultsMap[task]
	   for _, ch := range pendingTasks[task] {
		select {
		case ch <- BatchResult[ResultType]{Result: r, Err: err}:
		default:
			klog.V(7).InfoS("execute: ignoring channel with no receiver")
		}
	}
}
```

we don't see this in real environments because of EC2 API latency + the configured  max delay latency for batching. In tests, we create batcher instances with 0 max delay and mocks which are instant, narrowing the timing gap just enough to see this behavior in a small number of runs.

I think the right thing to do is make the send signal in batcher's execute always blocking because its job is to deliver the response , the default case is currently breaking that contract. From my studies, making the send blocking matches how Go's own standard library handles result delivery (guarantee rather than silently dropping). non-blocking sends are more meant for things like metrics.

Reproduced and validated fix here: https://gist.github.com/torredil/4dbc42b819d2fcf1a098d62bd56ea947 and manually audited all call sites in cloud.go to sanity check that every caller reads.

```release-note
NONE
```

### DIFF
--- a/pkg/batcher/batcher.go
+++ b/pkg/batcher/batcher.go
@@ -165,11 +165,7 @@ func (b *Batcher[InputType, ResultType]) execute(pendingTasks map[InputType][]ch
 	for _, task := range batch {
 		r := resultsMap[task]
 		for _, ch := range pendingTasks[task] {
-			select {
-			case ch <- BatchResult[ResultType]{Result: r, Err: err}:
-			default:
-				klog.V(7).InfoS("execute: ignoring channel with no receiver")
-			}
+			ch <- BatchResult[ResultType]{Result: r, Err: err}
 		}
 	}
 	klog.V(7).InfoS("execute: finished execution", "batchSize", len(batch))


ElijahQuinones (approved):
/lgtm

mdzraf (approved):
/lgtm 
/approve

PR #2929, opened 2026-05-05
DO NOT MERGE: ATTEMPTING TO CLEAN UP IMAGES IN ECR

mdzraf:
```release-note
NONE
```


mdzraf:
Cleanup succeeded, we should probably find a longterm solution to this.

/close

PR #2931, opened 2026-05-06
Fix possible nil pointer deref on clones path

ElijahQuinones:
#### What type of PR is this?

/kind bug 
This change ensures the source volume exists before checking its properties to avoid a nil pointer deference when the source volume does not exist.
CI

```release-note
Fix possible nil pointer deref on clones path
```


torredil (approved):

mdzraf (approved):
/lgtm 
/approve

ConnorJC3 (commented):
/lgtm

PR #2932, opened 2026-05-07
Fix init order tag unreserved bug

ElijahQuinones:
#### What type of PR is this?

/kind bug
This PR resolves an issue where the cluster tag as well as the IOPS tags where not being reserved properly.

This happens as `reservedTagKeys` a  denylist that was a package-level var map literal that captured `cloud.AwsEbsDriverTagKey` at package-init time, but it is itself a runtime-assigned var string that's only set inside cloud.NewCloud(). The map therefore captured "" as the value  instead of "ebs.csi.aws.com/cluster", and was never rebuilt. The same root cause left ebs.csi.aws.com/IOPSPerGb and ebs.csi.aws.com/AllowAutoIOPSIncreaseOnModify unprotected. The denylist for these three keys was effectively non-functional as such it has been converted to a function to get the runtime values to allow the reservations to work properly.
CI and unit tests

```release-note
Fix init order bug that allowed modifying tags reserved for use by the EBS CSI Driver
```

### DIFF
--- a/pkg/driver/validation.go
+++ b/pkg/driver/validation.go
@@ -22,18 +22,8 @@ import (
 	"strings"
 
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
+	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/util"
 	"k8s.io/klog/v2"
-)
-
-var (
-	reservedTagKeys = map[string]string{
-		cloud.VolumeNameTagKey:           "",
-		cloud.AwsEbsDriverTagKey:         "",
-		cloud.SnapshotNameTagKey:         "",
-		ClusterNameTagKey:                "",
-		IopsPerGBKey:                     "",
-		AllowAutoIOPSIncreaseOnModifyKey: "",
-	}
 )
 
 func ValidateDriverOptions(options *Options) error {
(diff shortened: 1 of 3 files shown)


pkg/driver/validation.go
```diff
@@ -54,12 +55,15 @@ func ValidateDriverOptions(options *Options) error {
...
-		if _, ok := reservedTagKeys[k]; ok {
+		if _, ok := reservedTagKeys()[k]; ok {
 			return fmt.Errorf("tag key '%s' is reserved", k)
 		}
 		if strings.HasPrefix(k, cloud.KubernetesTagKeyPrefix) {
 			return fmt.Errorf("tag key prefix '%s' is reserved", cloud.KubernetesTagKeyPrefix)
 		}
+		if strings.HasPrefix(k, util.GetDriverName()+"/") {
+			return fmt.Errorf("tag key prefix '%s/' is reserved", util.GetDriverName())
+		}
```
ConnorJC3, line 55: If we're adding this, do we even need the reserved key map?

pkg/driver/validation.go
```diff
@@ -22,19 +22,20 @@ import (
...
-var (
-	reservedTagKeys = map[string]string{
-		cloud.VolumeNameTagKey:           "",
-		cloud.AwsEbsDriverTagKey:         "",
-		cloud.SnapshotNameTagKey:         "",
-		ClusterNameTagKey:                "",
-		IopsPerGBKey:                     "",
-		AllowAutoIOPSIncreaseOnModifyKey: "",
+func reservedTagKeys() map[string]string {
+	return map[string]string{
```
ConnorJC3: If we're going to keep this, probably should use a `sync.Once` so we don't re-allocate a map every time

pkg/driver/validation_test.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 func TestValidateExtraTags(t *testing.T) {
+	const testDriverName = "test.ebs.csi.aws.com"
```
ConnorJC3: Why are we hardcoding this constant? And why are we not using the `cloud` version of the tag keys?

ConnorJC3 (approved):
/lgtm

mdzraf (approved):
/lgtm 
/approve

PR #2933, opened 2026-05-07
Add mutex to MetricRecorder

ElijahQuinones:
#### What type of PR is this?

/king bug
Added a mutex to the metrics recorder to prevent concurrent writing to metrics map
Added temp concurrency test 

```
 package metrics

  import (
        "fmt"
        "sync"
        "testing"

        "github.com/prometheus/client_golang/prometheus"
  )

  func TestRaceRepro(t *testing.T) {
        m := &MetricRecorder{
                registry: prometheus.NewRegistry(),
                metrics:  make(map[string]any),
        }

        const goroutines = 16
        const iterations = 100
        var wg sync.WaitGroup
        wg.Add(goroutines)
        for i := range goroutines {
                go func(id int) {
                        defer wg.Done()
                        for j := range iterations {
                                name := fmt.Sprintf("repro_counter_%d", j%4)
                                m.IncreaseCount(name, "help", map[string]string{"k": fmt.Sprintf("v%d", id)})
                                hist := fmt.Sprintf("repro_hist_%d", j%4)
                                m.ObserveHistogram(hist, "help", float64(j), map[string]string{"k": fmt.Sprintf("v%d", id)}, []float64{1, 5, 10})
                        }
                }(i)
        }
        wg.Wait()
  }
  
```

Confirmed this fails on master and passes with my fix.

```release-note
Add mutex to MetricRecorder to prevent crashes caused by concurrent access
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -56,6 +56,7 @@ var (
 
 type MetricRecorder struct {
 	registry        *prometheus.Registry
+	mu              sync.RWMutex
 	metrics         map[string]any
 	asyncEC2Metrics *AsyncEC2Collector
 }
@@ -128,7 +129,9 @@ func (m *MetricRecorder) IncreaseCount(name string, helpText string, labels map[
 		return // recorder is not initialized
 	}
 
+	m.mu.RLock()
 	metric, ok := m.metrics[name]
+	m.mu.RUnlock()
 
 	if !ok {
 		klog.V(4).InfoS("Metric not found, registering", "name", name, "labels", labels)
@@ -150,7 +153,10 @@ func (m *MetricRecorder) ObserveHistogram(name string, helpText string, value fl
 	if m == nil {
 		return // recorder is not initialized
 	}
+
+	m.mu.RLock()
 	metric, ok := m.metrics[name]
+	m.mu.RUnlock()
 
 	if !ok {
 		klog.V(4).InfoS("Metric not found, registering", "name", name, "labels", labels, "buckets", buckets)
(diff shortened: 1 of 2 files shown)


mdzraf (approved):
/lgtm

ConnorJC3 (commented):
I believe we also need to hold the lock in `initializeMetricWithOperations`. Otherwise lgtm

pkg/metrics/metrics.go
```diff
@@ -56,6 +56,7 @@ var (
 
 type MetricRecorder struct {
 	registry        *prometheus.Registry
+	mu              sync.RWMutex
```
torredil, line 59: Thoughts on using a [sync.Map](https://pkg.go.dev/sync#Map) instead?  its optimized for "when the entry for a given key is only ever written once but read many times", which fits well here, and would also naturally cover the unlocked read in initializeMetricWithOperations that @ConnorJC3 pointed out.

pkg/metrics/metrics.go
```diff
@@ -56,6 +56,7 @@ var (
 
 type MetricRecorder struct {
```
torredil, line 57: Suggest adding a `TestMetricRecorderConcurrentAccess` unit test in the metrics package for coverage.

torredil (commented):
2 non-blocking comments, fix lgtm, thanks!

pkg/metrics/metrics.go
ConnorJC3: `sync.Map` is not suitable here as we also need to guard against concurrent access to `registry` (as far as I can tell, the Prometheus interface does not require thread safety).

torredil: prometheus registry has its own mutex so concurrent access is fine. sync.Map works here with LoadOrStore, either way the mutex here is perfectly fine, I'm happy to leave as is.

torredil (approved):

PR #2940, opened 2026-05-19
Migrate to runtime FIPS

ConnorJC3:
#### What type of PR is this?

/kind feature

Migrate to runtime-based FIPS, see https://go.dev/doc/security/fips140

In order to do this, we do the following:
- Build with `GOFIPS140=certified`. This ensures we get the latest FIPS library that is certified, regardless of Go version.
- The above auto-enables FIPS in the built images. To avoid this, we turn it back off by default with `GODEBUG=fips140=off` in the `Dockerfile`.
- Change the `fips` Helm parameter to enable the FIPS library via `GODEBUG=fips140=on` (instead of swapping out the image like we did previously).
- Updated the docs to document the above (also added a short section on how to make the relevant changes in Kustomize).

`make e2e/external-fips` and similar

```release-note
WIP
```

### DIFF
--- a/hack/prow-e2e.sh
+++ b/hack/prow-e2e.sh
@@ -37,7 +37,6 @@ test-e2e-external)
   ;;
 test-e2e-external-fips)
   TEST="external-fips"
-  export FIPS_TEST="true"
   ;;
 test-e2e-external-arm64)
   TEST="external"
@@ -79,7 +78,6 @@ test-e2e-external-eks-windows)
   ;;
 test-e2e-external-eks-windows-fips)
   TEST="external-windows-fips"
-  export FIPS_TEST="true"
   export CLUSTER_TYPE="eksctl"
   export WINDOWS="true"
   ;;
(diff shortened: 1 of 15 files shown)


hack/prow-e2e.sh
```diff
@@ -37,7 +37,6 @@ test-e2e-external)
   ;;
 test-e2e-external-fips)
   TEST="external-fips"
-  export FIPS_TEST="true"
```
ConnorJC3, line 40: This file's diff looks weird - it's becauase we no longer need to build a special image, all of the `fips`-related configuration is handled in the `Makefile` so `FIPS_TEST` is no longer needed at all.

torredil (approved):
/lgtm

mdzraf (approved):
/hold
/lgtm
/approve

Holding until WIP is removed

mdzraf (approved):
/lgtm
/approve

mdzraf (approved):
/lgtm

PR #2947, opened 2026-06-03
Bumping Dependencies Ahead of v1.61.0 release

mdzraf:
#### What type of PR is this?

Bumping Dependencies Ahead of v1.61.0 release

```
make update
make test
make verify
make cluster/image
```

```release-note
NONE
```


deploy/kubernetes/base/clusterrolebinding-resizer.yaml
```diff
@@ -13,3 +13,4 @@ roleRef:
   kind: ClusterRole
   name: ebs-external-resizer-role
   apiGroup: rbac.authorization.k8s.io
+
```
ElijahQuinones, line 16: Where these spaces added by the make file targets ?

hack/tools/install.sh
```diff
@@ -19,25 +19,25 @@ set -euo pipefail
...
 # https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck
 GOVULNCHECK_VERSION="v1.3.0"
 # https://github.com/helm/helm
-HELM_VERSION="v4.1.4"
+HELM_VERSION="v4.2.0"
 # https://github.com/kubernetes/kops
 # Commit is preferred over version if non-empty, and can
 # be used to test new Kubernetes releases earlier
-KOPS_VERSION="v1.35.0"
+KOPS_VERSION="v1.35.1"
```
ElijahQuinones, line 40: Should we bump these to 1.36 ?

deploy/kubernetes/base/clusterrolebinding-resizer.yaml
mdzraf: Yes

hack/tools/install.sh
mdzraf: I left as is since they only have v1.36.0-alpha.1 right now given that we are doing a CVE fix, I did not want to hang up release on any issues that may come up when we upgrade to a the new K8s version

ElijahQuinones (approved):
/lgtm

PR #2954, opened 2026-06-15
Fixing warn-on-invalid-tag validation bug

mdzraf:
#### What type of PR is this?

/kind bug

The `validateExtraTags` function in `pkg/driver/validation.go` logs the tag violations and returns nil when `warnOnly=true`, but does not remove the invalid keys from the tag map. This allows for overriding values for driver-managed tags. This PR fixes this.

Added unit test to cover specific case.

```release-note
Disallow overriding driver-reserved tags when `--warn-on-invalid-tag` is `true`.
```


ConnorJC3:
I'm not a huge fan of the function signature or the test, but the code does work so in service of not blocking the release:

/lgtm

I will likely submit an additional PR of my own later to add extra tests, especially for the `warn=false` case

Issue #2828, opened 2026-01-14
Push helm chart to OCI registry


sylr:
Hi,

It would be nice if this project could publish its helm chart on a/several OCI registries such as registry.k8s.io/gchr.io ... etc.

This would allow users to use services like AWS ECR as pull through cache for fetching the chart and not be impacted when GitHub is down which happens on a regular basis.

Thank you.

ConnorJC3:
/kind feature

Hi - thanks for the request! I can't promise anything firm at this time but I'm going to bring it to the team for consideration.

PR #2957, opened 2026-06-23
Push Helm chart artifact to staging OCI registry on release

torredil:
#### What type of PR is this?

/kind feature

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2828

The driver image is published to both `registry.k8s.io` and `public.ecr.aws`, but the Helm chart still
ships only through github, which ties it to github availability and means users can't front the chart with ECR pull-through cache. This PR gets the chart into staging so that we can promote it to both GAR and ECR.

Triggered `aws-ebs-csi-driver-push-images` on a scratch `release-*` branch (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/release-helm-oci-smoke2):

```
Successfully packaged chart and saved it to: artifacts/aws-ebs-csi-driver-2.62.0.tgz
Pushed: gcr.io/k8s-staging-provider-aws/charts/aws-ebs-csi-driver:2.62.0
Digest: sha256:1b6c5de27d465b6b5ca11b33585ae7b21af81e892651c94bf7345254a9b27a31
```

See build logs: https://storage.googleapis.com/kubernetes-ci-logs/logs/aws-ebs-csi-driver-push-images/2069444720483897344/artifacts/build.log

```release-note
NONE
```

#### Note for reviewers

This is the first step in the release job to install a tarball based tool inside the build image, and that image (`gcb-docker-gcloud`) is alpine. BusyBox doesn't support GNU tar's `--transform`, hence the diff in install.sh - I've switched it to extract via stdout which is portable across both and produces a byte identical binary on GNU tar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -305,6 +305,14 @@ create-manifest: all-image-registry
 push-manifest: create-manifest
 	docker manifest push --purge $(IMAGE):$(TAG)
 
+## Helm chart
+# Package the Helm chart and push it to the staging OCI registry.
+# Invoked by hack/cloudbuild.sh on helm-chart tags.
+
+.PHONY: helm-chart-push
+helm-chart-push: bin/helm
+	./hack/helm-chart-package.sh
+
 ## Tools
 # Tools necessary to perform other targets
 
--- a/hack/cloudbuild.sh
+++ b/hack/cloudbuild.sh
@@ -44,3 +44,10 @@ export IMAGE=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver
 export TAG=${GIT_TAG:10}
 export VERSION=$PULL_BASE_REF
 IMAGE=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver make -j $(nproc) all-push
+
+# Push the Helm chart to the staging OCI registry when triggered by the
+# helm-chart tag that chart-releaser creates (e.g. helm-chart-aws-ebs-csi-driver-2.62.0).
+if [[ "${PULL_BASE_REF}" =~ ^helm-chart- ]]; then
+  loudecho "Push helm chart to staging GCR (${PULL_BASE_REF})"
+  make helm-chart-push
+fi
--- a/hack/helm-chart-package.sh
+++ b/hack/helm-chart-package.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Packages the Helm chart and pushes it to the staging OCI registry. Invoked by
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -305,6 +305,16 @@ create-manifest: all-image-registry
 push-manifest: create-manifest
 	docker manifest push --purge $(IMAGE):$(TAG)
 
+## Helm chart
+# Package the Helm chart and push it to the staging OCI registry.
+# Invoked by hack/cloudbuild.sh on release branches.
+
+HELM_CHART_REPO ?= gcr.io/k8s-staging-provider-aws/charts
```
ConnorJC3: We're setting this default twice, can we either drop it here (preferred) or drop it in `helm-chart-package.sh`?

hack/helm-chart-package.sh
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+
+# Packages the Helm chart and pushes it to the staging OCI registry. Invoked by
+# the `helm-chart-push` Makefile target, which runs from hack/cloudbuild.sh on
+# release branches.
+
+set -euo pipefail
+
+HELM=${HELM:-./bin/helm}
+YQ=${YQ:-./bin/yq}
```
ConnorJC3: np: can we do this based on the script's path like we do in other scripts: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b041945c196b5254e75478f88231bb168ead391e/hack/update-mockgen.sh#L19

This is both more consistent with the existing code, and more robust as it will still work if the script is run when the working directory is not the root of the repo.

hack/cloudbuild.sh
```diff
@@ -44,3 +44,8 @@ export IMAGE=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver
 export TAG=${GIT_TAG:10}
 export VERSION=$PULL_BASE_REF
 IMAGE=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver make -j $(nproc) all-push
+
+if [[ "${PULL_BASE_REF}" =~ ^release- ]]; then
```
ConnorJC3: If we do this, that means our staging charts get overwritten every time we push to a release branch, and it introduces a race during releases. Would it make more sense to instead base it on the `helm-chart` tags the existing chart releaser uses (that way we guarantee we're always getting a released version of a chart rather than an intermediate release branch).

(That change would probably also mean we could drop the yq dependency)

hack/cloudbuild.sh
```diff
@@ -44,3 +44,8 @@ export IMAGE=gcr.io/k8s-staging-provider-aws/aws-ebs-csi-driver
...
+  loudecho "Push helm chart to staging GAR (release branch ${PULL_BASE_REF})"
```
ConnorJC3: np: GAR? Meant to say GCR maybe?

ConnorJC3 (commented):
A few nitpicks, most important comment is about the `PULL_BASE_REF`. Overall looks very good

ElijahQuinones (approved):
/lgtm

hack/cloudbuild.sh
torredil: Good idea, this requires a config change in test-infra so that the post submit fires when the chart releaser action create the chart tag: https://github.com/kubernetes/test-infra/pull/37332

Makefile
torredil: Done.

hack/helm-chart-package.sh
torredil: Done.

ConnorJC3 (approved):
/lgtm

LGTM, and I intend to approve if there isn't any blocking feedback soon so we can have this engaged for the next release

mdzraf (approved):
/lgtm 
/approve

PR #2964, opened 2026-07-16
Unifying single/multi AZ tests into single suite.

mdzraf:
#### What type of PR is this?

/kind cleanup

This PR unifies what were previously single-az and multi-az test targets into one. 

This is PR # 1 out of three needed for this unification. In this PR we hijack 'single-az' target to actually run the functional target, keep 'multi-az' target as a no-op, and add in the new functional target. PR # 2 will be a 'test-infra' PR where we add in our configurations to run our new test and remove the single-az/multi-az targets. And lastly, PR # 3 will be a driver PR where we remove the 'single-az' and 'multi-az' targets and just leave the 'functional'.

- Created a kops cluster with 4 worker nodes (ensuring two of them are in the same AZ)
- Ran 'make e2e/functional' and 'make e2e/single-az' to test both the actual target and the hijack.
- Got test succeeded: 
```
Will run 90 of 100 specs
Running in parallel across 5 processes
SS••••••••••••••••••••••••••••••S••S••••••••••••••••••••••••••••••••••••••••••••••••••••SSSSSSSS••••
Ran 88 of 100 Specs in 797.618 seconds
SUCCESS! -- 88 Passed | 0 Failed | 0 Pending | 12 Skipped
```
- Results of CI run: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_aws-ebs-csi-driver/2964/pull-aws-ebs-csi-driver-e2e-single-az/2077808960680759296

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,21 +140,22 @@ e2e/parameters: bin/helm bin/ginkgo
 e2e/parameters-all: bin/helm bin/ginkgo test/helm-template
 	./hack/e2e/param-sets.sh run-all
 
-.PHONY: e2e/single-az
-e2e/single-az: bin/helm bin/ginkgo
-	AWS_AVAILABILITY_ZONES=us-west-2a \
+.PHONY: e2e/functional
+e2e/functional: bin/helm bin/ginkgo
 	TEST_PATH=./tests/e2e/... \
-	GINKGO_FOCUS="\[ebs-csi-e2e\] \[single-az\]" \
+	GINKGO_FOCUS="\[ebs-csi-e2e\] \[functional\]" \
 	GINKGO_PARALLEL=5 \
 	HELM_EXTRA_FLAGS="--set=controller.volumeModificationFeature.enabled=true,sidecars.provisioner.additionalArgs[0]='--feature-gates=VolumeAttributesClass=true',sidecars.resizer.additionalArgs[0]='--feature-gates=VolumeAttributesClass=true',node.enableMetrics=true" \
 	./hack/e2e/run.sh
 
+# Temporary aliases kept until test-infra migrates to e2e/functional, then
+# removed. single-az runs the merged suite; multi-az is a no-op (now included).
+.PHONY: e2e/single-az
+e2e/single-az: e2e/functional
+
 .PHONY: e2e/multi-az
-e2e/multi-az: bin/helm bin/ginkgo
-	TEST_PATH=./tests/e2e/... \
-	GINKGO_FOCUS="\[ebs-csi-e2e\] \[multi-az\]" \
-	GINKGO_PARALLEL=5 \
-	./hack/e2e/run.sh
+e2e/multi-az:
+	@echo "e2e/multi-az is a no-op: its tests are now part of e2e/functional."
 
 .PHONY: e2e/disruptive
 e2e/disruptive: bin/helm bin/ginkgo
--- a/docs/makefile.md
+++ b/docs/makefile.md
@@ -59,7 +59,8 @@ Creates a cluster for running E2E tests against. There are many parameters that 
 - `AMI_FAMILY`: Which ami family to create a linux node group with for the cluster (`eksctl` clusters only) - defaults to `AmazonLinux2023`
 - `WINDOWS`: Whether or not to create a Windows node group for the cluster (`eksctl` clusters only) - defaults to `false`
 - `AWS_REGION`: Which region to create the cluster in - defaults to `us-west-2`
-- `AWS_AVAILABILITY_ZONES`: Which AZs to create nodes for the cluster in - defaults to `us-west-2a,us-west-2b,us-west-2c`
+- `AWS_AVAILABILITY_ZONES`: Which AZs to create nodes for the cluster in - defaults to the first `NUM_AZS` non-opt-in AZs in the region
+- `NUM_AZS`: How many AZs to auto-detect when `AWS_AVAILABILITY_ZONES` is unset - defaults to `2`
 - `OUTPOST_ARN`: If set, create an additional nodegroup on an [outpost](https://aws.amazon.com/outposts/) (`eksctl clusters only)
 - `OUTPOST_INSTANCE_TYPE`: The instance type to use for the outpost nodegroup (only used when `OUTPOST_ARN` is non-empty) - defaults to `INSTANCE_TYPE`
 
@@ -201,13 +202,9 @@ Alternatively, you may run on an externally created cluster by passing `CLUSTER_
 
 Run the Kubernetes upstream [external storage E2E tests](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/README.md). This is the most comprehensive E2E test, recommended for local development.
 
-### `make e2e/single-az`
+### `make e2e/functional`
 
-Run the single-AZ EBS CSI E2E tests. Requires a cluster with only one Availability Zone.
-
-### `make e2e/multi-az`
-
-Run the multi-AZ EBS CSI E2E tests. Requires a cluster with at least two Availability Zones.
+Run the EBS CSI functional E2E tests (the `[functional]` labeled specs) against a single multi-AZ cluster. `AWS_AVAILABILITY_ZONES` is derived automatically from the cluster's worker nodes. The multi-attach specs pin their volume to an AZ with at least two schedulable worker nodes and place two pods on different nodes sharing it, so the cluster must have at least two nodes in one AZ (the default `make cluster/create` satisfies this).
 
 ### `make e2e/external-windows`
 
--- a/hack/e2e/config.sh
+++ b/hack/e2e/config.sh
@@ -28,8 +28,9 @@ KUBECONFIG=${KUBECONFIG:-"${TEST_DIR}/${CLUSTER_NAME}.${CLUSTER_TYPE}.kubeconfig
 # Use AWS_REGION as priority, fallback to region from awscli config, fallback to us-west-2
 REGION_FROM_CONFIG="$(${BIN}/aws configure get region || echo '')"
 export AWS_REGION=${AWS_REGION:-${REGION_FROM_CONFIG:-us-west-2}}
-# If zones are not provided, auto-detect the first 3 AZs that are not opt in
-ZONES=${AWS_AVAILABILITY_ZONES:-$(${BIN}/aws ec2 describe-availability-zones --region "${AWS_REGION}" | jq -r '[.AvailabilityZones[] | select(.OptInStatus == "opt-in-not-required") | .ZoneName][:3] | join(",")')}
+# If zones are not provided, auto-detect the first NUM_AZS AZs that are not opt in.
+NUM_AZS=${NUM_AZS:-2}
+ZONES=${AWS_AVAILABILITY_ZONES:-$(${BIN}/aws ec2 describe-availability-zones --region "${AWS_REGION}" | jq -r "[.AvailabilityZones[] | select(.OptInStatus == \"opt-in-not-required\") | .ZoneName][:${NUM_AZS}] | join(\",\")")}
 FIRST_ZONE=$(echo "${ZONES}" | cut -d, -f1)
 NODE_COUNT=${NODE_COUNT:-3}
 INSTANCE_TYPE=${INSTANCE_TYPE:-c5.large}
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -41,20 +41,20 @@ else
   exit 1
 fi
 
-# Fail single-az tests early if we know cluster is multi-az.
-IGNORE_SINGLE_AZ_ERR=${IGNORE_SINGLE_AZ_ERR:="false"}
-if [[ $IGNORE_SINGLE_AZ_ERR != "true" && "$GINKGO_FOCUS" =~ "single-az" ]]; then
-  # Get unique AZs of non-control-plane nodes
-  azs=$(kubectl get nodes \
+# [env]-gated tests (pre-provisioned, topology-aware) read AWS_AVAILABILITY_ZONES
+# to target real AZs. When unset, derive it from the live worker nodes.
+if [[ -z "${AWS_AVAILABILITY_ZONES:-}" ]]; then
+  AWS_AVAILABILITY_ZONES=$(kubectl get nodes \
     --kubeconfig "${KUBECONFIG}" \
     --selector '!node-role.kubernetes.io/control-plane' \
-    -o jsonpath='{.items[*].metadata.labels.topology\.kubernetes\.io/zone}' | tr " " "\n" | sort -u)
-
-  # Check if there's exactly one AZ and it matches $AWS_AVAILABILITY_ZONES
-  if [[ $(echo "$azs" | wc -w) -gt 1 ]] || [[ "$azs" != "$AWS_AVAILABILITY_ZONES" ]]; then
-    loudecho "ERROR. single-az tests require all worker nodes to be in a single availability zone (AZ) that matches env var \$AWS_AVAILABILITY_ZONES (Currently set as \"$AWS_AVAILABILITY_ZONES\"). Please delete nodes in other AZs. If you want to bypass this error, set env var IGNORE_SINGLE_AZ_ERR='true'"
+    --field-selector 'spec.unschedulable=false' \
+    -o jsonpath='{.items[*].metadata.labels.topology\.kubernetes\.io/zone}' | tr " " "\n" | sort -u | paste -sd, -)
+  if [[ -z "${AWS_AVAILABILITY_ZONES}" ]]; then
+    loudecho "ERROR. Could not derive AWS_AVAILABILITY_ZONES from cluster worker nodes. Set it explicitly or ensure worker nodes are labeled with topology.kubernetes.io/zone."
     exit 1
   fi
+  export AWS_AVAILABILITY_ZONES
+  loudecho "Derived AWS_AVAILABILITY_ZONES from cluster worker nodes: ${AWS_AVAILABILITY_ZONES}"
 fi
 
 if [[ "$WINDOWS" == true ]]; then
--- a/hack/prow-e2e.sh
+++ b/hack/prow-e2e.sh
@@ -22,12 +22,14 @@
 make bin/aws
 
 case ${1} in
-test-e2e-single-az)
-  TEST="single-az"
-  export AWS_AVAILABILITY_ZONES="us-west-2a"
+test-e2e-functional | test-e2e-single-az)
+  # test-e2e-single-az is a temporary alias that runs the merged [functional]
+  # suite until the test-infra jobs migrate to test-e2e-functional.
+  TEST="functional"
   ;;
 test-e2e-multi-az)
-  TEST="multi-az"
+  echo "test-e2e-multi-az is a no-op: its tests are now part of test-e2e-functional."
+  exit 0
   ;;
 test-e2e-disruptive)
   TEST="disruptive"
--- a/tests/e2e/driver/driver.go
+++ b/tests/e2e/driver/driver.go
@@ -41,7 +41,7 @@ type DynamicPVTestDriver interface {
 // PreProvisionedVolumeTestDriver represents an interface for a CSI driver that supports pre-provisioned volume.
 type PreProvisionedVolumeTestDriver interface {
 	// GetPersistentVolume returns a PersistentVolume with pre-provisioned volumeHandle
-	GetPersistentVolume(volumeID string, fsType string, size string, reclaimPolicy *v1.PersistentVolumeReclaimPolicy, namespace string, accessMode v1.PersistentVolumeAccessMode, volumeMode v1.PersistentVolumeMode) *v1.PersistentVolume
+	GetPersistentVolume(volumeID string, fsType string, size string, reclaimPolicy *v1.PersistentVolumeReclaimPolicy, namespace string, accessMode v1.PersistentVolumeAccessMode, volumeMode v1.PersistentVolumeMode, availabilityZone string) *v1.PersistentVolume
 }
 
 type VolumeSnapshotTestDriver interface {
--- a/tests/e2e/driver/ebs_csi_driver.go
+++ b/tests/e2e/driver/ebs_csi_driver.go
@@ -68,7 +68,7 @@ func (d *ebsCSIDriver) GetVolumeSnapshotClass(namespace string, parameters map[s
 	return getVolumeSnapshotClass(generateName, provisioner, parameters)
 }
 
-func (d *ebsCSIDriver) GetPersistentVolume(volumeID string, fsType string, size string, reclaimPolicy *v1.PersistentVolumeReclaimPolicy, namespace string, accessMode v1.PersistentVolumeAccessMode, volumeMode v1.PersistentVolumeMode) *v1.PersistentVolume {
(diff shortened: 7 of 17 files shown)


mdzraf:
/retest 

Looks like a flake

mdzraf:
/retest

One last time

docs/makefile.md
```diff
@@ -201,13 +201,9 @@ Alternatively, you may run on an externally created cluster by passing `CLUSTER_
 
 Run the Kubernetes upstream [external storage E2E tests](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/README.md). This is the most comprehensive E2E test, recommended for local development.
 
-### `make e2e/single-az`
```
ElijahQuinones, line 204: This will break prow and the internal tests as it will be looking for single-az make file target can we just make a new target for the combined tests ? Once that is stable we can move off on the separated tests?

ElijahQuinones (dismissed):

docs/makefile.md
mdzraf: Right, that is why we are breaking this into three PRs, see the [description ](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2964#issue-4904510314)

mdzraf: Or are you saying that we do not remove the single-az and multi-az tests for now?

ElijahQuinones: In my opinion we should 

1. Release PR with new makefile target of combined tests  
2. Update prow to enable this test without reporting (as we do with all new tests )
3. once it is stable cut a CR to have the combined tests report to GitHub and remove the single and multi az tests from prow.
4. If we feel like we want to, cut out the single and multi az tests from our makefile.

But if we have a reason to not do so Id be open to discussing.

mdzraf: Sure, not opposed to being extra careful and have these run as optional for a bit. @torredil & @ConnorJC3 any thoughts?

hack/e2e/config.sh
```diff
@@ -31,7 +31,7 @@ export AWS_REGION=${AWS_REGION:-${REGION_FROM_CONFIG:-us-west-2}}
 # If zones are not provided, auto-detect the first 3 AZs that are not opt in
 ZONES=${AWS_AVAILABILITY_ZONES:-$(${BIN}/aws ec2 describe-availability-zones --region "${AWS_REGION}" | jq -r '[.AvailabilityZones[] | select(.OptInStatus == "opt-in-not-required") | .ZoneName][:3] | join(",")')}
 FIRST_ZONE=$(echo "${ZONES}" | cut -d, -f1)
-NODE_COUNT=${NODE_COUNT:-3}
+NODE_COUNT=${NODE_COUNT:-4}
```
ConnorJC3: Why? Surely the literal one extra test from the `multi-az` test suite doesn't necessitate an entire extra node?

tests/e2e/driver/driver.go
```diff
@@ -63,7 +66,14 @@ func getStorageClass(
...
 	if bindingMode == nil {
-		defaultBindingMode := storagev1.VolumeBindingImmediate
+		// Default to WaitForFirstConsumer so that dynamically provisioned
+		// volumes are created in the same availability zone as the pod that
+		// consumes them. This lets the suite run against a multi-AZ cluster:
+		// with Immediate binding a volume can be provisioned in a different AZ
+		// than its pod, leaving the pod unschedulable. Tests that specifically
+		// exercise Immediate binding (e.g. the reclaim-policy tests, which
+		// provision a PV without a consuming pod) set it explicitly.
+		defaultBindingMode := storagev1.VolumeBindingWaitForFirstConsumer
```
ConnorJC3: Why are we changing the default binding mode? Isn't this a fairly substantial change in test behavior/coverage?

tests/e2e/driver/ebs_csi_driver.go
```diff
@@ -106,6 +106,30 @@ func (d *ebsCSIDriver) GetPersistentVolume(volumeID string, fsType string, size
 			VolumeMode: &volumeMode,
 		},
 	}
+
+	// Pin the PV to the AZ the pre-provisioned volume was created in. Without
+	// this, on a multi-AZ cluster the consuming pod may be scheduled in a
+	// different AZ than the EBS volume, so the volume can never attach and the
+	// pod hangs in ContainerCreating.
+	if availabilityZone != "" {
```
ConnorJC3: Is there a case where we expect `availabilityZone` to be an empty string? Is that even a valid input to this function?

hack/e2e/run.sh
```diff
@@ -41,20 +41,28 @@ else
...
-  if [[ $(echo "$azs" | wc -w) -gt 1 ]] || [[ "$azs" != "$AWS_AVAILABILITY_ZONES" ]]; then
-    loudecho "ERROR. single-az tests require all worker nodes to be in a single availability zone (AZ) that matches env var \$AWS_AVAILABILITY_ZONES (Currently set as \"$AWS_AVAILABILITY_ZONES\"). Please delete nodes in other AZs. If you want to bypass this error, set env var IGNORE_SINGLE_AZ_ERR='true'"
+    -o jsonpath='{.items[*].metadata.labels.topology\.kubernetes\.io/zone}' | tr " " "\n" | sort -u | paste -sd, -)
+  # Fail fast if the derivation produced nothing (e.g. nodes not yet labeled
+  # with topology.kubernetes.io/zone, wrong KUBECONFIG, or the selector
+  # excluded every node). Otherwise AWS_AVAILABILITY_ZONES would be exported
+  # empty and every [env]-gated test (pre-provisioned, topology-aware) would
+  # silently Skip, letting the suite report success without running them.
+  if [[ -z "${AWS_AVAILABILITY_ZONES}" ]]; then
+    loudecho "ERROR. Could not derive AWS_AVAILABILITY_ZONES from cluster worker nodes. Ensure the cluster has worker nodes labeled with topology.kubernetes.io/zone and that KUBECONFIG points at it, or set AWS_AVAILABILITY_ZONES explicitly."
```
ConnorJC3: What is the purpose of this diff? Don't we already set `AWS_AVAILABILITY_ZONES` in `config.sh`?

tests/e2e/testsuites/testsuites.go
```diff
@@ -400,28 +400,33 @@ func (t *TestPersistentVolumeClaim) ValidateProvisionedPersistentVolume() {
 				To(HaveLen(1))
 		}
 		if len(t.storageClass.AllowedTopologies) > 0 {
-			// Since we're chaging our topology key, assume we have the values below to compare:
```
ConnorJC3, line 403: Why are we changing this file at all? Maybe there's a good reason but it wasn't obvious to me, can someone dumb down why this change is necessary for me?

hack/prow-e2e.sh
```diff
@@ -22,12 +22,27 @@
...
+  # the test-infra jobs migrate to test-e2e-functional, then it will be removed.
+  TEST="functional"
+  # Multi-AZ cluster: leave AWS_AVAILABILITY_ZONES unset so config.sh
+  # auto-detects the first 3 AZs. The multi-attach specs need an AZ with at
+  # least two schedulable worker nodes (they place two pods on different nodes
+  # sharing one AZ-local volume); they detect such an AZ at runtime and fail if
+  # none exists. NODE_COUNT=4 across 3 AZs gives one AZ two nodes on the default
+  # kops path (kops.sh passes --node-count). NOTE: this only takes effect for
+  # kops; the eksctl path uses a fixed desiredCapacity in cluster.yaml and
+  # ignores NODE_COUNT.
```
ConnorJC3: Can we clean up some of these super long verbose comments? Also, this comment explains the purpose of the NODE_COUNT at 4 is to force a second node into the same AZ, but it doesn't explain why that is important.

docs/makefile.md
ConnorJC3: We're not running "new" tests. We should just get this PR to the point where we are comfortable it works and just merge it in if you want my opinion.

For example, every time we bugfix or add a new test case we don't turn the test suite back to non-reporting.

Makefile
```diff
@@ -140,21 +140,38 @@ e2e/parameters: bin/helm bin/ginkgo
...
+# and deliberately excludes the [param:...] specs, which are a separate suite
+# run via `make e2e/parameters` with per-parameter Helm configuration. run.sh
+# derives AWS_AVAILABILITY_ZONES from the cluster's worker nodes so
+# topology-dependent tests target real AZs.
+#
+# NOTE: the multi-attach specs place two pods on different nodes sharing one
+# AZ-local volume, so they need an AZ with at least two schedulable worker
+# nodes. They detect such an AZ at runtime and fail if none exists. On the
+# default kops path, NODE_COUNT=4 across 3 AZs gives one AZ two nodes; see
+# hack/e2e/config.sh and hack/prow-e2e.sh.
```
ConnorJC3: Ditto above feedback about unnecessarily long comments

hack/e2e/config.sh
mdzraf: We can keep it at 3 and just have to make sure we have at least two of those on the same AZ to ensure the multi-attach tests can still run. Will be adjust in next rev ✅

tests/e2e/driver/ebs_csi_driver.go
mdzraf: Removing check in next rev ✅

tests/e2e/driver/driver.go
mdzraf: This happened when I was trying to debug a completely unrelated issue (which was not fixed by this), will revert in next rev ✅

hack/prow-e2e.sh
mdzraf: > Can we clean up some of these super long verbose comments? 

Will fix next rev ✅ 

> Also, this comment explains the purpose of the NODE_COUNT at 4 is to force a second node into the same AZ, but it doesn't explain why that is important.

Next rev reverts the node count back to 3 and just forces two of those on the same AZ to make sure multi attach tests can run, new comment will explain that.

tests/e2e/testsuites/testsuites.go
mdzraf: The `[multi-az] [env] allowedTopologies` test was skipped because the old `e2e/multi-az` target never set `AWS_AVAILABILITY_ZONES`. But after unifying the tests, this test actually runs and was failing since this function was checking for the wrong key which lead to the assertion never passing.

Makefile
mdzraf: Fixed in next rev ✅

hack/e2e/run.sh
mdzraf: No it does not, it reads it into `ZONES (${AWS_AVAILABILITY_ZONES:-<auto-detect>})` just to pass zones at cluster creation. The `[env]` specs read A`WS_AVAILABILITY_ZONES` at runtime and the old `single-az` target set it explicitly in the make target, but the unified target intentionally leaves it unpinned, so `run.sh` gets it from the live worker node zone labels and exports it. The guard here is just to make sure those tests don't get silently skipped if for some reason it comes back empty when trying to get it from the worker nodes.

hack/e2e/config.sh
```diff
@@ -28,8 +28,9 @@ KUBECONFIG=${KUBECONFIG:-"${TEST_DIR}/${CLUSTER_NAME}.${CLUSTER_TYPE}.kubeconfig
 # Use AWS_REGION as priority, fallback to region from awscli config, fallback to us-west-2
 REGION_FROM_CONFIG="$(${BIN}/aws configure get region || echo '')"
 export AWS_REGION=${AWS_REGION:-${REGION_FROM_CONFIG:-us-west-2}}
-# If zones are not provided, auto-detect the first 3 AZs that are not opt in
-ZONES=${AWS_AVAILABILITY_ZONES:-$(${BIN}/aws ec2 describe-availability-zones --region "${AWS_REGION}" | jq -r '[.AvailabilityZones[] | select(.OptInStatus == "opt-in-not-required") | .ZoneName][:3] | join(",")')}
+# If zones are not provided, auto-detect the first NUM_AZS AZs that are not opt in
+NUM_AZS=${NUM_AZS:-3}
```
ElijahQuinones: Doesn't this default config  node count of 3 with num_azs 3 fail the multi attach test as no one AZ will have 2 nodes ? IE if I run make cluster/create and then run this test it will fail right ?

mdzraf: Yes a plain `make cluster/create && make e2e/functional` would fail multi-attach, but that is the same as prior to this change, if you just ran `make cluster/create && make e2e/single-az` without `export AWS_AVAILABILITY_ZONES=<xyz>` you would fail multi-attach tests. So I think that it is best to keep it as is with the unified test suite so that we don't change any cluster configuration on the default `make cluster/create` that we often use to run `e2e/external` and `e2e/external-windows` manually.

tests/e2e/dynamic_provisioning.go
```diff
@@ -830,6 +838,38 @@ var _ = Describe("[ebs-csi-e2e] [multi-az] Dynamic Provisioning", func() {
...
+// creation independently of this test.
+func multiAttachZone(cs clientset.Interface) string {
+	nodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		Fail(fmt.Sprintf("could not list nodes to find a multi-node AZ: %v", err))
+	}
+	nodesPerZone := map[string]int{}
+	for i := range nodes.Items {
+		node := &nodes.Items[i]
+		// Skip control-plane nodes; only workers run test pods.
```
AndrewSirenko: Should we also skip unschedulable nodes  here too (unless we do it elsewhere)? I know I've been burnt by this before when cordoning nodes in other AZs hoping to run single az tests.

```
if node.Spec.Unschedulable {
      continue
  }
```

hack/e2e/config.sh
ElijahQuinones: Yeah but that no longer needs to be true right ? Ideally as someone who runs manual testing (especially the case for external contributors). We shouldn't have to set random parameters to have the default set of tests pass.

tests/e2e/dynamic_provisioning.go
```diff
@@ -282,8 +289,9 @@ var _ = Describe("[ebs-csi-e2e] [single-az] Dynamic Provisioning", func() {
 							NameGenerate: "test-block-volume-",
 							DevicePath:   "/dev/xvda",
 						},
-						AccessMode:        v1.ReadWriteMany,
-						VolumeBindingMode: &volumeBindingMode,
+						AccessMode:            v1.ReadWriteMany,
+						VolumeBindingMode:     &volumeBindingMode,
+						AllowedTopologyValues: allowedTopologyValues,
```
ConnorJC3, line 294: The new `AllowedTopologyValues` needs to be mirrored to the below test "should fail to multi-attach dynamically provisioned IO2 block device - not enabled"

(right now it's essentially passing for the wrong reason)

tests/e2e/driver/ebs_csi_driver.go
```diff
@@ -106,6 +106,27 @@ func (d *ebsCSIDriver) GetPersistentVolume(volumeID string, fsType string, size
 			VolumeMode: &volumeMode,
 		},
 	}
+
+	// Pin the PV to the AZ the pre-provisioned volume was created in so the
+	// consuming pod schedules there; otherwise on a multi-AZ cluster the pod may
+	// land in a different AZ than the EBS volume and never attach.
+	pv.Spec.NodeAffinity = &v1.VolumeNodeAffinity{
```
ConnorJC3: Could we not just put this in the above object? Why the dance of assigning it to a new `pv` variable?

hack/e2e/config.sh
ConnorJC3: +1 why would we make the default configuration broken?

Makefile
```diff
@@ -140,21 +140,24 @@ e2e/parameters: bin/helm bin/ginkgo
 e2e/parameters-all: bin/helm bin/ginkgo test/helm-template
 	./hack/e2e/param-sets.sh run-all
 
-.PHONY: e2e/single-az
-e2e/single-az: bin/helm bin/ginkgo
-	AWS_AVAILABILITY_ZONES=us-west-2a \
+# e2e/functional runs the [functional] specs (the merged single-az + multi-az
+# suites) against a multi-AZ cluster.
```
ConnorJC3: Drop this comment

tests/e2e/driver/driver.go
```diff
@@ -40,8 +40,11 @@ type DynamicPVTestDriver interface {
 
 // PreProvisionedVolumeTestDriver represents an interface for a CSI driver that supports pre-provisioned volume.
 type PreProvisionedVolumeTestDriver interface {
-	// GetPersistentVolume returns a PersistentVolume with pre-provisioned volumeHandle
-	GetPersistentVolume(volumeID string, fsType string, size string, reclaimPolicy *v1.PersistentVolumeReclaimPolicy, namespace string, accessMode v1.PersistentVolumeAccessMode, volumeMode v1.PersistentVolumeMode) *v1.PersistentVolume
+	// GetPersistentVolume returns a PersistentVolume with pre-provisioned volumeHandle.
+	// availabilityZone is set as node affinity on the PV so the consuming pod is
+	// scheduled in the same AZ as the pre-provisioned volume (required on
+	// multi-AZ clusters).
```
ConnorJC3: Don't expand this comment

tests/e2e/pre_provsioning.go
```diff
@@ -241,11 +247,10 @@ var _ = Describe("[ebs-csi-e2e] [single-az] Pre-Provisioned with Multi-Attach",
 		ns = f.Namespace
 		ebsDriver = driver.InitEbsCSIDriver()
 
-		if os.Getenv(awsAvailabilityZonesEnv) == "" {
-			Skip(fmt.Sprintf("env %q not set", awsAvailabilityZonesEnv))
+		availabilityZone = multiAttachZone(cs)
+		if availabilityZone == "" {
+			Fail("no availability zone has two or more schedulable worker nodes; multi-attach requires a cluster with at least two nodes in one AZ")
```
ConnorJC3, line 252: If you're going to add a `Fail` here, you need to add a check for if `cloud == nil` in the `AfterEach`, because right now if this fail is hit it's going to panic because the test was never run and thus `cloud` is nil

hack/prow-e2e.sh
```diff
@@ -22,12 +22,20 @@
...
-test-e2e-single-az)
-  TEST="single-az"
-  export AWS_AVAILABILITY_ZONES="us-west-2a"
+test-e2e-functional | test-e2e-single-az)
+  # test-e2e-single-az is a temporary alias that runs the merged [functional]
+  # suite until the test-infra jobs migrate to test-e2e-functional.
+  TEST="functional"
+  # 2 AZs with 3 nodes gives a 2/1 split, so one AZ has the two nodes multi-attach
+  # needs (it runs two pods on different nodes sharing one AZ-local volume) while
+  # still exercising cross-AZ topology scheduling. Only honored on kops.
```
ConnorJC3: Pretty sure "Only honored on kops" is wrong - `NUM_AZS` always touchs the variable and thus will affect `eksctl`.

Honestly the comments in this file are way too verbose and need to be cut down - why are we documenting a `config.sh` option in `prow-e2e.sh`?

docs/makefile.md
```diff
@@ -201,13 +201,9 @@ Alternatively, you may run on an externally created cluster by passing `CLUSTER_
...
+### `make e2e/functional`
 
-Run the single-AZ EBS CSI E2E tests. Requires a cluster with only one Availability Zone.
-
-### `make e2e/multi-az`
-
-Run the multi-AZ EBS CSI E2E tests. Requires a cluster with at least two Availability Zones.
+Run the EBS CSI functional E2E tests (the `[functional]` labeled specs) against a single multi-AZ cluster. `AWS_AVAILABILITY_ZONES` is derived automatically from the cluster's worker nodes. The multi-attach specs pin their volume to the first AZ and place two pods on different nodes sharing it, so that AZ must contain at least two worker nodes (e.g. create the cluster with `NODE_COUNT=4` across 3 AZs).
```
ConnorJC3: "pin their volume to the first AZ" - go map order is random, so this is wrong

this should probably just say something like "pin their volume to an AZ with at least two nodes"

ConnorJC3: Also also, drop the mention of `NODE_COUNT=4` as we're not doing that anymore

docs/makefile.md
ConnorJC3, line 65: Can we add `NUM_AZS` here?

ConnorJC3 (commented):
I tried to point out instances of this issue where I saw it, but I'd suggest going over this PR and ensuring all the comments truly add value. Otherwise, see my comments below for more important stuff.

tests/e2e/dynamic_provisioning.go
ConnorJC3: This is my most important comment because the test is objectively broken right now, so if you fix anything make it this.

docs/makefile.md
mdzraf: Fixed in new rev ✅

hack/e2e/config.sh
mdzraf: Fixed in new rev ✅

hack/prow-e2e.sh
mdzraf: removed in new rev ✅

tests/e2e/driver/driver.go
mdzraf: Fixed in new rev ✅

tests/e2e/driver/ebs_csi_driver.go
mdzraf: Fixed in new rev ✅

tests/e2e/dynamic_provisioning.go
mdzraf: Fixed in new rev ✅

tests/e2e/dynamic_provisioning.go
mdzraf: Done in new rev ✅

tests/e2e/pre_provsioning.go
mdzraf: Fixed in new rev ✅

Makefile
mdzraf: Done in new rev ✅

docs/makefile.md
mdzraf: ✅  Done in new rev

hack/e2e/run.sh
```diff
@@ -41,20 +41,19 @@ else
...
-IGNORE_SINGLE_AZ_ERR=${IGNORE_SINGLE_AZ_ERR:="false"}
-if [[ $IGNORE_SINGLE_AZ_ERR != "true" && "$GINKGO_FOCUS" =~ "single-az" ]]; then
-  # Get unique AZs of non-control-plane nodes
-  azs=$(kubectl get nodes \
+# [env]-gated tests (pre-provisioned, topology-aware) read AWS_AVAILABILITY_ZONES
+# to target real AZs. When unset, derive it from the live worker nodes.
+if [[ -z "${AWS_AVAILABILITY_ZONES:-}" ]]; then
+  AWS_AVAILABILITY_ZONES=$(kubectl get nodes \
     --kubeconfig "${KUBECONFIG}" \
     --selector '!node-role.kubernetes.io/control-plane' \
```
ConnorJC3, line 49: np: can we add: `,!node.kubernetes.io/unschedulable` to this filter? Because otherwise the selector can select AZs in which all nodes are cordoned (e.g. cordoning windows nodes for linux test).

(I know this line isn't from your PR, but the changes to the multi-attach tests bring up this potential failure case)

PR otherwise lgtm

mdzraf: Fixed ✅

PR #2965, opened 2026-07-20
helm: add optional NetworkPolicy for controller and node

### LINKED ISSUE #2963: Helm chart: add network policy
I follow best practices ([enable global deny network policy ](https://docs.tigera.io/calico/3.29/network-policy/get-started/kubernetes-default-deny#create-a-global-default-deny-policy)) and force each application to explicitly define network policy.

`aws-ebs-csi-driver` chart has no built-in network policy and no `extraObjects`. I need to create my own chart just to add a single network policy.

/feature

Built-in (k8s native) Network Policy that I can enable.

Only 1 available: create my wrapper chart to add a network policy

**Additional context**
```
  - name: aws-ebs-csi-driver
    chart: aws-ebs-csi-driver/aws-ebs-csi-driver
    version: 2.60.0
    namespace: aws-ebs-csi-driver
```

https://docs.tigera.io/calico/3.29/network-policy/get-started/kubernetes-default-deny#create-a-global-default-deny-policy

mbmtbtbc:
#### What type of PR is this?
/kind feature

Adds optional, off-by-default NetworkPolicy resources for the controller and node components, gated by `controller.networkPolicy.enabled` / `node.networkPolicy.enabled`.

Users enforcing default-deny NetworkPolicy currently have no way to allow this driver's traffic without a wrapper chart. Default rules cover health checks, metrics, and IMDS/kube-apiserver/AWS API egress. Ingress/egress are extensible via `controller.networkPolicy.ingress`/`egress` and `node.networkPolicy.ingress`/`egress`, appended to the defaults.

Fixes #2963

- `helm lint` / `helm template` with both flags enabled
- `tests/helm-template` covers: default ports present, custom rules appended,
  metrics ports absent when `enableMetrics=false`, disabled by default, `additionalDaemonSets` support
- Not yet tested against live enforcement

```release-note
Added optional NetworkPolicy resources for controller and node, disabled by default.
Enable via `controller.networkPolicy.enabled` / `node.networkPolicy.enabled`.
```

### DIFF
--- a/charts/aws-ebs-csi-driver/templates/_networkpolicy_node.tpl
+++ b/charts/aws-ebs-csi-driver/templates/_networkpolicy_node.tpl
@@ -1,0 +1,75 @@
+{{- define "networkpolicy-node" }}
+{{- if .Values.node.networkPolicy.enabled }}
+{{- if .Values.node.hostNetwork }}
+{{- fail (printf "node.networkPolicy.enabled=true has no effect on %s because node.hostNetwork=true: NetworkPolicy does not apply to pods using the host network namespace. Set node.hostNetwork=false to use NetworkPolicy enforcement on the node DaemonSet." .NodeName) }}
+{{- end }}
+{{- if .Values.node.enableWindows }}
+{{- fail (printf "node.networkPolicy.enabled=true has no effect on the %s Windows DaemonSet because Windows node pods always run with hostNetwork=true (NetworkPolicy is not enforced against hostNetwork pods). Set node.enableWindows=false, or do not enable node.networkPolicy, until Windows NetworkPolicy support is added." .NodeName) }}
+{{- end }}
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ .NodeName }}
+  namespace: {{ .Values.node.namespaceOverride | default .Release.Namespace }}
+  labels:
+    {{- include "aws-ebs-csi-driver.labels" . | nindent 4 }}
+spec:
+  podSelector:
+    matchLabels:
+      app: {{ .NodeName }}
+      {{- include "aws-ebs-csi-driver.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    # Always allow kubelet health checks (ebs-plugin liveness-probe + node-driver-registrar),
+    # regardless of any additional rules below.
+    - ports:
+        - protocol: TCP
+          port: 9808
+        - protocol: TCP
+          port: {{ .Values.sidecars.nodeDriverRegistrar.healthPort }}
+    {{- if .Values.node.enableMetrics }}
+    # Always allow metrics scraping when enabled, regardless of any additional rules below.
+    - ports:
+        - protocol: TCP
+          port: 3302
+    {{- end }}
+    {{- with .Values.node.networkPolicy.ingress }}
+    # Additional user-supplied ingress rules, appended to the defaults above.
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  egress:
+    # Always allow DNS resolution, HTTPS (kube-apiserver, etc.), and IMDS,
+    # regardless of any additional rules below.
+    - ports:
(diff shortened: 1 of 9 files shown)


charts/aws-ebs-csi-driver/templates/_networkpolicy_node.tpl
```diff
@@ -0,0 +1,59 @@
+{{- define "networkpolicy-node" }}
+{{- if .Values.node.networkPolicy.enabled }}
```
YuryHrytsuk, line 2: Does it actually come into effect if `aws-ebs-csi-driver-node` pod runs with `hostNetwork: true`?

YuryHrytsuk: if not, it might be misleading

charts/aws-ebs-csi-driver/templates/networkpolicy-controller.yaml
```diff
@@ -0,0 +1,69 @@
...
+    # Always allow DNS resolution and HTTPS (EC2/EBS/KMS/STS API calls,
+    # kube-apiserver, etc.), regardless of any additional rules below.
+    - ports:
+        - protocol: UDP
+          port: 53
+        - protocol: TCP
+          port: 53
+    - ports:
+        - protocol: TCP
+          port: 443
```
YuryHrytsuk, line 56: what is a more reasonable default `443` or `6443`?

YuryHrytsuk:
Thanks! :pray:

charts/aws-ebs-csi-driver/templates/_networkpolicy_node.tpl
mbmtbtbc: thanks for catching this. 

i verified that NetworkPolicies dont apply to hostNetwork pods since they lack cluster pod IPs. The template currently renders successfully as a silent no-op

i will add a fail check to block template rendering with a clear error if hostNetwork=true and networkPolicy.enabled=true are set together (covering both the default DaemonSet and additionalDaemonSets)
also adding a test case for this failure.

charts/aws-ebs-csi-driver/templates/networkpolicy-controller.yaml
mbmtbtbc: i chose 443 because it matches the kubernetes.default.svc Service port that virtually all pods actually dial for apiserver access, and it also covers AWS API traffic (EC2/KMS/STS etc) and it also matches this project's own e2e reference clusters (as kOps sets --secure-port=443 by convention)

but yes this driver isn't EKS-exclusive and self-managed/kubeadm-based clusters commonly expose kube-apiserver on 6443. so i'll update the default to allow both 443 and 6443. users on a nonstandard apiserver port can still override via `controller.networkPolicy.egress`

ConnorJC3:
Hi, thanks for the contribution! To be clear this is NOT a full review, as I haven't had time to do that yet (when I do I'll submit it as a GitHub review) so I'm almost certainly missing more minor issues, but a few things that stick out to me skimming the PR you'll definitely need to fix:

- You're failing CI `make verify`
- Most/all of the files are missing a trailing newline, we generally prefer to have trailing newlines in our repo
- You're missing ipv6/pod identity in the default config
- I don't think the node network policy renders for the windows DaemonSet?

mbmtbtbc:
thanks for pointing out the issues @ConnorJC3!
i'll implement all these points asap

mbmtbtbc:
@YuryHrytsuk , @ConnorJC3 i have tried to implement all the required changes

for now, i went with **failing the render** when `node.networkPolicy.enabled=true` and `node.enableWindows=true` (the chart default). also added a companion test.

I think we can give the Windows daemonset its own distinct selector label instead, so the Linux policy would just exclude Windows pods cleanly rather than erroring out. I didn't do that here since it's a breaking label change for anyone with existing Services/monitoring/policies keyed on `app: ebs-csi-node` for the Windows daemonset and felt out of scope for this PR.

please check it out

YuryHrytsuk (approved):
Amazing! Thank you 🙏

PR #2968, opened 2026-07-20
Add note in install.md about CNI plugins that require a hop limit >2

### LINKED ISSUE #2966: Clarify IMDSv2 hop limit requirements for advanced CNI plugins
In the installation docs it is stated that when using IMDSv2, a hop limit of _at least 2_ is required. However i think it would be good to be more clear about when you need to go higher than 2.
/feature

Add a note that says something like: If you are using a CNI plugin that adds extra routing layers, like eBPF overlays, a hop limit of 2 may lead to dropped packets. A hop limit of 3 or higher is often required for the IMDS return packets to reach the pods. 

N/A

**Additional context**
Tested on v1.36.2, increasing the hop limit from 2 to 3 in the metadata fixed the problem of "context deadline exceeded" error when starting up the ebs driver pods.

ConnorJC3:
#### What type of PR is this?

/kind documentation

Fixes #2966 

N/A

```release-note
NONE
```


torredil (approved):
/lgtm
/approve

mdzraf (approved):
/lgtm

PR #2970, opened 2026-07-21
Add extra safeguards for volume limits table scripts

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Add some additional safeguards to the volume limits table scripts:
- Bomb out of `generate-volume-limits-table` if we have less than the expected number of AWS regions. This has multiple times in the past resulted in accidentally dropping instance types that we should not drop.
- Add `detect-potentially-invalid-limits` to `make verify` so it's run in CI. This script is yet to misidentify a instance type, and if it does we can just add a hardcoded exception. The protection we get from running this automatically is much greater than once in a blue moon adding a manual exception.
- Fix file path in `generate-volume-limits-table` as we've since moved that file into its own sub-package.

Test account with all opt-in regions enabled:
```
$ go run ./hack/generate-volume-limits-table
2026/07/21 16:03:24 Discovered 34 regions
2026/07/21 16:03:24 Getting volume limits for ap-south-2...
...
```

Test account WITHOUT all opt-in regions enabled:
```
$ go run ./hack/generate-volume-limits-table
2026/07/21 16:03:38 Failed to discover regions: expected at least 34 regions but got 29 (account is probably missing opt-in regions)
```

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ update: update/gofix update/gofmt update/golangci-fix update/kustomize update/mo
 	@echo "All updates succeeded!"
 
 .PHONY: verify
-verify: verify/govet verify/golangci-lint verify/update
+verify: verify/govet verify/golangci-lint verify/update verify/volume-limits
 	@echo "All verifications passed!"
 
 .PHONY: cluster/create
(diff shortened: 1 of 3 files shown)


ConnorJC3:
/retest
/hold

Holding so it doesn't interfere with the release

ElijahQuinones (approved):
/lgtm
/approve

mdzraf (approved):
/lgtm
/approve

ConnorJC3:
/remove-hold

Release is done now

PR #2972, opened 2026-07-22
Switch Windows tests to m7a family instances

ConnorJC3:
#### What type of PR is this?

/kind cleanup

We're seeing an excessive number of Windows E2E flakes that look like:
```
MountVolume.MountDevice failed for volume "pvc-796c8723-d264-4ec8-a8f8-376194a4b00d" : rpc error: code = NotFound desc = Failed to find device path /dev/xvdaa. disk number for device path "/dev/xvdaa" volume id "vol-0b21011e877c74079" not found
```

To the best of our knowledge, this is a Microsoft and/or AWS-side issue that cannot be fixed or worked around in the driver. The relevant teams have been informed/engaged on this issue, but in the meantime we can work around it in our tests by using a different instance type. Testing indicates `m7a` either is not susceptible to this issue, or at the very least it is much rarer.

This will slightly increase the cost of the Windows E2E job but should in theory result in a net cost savings by reducing the number of times the job has to be rerun in entirety.

```release-note
NONE
```

### DIFF
--- a/hack/e2e/eksctl/cluster.yaml
+++ b/hack/e2e/eksctl/cluster.yaml
@@ -58,7 +58,7 @@ managedNodeGroups:
     amiFamily: WindowsServer2025CoreContainer
     desiredCapacity: 3
     disablePodIMDS: true
-    instanceTypes: [m5.2xlarge]
+    instanceTypes: [m7a.2xlarge]
     ssh:
       allow: false
       enableSsm: true
@@ -87,7 +87,7 @@ nodeGroups:
     ami: {{ .Env.WINDOWS_AMI }}
     desiredCapacity: 3
     disablePodIMDS: true
-    instanceType: m5.2xlarge
+    instanceType: m7a.2xlarge
     ssh:
       allow: false
       enableSsm: true


ConnorJC3:
/hold

Don't want this to merge in the middle of the release

ElijahQuinones (approved):
/lgtm

hack/e2e/eksctl/cluster.yaml
```diff
@@ -58,7 +58,7 @@ managedNodeGroups:
     amiFamily: WindowsServer2025CoreContainer
     desiredCapacity: 3
     disablePodIMDS: true
-    instanceTypes: [m5.2xlarge]
+    instanceTypes: [m7a.2xlarge]
```
torredil, line 61: Should also update it for this node group https://github.com/ConnorJC3/aws-ebs-csi-driver/blob/71266f21dc6e0ff8e58bba7058de8bbb44fd91b7/hack/e2e/eksctl/cluster.yaml#L90

ConnorJC3: Duh, can't believe I missed that. We should probably have an environment variable for this honestly but I'm not going to over-complicate this PR.

ConnorJC3:
/remove-hold

@torredil's comment fixed and release is done so ready for reviews/merge

PR #2994, opened 2026-08-20
Auto-detect VAC server version in E2E tests

ConnorJC3:
#### What type of PR is this?

/kind cleanup

Currently, our internal VAC tests hardcode the `v1beta1` API version for VAC tests. VAC `v1beta1` is disabled in Kubernetes 1.37, so it fails there.

Switch instead to auto-detecting if `v1` is available, using that if so, and falling back to `v1beta1` if it isn't (we can't drop `v1beta1` entirely because it would break the oldest EKS version test as 1.31 doesn't have the GA version of VAC).

CI + manually executed against a 1.37 cluster with `make e2e/functional`.

```release-note
NONE
```


ConnorJC3:
/cherrypick release-1.64

Pre-loading this onto our release branch in case we end up needing to do a patch release after 1.37 is GA.

torredil (approved):
/lgtm

mdzraf (approved):
/lgtm 
/approve

PR #3001, opened 2026-09-01
fix: serviceMonitor URL in Windows example

cdirubbio:
#### What type of PR is this?
/kind documentation

Fixing servicemonitor URL from deprecated `dotnetbinaries` URL, see https://github.com/dotnet/announcements/issues/351. Just making this PR after making the same change/fix here https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/704

Note that the version had to be bumped since microsoft only added the latest version `.10` to the new download location, rather than all historic versions, such as `.6`

Deployed the manifest and it runs as expected.

```release-note
NONE
```

### DIFF
--- a/examples/kubernetes/windows/manifests/windows.yaml
+++ b/examples/kubernetes/windows/manifests/windows.yaml
@@ -40,7 +40,7 @@ spec:
           command:
             - powershell.exe
             - -command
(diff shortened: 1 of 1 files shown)


AndrewSirenko (approved):
/lgtm

PR #3017, opened 2026-09-10
Use PULL_BASE_REF for image tag and version in cloudbuild.sh

ConnorJC3:
#### What type of PR is this?

/kind bug

Fixes an issue we saw during the 1.66 release - for some reason, `cloudbuild.sh` takes the image tag from a different place than the version which can lead to a mismatch. We ALWAYS want to use the `PULL_BASE_REF`, there's no reason to use another variable.

N/A

```release-note
NONE
```


ElijahQuinones (approved):
/lgtm

torredil (approved):
/lgtm
/approve

PR #3019, opened 2026-09-16
Enable preventPodSchedulingIfMissing on k8s 1.37+

ConnorJC3:
#### What type of PR is this?

/kind feature

Enable Kubernetes feature to prevent pods from being scheduled to nodes without a `CSIDriver` on k8s 1.37 and later (where the feature is enabled in beta).

Unfortunately, there's some negative interactions with this feature and some rarer configs of `cluster-autoscaler`. To avoid issues, I have added a parameter to allow users to opt out of the feature and added some wording about it in `install.md`.

- Ran `make`, `make update`, `make verify`, `make test`, and `make test/helm-template` (also added new tests).
- Manually rendered on a few different k8s versions.
- Tested the CA issue doesn't affect Karpenter on a Karpenter cluster.

```release-note
NOTE FOR RELEASE AUTHOR: Create important changes section and move this to that (see previous releases and copy the markdown for them. This release note should be in the Helm changelog, NOT the main driver changelog.

Enable `preventPodSchedulingIfMissing` by default on Kubernetes 1.37 and later. **If you use Cluster Autoscaler, you MUST validate `--enable-csi-node-aware-scheduling=true` is enabled** (the default in Cluster Autoscaler 1.37.0 and later) or disable this feature by passing `false` to parameter `preventPodSchedulingIfMissing`. See [the relevant Kubernetes documentation](https://kubernetes.io/docs/concepts/storage/storage-limits/#csi-volume-attach-limits-and-cluster-autoscaler) for more information.
```


AndrewSirenko (commented):
/lgtm

/hold

docs/install.md
```diff
@@ -183,6 +183,11 @@ helm upgrade --install aws-ebs-csi-driver \
...
+
+On Kubernetes 1.37 and later, the Helm chart [prevents Pods that use EBS volumes from being scheduled to nodes where the EBS CSI driver is not registered](https://kubernetes.io/docs/concepts/storage/storage-limits/#preventing-pod-placement-without-csi-driver). This is disabled by default on earlier Kubernetes versions. Set `preventPodSchedulingIfMissing` to `true` or `false` to override the version-based default regardless of the Kubernetes version.
+
+When you use [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) 1.35.0 or later, ensure `--enable-csi-node-aware-scheduling=true` (the default in Cluster Autoscaler 1.37.0 and later) before you enable `preventPodSchedulingIfMissing`. Cluster Autoscaler 1.35 and 1.36 require this flag to be set explicitly. Without this option, Cluster Autoscaler might fail to scale up for pending Pods that use EBS volumes because its scheduling simulations do not include the required `CSINode` information. For more information, see [CSI volume attach limits and cluster autoscaler](https://kubernetes.io/docs/concepts/storage/storage-limits/#csi-volume-attach-limits-and-cluster-autoscaler).
```
torredil, line 189: Should we also mention that scale-from-zero needs an explicit declaration of the CSI drivers a node group will register? the tag key for us will be:

```
k8s.io/cluster-autoscaler/node-template/csi-driver=ebs.csi.aws.com
```

charts/aws-ebs-csi-driver/templates/csidriver.yaml
```diff
@@ -8,6 +8,11 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: false
+  {{- if ne .Values.preventPodSchedulingIfMissing nil }}
+  preventPodSchedulingIfMissing: {{ .Values.preventPodSchedulingIfMissing }}
+  {{- else if semverCompare ">=1.37.0-0" .Capabilities.KubeVersion.Version }}
+  preventPodSchedulingIfMissing: true
```
torredil, line 14: If this parameter is mutable, I'm wondering if it would be best / safer to default false for now until we have a CA release where the scale from zero fix https://github.com/kubernetes/autoscaler/pull/10133 is available.

docs/install.md
ConnorJC3: This PR doesn't change that fact right? Like even if you have `preventPodSchedulingIfMissing: false`, I believe that is already the case with CAS 1.37.0 and later?

If I'm right about that, I'd prefer not to re-document CAS's requirements if this change doesn't affect them. If I'm wrong and this change somehow triggers that requirement I think we can consider it.

Edit: Going to go test that assumption and will report back.

charts/aws-ebs-csi-driver/templates/csidriver.yaml
ConnorJC3: I don't have a strong opinion, personally I think this is such a niche usecase the changelog warning and ability to turn it off probably suffices, but if other people want to wait I don't mind.

I think if we decide as a team to wait, the better course of action is probably just to hold this PR until we are comfortable with merging it rather than change the default behavior of the parameter in a later release.

torredil: I'm OK with default on and changelog warning.

docs/install.md
torredil: Seems like a fairly important detail worth documenting IMO, but I will let you decide.

ConnorJC3: Ok I tested and for some insane reason I don't understand `preventPodSchedulingIfMissing` actually DOES change the CAS behavior here, so I will mention it both in the docs and changelog. As a result we are 100% going to need to wait on this CR for github.com/kubernetes/autoscaler/pull/10133 to merge I think.

ConnorJC3:
/hold

Can't merge this until github.com/kubernetes/autoscaler/pull/10133 is merged

docs/install.md
torredil: OK thanks for testing and catching that. FWIW, since this is mutable I think it would be nice to still provide the toggle for non-CAS users and then switch default enabled once the CAS issues are sorted out.

PR #3021, opened 2026-09-17
Properly cap iops for volumes below minimum iops below minimum iops

ConnorJC3:
#### What type of PR is this?

/kind bug

Today, the driver fails to cap the iops of a volume when BOTH of the following are true:
- The requested iops exceeds the maximum per-GB volume iops limit
- The calculated iops cap is below the volume's minimum iops

In this circumstance, the volume should be capped at exactly the minimum iops for the volume type. For example, requesting a GP3 volume with `iopsPerGb` of 3001 at a size of 1GiB must cap to 3000 iops, but doesn't without this fix.

Manually tested in cases that errored before (e.g. gp3/iopsPerGb 3001/1 GiB). Also added some unit tests for this code path as none exist today.

```release-note
Properly cap volume iops when the requested iops are below the volume type's minimum allowed iops
```


ElijahQuinones (approved):
/lgtm

mdzraf (approved):
/lgtm 
/approve

